PR #4, opened 2020-05-19
First version of the build command.

facundobatista:
Still in draft, let's discuss several points risen here.

### DIFF
--- a/charmcraft/commands/build.py
+++ b/charmcraft/commands/build.py
@@ -77,8 +77,12 @@ def link(srcpath, destdir):
     return destpath
 
 
-def _build(charmdir, entrypoint):
+def build(args):
     """Main entry point."""
+    charmdir = args['from']
+    entrypoint = args['entrypoint']
+    requirement_paths = args['requirement']
+
     buildpath = charmdir / BUILD_DIRNAME
     logger.debug("Building charm in %r", str(buildpath))
 
@@ -117,13 +121,13 @@ def _build(charmdir, entrypoint):
     # XXX Facundo 2020-05-18: we may want to be flexible with how to include the
     # dependencies, e.g. respecting current lib directory, or not having a requirements file
     libpath = buildpath / 'lib'
-    reqspath = charmdir / 'requirements.txt'
     cmd = [
         'pip3', 'install',  # base command
         '--system',  # indicates to use the system file structure
         '--target={}'.format(libpath),  # put all the resulting files in that specific dir
-        '--requirement={}'.format(reqspath),  # the dependencies file
     ]
+    for reqspath in requirement_paths:
+        cmd.append('--requirement={}'.format(reqspath))  # the dependencies file(s)
     retcode = polite_exec(cmd)
     if retcode:
         raise CommandError("problems installing the dependencies")
@@ -142,6 +146,71 @@ def _build(charmdir, entrypoint):
     logger.info("Done, charm left in %r", zipname)
 
 
+class Validator:
+    """A validator of all received options."""
+
+    _options = [
+        'from',  # this needs to be processed first, as it's a base dir to find other files
+        'entrypoint',
+        'requirement',
+    ]
+
+    def __init__(self):
+        self.basedir = None  # this will be fulfilled when processing 'from'
+
+    def process(self, parsed_args):
+        """Process the received options."""
+        result = {}
+        for opt in self._options:
+            meth = getattr(self, 'validate_' + opt)
+            result[opt] = meth(getattr(parsed_args, opt, None))
+        return result
+
+    def validate_from(self, arg):
+        """Validate that the charm dir is there and yes, a directory."""
+        if arg is None:
+            arg = '.'
+        arg = pathlib.Path(arg).expanduser()
+
+        if not arg.exists():
+            raise CommandError("the charm directory was not found: {!r}".format(str(arg)))
+        if not arg.is_dir():
+            raise CommandError(
+                "the charm directory is not really a directory: {!r}".format(str(arg)))
+
+        self.basedir = arg
+        return arg
+
+    def validate_entrypoint(self, arg):
+        """Validate that the entrypoint exists and is executable."""
+        if arg is None:
+            arg = self.basedir / 'src' / 'charm.py'
+        arg = pathlib.Path(arg).expanduser()
+
+        if not arg.exists():
+            raise CommandError("the charm entry point was not found: {!r}".format(str(arg)))
+        if not os.access(arg, os.X_OK):
+            raise CommandError("the charm entry point must be executable: {!r}".format(str(arg)))
+        return arg
+
+    def validate_requirement(self, arg):
+        """Validate that the given requirement(s) (if any) exist.
+
+        If not specified, default to requirements.txt if there.
+        """
+        if arg is None:
+            arg = self.basedir / 'requirements.txt'
+            if arg.exists() and os.access(arg, os.R_OK):
+                return [arg]
+            return []
+
+        arg = [pathlib.Path(x) for x in arg]
+        for fpath in arg:
+            if not fpath.exists():
+                raise CommandError("the requirements file was not found: {!r}".format(str(fpath)))
+        return arg
+
+
 class BuildCommand(BaseCommand):
     """Show the version."""
     name = 'build'
@@ -150,29 +219,21 @@ class BuildCommand(BaseCommand):
     def fill_parser(self, parser):
         """Add own parameters to the general parser."""
         parser.add_argument(
-            'charmdir', metavar='charm-dir',
-            help='The directory where the charm project is located')
+            '-f', '--from',
+            help="the directory where the charm project is located, from where the build "
+                 "is done; defaults to '.'")
         parser.add_argument(
-            'entrypoint', metavar='charm-entrypoint',
-            help='The executable script or program which is the entry point to all the charm code')
+            '-e', '--entrypoint',
+            help="the executable script or program which is the entry point to all the "
+                 "charm code; defaults to 'src/charm.py'")
+        parser.add_argument(
+            '-r', '--requirement', action='append',
+            help="the file(s) with the needed dependencies (this option can be used multiple "
+                  "times); defaults to 'requirements.txt'")
 
     def run(self, parsed_args):
         """Run the command."""
-        charmdir = pathlib.Path(parsed_args.charmdir)
-        if not charmdir.exists():
-            raise CommandError(
-                "indicated charm directory not found: {!r}".format(parsed_args.charmdir))
-        if not charmdir.is_dir():
-            raise CommandError(
-                "indicated charm directory is not a directory: {!r}".format(parsed_args.charmdir))
-
-        entrypoint = pathlib.Path(parsed_args.entrypoint)
-        if not entrypoint.exists():
-            raise CommandError(
-                "indicated charm entry point not found: {!r}".format(parsed_args.entrypoint))
-        if not os.access(entrypoint, os.X_OK):
-            raise CommandError(
-                "indicated charm entry point must be executable: {!r}".format(
-                    parsed_args.entrypoint))
-
-        _build(charmdir, entrypoint)
+        validator = Validator()
+        args = validator.process(parsed_args)
+        logger.debug("working arguments: %s", args)
+        build(args)
--- a/charmcraft/commands/build.py
+++ b/charmcraft/commands/build.py
@@ -65,16 +65,6 @@ def polite_exec(cmd):
     return retcode
 
 
-def link(srcpath, destdir):
-    """Link a file."""
-    if not destdir.exists():
-        os.makedirs(destdir)
-
-    destpath = destdir / srcpath.name
-    destpath.symlink_to(srcpath)
-    return destpath
-
-
 class Builder:
     """The package builder."""
 
@@ -101,20 +91,32 @@ class Builder:
         logger.info("Done, charm left in %r", zipname)
         return zipname
 
+    def _link(self, srcpath, destdir=None):
+        """Link a file."""
+        if destdir is None:
+            destdir = self.buildpath
+        else:
+            if not destdir.exists():
+                destdir.mkdir()
+
+        destpath = destdir / srcpath.name
+        destpath.symlink_to(srcpath)
+        return destpath
+
     def handle_code(self):
         """Handle basic files and the charm source code."""
         # basic files
         logger.debug("Linking in basic files and charm code")
-        link(self.charmdir / CHARM_METADATA, self.buildpath)
+        self._link(self.charmdir / CHARM_METADATA)
 
         # the whole dir/tree if entry point is in a project's subdir, itself alone otherwise
         if self.charmdir in self.entrypoint.parents and self.charmdir != self.entrypoint.parent:
             # link the whole dir
-            linked_subdir = link(self.entrypoint.parent, self.buildpath)
+            linked_subdir = self._link(self.entrypoint.parent)
             linked_entrypoint = linked_subdir / self.entrypoint.name
         else:
             # just the entry point, putting it under a src dir for juju compliance
-            linked_entrypoint = link(self.entrypoint, self.buildpath / 'src')
+            linked_entrypoint = self._link(self.entrypoint, self.buildpath / 'src')
         return linked_entrypoint
 
     def handle_dispatcher(self, linked_entrypoint):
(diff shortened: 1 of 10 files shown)


charmcraft/commands/build.py
```diff
@@ -0,0 +1,178 @@
...
+    # XXX Facundo 2020-05-18: for now, one file, may we copy the entrypoint and its whole dir/tree
+    # (if not charmdir); also we need to understand if we want to ignore some files (e.g. .pyc)
+    linked_entrypoint = link(entrypoint, buildpath / 'src')
+
+    # dispatch mechanism
+    logger.debug("Creating the dispatch mechanism")
+    dispatch_content = DISPATCH.format(
+        entrypoint_relative_path=linked_entrypoint.relative_to(buildpath))
+    dispatch_path = buildpath / DISPATCH_FILENAME
+    with open(dispatch_path, "wt", encoding="utf8") as fh:
```
jameinel, line 105: its a shame that os.open() supports a mask but that isn't exposed in any way for open().
I don't think it matters much here, though. Usually the open-with-mask is because you don't want to create the file readable by others before you write to it, because then other people can spin and grab the file handle before you chmod it.

charmcraft/commands/build.py
```diff
@@ -0,0 +1,178 @@
...
+    hookpath.mkdir()
+    for hookname in HOOK_NAMES:
+        (hookpath / hookname).symlink_to(dispatch_path)
+
+    # dependencies
+    logger.debug("Installing dependencies")
+    # XXX Facundo 2020-05-18: we may want to be flexible with how to include the
+    # dependencies, e.g. respecting current lib directory, or not having a requirements file
+    libpath = buildpath / 'lib'
+    reqspath = charmdir / 'requirements.txt'
```
jameinel, line 120: I feel like we can be explicit about 'charmcraft.yaml', but everything else could be dependent on that. 1 file *must* be accessible to get started.

charmcraft/commands/build.py
```diff
@@ -0,0 +1,178 @@
...
+class BuildCommand(BaseCommand):
+    """Show the version."""
+    name = 'build'
+    help_msg = "build the charm"
+
+    def fill_parser(self, parser):
+        """Add own parameters to the general parser."""
+        parser.add_argument(
+            'charmdir', metavar='charm-dir',
+            help='The directory where the charm project is located')
```
jameinel, line 154: Couldn't this have a default of '.' ?
And entrypoint could have a default of 'src/charm.py' ?

charmcraft/commands/build.py
```diff
@@ -0,0 +1,178 @@
...
+
+# Some constants that are used through the code.
+CHARM_METADATA = 'metadata.yaml'
+BUILD_DIRNAME = 'build'
+DISPATCH_FILENAME = 'dispatch'
+
+# The template for the dispatch script
+DISPATCH = """#!/bin/sh
+
+PYTHONPATH=lib {entrypoint_relative_path}
```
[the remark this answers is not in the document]
facundobatista: You don't really need the `export` if setting the envvar in the same line than the call, right?

charmcraft/commands/build.py
```diff
@@ -0,0 +1,178 @@
...
+
+def polite_exec(cmd):
+    """Execute a command, only showing output if error."""
+    p = subprocess.Popen(
+        cmd, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, universal_newlines=True)
+
+    stdout = []
+    for line in p.stdout:
+        stdout.append(line)
+    retcode = p.wait()
```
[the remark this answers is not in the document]
facundobatista: The idea here was to present some progress. But I deferred that to the future.

With latest news about always logging to a file, I will refactor this a little in the next commit.

charmcraft/commands/build.py
facundobatista: Yes, added that in later commit.

charmcraft/commands/build.py
```diff
@@ -0,0 +1,236 @@
...
+
+def link(srcpath, destdir):
+    """Link a file."""
+    if not destdir.exists():
+        os.makedirs(destdir)
+
+    # XXX Facundo 2020-05-19: we could make hard links here, with the benefit of `juju deploy`
+    # being happy about these links, while at the same time getting the benefit of "changing
+    # a file in the project and being ready to deploy" (however, TRICKY!); that said, we would
+    # need to handle directories by hand (but currently we're not linking dirs...)
```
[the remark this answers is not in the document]
facundobatista: You're right about editors breaking hardlinks, we can not use that really, as would be a debugging mess.

So, if you agree, I'd remove that XXX.

charmcraft/commands/build.py
```diff
@@ -0,0 +1,236 @@
...
+    linked_entrypoint = link(entrypoint, buildpath / 'src')
+
+    # dispatch mechanism
+    logger.debug("Creating the dispatch mechanism")
+    dispatch_content = DISPATCH.format(
+        entrypoint_relative_path=linked_entrypoint.relative_to(buildpath))
+    dispatch_path = buildpath / DISPATCH_FILENAME
+    with open(dispatch_path, "wt", encoding="utf8") as fh:
+        fh.write(dispatch_content)
+    os.chmod(dispatch_path, dispatch_path.stat().st_mode | stat.S_IXUSR)
```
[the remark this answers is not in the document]
facundobatista: Awesome, thanks.

charmcraft/commands/build.py
facundobatista: Yes, I improved it.

charmcraft/commands/build.py
```diff
@@ -0,0 +1,236 @@
...
+    for line in p.stdout:
+        logger.debug(":: %s", line.rstrip())
+    retcode = p.wait()
+
+    if retcode:
+        logger.error("Execution ended in %d for cmd %s", retcode, cmd)
+    return retcode
+
+
+def link(srcpath, destdir):
```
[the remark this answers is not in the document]
facundobatista: Nothing is tested! We need to agree that all this makes sense before jumping into that :)

BTW, maybe we need to make all absolute paths when we read from the cmdline? it would make everything way more explicit in debug logs and everything...

charmcraft/commands/build.py
```diff
@@ -0,0 +1,236 @@
...
+    # bunch of symlinks, to support old juju
+    hookpath = buildpath / 'hooks'
+    hookpath.mkdir()
+    for hookname in HOOK_NAMES:
+        (hookpath / hookname).symlink_to(dispatch_path)
+
+    # dependencies
+    logger.debug("Installing dependencies")
+    # XXX Facundo 2020-05-18: we may want to be flexible with how to include the
+    # dependencies, e.g. respecting current lib directory, or not having a requirements file
```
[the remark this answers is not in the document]
facundobatista: It already defaults to requirement.txt

charmcraft/commands/build.py
```diff
@@ -0,0 +1,236 @@
...
+
+    # z-z-zip it!
+    logger.debug("Creating the package itself")
+    zipname = charmdir.name + '.charm'
+    zipfh = zipfile.ZipFile(zipname, 'w', zipfile.ZIP_DEFLATED)
+    for dirpath, dirnames, filenames in os.walk(buildpath):
+        dirpath = pathlib.Path(dirpath)
+        for filename in filenames:
+            filepath = dirpath / filename
+            zipfh.write(str(filepath), str(filepath.relative_to(buildpath)))
```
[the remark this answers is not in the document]
facundobatista: Why? Just to make the file smaller? I was happy with just getting the whole file in, it's simpler (specially when running this in windows).

charmcraft/commands/build.py
chipaca: "best" is subjective of course, and if it were particularly sensitive I'd suggest using os.open directly :-) (`open()` can take a fd instead of a path)

charmcraft/commands/build.py
```diff
@@ -0,0 +1,281 @@
...
+    def handle_code(self):
+        """Handle basic files and the charm source code."""
+        # basic files
+        logger.debug("Linking in basic files and charm code")
+        link(self.charmdir / CHARM_METADATA, self.buildpath)
+
+        # the whole dir/tree if entry point is in a project's subdir, itself alone otherwise
+        if self.charmdir in self.entrypoint.parents and self.charmdir != self.entrypoint.parent:
+            # link the whole dir
+            linked_subdir = link(self.entrypoint.parent, self.buildpath)
```
jameinel: from a functional perspective, I find using link() everywhere a bit odd. As it means you cannot then use actual symlinks in the .charm because half of the things that are links are meant to be actual content, but some things would be meant as actual links. (eg build/hooks/install should likely be a symlink, but build/src/charm.py shouldn't)

charmcraft/commands/build.py
```diff
@@ -0,0 +1,281 @@
...
+            linked_entrypoint = linked_subdir / self.entrypoint.name
+        else:
+            # just the entry point, putting it under a src dir for juju compliance
+            linked_entrypoint = link(self.entrypoint, self.buildpath / 'src')
```
jameinel: what is the actual use case we are trying to support here? (Juju doesn't actually require the entrypoint to be in src, that was a convention from Gustavo, IIRC.)
This feels like it moves us farther away from the source tree looking like the built charm layout, which I want to put a lot of backpressure on, as it makes debugging much harder than it needs to be.
If our statement is "we want the entry point to look like src/charm.py, even if they don't lay it out that way" then I'd rather say "you have to lay it out this way".
It would make a lot more sense to me if this was:
```
linked_entrypoint = link(self.entrypoint, self.buildpath)
```

IOW, if they declare a different entry point, use it, but leave it in the same relative path as how they defined it.

charmcraft/commands/build.py
```diff
@@ -0,0 +1,284 @@
...
+            with dispatch_path.open("wt", encoding="utf8") as fh:
+                fh.write(dispatch_content)
+                fileno = fh.fileno()
+                os.fchmod(fileno, os.fstat(fileno).st_mode | stat.S_IXUSR)
+
+        # bunch of symlinks, to support old juju
+        current_hookpath = self.charmdir / 'hooks'
+        dest_hookpath = self.buildpath / 'hooks'
+        dest_hookpath.mkdir()
+        for hookname in HOOK_NAMES:
```
jameinel, line 143: I'm a bit surprised with this being "check to see if the charm declares a hooks/* object and then symlink it", but only do that for the 3 hooks we must have. It also isn't checking if 'hooks/install' is just a symlink to src/charm.py, and this would do 
in build/hooks/install => ../../hooks/install => ../src/charm.py

charmcraft/commands/build.py
```diff
@@ -0,0 +1,284 @@
...
+            cmd = [
+                'pip3', 'install',  # base command
+                '--system',  # indicates to use the system file structure
+                '--target={}'.format(venvpath),  # put all the resulting files in that specific dir
+            ]
+            for reqspath in self.requirement_paths:
+                cmd.append('--requirement={}'.format(reqspath))  # the dependencies file(s)
+            retcode = polite_exec(cmd)
+            if retcode:
+                raise CommandError("problems installing the dependencies")
```
jameinel, line 175: problems installing dependencies

tests/commands/test_build.py
```diff
@@ -0,0 +1,692 @@
...
+    validator = Validator()
+    resp = validator.validate_entrypoint('dirX/file.py')
+    testfile = pathlib.Path(str(testfile))  # comparison below don't work well in Py3.5
+    assert resp == testfile
+
+
+def test_validator_entrypoint_expanded():
+    """'entrypoint' param: expands the user-home prefix."""
+    home = pathlib.Path("~").expanduser()
+    testfd, testpath = tempfile.mkstemp(dir=str(home))
```
jameinel, line 176: I'd *really* prefer if test suites don't create files outside of the code directory or $TEMP. creating something in my home dir has a high likelyhood of leaving garbage due to failed test, etc.
I'd actually rather not have the test than have one that touches $HOME.
If there is a different way to test the functionality (set $HOME to somewhere in $TEMP), I'd be ok with it.

tests/commands/test_build.py
```diff
@@ -0,0 +1,692 @@
...
+    cmd = ['echo', 'HELO']
+    polite_exec(cmd)
+    assert [":: HELO"] == [rec.message for rec in caplog.records]
+
+
+def test_politeexec_stderr_logged(caplog):
+    """The standard error is logged in debug."""
+    caplog.set_level(logging.DEBUG, logger="charmcraft")
+
+    cmd = ['python3', '-c', "import sys; print('weird, huh?', file=sys.stderr)"]
```
jameinel, line 325: sys.executable

tests/commands/test_build.py
```diff
@@ -0,0 +1,692 @@
...
+# --- Polite Executor tests
+
+def test_politeexec_base(caplog):
+    """Basic execution."""
+    caplog.set_level(logging.ERROR, logger="charmcraft")
+
+    cmd = ['echo', 'HELO']
+    retcode = polite_exec(cmd)
+    assert retcode == 0
+    assert not any("Execution ended" in rec.message for rec in caplog.records)
```
jameinel, line 309: This seems like a very specific test that isn't going to fail when something gets changed. I'm not sure what a fix would be. Maybe assert that there are no records at ERROR level?

tests/commands/test_build.py
```diff
@@ -0,0 +1,692 @@
...
+
+
+def test_politeexec_failed(caplog):
+    """It's logged in error if cmd fails."""
+    caplog.set_level(logging.ERROR, logger="charmcraft")
+
+    cmd = ['python3', '-c', "exit(3)"]
+    retcode = polite_exec(cmd)
+    assert retcode == 3
+    assert any("Execution ended in 3" in rec.message for rec in caplog.records)
```
jameinel, line 337: "Execution ended in 3" doesn't feel very useful to give as information to users. Maybe its 

"Executing {cmd} failed with return code {retcode}" 
?

tests/commands/test_build.py
```diff
@@ -0,0 +1,692 @@
...
+        'entrypoint': entrypoint,
+        'requirement': [],
+    })
+    linked_entrypoint = builder.handle_code()
+
+    built_metadata = build_dir / CHARM_METADATA
+    assert built_metadata.is_symlink()
+    assert built_metadata.resolve() == metadata
+
+    built_entrypoint = build_dir / 'src' / 'crazycharm.py'  # note it's inside 'src'
```
jameinel, line 419: as mentioned before, I don't prefer this.

charmcraft/commands/build.py
facundobatista: For now this will work, let's discuss in the future if we want some stuff to be copied and other symlinks, or what.

charmcraft/commands/build.py
facundobatista: I was wrong regarding src/!!

Great, now I just put the entrypoint in the builddir, and that's all.

charmcraft/commands/build.py
facundobatista: Improved.

tests/commands/test_build.py
facundobatista: Nice catch, thanks

tests/commands/test_build.py
facundobatista: Good idea! Changed to verify that there are no ERROR records.

tests/commands/test_build.py
facundobatista: Nice! I also improved the message when crashing...

tests/commands/test_build.py
facundobatista: Already changed, yes, thanks

tests/commands/test_build.py
facundobatista: Ah, awesome idea of touching $HOME!!! 

Fixed the code to use that, so no more files outside temp dir. 

Thanks!!

charmcraft/commands/build.py
facundobatista: As we talked on IRC:

1. You're right, I'm not checking "all the other stuff under hooks"... changed the code to "respect" everything the user has under hooks, and also verify that the "mandatory" ones are present, else create them

2. That symlink chain works for now, as the zip is putting just the content; but let's talk about this in the future when we revisit this "why all links?" -- as said above, please open an issue with your ideas

charmcraft/commands/build.py
```diff
@@ -131,19 +127,28 @@ def handle_dispatcher(self, linked_entrypoint):
...
-        # bunch of symlinks, to support old juju
+        # bunch of symlinks, to support old juju; whatever is in the charm's hooks directory
+        # is respect, but also the mandatory ones are created if missing
         current_hookpath = self.charmdir / 'hooks'
         dest_hookpath = self.buildpath / 'hooks'
         dest_hookpath.mkdir()
-        for hookname in HOOK_NAMES:
-            current_hook = current_hookpath / hookname
+        if current_hookpath.exists():
+            current_hooks = set(current_hookpath.iterdir())
```
jameinel, line 136: I'm curious why you use a set here instead of a list.
Given you have to switch to a set later because it is just x.name you want to keep.

facundobatista: Originally I was using them directly, then I found that I had to get the `.name` of each :/.

Moving them to a list for simplicity.

charmcraft/commands/build.py
```diff
@@ -0,0 +1,284 @@
...
+            dispatch_content = DISPATCH_CONTENT.format(
+                entrypoint_relative_path=linked_entrypoint.relative_to(self.buildpath))
+            dispatch_path = self.buildpath / DISPATCH_FILENAME
+            with dispatch_path.open("wt", encoding="utf8") as fh:
+                fh.write(dispatch_content)
+                fileno = fh.fileno()
+                os.fchmod(fileno, os.fstat(fileno).st_mode | stat.S_IXUSR)
+
+        # bunch of symlinks, to support old juju; whatever is in the charm's hooks directory
+        # is respect, but also the mandatory ones are created if missing
```
chipaca, line 131:
```suggestion
        # is respected, but also the mandatory ones are created if missing
```

facundobatista: Fixed.

PR #8, opened 2020-05-22
Have a version in the project, show it in the command and on startup.

facundobatista:
In the future, we could also build the version shown using what's in the file, plus the result of `git describe`, but first we'd need to have tags and (maybe) releases.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,7 +16,17 @@
 #
 # For further info, check https://github.com/canonical/charmcraft
 
+import ast
 import setuptools
+
+
+def get_version():
+    """Extract the version string from project's init."""
+    with open("charmcraft/__init__.py", "rt", encoding='utf8') as fh:
+        for line in fh:
+            if '__version__' in line:
+                return ast.literal_eval(line.split('=', 1)[1].strip())
+
 
 with open("README.md", "rt", encoding='utf8') as fh:
     long_description = fh.read()
@@ -26,7 +36,7 @@ with open("requirements.txt", "rt", encoding='utf8') as fh:
 
 setuptools.setup(
     name="charmcraft",
-    version="0.1.1",
+    version=get_version(),
     author="Facundo Batista",
     author_email="facundo.batista@canonical.com",
     description="The main tool to build, upload, and develop in general the Juju charms.",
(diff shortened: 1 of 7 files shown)


setup.py
```diff
@@ -16,8 +16,18 @@
 #
 # For further info, check https://github.com/canonical/charmcraft
 
+import ast
 import setuptools
 
+
+def get_version():
```
chipaca, line 23: why not `import charmcraft; return charmcraft.__version__`?

jameinel: Agreed.
finding a line in an __init__ script to literal_eval it seems odd here.

charmcraft/commands/version.py
```diff
@@ -29,35 +30,4 @@ class VersionCommand(BaseCommand):
...
-
-class CommandExampleError(BaseCommand):
-    """Just an example."""
-
-    name = 'example-error'
-    help_msg = "show msg in error"
-
-    def run(self, parsed_args):
-        logger.error("Example showing log in ERROR.")
+        logger.info('Version: %s', __version__)
```
jameinel, line 33: I really appreciate this, many tools have come to have this added, as it really aids in debugging when someone is reporting an issue.
In Juju we went as far as updating the build script to embed "git describe --dirty", which gives a version string and the exact hash, and a field to tell you if it might be including local changes as well.

tests/test_infra.py
```diff
@@ -74,3 +77,11 @@ def test_ensure_copyright():
     if issues:
         msg = "Please add copyright headers to the following files:\n" + "\n".join(issues)
         pytest.fail(msg, pytrace=False)
+
+
+def test_setup_version():
+    """Verify that setup.py is picking up the version correctly."""
+    cmd = [os.path.abspath('setup.py'), '--version']
```
jameinel, line 84: do you want
cmd = [sys.executable, os.path.abspath('setup.py')...
?

charmcraft/commands/version.py
facundobatista: Great! We could discuss in the future more specific versions, indeed!

tests/test_infra.py
facundobatista: setup.py is already executable, so just used it...

PR #31, opened 2020-06-11
Include config.yaml in built charm

charness:
Include config.yaml in the zipped charm if present in the source.

Test both with and without a config.yaml in the simple-build case.

### DIFF
--- a/tests/commands/test_build.py
+++ b/tests/commands/test_build.py
@@ -29,6 +29,7 @@ from charmcraft.cmdbase import CommandError
 from charmcraft.commands.build import (
     BUILD_DIRNAME,
     Builder,
+    CHARM_CONFIG,
     CHARM_METADATA,
     DISPATCH_CONTENT,
     DISPATCH_FILENAME,
@@ -405,13 +406,17 @@ def test_build_basic_complete_structure(tmp_path, monkeypatch):
     assert zf.read('lib/ops/stuff.txt') == b"ops stuff"
 
 
-def test_build_code_simple(tmp_path):
+@pytest.mark.parametrize('has_config', [False, True], ids=lambda val: f'has_config={val}')
+def test_build_code_simple(tmp_path, has_config):
     """Check transferred metadata and simple entrypoint."""
     build_dir = tmp_path / BUILD_DIRNAME
     build_dir.mkdir()
 
     metadata = tmp_path / CHARM_METADATA
     entrypoint = tmp_path / 'crazycharm.py'
+    if has_config:
+        config = tmp_path / CHARM_CONFIG
+        config.touch()
 
     builder = Builder({
         'from': tmp_path,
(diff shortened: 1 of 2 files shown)


chipaca:
Actually a better link would probably be [this one](https://ubuntu.com/legal/contributors), that gives you a bit more context before you click through to the form.

tests/commands/test_build.py
```diff
@@ -404,13 +406,17 @@ def test_build_basic_complete_structure(tmp_path):
     assert zf.read('lib/ops/stuff.txt') == b"ops stuff"
 
 
-def test_build_code_simple(tmp_path):
+@pytest.mark.parametrize('has_config', [False, True], ids=lambda val: 'has_config={}'.format(val))
```
facundobatista, line 409: We try to keep the individual tests as simple as possible, so to test the "there's a config" case, please create another test (which would be very similar to this one, but only will check that the config file is included).

charness: My own preference is for reusing the code instead of copying it, but I don't mind conforming to local practice.

tests/commands/test_build.py
```diff
@@ -430,6 +432,38 @@ def test_build_code_simple(tmp_path):
...
+    })
+    linked_entrypoint = builder.handle_code()
+
+    built_metadata = build_dir / CHARM_METADATA
+    assert built_metadata.is_symlink()
+    assert built_metadata.resolve() == metadata
+
+    built_entrypoint = build_dir / 'crazycharm.py'
+    assert built_entrypoint.is_symlink()
+    assert built_entrypoint.resolve() == entrypoint
```
facundobatista, line 458: Sorry, I should have been more explicit in the last comment: This test should test only that , as it's present, the "config" is included (no need to test again metadata, entry point, etc.).

Thanks!!!

chipaca:
@charness any news on the contributor agreement?

charness:
I hope to have clearance from my employer on the contributor agreement in the next couple days. Sorry for the delay there.

chipaca:
No problem. Thank you.

chipaca:
Note #36 also fixes the titular issue here, but this pr does more than that so it is still relevant.

charness:
> Note #36 also fixes the titular issue here, but this pr does more than that so it is still relevant.

Happy to see the more comprehensive fix in #36.

I think that leaves relevant only the first commit on this branch. I had put that change in its own commit because it was separable.

Would you prefer I keep this PR open, with only that commit's content and new summary/description, or close it and potentially open a new PR for just that change when I get cleared for the contributor agreement?

Or perhaps I should just file an issue for the unit test's stray zip so you can fix however you see fit?

chipaca:
I'm happy to keep this open for a while longer until the paperwork goes through, and then we can cherry-pick the first commit (or you can). I did notice it was separate and useful, yes, thank you again :slightly_smiling_face: 

However if you'd rather close it for workflow purposes of your own, that's ok.

If the paperwork does take a _lot_ longer to go through we might do the cleanup ourselves, but we do have our plates full for a couple more weeks still...

chipaca:
Still no rush, but wondering if there was any news.

charness:
Just got cleared for the agreement and submitted.

charness:
With my contributor agreement signed, I think your cherry-picking d6d510652e33ba671176c95fb4b9f50dde83b452 is simplest, but please let me know if you'd prefer a force-push on my branch and updated prose in this PR.

chipaca:
woo! the paperwork finally came through at my end as well. Great!

I'll cherry-pick d6d5106 at some point today. Thank you again!

chipaca:
This is now #60.

PR #36, opened 2020-06-16
'charmcraft build' should pull in actions.yaml and config.yaml

chipaca:


charmcraft/commands/build.py
```diff
@@ -31,6 +31,7 @@
 
 # Some constants that are used through the code.
 CHARM_METADATA = 'metadata.yaml'
+CHARM_OPTIONAL = ['config.yaml', 'actions.yaml']  # copy these if they exist
```
jameinel, line 34: The other optional files according to https://github.com/juju/charm/blob/master/charmdir.go#L69
are:
metadata.yaml # mandatory
config.yaml
metrics.yaml
actions.yaml
revision
lxd-profile.yaml
version

facundobatista: Probably the test case should be transformed into "patching CHARM_OPTIONAL and verifying that the code uses it correctly", instead of checking every combination.

chipaca: that sounds a lot less like testing the wanted behaviour, and a lot more like testing the code as written. I'd rather the first :slightly_smiling_face: but yes it'd be hard to write the code to cover all the things in CHARM_OPTIONAL now. On the other hand, leaving the test as it stands still passes, and still checks that things in optional get copied if they exist (and not if they don't), which is what we want. Is that good enough?

chipaca: (I've not included `revision` as it's not going to be present at build time)

auria: If `revision` is not included at build time, what is the way to know which version of a charm an environment is using? For instance, when a charm has a bug, a Cloud operator may check if a charm is up to date by checking its version, verify it the latest version fixes the found bug or will file a new bug before upgrading to a newer version that doesn't fix their problem.

Without a `revision` file on a production environment, it seems that we can only upgrade to the latest version and file bugs if the upgrade didn't fix a problem.

chipaca: I think you're confusing `version` (for which we have #37) with `revision` (set by the store)...?

auria: Interesting. Thank you for the clarification. In the past, I've seen files such as `revision` and `repo-info` to actually track what `version` is meant for. On the other hand, I'm not familiar with charmcraft but shouldn't CHARM_OPTIONAL track all the files mentioned by jam? They're probably tracked somewhere else.

PR #41, opened 2020-06-17
package it as a snap

chipaca:


snap/snapcraft.yaml
```diff
@@ -0,0 +1,64 @@
...
+grade: stable
+confinement: strict
+
+parts:
+  charmcraft:
+    source: .
+    plugin: python
+    # snapcraft uses venv, which doesn't pull in wheel (as opposed to virtualenv)
+    # so then 'pip install PyYAML' gets cross.
+    python-packages: [wheel]
```
jameinel, line 40: if you want wheel, don't you also then want Cython for PyYAML? (and pythonX-dev and libyaml-dev, and...)

snap/snapcraft.yaml
```diff
@@ -0,0 +1,64 @@
...
+      - pyvenv.cfg
+      - lib/python3.8/site-packages/charmcraft
+      - lib/python3.8/site-packages/charmcraft-*.egg-info
+      - lib/python3.8/site-packages/pip
+      - lib/python3.8/site-packages/pip-*.dist-info
+      - lib/python3.8/site-packages/wheel
+      - lib/python3.8/site-packages/wheel-*.dist-info
+      # this effectively means we're shipping pip and wheel twice (once above,
+      # and once below), but untangling that is probably not worth it.
+      - share/python-wheels
```
jameinel, line 64: I agree that it probably isn't worth worrying about. It is interesting that currently the size of charmcraft is dominated by this duplication, though:
```
$ du . | sort -n | tail
1365    ./lib/python3.8/site-packages/pip/_internal
1376    ./lib/python3.8/site-packages/pip
1577    ./lib
1577    ./lib/python3.8
1577    ./lib/python3.8/site-packages
2187    ./share
2187    ./share/python-wheels
3767    .

$ ls -Slh share/python-wheels | head
total 2.2M
-rw-r--r-- 1 root root 467K Jun 17 14:03 setuptools-44.0.0-py2.py3-none-any.whl
-rw-r--r-- 1 root root 257K Jun 17 14:03 pip-20.0.2-py2.py3-none-any.whl
-rw-r--r-- 1 root root 161K Jun 17 14:03 certifi-2019.11.28-py2.py3-none-any.whl
-rw-r--r-- 1 root root 149K Jun 17 14:03 distlib-0.3.0-py2.py3-none-any.whl
```

That said, ~3-4MB for charmcraft doesn't feel onerous. (juju's snap is 392MB because it has a copy of both 'juju' and 'jujud'. Thinking about it, that means snap juju is *most* of a CD ISO...)

snap/snapcraft.yaml
```diff
@@ -0,0 +1,64 @@
...
+adopt-info: charmcraft  # look for 'snapcraftctl set-*' in the charmcraft part
+architectures: [all]    # nothing platform-dependent here (for now at least)
+
+apps:
+  charmcraft:
+    command: bin/charmcraft
+    plugs:
+      - home            # so it can acess files under the user's home
+      - removable-media # so it can access things in /media, etc. (manually connected)
+      - network         # so 'pip install' can download things
```
jameinel, line 24: Do any of these need official approval? Or are they always available to any charm since they are likely needed by most?

snap/snapcraft.yaml
chipaca: we would, if we were using the yaml that this process builds. But we're not, because core20 already includes it (complete with libyaml), huzzah

snap/snapcraft.yaml
chipaca: These in particular don't. When we pull in git and/or juju we'd need to ask for read permission on the config files and such, which is a reasonably straightforward process (I did it for `icdiff` for example).

PR #43, opened 2020-06-19
Store authentication related commands

facundobatista:


charmcraft/commands/store/__init__.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+class LoginCommand(BaseCommand):
+    """Log into the store."""
+    name = 'login'
+    help_msg = "login with your Ubuntu One e-mail address and password"
+
+    def run(self, parsed_args):
+        """Run the command."""
+        # FIXME: implement! call whoami (to trigger login) AFTER removing current credentials
```
chipaca, line 34: why would `whoami` trigger `login`?

charmcraft/commands/store/client.py
```diff
@@ -0,0 +1,112 @@
...
+    def _load_credentials(self):
+        self._client = httpbakery.Client(cookies=MozillaCookieJar(COOKIE_JAR_FILE))
+
+        if os.path.exists(COOKIE_JAR_FILE):
+            logger.debug("Loading credentials from file: %r", COOKIE_JAR_FILE)
+            try:
+                self._client.cookies.load()
+            except Exception as err:
+                # alert and continue processing (without having credentials, of course, the user
+                # will be asked to authenticate
```
chipaca, line 65:
```suggestion
                # will be asked to authenticate)
```

charmcraft/commands/store/__init__.py
facundobatista: Because ALL server requests may trigger login. It's the only way of triggering a login, BTW.

chipaca:
One disadvantage of the current layout is logs that look like this:
```
2020-06-22 11:44:57,655  charmcraft.main                DEBUG    Parsed arguments: Namespace(_command=<charmcraft.commands.store.WhoamiCommand object at 0x7f21742d1d90>, quiet=False,>
2020-06-22 11:44:57,655  charmcraft.commands.store.client DEBUG    Hitting the store: GET https://api.staging.snapcraft.io/publisher/api/v1/whoami
2020-06-22 11:44:57,655  charmcraft.commands.store.client DEBUG    Credentials file not found: '.charmcraft_credentials'
2020-06-22 11:44:58,136  charmcraft                     ERROR    charmcraft internal error! PermissionError: [Errno 13] Permission denied: 'xdg-settings' (full execution logs in /tmp>
```

(the extra `.client` in the logs could be pruned with no loss of information)

chipaca:
don't forget to add charmcraft.commands.store to setup.py's packages

chipaca:
also the exception when hitting ^C when waiting for the store is ugly

facundobatista:
Trimmed the `.client` part, improved `setup.py`, thanks!

charmcraft/commands/store/client.py
facundobatista: Improved

chipaca:
I think `save_credentials` needs to create the directory before trying to save the file in it:
```
2020-06-23 14:09:14,658  charmcraft.commands.store      DEBUG    Saving credentials to file: '/home/john/snap/charmcraft/x8/.config/charmcraft.credentials'
2020-06-23 14:09:14,659  charmcraft                     ERROR    charmcraft internal error! FileNotFoundError: [Errno 2] No such file or directory: '/home/john/snap/charmcraft/x8/.config/charmcraft.credentials' (full execution logs in /tmp/charmcraft-log-bn9zj38a)
2020-06-23 14:09:14,659  charmcraft.guard               ERROR    
Traceback (most recent call last):
  File "/snap/charmcraft/x8/lib/python3.8/site-packages/charmcraft/main.py", line 167, in main
    dispatcher.run()
  File "/snap/charmcraft/x8/lib/python3.8/site-packages/charmcraft/main.py", line 107, in run
    command.run(self.parsed_args)
  File "/snap/charmcraft/x8/lib/python3.8/site-packages/charmcraft/commands/store/__init__.py", line 64, in run
    result = client.get('/v1/whoami')
  File "/snap/charmcraft/x8/lib/python3.8/site-packages/charmcraft/commands/store/client.py", line 140, in get
    return self._hit('GET', urlpath)
  File "/snap/charmcraft/x8/lib/python3.8/site-packages/charmcraft/commands/store/client.py", line 130, in _hit
    resp = self._auth_client.request(method, url)
  File "/snap/charmcraft/x8/lib/python3.8/site-packages/charmcraft/commands/store/client.py", line 112, in request
    self._save_credentials()
  File "/snap/charmcraft/x8/lib/python3.8/site-packages/charmcraft/commands/store/client.py", line 75, in _save_credentials
    self._cookiejar.save()
  File "/usr/lib/python3.8/http/cookiejar.py", line 2085, in save
    with open(filename, "w") as f:
FileNotFoundError: [Errno 2] No such file or directory: '/home/john/snap/charmcraft/x8/.config/charmcraft.credentials'
```

chipaca:
I think we need to make the `whoami` more like the output of `snap` and `snapcraft`:
```
~$ snap whoami
e-mail: ██████@gmail.com
~$ snapcraft whoami
email:        ██████@gmail.com
developer-id: H5KLL███s3naD3███B1IAW3aBZv
~$ charmcraft whoami
You are John Lenton (username='chipaca', id='lCm███2eedB█████VzjILRxU')
```

chipaca:
(it's essentially yaml, but with the fields aligned to make reading easier)

charmcraft/commands/store/__init__.py
```diff
@@ -0,0 +1,74 @@
...
+    _titles = [
+        ('name:', 'display-name'),
+        ('username:', 'username'),
+        ('id:', 'id'),
+    ]
+
+    def run(self, parsed_args):
+        """Run the command."""
+        client = Client()
+        result = client.get('/v1/whoami')
```
jameinel, line 70: Ultimately it feels like there should be a Python class that knows how to talk to the API and URLs without being the command line code.

iow this would be Client.whoami() which then does a URL /v1/whoami. It feels odd to have the run() methods encoding how to talk to the charmstore API.

Then the commonality between this and Login would be pulled out.

Generally python class talking to the API is very shareable and reusable code that should exist independent of the CLI code.

I'm ok to start here, but I feel like we either want Client to grow more, or Client is a raw client and we grow something above it.

charmcraft/commands/store/client.py
```diff
@@ -0,0 +1,144 @@
...
+import appdirs
+from macaroonbakery import httpbakery
+
+from charmcraft.cmdbase import CommandError
+
+logger = logging.getLogger('charmcraft.commands.store')
+
+# XXX Facundo 2020-06-19: only staging for now; will make it "multi-server" when we have proper
+# functionality in Store's production
+BASE_URL = 'https://api.staging.snapcraft.io/publisher/api'
```
jameinel, line 33: This is definitely something that you'll want to expose as config/global option. Since it makes it easier for us doing dev to iterate. And in the future of potentially enterprise stores.

charmcraft/commands/store/client.py
```diff
@@ -0,0 +1,144 @@
...
+
+    def clear_credentials(self):
+        """Clear stored credentials."""
+        if os.path.exists(self._cookiejar_filepath):
+            os.unlink(self._cookiejar_filepath)
+            logger.debug("Credentials cleared: file %r removed", self._cookiejar_filepath)
+        else:
+            logger.debug("Credentials file not found to be removed: %r", self._cookiejar_filepath)
+
+    def _save_credentials(self):
```
jameinel, line 71: I prefer a name that hints that it might not be done. ensure or _check_save... are helpful to indicate it doesnt always actually perform the action.

charmcraft/commands/store/client.py
```diff
@@ -0,0 +1,144 @@
...
+            except Exception as err:
+                # alert and continue processing (without having credentials, of course, the user
+                # will be asked to authenticate)
+                logger.warning("Failed to read credentials: %r", err)
+        else:
+            logger.debug("Credentials file not found: %r", self._cookiejar_filepath)
+
+        # iterates the cookiejar (which is mutable, may change later) and get the cookies
+        # for comparison after hitting the endpoint
+        self._old_cookies = list(self._cookiejar)
```
jameinel, line 100: if it is mutable do we need a deep copy?

tests/commands/test_store_client.py
```diff
@@ -0,0 +1,321 @@
...
+
+    In detail:
+
+    - writes into a temp dir
+
+    - with a mocked webbrowser module
+    """
+
+    ah = _AuthHolder()
+    ah._cookiejar_filepath = str(tmp_path / 'test.credentials')
```
jameinel, line 44: any reason we couldn't pass this in, instead of patching it after the fact?

tests/commands/test_store_client.py
```diff
@@ -0,0 +1,321 @@
...
+    """Clear credentials removes the file."""
+    caplog.set_level(logging.DEBUG, logger="charmcraft.commands")
+
+    with open(auth_holder._cookiejar_filepath, 'wb') as fh:
+        fh.write(b'fake credentials')
+
+    auth_holder.clear_credentials()
+
+    assert not os.path.exists(auth_holder._cookiejar_filepath)
+    expected = "Credentials cleared: file {!r} removed".format(auth_holder._cookiejar_filepath)
```
jameinel, line 79: is !r what we want? IIRC chipaca mentioned that it is close but not always correct

tests/commands/test_store_client.py
```diff
@@ -0,0 +1,321 @@
...
+    with open(auth_holder._cookiejar_filepath, 'rb') as fh:
+        prv_file_content = fh.read()
+
+    other_cookie = get_cookie(value='different')
+    auth_holder._cookiejar.set_cookie(other_cookie)
+    auth_holder._save_credentials()
+
+    with patch.object(auth_holder._cookiejar, 'save') as mock:
+        auth_holder._save_credentials()
+    assert mock.call_count == 1
```
jameinel, line 178: I'm a bit confused by this test. As we call save multiple times but mock it out some of the time but then assert that the content is actually updated.

tests/commands/test_store_client.py
```diff
@@ -0,0 +1,321 @@
...
+    # save a cookie to be used
+    fake_cookiejar = MozillaCookieJar(auth_holder._cookiejar_filepath)
+    fake_cookiejar.set_cookie(get_cookie())
+    fake_cookiejar.save()
+
+    with patch('macaroonbakery.httpbakery.Client.request') as mock:
+        mock.side_effect = httpbakery.InteractionError('bad auth!!')
+        expected = "Authentication failure: cannot start interactive session: bad auth!!"
+        with pytest.raises(CommandError, match=expected):
+            auth_holder.request('testmethod', 'testurl')
```
jameinel, line 259: should we be testing that this doesn't have a side effect of breaking/deleting the existing cookie file?

charmcraft/commands/store/__init__.py
```diff
@@ -0,0 +1,74 @@
...
+    help_msg = "login in the Ubuntu Single Sign On"
+
+    def run(self, parsed_args):
+        """Run the command.
+
+        The login happens on every request to the Store (if current credentials were not
+        enough), so here we just exercise the simplest command regarding developer identity.
+        """
+        client = Client()
+        client.clear_credentials()
```
jameinel, line 39: do people expect login to clear credentials? or is login while logged in more of "already logged in as..."

eg 'ensure I'm logged in.

charmcraft/commands/store/__init__.py
```diff
@@ -0,0 +1,67 @@
...
+from charmcraft.cmdbase import BaseCommand
+from charmcraft.commands.store.client import Client
+
+logger = logging.getLogger('charmcraft.commands.store')
+
+
+class LoginCommand(BaseCommand):
+    """Log into the store."""
+    name = 'login'
+    help_msg = "login in the Ubuntu Single Sign On"
```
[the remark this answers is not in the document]
facundobatista: improved

charmcraft/commands/store/__init__.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def run(self, parsed_args):
+        """Run the command.
+
+        The login happens on every request to the Store (if current credentials were not
```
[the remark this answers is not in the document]
facundobatista: You're right, changed.

charmcraft/commands/store/__init__.py
facundobatista: Yes, I'll add an `api` layer that knows how to talk to the Store, and remove from this particular file any "magic string". But I wanted to leave that to the next branch when adding more complex commands, as this branch introduced already too many moving parts.

charmcraft/commands/store/client.py
facundobatista: Yes, I plan to do that when we have a second store to aim to ("production").

charmcraft/commands/store/client.py
facundobatista: Nop, the cookiejar is the mutable one, which may change having different cookies after the requests. Each internal cookie will not mutate.

tests/commands/test_store_client.py
facundobatista: It would be something the code would instrument only to the test.

We would need to test what happens if it receives it or if it doesn't, etc.

tests/commands/test_store_client.py
facundobatista: repr() will also have some (rare) corners cases where would be ambiguous, but at least it's not confusing when spaces are used (which is a lot more common)

charmcraft/commands/store/client.py
```diff
@@ -0,0 +1,144 @@
...
+        else:
+            logger.debug("Credentials file not found to be removed: %r", self._cookiejar_filepath)
+
+    def _save_credentials(self):
+        """Save credentials if changed."""
+        if list(self._cookiejar) != self._old_cookies:
+            logger.debug("Saving credentials to file: %r", self._cookiejar_filepath)
+            dirpath = os.path.dirname(self._cookiejar_filepath)
+            if not os.path.exists(dirpath):
+                os.makedirs(dirpath)
```
chipaca, line 77:
```suggestion
            os.makedirs(dirpath, exist_ok=True)
```

tests/commands/test_store_client.py
facundobatista: It's called twice to check that the file was changed, and that it was calling the cookiejar's save. I refactored the test method to make it more clear, and added comments.

tests/commands/test_store_client.py
facundobatista: We don't care, really... if the auth failed, which makes the auth processes to be attempted, which failed, the original file was already of no use.

charmcraft/commands/store/__init__.py
facundobatista: `snapcraft login` asks your credentials twice if you run it twice, so I guess this is normal for these Canonical's tools.

facundobatista: IOW, it doesn't force you to logout first if you want to change your "personality".

facundobatista: All that said, this is something UX will have a final word about, so I wouldn't care too much right now.

charmcraft/commands/store/client.py
facundobatista: Opened #51 for this

charmcraft/commands/store/client.py
facundobatista: Renamed to `_save_credentials_if_changed`.

charmcraft/commands/store/client.py
facundobatista: Improved, thanks!

PR #46, opened 2020-06-23
handle ^C better. Also exit with 1 on error, not -1.

chipaca:


charmcraft/logsetup.py
```diff
@@ -82,6 +82,11 @@ def ended_ok(self):
         """Cleanup after successful execution."""
         os.unlink(self._log_filepath)
 
+    def ended_interrupt(self):
```
facundobatista, line 85: Please provide a test for this method. Thanks!

chipaca: done :-)

charmcraft/commands/build.py
```diff
@@ -65,7 +65,7 @@ def polite_exec(cmd):
             cmd, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, universal_newlines=True)
     except Exception as err:
         logger.error("Executing %s crashed with %r", cmd, err)
-        return -1
+        return 1
```
jameinel, line 68: should this be a different return code for an unknown error?
A lot of tools use a 1 vs 2

charmcraft/logsetup.py
```diff
@@ -82,6 +82,11 @@ def ended_ok(self):
...
+        """Clean up on keyboard interrupt."""
+        _logger.error("Exiting on keyboard interrupt.")
```
jameinel, line 87: this doesn't feel like a great user error. What about just "Interrupted"
And I would log a trace back but not show a trace back to the user. But I think that is already true?

tests/test_logsetup.py
```diff
@@ -112,6 +112,19 @@ def test_ended_success(caplog, create_message_handler):
...
+    """Reports ^C, removes log file."""
+    caplog.set_level(logging.DEBUG, logger="charmcraft")
+
+    mh = create_message_handler()
+    mh.init(mh.NORMAL)
+    mh.ended_interrupt()
+
+    # file is removed, no particular message emitted
+    assert not os.path.exists(mh._log_filepath)
+    assert "Exiting on keyboard interrupt." in [rec.message for rec in caplog.records]
```
jameinel, line 125: do we care about the log level?
should we be asserting anything about a traceback?

tests/test_main.py
```diff
@@ -299,5 +299,16 @@ def test_main_crash():
             d_mock.side_effect = simulated_exception
             retcode = main(['charmcraft', 'version'])
 
-    assert retcode == -1
+    assert retcode == 1
     assert mh_mock.ended_crash.called_once_with(simulated_exception)
+
+
+def test_main_interrupted():
+    """Work crashed: message handler notified properly, return code in -1."""
```
jameinel, line 307: docstring is wrong

charmcraft/logsetup.py
chipaca: my apologies for merging as it stood, i'll be addressing your review comments in a followup :-(

as it stands it _doesn't_ log a trace. The logfile is getting deleted anyway so I didn't think it was worth it.
On the other hand, if `--verbose` was requested, we _could_ log the traceback. Sounds reasonable to me.

tests/test_main.py
chipaca: will fix

charmcraft/commands/build.py
chipaca: python's doc say,

> zero is considered “successful termination” and any nonzero value is considered “abnormal termination” by shells and the like. Most systems require it to be in the range 0-127, and produce undefined results otherwise. Some systems have a convention for assigning specific meanings to specific exit codes, but these are generally underdeveloped; Unix programs generally use 2 for command line syntax errors and 1 for all other kind of errors.

But then [other] [sources] say 0-255 is fine, and 128+n is «Fatal error signal “n”». As SIGINT is 2, we *could* use 130 for that; wdyt?

[other]: https://tldp.org/LDP/abs/html/exitcodes.html
[sources]: https://shapeshed.com/unix-exit-codes/

jameinel:
I don't know how Python handles it. In many runtimes if you return -1 you
actually get 255 or 127:
$ py -c 'import sys; sys.exit(-1)'; echo $?
255
$ py -c 'import sys, time; time.sleep(10)'; echo $?
^CTraceback (most recent call last):
  File "<string>", line 1, in <module>
KeyboardInterrupt
1
$ py -c 'import sys, time; time.sleep(20)'; echo $?
Terminated
143

128+15 (SIGTERM) = 143

So Python's standard is to map ^C to 1, but otherwise SIGNAL gets mapped to
128+SIGNAL.

I don't particularly care. I don't find it super useful either way, but I
know that some people like to interpret return codes.

chipaca:
> I don't particularly care. I don't find it super useful either way, but I know that some people like to interpret return codes.

I say we mess with those people by returning `randint(1, 255)`

jameinel:
nice.

PR #54, opened 2020-06-24
address issues raised in review of #46 that i merged too fast

chipaca:


charmcraft/logsetup.py
```diff
@@ -84,7 +84,10 @@ def ended_ok(self):
 
     def ended_interrupt(self):
         """Clean up on keyboard interrupt."""
-        _logger.error("Exiting on keyboard interrupt.")
+        if self.mode == self.VERBOSE:
+            _logger.exception("Interrupted.")
+        else:
+            _logger.error("Interrupted.")
```
facundobatista, line 90: I'm fine with it, but note that you're not sending the exception to the file in this case.

chipaca: I'm unlinking the file in the next line so I didn't think it was worth sending it there

PR #57, opened 2020-06-25
Make pip list current packages before proceeding with the install.

facundobatista:


tests/commands/test_build.py
```diff
@@ -319,7 +319,11 @@ def test_politeexec_stdout_logged(caplog):
 
     cmd = ['echo', 'HELO']
     polite_exec(cmd)
-    assert [":: HELO"] == [rec.message for rec in caplog.records]
+    expected = [
+        "Running external command ['echo', 'HELO']",
```
jameinel, line 323: This isn't the nicest formatting for a user facing output, but if it is debug logging, then I suppose it is more precise.

tests/commands/test_build.py
```diff
@@ -328,7 +332,11 @@ def test_politeexec_stderr_logged(caplog):
 
     cmd = [sys.executable, '-c', "import sys; print('weird, huh?', file=sys.stderr)"]
     polite_exec(cmd)
-    assert [":: weird, huh?"] == [rec.message for rec in caplog.records]
+    expected = [
+        "Running external command " + str(cmd),
+        ":: weird, huh?",
```
jameinel, line 337: no distinguishing of stdout vs stderr?

tests/commands/test_build.py
chipaca: I was thinking about this, and I think we can print a more user-friendly error without losing the debugging ability by joining the commandline with some way that highlights the difference between spaces between arguments and spaces within arguments. It doesn't have to be perfect as it's just for logging. It could be something as simple as joining the command list with middle dots, so that `['echo', 'HELO WORLD']` becomes `'echo·HELO WORLD'`, or, given that space within args is rarer, it could be just `'echo HELO␠WORLD'`. Both of these are subtle so that they're easy to read if you don't need to know, and clear if you do. IMHO.

chipaca: FWIW that's `"·".join(['echo', 'HELO WORLD'])` vs `" ".join("␠".join(i.split()) for i in ['echo', 'HELO WORLD'])`

chipaca: (to be clear: this is something to be done later, across all places where we log commands)

tests/commands/test_build.py
facundobatista: Not yet, we still need to do that in general for the tool.

facundobatista:
> I think the debug log is intended to be overly precise and technical to aid debugging, rather than helpful to guide users that don't understand. Is that correct?

Yes, I think we need to be super precise, even if it's ugly (that's why, for example, I'm -0 about mangling the logged command...).

Issue #59, opened 2020-06-25
tab completion in-snap


chipaca:
completion's gotta tab

PR #61, opened 2020-06-26
a first pass at tab completion for the snap (hand-crafted)

chipaca:
fixes: #59

### DIFF
--- a/completion.bash
+++ b/completion.bash
@@ -67,8 +67,11 @@ _charmcraft()
     case "$cmd" in
         "build")
             case "$prev" in
-                -r|--requirement|-e|--entrypoint)
-                    _filedir
+                -r|--requirement)
+                    _filedir txt
+                    ;;
+                -e|--entrypoint)
+                    _filedir py
                     ;;
                 -f|--from)
                     _filedir -d
--- a/completion.bash
+++ b/completion.bash
@@ -1,0 +1,80 @@
+# -*- mode: sh; sh-shell: bash; -*-
+
+# Copyright 2020 Canonical Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# For further info, check https://github.com/canonical/charmcraft
+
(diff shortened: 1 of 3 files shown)


completion.bash
```diff
@@ -0,0 +1,80 @@
...
+        --verbose|--quiet)
+            COMPREPLY=( $(compgen -W "${cmds[*]}" -- "$cur") )
+            return
+            ;;
+        login|logout|whoami|version)
+            COMPREPLY=( $(compgen -W "--help" -- "$cur") )
+            return
+            ;;
+        build)
+            COMPREPLY=( $(compgen -W "--help --from --entrypoint --requirement" -- "$cur") )
```
facundobatista, line 46: Please make --entrypoint to filter by "*.py" and --requirement to filter by "*.txt", thanks!

chipaca: done :-)

completion.bash
```diff
@@ -0,0 +1,83 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# For further info, check https://github.com/canonical/charmcraft
+
+_charmcraft()
```
jameinel, line 19: TBH I don't know much about how to write tab completion scripts.  So it is hard for me to know if this is doing the right thing.

completion.bash
```diff
@@ -0,0 +1,83 @@
...
+    # only offer long options, as they should be self-explanatory (and
+    # it's not like it's more typing for the user)
+
+    if [ "$cword" -eq 1 ]; then
+        COMPREPLY=( $(compgen -W "--help --verbose --quiet ${cmds[*]}" -- "$cur") )
+        return
+    fi
+
+    case "$prev" in
+        --help)
```
jameinel, line 34: I see that you are declaring "prev" to be local earlier, and then you are evaluating it here, but I don't see anywhere that you are setting prev. Is it something done by _init_completion?

chipaca: yup,
```bash
# Initialize completion and deal with various general things: do file
# and variable completion where appropriate, and adjust prev, words,
# and cword as if no redirections exist so that completions do not
# need to deal with them.  Before calling this function, make sure
# cur, prev, words, and cword are local, ditto split if you use -s.
#
# Options:
#     -n EXCLUDE  Passed to _get_comp_words_by_ref -n with redirection chars
#     -e XSPEC    Passed to _filedir as first arg for stderr redirections
#     -o XSPEC    Passed to _filedir as first arg for other output redirections
#     -i XSPEC    Passed to _filedir as first arg for stdin redirections
#     -s          Split long options with _split_longopt, implies -n =
# @return  True (0) if completion needs further processing,
#          False (> 0) no further processing is necessary.
#
_init_completion()
```

PR #62, opened 2020-06-26
Command to interact with store for names listing and registration

facundobatista:


charmcraft/commands/store/__init__.py
```diff
@@ -57,16 +56,60 @@ class WhoamiCommand(BaseCommand):
...
-        client = Client()
-        result = client.get('/v1/whoami')
+        store = Store()
+        result = store.whoami()
 
         longest_title = max(len(t[0]) for t in self._titles)
-        for title, key in self._titles:
-            logger.info("%-*s %s", longest_title, title, result[key])
+        for title, attr in self._titles:
+            logger.info("%-*s %s", longest_title, title, getattr(result, attr))
```
chipaca, line 71: as you're touching this code anyway, and have pulled in `tabulate`, please switch this to use tabulate.
it's a lot easier to follow; the pesky `_titles` goes away and the meat of the above becomes just a call to `tabulate`,
```python
logger.info(tabulate([
    ['name:', result.name],
    ['username:', result.username],
    ['id:', result.userid],
], tablefmt='plain')
```
and that's it!  :grin:

facundobatista: Great idea! I had to get the table and log each line separatedly, of course, but the "table building" is done now by tabular. Thanks!

charmcraft/commands/store/__init__.py
```diff
@@ -18,8 +18,10 @@
 
 import logging
 
+from tabulate import tabulate
+
 from charmcraft.cmdbase import BaseCommand
-from charmcraft.commands.store.client import Client
+from charmcraft.commands.store.store import Store
```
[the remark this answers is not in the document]
facundobatista: Nice! Improved.

charmcraft/commands/store/client.py
```diff
@@ -126,18 +126,41 @@ def clear_credentials(self):
...
+        """Get the proper error from the Store response."""
+        default_msg = "Failure working with the Store: [{}] {!r}".format(
+            response.status_code, response.content)
+        try:
+            error_data = response.json()
+        except ValueError:
+            return default_msg
+
+        try:
+            error = error_data['error-list'][0]
```
chipaca, line 139: why's it just grabbing the first one?

charmcraft/commands/store/store.py
```diff
@@ -59,3 +60,15 @@ def whoami(self):
...
+    def register_name(self, name):
+        """Register the specified name for the authenticated user."""
+        self._client.post('/v1/charm', {'name': name})
+
+    def list_registered_names(self):
+        """Return names registered by the authenticated user."""
+        response = self._client.get('/v1/charm')
+        result = []
+        for item in response['charms']:
+            result.append(Charm(name=item['name'], private=item['private'], status=item['status']))
```
chipaca, line 73: reading the code I have no idea what to expect from `private` and `status`. Are they booleans? integers? strings?

I then go down and look at the tests, and discover that `private` is a boolean, and `status` is a string. Is it an arbitrary string, or is it expected to take on discrete values?

facundobatista: This is the [source of truth](https://api.staging.snapcraft.io/publisher/api/docs/default.html) (I now included this in the class' docstring).

`status` is an arbitrary string in the sense it can have multiple values defined by the Store at anytime...

charmcraft/commands/store/client.py
facundobatista: There I handled multiple errors.

charmcraft/commands/store/store.py
```diff
@@ -59,3 +60,15 @@ def whoami(self):
             userid=response['id'],
         )
         return result
+
+    def register_name(self, name):
+        """Register the specified name for the authenticated user."""
+        self._client.post('/v1/charm', {'name': name})
```
jameinel, line 66: I'm rather surprised that all you do is post a name to charm. It seems like it should be more than just that. But if this is the API, then it is what it is.

charmcraft/commands/store/store.py
jameinel: Would it be better to make Charm an actual type instead of a namedtuple and then you can more easily docstring/type annotate it?

tests/commands/test_store_client.py
```diff
@@ -327,3 +354,50 @@ def test_client_clear_credentials():
...
+
+def test_client_errorparsing_nojson():
+    """Produce a default message if response is not a json."""
+    response = FakeResponse(content='this is not a json', status_code=404)
+    result = Client()._parse_store_error(response)
+    assert result == "Failure working with the Store: [404] 'this is not a json'"
+
+
+def test_client_errorparsing_no_errors_inside():
+    """Produce a default message if response is not a json."""
```
jameinel: the comment is wrong here as are the next few tests

charmcraft/commands/store/store.py
facundobatista: Yeap, that is it.

tests/commands/test_store_client.py
facundobatista: Fixed, thanks

charmcraft/commands/store/store.py
facundobatista: Probably yes, but also to be stricter on the response from the Store (which would be the main benefit). After trying different things and talking about it with @chipaca I ended up opening #70. So, in the future.

PR #63, opened 2020-06-29
Added 'src' to PYTHONPATH.

facundobatista:


facundobatista:
The script dir is already in the PTYHONPATH! Python itself handles that (unless you run it in [isolated mode](https://docs.python.org/3/using/cmdline.html#id2)).

PR #67, opened 2020-07-02
A small refactor creating a "store api" layer between commands and the client

facundobatista:
Note that this new layer returns hard-defined structures to the commands (not leaking whatever the store returns). IOW, all dealing-with-store issues are isolated from the commands.

Also I started using `tabulate` for the Whoami command; it may look as overkill for just that, but will use tabulate for other commands in subsequent branches, so let's start here...

This is a work preparing for bringing more commands later (#62).

### DIFF
--- a/charmcraft/commands/store/__init__.py
+++ b/charmcraft/commands/store/__init__.py
@@ -18,8 +18,11 @@
 
 import logging
 
+from tabulate import tabulate
+
 from charmcraft.cmdbase import BaseCommand
-from charmcraft.commands.store.client import Client
+
+from .store import Store
 
 logger = logging.getLogger('charmcraft.commands.store')
 
@@ -31,11 +34,8 @@ class LoginCommand(BaseCommand):
 
     def run(self, parsed_args):
         """Run the command."""
-        # The login happens on every request to the Store (if current credentials were not
-        # enough), so here we just exercise the simplest command regarding developer identity.
-        client = Client()
-        client.clear_credentials()
-        client.get('/v1/whoami')
+        store = Store()
+        store.login()
         logger.info("Login successful")
 
 
@@ -46,8 +46,8 @@ class LogoutCommand(BaseCommand):
 
     def run(self, parsed_args):
         """Run the command."""
-        client = Client()
-        client.clear_credentials()
+        store = Store()
+        store.logout()
         logger.info("Credentials cleared")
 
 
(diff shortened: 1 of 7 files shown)


charmcraft/commands/store/__init__.py
```diff
@@ -57,16 +57,17 @@ class WhoamiCommand(BaseCommand):
...
-        result = client.get('/v1/whoami')
+        store = Store()
+        result = store.whoami()
 
-        longest_title = max(len(t[0]) for t in self._titles)
-        for title, key in self._titles:
-            logger.info("%-*s %s", longest_title, title, result[key])
+        data = [(title, getattr(result, attr)) for title, attr in self._titles]
+        table = tabulate(data, tablefmt='plain')
+        for line in table.split('\n'):
```
chipaca, line 72:
```suggestion
        for line in table.splitlines():
```
?

charmcraft/commands/store/__init__.py
```diff
@@ -57,16 +57,17 @@ class WhoamiCommand(BaseCommand):
...
         """Run the command."""
-        client = Client()
-        result = client.get('/v1/whoami')
+        store = Store()
+        result = store.whoami()
 
-        longest_title = max(len(t[0]) for t in self._titles)
-        for title, key in self._titles:
-            logger.info("%-*s %s", longest_title, title, result[key])
+        data = [(title, getattr(result, attr)) for title, attr in self._titles]
```
chipaca, line 70: why the indirection via `_titles`? I don't get what value it adds.

charmcraft/commands/store/store.py
```diff
@@ -0,0 +1,61 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# For further info, check https://github.com/canonical/charmcraft
+
+"""The Store API handling."""
+
+from collections import namedtuple
```
chipaca, line 19: could you use `typing.NamedTuple` instead?
The 3.5-compatible version is very similar to your usage here (3.6+ is even more interesting but alas)

charmcraft/commands/store/__init__.py
```diff
@@ -18,8 +18,11 @@
 
 import logging
 
+from tabulate import tabulate
```
jameinel, line 21: Shouldn't this be grouped with logging?

charmcraft/commands/store/__init__.py
```diff
@@ -57,16 +57,17 @@ class WhoamiCommand(BaseCommand):
...
-        ('id:', 'id'),
+        ('id:', 'userid'),
     ]
 
     def run(self, parsed_args):
         """Run the command."""
-        client = Client()
-        result = client.get('/v1/whoami')
+        store = Store()
+        result = store.whoami()
```
jameinel, line 68: I like this quite a bit.

charmcraft/commands/store/__init__.py
facundobatista: Nice, done.

charmcraft/commands/store/__init__.py
facundobatista: Nop, logging is stdlib, tabulate is an external module.

charmcraft/commands/store/store.py
facundobatista: I ended up writing #70 for this and related

charmcraft/commands/store/__init__.py
facundobatista: indeed

PR #68, opened 2020-07-02
Totally avoid dealing with paths as strings.

facundobatista:
After this, the code ALWAYS handles Path objects.

### DIFF
--- a/charmcraft/commands/build.py
+++ b/charmcraft/commands/build.py
@@ -237,51 +237,52 @@ class Validator:
             result[opt] = meth(getattr(parsed_args, opt, None))
         return result
 
-    def validate_from(self, arg):
+    def validate_from(self, dirpath):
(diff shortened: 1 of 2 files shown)


charmcraft/commands/build.py
```diff
@@ -240,8 +240,9 @@ def process(self, parsed_args):
     def validate_from(self, arg):
```
jameinel, line 240: Should we add annotations to be clear that arg isn't a str?

facundobatista: Nothing is a string! :p Anyway, we could bring annotations in the future for everything, but that's a separate talk,,,

facundobatista:
Improved the validator arg names. `arg` was something left from a previous structure, now I used names that reflect what is received (a file, a dir, etc).

PR #72, opened 2020-07-08
Send user agent to the store, better log and messages, and other small improvements.

facundobatista:


charmcraft/commands/store/__init__.py
```diff
@@ -54,7 +54,7 @@ def run(self, parsed_args):
 class WhoamiCommand(BaseCommand):
     """Show login information."""
     name = 'whoami'
-    help_msg = "returns your login information relevant to the store"
+    help_msg = "returns your login information relevant to the Store"
```
jameinel, line 57: It feels too strong to use a Proper Noun.
```
$ snapcraft --help
...
  export-login       Save login configuration for a store account in FILE.
...
  whoami             Returns your login information relevant to the store.
```

I wouldn't capitalize them unless there is a design guideline that says we should.
That said, we probably *should* use full Sentence case (with final stop).
So:
```
help_msg = "Returns your login information relevant to the store."
```

charmcraft/commands/store/client.py
```diff
@@ -24,13 +24,17 @@
...
 
 logger = logging.getLogger('charmcraft.commands.store')
 
 # XXX Facundo 2020-06-19: only staging for now; will make it "multi-server" when we have proper
 # functionality in Store's production (related: issue #51)
-BASE_URL = 'https://api.staging.snapcraft.io/publisher/api'
+API_BASE_URL = 'https://api.staging.snapcraft.io/publisher/api'
+
+# XXX Facundo 2020-06-29: we need to include here at least the platform, maybe architecture, etc
+USER_AGENT = "charmcraft/{}".format(__version__)
```
jameinel, line 37: I very much appreciate this
Also relevant for future KPIs

tests/commands/test_store_client.py
```diff
@@ -213,11 +219,12 @@ def test_authholder_request_simple(auth_holder):
 
     other_cookie = get_cookie(value='different')
 
-    def fake_request(self, method, url, json):
+    def fake_request(self, method, url, json, headers):
         # check it was properly called
         assert method == 'testmethod'
         assert url == 'testurl'
         assert json == 'testbody'
+        assert headers == {'User-Agent': USER_AGENT}
```
jameinel, line 227: I do understand Don't Repeat Yourself.
I worry a bit about tests that do this, and become sort of tautologies vs having a useful "the user agent string is something that is a nice useful value"
assert headers == {'User-Agent': 'charmcraft/0.1.0'}

I'm a bit mixed here, but the above test leads to user agent strings that humans never see and you don't realize they've gotten out of hand and aren't properly meaningful.

We don't have to change this, but we should discuss it a bit and decide.

charmcraft/commands/store/__init__.py
facundobatista: That initial capitalization is still on discussion (particularly, I asked about these details to Lilyana, she said still to be defined).

I can go back to lowecase `store`, I just wanted to emphasize that is not "any store", but "THE Store".

charmcraft/commands/store/client.py
facundobatista: BTW, I opened #74 for this.

tests/commands/test_store_client.py
facundobatista: I think those tests should just use the user agent without worrying their format (as the test asserts that it's *using* the UA, nothing else).

That said, you have a point about the format of the UA itself, and never seeing "the final form". That's why I added a test specifically for the UA, which currently is very simple but may grow in complexity when we solve #74 (had to stop using a module constant in favor of a function, but it's fine).

PR #75, opened 2020-07-09
Support for pushing bytes to the Storage.

facundobatista:
This is in preparation for a next branch about the whole Upload command.

### DIFF
--- a/charmcraft/commands/store/client.py
+++ b/charmcraft/commands/store/client.py
@@ -41,7 +41,7 @@ logger = logging.getLogger('charmcraft.commands.store')
 # XXX Facundo 2020-06-19: only staging for now; will make it "multi-server" when we have proper
 # functionality in Store's production (related: issue #51)
 API_BASE_URL = 'https://api.staging.snapcraft.io/publisher/api'
-STORAGE_BASE_URL = 'https://storage.staging.snapcraftcontent.com/'
+STORAGE_BASE_URL = 'https://storage.staging.snapcraftcontent.com'
 
 
 def build_user_agent():
@@ -138,21 +138,22 @@ class _AuthHolder:
 
 def _storage_push(monitor):
     """Push bytes to the storage."""
-    session = requests.Session()
-    retries = Retry(total=5, backoff_factor=2, status_forcelist=[500, 502, 503, 504])
-    session.mount("https://", HTTPAdapter(max_retries=retries))
-
-    url = STORAGE_BASE_URL + 'unscanned-upload/'
+    url = STORAGE_BASE_URL + '/unscanned-upload/'
     headers = {
         'Content-Type': monitor.content_type,
         'Accept': 'application/json',
         'User-Agent': build_user_agent(),
     }
-    try:
-        response = session.request('POST', url, headers=headers, data=monitor)
-    except RequestException as err:
-        raise CommandError("Network error when pushing file: {}({!r})".format(
-            err.__class__.__name__, str(err)))
+    retries = Retry(total=5, backoff_factor=2, status_forcelist=[500, 502, 503, 504])
+
+    with requests.Session() as session:
+        session.mount("https://", HTTPAdapter(max_retries=retries))
+
+        try:
+            response = session.post(url, headers=headers, data=monitor)
+        except RequestException as err:
+            raise CommandError("Network error when pushing file: {}({!r})".format(
+                err.__class__.__name__, str(err)))
 
     return response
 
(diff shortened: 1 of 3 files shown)


tests/commands/test_store_client.py
```diff
@@ -434,3 +438,116 @@ def test_client_errorparsing_bad_structure():
...
+    test_filepath = tmp_path / 'supercharm.bin'
+    with test_filepath.open('wb') as fh:
+        fh.write(b"abcdefgh")
+
+    with patch('charmcraft.commands.store.client._storage_push') as mock:
+        raw_content = dict(successful=False, upload_id=None)
+        mock.return_value = FakeResponse(content=json.dumps(raw_content), status_code=200)
+        with pytest.raises(CommandError) as cm:
+            Client().push(test_filepath)
+        assert str(cm.value) == "Server error when pushing file: " + repr(raw_content)
```
jameinel, line 513: Server error while sending file? while pushing file?

tests/commands/test_store_client.py
```diff
@@ -434,3 +438,116 @@ def test_client_errorparsing_bad_structure():
...
+        'Accept': 'application/json',
+        'User-Agent': build_user_agent(),
+    }
+    assert mock().request.called_once_with('POST', url, headers=headers, data=test_monitor)
+
+    # check the retries were properly setup
+    (protocol, adapter), _ = mock().mount.call_args
+    assert protocol == 'https://'
+    assert isinstance(adapter, HTTPAdapter)
+    assert adapter.max_retries.backoff_factor == 2
```
jameinel, line 538: Did you want a test that we actually do a retry?
Should we be logging a warning of some sort if we have to retry?
What happens with the progress bar if we get a retry after uploading 25% of the file?

charmcraft/commands/store/client.py
```diff
@@ -126,6 +136,27 @@ def request(self, method, url, body):
         return resp
 
 
+def _storage_push(monitor):
+    """Push bytes to the storage."""
+    session = requests.Session()
```
jameinel, line 141: you don't seem to be closing the Session.
Looking over:
https://requests.readthedocs.io/en/master/user/advanced/

it seems to recommend using them as a context manager so you can guarantee their cleanup.

charmcraft/commands/store/client.py
```diff
@@ -126,6 +136,27 @@ def request(self, method, url, body):
...
+    retries = Retry(total=5, backoff_factor=2, status_forcelist=[500, 502, 503, 504])
+    session.mount("https://", HTTPAdapter(max_retries=retries))
+
+    url = STORAGE_BASE_URL + 'unscanned-upload/'
```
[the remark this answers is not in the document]
facundobatista: We can not use `path.join` as it depends on the platform (it's for directories!).

The one I intended to use is `urllib.urljoin`, though, but has some weird semantics:

```
>>> parse.urljoin('http://foo.com/bar/baz/', 'unscanned-upload/')
'http://foo.com/bar/baz/unscanned-upload/'
>>> parse.urljoin('http://foo.com/bar/baz', 'unscanned-upload/')
'http://foo.com/bar/unscanned-upload/'
```

In any case, your worrying is valid. So I added a slash here and removed it in the base url:

```
STORAGE_BASE_URL = 'https://storage.staging.snapcraftcontent.com'
    url = STORAGE_BASE_URL + '/unscanned-upload/'
```

The worst thing it could happen is that somebody for testing uses a slash at the end of that base url (`http:localhost:8080/`), which would create an url with a double slash (`http:localhost:8080//unscanned-upload/`), which is valid anyway.

charmcraft/commands/store/client.py
```diff
@@ -182,3 +213,33 @@ def get(self, urlpath):
...
+
+        def _progress(monitor):
+            # XXX Facundo 2020-07-01: use a real progress bar
+            if monitor.bytes_read <= monitor.len:
+                progress = 100 * monitor.bytes_read / monitor.len
+                print("Uploading... {:.2f}%\r".format(progress), end='', flush=True)
+
+        with filepath.open('rb') as fh:
+            encoder = MultipartEncoder(
+                fields={"binary": ("filename", fh, "application/octet-stream")})
```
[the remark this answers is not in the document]
facundobatista: I thought it was a kind of flag, but I was wrong. From `requests_toolbelt` source code:

```
                elif len(v) == 3:
                    file_name, file_pointer, file_type = v
```

So I changed it to just use the real filename

tests/commands/test_store_client.py
facundobatista: Changed this and other errors to use "while", thanks. I like "pushing" as it represents the used method.

tests/commands/test_store_client.py
facundobatista: I don't want to test that "retry works", as I'm not implementing that (just using the lib).

There's no risk for the progress bar, as it uses "total bytes read" (it's not advancing the bar as "per ticks").

charmcraft/commands/store/client.py
facundobatista: Oh, nice! Thanks

charmcraft/commands/store/client.py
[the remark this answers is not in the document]
jameinel: hm. very strange behavior for urljoin. I think I was confused with Golang which has path.join() and filepath.join(), where the latter is platform specific and the former is always '/'.

PR #76, opened 2020-07-09
Store upload

facundobatista:
This is the next branch after #75

### DIFF
--- a/charmcraft/commands/store/__init__.py
+++ b/charmcraft/commands/store/__init__.py
@@ -152,7 +152,7 @@ class UploadCommand(BaseCommand):
         else:
             # the path is given, asume the charm name is part of the file name
             # XXX Facundo 2020-06-30: Actually, we need to open the ZIP file, extract the
-            # included metadata.yaml file, and read the name from there
+            # included metadata.yaml file, and read the name from there. Issue: #77.
             charm_filepath = charm_filepath.expanduser()
             if not os.access(str(charm_filepath), os.R_OK):  # access doesnt support pathlib in 3.5
                 raise CommandError(
@@ -180,5 +180,5 @@ class UploadCommand(BaseCommand):
             logger.info("Revision %s of %r created", result.revision, str(name))
         else:
             # XXX Facundo 2020-06-30: at some point in the future the Store will give us also a
-            # reason why it failed, to improve the message.
+            # reason why it failed, to improve the message. Issue: #78.
             logger.info("Upload failed: got status %r", result.status)
--- a/charmcraft/commands/store/__init__.py
+++ b/charmcraft/commands/store/__init__.py
@@ -144,7 +144,7 @@ class UploadCommand(BaseCommand):
                     "the --charm-file option.")
 
             charm_filepath = pathlib.Path(charm_name + '.charm').absolute()
-            if not os.access(charm_filepath, os.R_OK):  # access does not support pathlib in 3.5
+            if not os.access(str(charm_filepath), os.R_OK):  # access doesnt support pathlib in 3.5
                 raise CommandError(
                     "Can't access charm file {!r}. You can indicate a charm file with "
                     "the --charm-file option.".format(str(charm_filepath)))
@@ -154,7 +154,7 @@ class UploadCommand(BaseCommand):
             # XXX Facundo 2020-06-30: Actually, we need to open the ZIP file, extract the
             # included metadata.yaml file, and read the name from there
             charm_filepath = charm_filepath.expanduser()
-            if not os.access(charm_filepath, os.R_OK):  # access does not support pathlib in 3.5
+            if not os.access(str(charm_filepath), os.R_OK):  # access doesnt support pathlib in 3.5
                 raise CommandError(
                     "Can't access the indicated charm file: {!r}".format(str(charm_filepath)))
             if not charm_filepath.is_file():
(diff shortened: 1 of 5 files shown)


charmcraft/commands/store/__init__.py
```diff
@@ -113,3 +116,69 @@ def run(self, parsed_args):
...
+                    metadata = yaml.safe_load(fh)
+                charm_name = metadata['name']
+            except (yaml.error.YAMLError, OSError, KeyError):
+                raise CommandError(
+                    "Can't access name in 'metadata.yaml' file. The 'upload' command needs to be "
+                    "executed in a valid project's directory, or point to a charm file with "
+                    "the --charm-file option.")
+
+            charm_filepath = pathlib.Path(charm_name + '.charm').absolute()
+            if not os.access(charm_filepath, os.R_OK):  # access does not support pathlib in 3.5
```
chipaca, line 147: you added the comment, but you didn't `str()` the `Path` :-)

chipaca: You could also do `charm_filepath.exists()` tho

charmcraft/commands/store/__init__.py
```diff
@@ -113,3 +116,69 @@ def run(self, parsed_args):
...
+                raise CommandError(
+                    "Can't access charm file {!r}. You can indicate a charm file with "
+                    "the --charm-file option.".format(str(charm_filepath)))
+
+        else:
+            # the path is given, asume the charm name is part of the file name
+            # XXX Facundo 2020-06-30: Actually, we need to open the ZIP file, extract the
+            # included metadata.yaml file, and read the name from there
+            charm_filepath = charm_filepath.expanduser()
+            if not os.access(charm_filepath, os.R_OK):  # access does not support pathlib in 3.5
```
chipaca, line 157: str the charm_filepath then! :-)

charmcraft/commands/store/store.py
```diff
@@ -76,3 +88,29 @@ def list_registered_names(self):
...
         return result
+
+    def upload(self, name, filepath):
+        """Upload the content of filepath to the indicated charm."""
+        upload_id = self._client.push(filepath)
+
+        endpoint = '/v1/charm/{}/revisions'.format(name)
+        response = self._client.post(endpoint, {'upload-id': upload_id})
+        status_url = response['status-url']
+        logger.debug("Upload id indicated to the Store, got status url %s", status_url)
```
chipaca, line 99: I'm not sure what `Upload id indicated` means. Also "to the store" seems to me to not add anything to the message. How about `"Upload %s started, got status url %s", upload_id, status_url`?

chipaca: of course it's not 'started'. Hrmph.

charmcraft/commands/store/__init__.py
facundobatista: That it exists doesn't mean that you can read it.

Fixed it.

charmcraft/commands/store/__init__.py
facundobatista: Yes :)

charmcraft/commands/store/store.py
facundobatista: I like it! Improved.

Also found that I forgot to test logs, did that now.

charmcraft/commands/store/store.py
```diff
@@ -76,3 +88,29 @@ def list_registered_names(self):
...
+            (revision,) = response['revisions']
+            status = revision['status']
+
+            if status in UPLOAD_ENDING_STATUSES:
+                return Uploaded(
+                    ok=UPLOAD_ENDING_STATUSES[status],
+                    status=status, revision=revision['revision'])
+
+            # XXX Facundo 2020-06-30: Implement a slight backoff algorithm and fallout after
+            # N attempts (which should be big, as per snapcraft experience). Issue: #79.
```
jameinel, line 115: I feel like this should also eventually stop retrying. What happens if the status ends up as ERROR or some other string that isn't strictly 'rejected' or 'accepted' ?
If anything, I would think that we would only retry if we understand the status and it indicates we aren't done yet. (instead of always retrying unless we are sure we shouldn't)

charmcraft/commands/store/__init__.py
```diff
@@ -113,3 +116,69 @@ def run(self, parsed_args):
...
+            charm_filepath = pathlib.Path(charm_name + '.charm').absolute()
+            if not os.access(charm_filepath, os.R_OK):  # access does not support pathlib in 3.5
+                raise CommandError(
+                    "Can't access charm file {!r}. You can indicate a charm file with "
+                    "the --charm-file option.".format(str(charm_filepath)))
+
+        else:
+            # the path is given, asume the charm name is part of the file name
+            # XXX Facundo 2020-06-30: Actually, we need to open the ZIP file, extract the
+            # included metadata.yaml file, and read the name from there
```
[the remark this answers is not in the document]
facundobatista: Yes. Opened #77 for this.

charmcraft/commands/store/__init__.py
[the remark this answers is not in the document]
facundobatista: Indeed, already fixed, thanks

charmcraft/commands/store/store.py
facundobatista: Yes, the #79 description involves a timeout.

Regarding on "what check", I talked with Matias (from the Store) and we agreed on check on those both tokens to flag "verification ended", and leave the rest for other possible intermediate verification steps (that may grow/shrink in time). 

I will ask for more certainties about this, anyway.

PR #82, opened 2020-07-15
Work in preparation for the list-revisions branch.

facundobatista:
I just isolated a helper function to get the charm name from the metadata file, and improved the tests history around that.

Also sneaked in the new domain for the Store API base url.

### DIFF
--- a/charmcraft/commands/store/__init__.py
+++ b/charmcraft/commands/store/__init__.py
@@ -30,6 +30,17 @@ from .store import Store
 logger = logging.getLogger('charmcraft.commands.store')
 
 
+def get_name_from_metadata():
+    """Return the name (if present) from metadata file (if there and readable and sane)."""
+    try:
+        with open('metadata.yaml', 'rb') as fh:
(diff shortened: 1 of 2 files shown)


charmcraft/commands/store/__init__.py
```diff
@@ -30,6 +30,17 @@
 logger = logging.getLogger('charmcraft.commands.store')
 
 
+def get_name_from_metadata():
+    """Return the name (if present) from metadata file (if there and readable and sane)."""
```
jameinel, line 34: this is ok, but would it be better to actually just load all of metadata, and keep it around, so that things wouldn't have to parse it multiple times to get a single piece from it?
I suppose that could come in a later refactoring.

facundobatista: We only get the name from it, so far. And yes, we get from two different places, but as those are different commands, they are not run "at the same time". 

IOW, we only parse the metadata once per charmcraft process execution.

PR #84, opened 2020-07-15
Renamed 'list' command to 'names' in sync with latest UX decisions.

facundobatista:


tests/commands/test_store_commands.py
```diff
@@ -151,7 +151,7 @@ def test_list_registered_one_public(caplog, store_mock):
     ]
     store_mock.list_registered_names.return_value = store_response
 
-    ListRegisteredCommand('group').run(noargs)
+    ListNamesCommand('group').run(noargs)
 
     assert store_mock.mock_calls == [
         call.list_registered_names(),
```
chipaca, line 157: shouldn't we also rename `list_registered_names`, dropping `list_`?

facundobatista: I tend to have separated the "semantics to the user" from the "semantics to the store". That method remains with that name because of the Store's API: `https://api.staging.snapcraft.io/publisher/api/docs/default.html#list_registered_names`

PR #85, opened 2020-07-16
When building replace current hooks that are links to the charm.

facundobatista:


charmcraft/commands/build.py
```diff
@@ -147,25 +147,35 @@ def handle_dispatcher(self, linked_entrypoint):
...
             logger.debug("Including current %r hook", current_hook.name)
             dest_hook = dest_hookpath / current_hook.name
             dest_hook.symlink_to(current_hook)
 
-        # include mandatory ones if missing
-        missing_mandatory = MANDATORY_HOOK_NAMES - {x.name for x in current_hooks}
-        for hookname in missing_mandatory:
+        # include the mandatory ones (if missing) and those we need to replace
+        missing = MANDATORY_HOOK_NAMES - {x.name for x in current_hooks_ok}
+        missing |= {x.name for x in current_hooks_to_replace}
```
jameinel, line 177: Unfortunately this doesn't give you the logging information that one is replacing a symlink while the others are being populated as 'necessary' hooks.
It would probably good to log that at some point.

charmcraft/commands/build.py
```diff
@@ -147,25 +147,35 @@ def handle_dispatcher(self, linked_entrypoint):
...
+        # environment in the middle)
+        current_hooks_ok = []
+        current_hooks_to_replace = []
         if current_hookpath.exists():
-            current_hooks = list(current_hookpath.iterdir())
-        else:
-            current_hooks = []
+            for node in current_hookpath.iterdir():
+                if node.resolve() == self.entrypoint:
+                    current_hooks_to_replace.append(node)
```
chipaca, line 165: further to @jameinel's comment about logging, I think something like
```suggestion
                    current_hooks_to_replace.append(node)
                    logger.debug("Ignoring existing hook %r as it's a symlink to the entrypoint", node)
```
would do the trick

PR #86, opened 2020-07-16
The "list revisions from the Store" command

facundobatista:


tests/commands/test_store_commands.py
```diff
@@ -370,3 +372,169 @@ def test_upload_discover_default_no_charm_file(tmp_path, monkeypatch):
...
+
+def test_revisions_ordered_by_revision(caplog, store_mock):
+    """Results are presented ordered by revision in the table."""
+    caplog.set_level(logging.INFO, logger="charmcraft.commands")
+
+    # three Revisions with all values weirdly similar, the only difference is revision, so
+    # we really assert later that it was used for ordering
+    store_response = [
+        Revision(
+            revision=1, version='v1', created_at=datetime.datetime(2020, 7, 3, 20, 30, 40),
```
chipaca, line 454: I would've stored that timestamp in a `t0` and used that, so i didn't have to read all of them to be happy that they're the same :-)

jameinel: :+1:

charmcraft/commands/store/store.py
```diff
@@ -114,3 +118,19 @@ def upload(self, name, filepath):
...
+    def list_revisions(self, name):
+        """Return charm revisions for the indicated charm."""
+        response = self._client.get('/v1/charm/{}/revisions'.format(name))
+        result = []
+        for item in response['revisions']:
+            errors = [Error(message=e['message'], code=e['code']) for e in (item['errors'] or [])]
+            result.append(Revision(
+                revision=item['revision'],
+                version=item['version'],
+                # we could use `datetime.datetime.fromisoformat` after we drop Py35 support
```
chipaca, line 131: > classmethod datetime.fromisoformat(date_string)
> [...]
>    New in version 3.7.

tests/commands/test_store_commands.py
facundobatista: Improved!

charmcraft/commands/store/__init__.py
```diff
@@ -190,3 +191,52 @@ def run(self, parsed_args):
             # XXX Facundo 2020-06-30: at some point in the future the Store will give us also a
             # reason why it failed, to improve the message. Issue: #78.
             logger.info("Upload failed: got status %r", result.status)
+
+
+class ListRevisionsCommand(BaseCommand):
```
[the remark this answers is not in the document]
facundobatista: I like it this way, it's more verbose but expresses better what it does.

charmcraft/commands/store/store.py
facundobatista: Oh, right! Improved this message and removed the one in `requirements.txt`

PR #87, opened 2020-07-17
get version from git

chipaca:
Also drop unused and untested VERSION. It doesn't work well with the
PEP440 suffixes, and if we need something like that using the standard
library's packaging.version.Version directly would be better (but as
long as our versions are pep-compliant users can simply
packaging.version.parse() our version and things work).

### DIFF
--- a/charmcraft/version.py
+++ b/charmcraft/version.py
@@ -27,7 +27,7 @@ def _get_version():
     if (p.parent / '.git').exists():
         try:
             proc = subprocess.run(
-                ['git', 'describe', '--tags', '--long', '--dirty'],
+                ['git', 'describe', '--tags', '--dirty'],
                 stdout=subprocess.PIPE,
                 stderr=subprocess.DEVNULL,
                 cwd=p,
--- a/charmcraft/version.py
+++ b/charmcraft/version.py
@@ -1,0 +1,50 @@
+# Copyright 2020 Canonical Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


charmcraft/version.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+def _get_version():
+    version = _FALLBACK + ".dev0+unknown"
+
+    p = Path(__file__).parent
+    if (p.parent / '.git').exists():
+        try:
+            proc = subprocess.run(
+                ['git', 'describe', '--tags', '--long', '--dirty'],
```
jameinel, line 30: Didn't we decide we *didn't* want --long?

setup.py
```diff
@@ -27,26 +28,40 @@
...
+version_path = Path("charmcraft/version.py")
+version_backup = Path("charmcraft/version.py~")
+version_path.rename(version_backup)
+try:
+    with version_path.open("wt", encoding="utf8") as fh:
+        fh.write('''\
+# this is a generated file
+
+version = {!r}
+'''.format(charmcraft.__version__))
```
jameinel, line 40: Don't you need to grab `charmcraft.__version__` before you rename version.py ? or is it already done by `__init__`?

charmcraft/version.py
chipaca: hah! yes indeed. I grabbed it from the wrong branch of `ops` :-)

setup.py
chipaca: `__init__` does it (and we can use it here, and not in `ops`, because here `__init__` doesn't import other things)

setup.py
```diff
@@ -27,26 +28,40 @@
...
-    },
-    python_requires='>=3',
-    install_requires=requirements,
-)
+version_path = Path("charmcraft/version.py")
+version_backup = Path("charmcraft/version.py~")
+version_path.rename(version_backup)
+try:
+    with version_path.open("wt", encoding="utf8") as fh:
+        fh.write('''\
```
facundobatista, line 36: Please move this text outside (to module level) to a "template" or something, so it doesn't break visual indentation. Thanks!

chipaca: I did it differently, please confirm it's alright :-)

facundobatista:
It's fine, go for it.

PR #89, opened 2020-07-20
Store release

facundobatista:


charmcraft/commands/store/__init__.py
```diff
@@ -240,3 +240,49 @@ def run(self, parsed_args):
...
+    name = 'release'
+    help_msg = "relase a charm revision to one or more channels"
+
+    def fill_parser(self, parser):
+        """Add own parameters to the general parser."""
+        parser.add_argument(
+            'channels', metavar='channel', nargs='+',
+            help="the channel(s) to release to")
+        parser.add_argument('--name', help="the name of the charm to get revisions")
+        parser.add_argument(
```
jameinel, line 256: 'snapcraft release --help' says:
```
snapcraft release [OPTIONS] <snap-name> <revision> <channels>
```

Which means you must always supply the positional revision.
Are we intentionally diverging from that? Is it part of the CLI review?

charmcraft/commands/store/__init__.py
```diff
@@ -240,3 +240,49 @@ def run(self, parsed_args):
...
+class ReleaseCommand(BaseCommand):
+    """Release a charm revision to specific channels."""
+    name = 'release'
+    help_msg = "relase a charm revision to one or more channels"
+
+    def fill_parser(self, parser):
+        """Add own parameters to the general parser."""
+        parser.add_argument(
+            'channels', metavar='channel', nargs='+',
+            help="the channel(s) to release to")
```
jameinel, line 254: The snapcraft release help text at least says it is:
```
      snapcraft release my-snap 9 beta,edge
```

Which is a comma separated list, rather than separat args. I like your way better, but I'm not sure whether we are going by a new design or not.

charmcraft/commands/store/__init__.py
```diff
@@ -240,3 +240,49 @@ def run(self, parsed_args):
...
+                    "the --name option.")
+
+        if parsed_args.revision:
+            revision = parsed_args.revision
+        else:
+            # find out which is the latest revision for the charm
+            revisions = store.list_revisions(charm_name)
+            if not revisions:
+                raise CommandError(
+                    "The charm {!r} doesn't have any uploaded revision to release."
```
jameinel, line 281: "The charm "foo" doesn't have any uploaded revisions to release."

or

"The charm "foo" doesn't have an uploaded revision to release."

(any is plural)

charmcraft/commands/store/__init__.py
```diff
@@ -240,3 +240,49 @@ def run(self, parsed_args):
...
+                    .format(charm_name))
+            revision = max(rev.revision for rev in revisions)
+
+        store.release(charm_name, revision, parsed_args.channels)
+        logger.info(
+            "Revision %d of charm %r released ok to %s",
```
jameinel, line 287: 'released ok to' is a little clumsy. Is it ok if we just did "released to" ?

charmcraft/commands/store/store.py
```diff
@@ -134,3 +134,9 @@ def list_revisions(self, name):
                 errors=errors,
             ))
         return result
+
+    def release(self, name, revision, channels):
+        """Release one or more revisions for a package."""
+        endpoint = '/v1/charm/{}/releases'.format(name)
+        items = [{'revision': revision, 'channel': channel} for channel in channels]
+        self._client.post(endpoint, items)
```
jameinel, line 142: Interesting that the API supports arbitrary revisions to arbitrary channels.
Is this for the "release all the different Architectures simultaneously" support?
Do we have a design for how we might do that in our CLI?
(AIUI, this is around 'i want to release the Juju 2.8.1 snap, but that involves revision 1234 for amd64, 1235 for arm64, 1239 for ppc64le, etc.)

charmcraft/commands/store/__init__.py
```diff
@@ -240,3 +240,49 @@ def run(self, parsed_args):
...
+                raise CommandError(
+                    "Can't access name in 'metadata.yaml' file. The 'release' command needs to "
+                    "be executed in a valid project's directory, or indicate the charm name with "
+                    "the --name option.")
+
+        if parsed_args.revision:
+            revision = parsed_args.revision
+        else:
+            # find out which is the latest revision for the charm
+            revisions = store.list_revisions(charm_name)
```
jameinel, line 278: Over time, the list of all revisions grows without bound. Is there a way to have the store tell us the latest revisions rather than telling us all of them?
I'm concerned when you've released 1000 versions of a charm that it takes a while and generates a lot of data we don't care about.

charmcraft/commands/store/__init__.py
facundobatista: We're intentionally diverging from that: with @chipaca we thought it was a good idea to default to "last revision", as it's what you normally release.

charmcraft/commands/store/__init__.py
facundobatista: We choose what we liked the best, let's see if UX people have opinions on this...

charmcraft/commands/store/__init__.py
chipaca: to be clear, we didn't intentionally diverge on a whim; after talking with snapcraft and ux people, the current snapcraft ux for release is Bad, so we're trying something different

charmcraft/commands/store/__init__.py
chipaca: or just "[...] doesn't have any uploaded revisions."

charmcraft/commands/store/store.py
chipaca: yes, that's for this. We're yet to discuss how to do it in the CLI, but it got messy so I suggested we move ahead without it and revisit when we know more

charmcraft/commands/store/__init__.py
facundobatista: Improved.

charmcraft/commands/store/__init__.py
facundobatista: Improved.

charmcraft/commands/store/store.py
facundobatista: Yeap, specially in the "not only arch but also OS and series" world.

charmcraft/commands/store/__init__.py
facundobatista: Processing a list of a 1000 items should be fast. Currently there are no ways of requesting that to the store, but they may grow something for this, as the processing time for them would be worse.

IOW, I'm more concerned on how much time the Store endpoint will take to answer, than in the time we take to process the response...

chipaca: it's not about where the slowness is, but that it's slow for the user. Let's ask the store, maybe they agree and make it a priority (hopefully its not a massive amount of work)

facundobatista: Will do

charmcraft/commands/store/__init__.py
```diff
@@ -240,3 +240,48 @@ def run(self, parsed_args):
...
+            if not revisions:
+                raise CommandError(
+                    "The charm {!r} doesn't have any uploaded revisions.".format(charm_name))
+            revision = max(rev.revision for rev in revisions)
+
+        store.release(charm_name, revision, parsed_args.channels)
+        logger.info(
+            "Revision %d of charm %r released to %s",
+            revision, charm_name, ", ".join(parsed_args.channels))
```
chipaca, line 287: I was going to say "hey why don't you use `_join_and` that I just wrote" only to realise it's in `ops`, not here :-D

PR #90, opened 2020-07-21
charmcraft/jujuignore.py: ability to parse .jujuignore

jameinel:
This adds some tests around functionality for handling .jujuignore rules
so that we can build charms including files based on how users
explicitly state what they want included vs ignored.

This passes the somewhat limited test suite that Charm's jujuignore
handling supported.


charmcraft/jujuignore.py
```diff
@@ -0,0 +1,215 @@
...
+            # humans like line numbers to start from 1
+            line_num = line_num + 1
+            orig_rule = rule
+            rule = rule.lstrip()
+            if not rule or rule.startswith('#'):
+                continue
+            invert = False
+            if rule.startswith('!'):
+                invert = True
+                rule = rule[1:]
```
chipaca, line 178: AFAIR if the rule starts with multiple bangs juju treats it as a single one, so maybe what we want here is `rule = rule.lstrip('!')`?

charmcraft/jujuignore.py
```diff
@@ -0,0 +1,215 @@
...
+_escapes = {
+    r'\!': '!',
+    r'\ ': ' ',
+    r'\#': '#',
+
+}
+
+
+def _rstrip_whitespace(rule):
+    """Remove trailing whitespace that isn't escaped"""
```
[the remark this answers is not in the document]
jameinel: I just decided to fix it so that it handles it, rather than trying to fix it with documentation.

charmcraft/jujuignore.py
```diff
@@ -0,0 +1,215 @@
...
+
+_escapes = {
+    r'\!': '!',
+    r'\ ': ' ',
+    r'\#': '#',
+
+}
+
+
+def _rstrip_whitespace(rule):
```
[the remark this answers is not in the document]
jameinel: :+1:

charmcraft/jujuignore.py
```diff
@@ -0,0 +1,215 @@
...
+import typing
+
+logger = logging.getLogger(__name__)
+
+KEEP = 'keep'
+SKIP = 'skip'
+FORCEKEEP = 'forcekeep'
+
+
+_escapes = {
```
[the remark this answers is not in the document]
jameinel: 👍

charmcraft/jujuignore.py
```diff
@@ -0,0 +1,215 @@
...
+
+    i, n = 0, len(rule)
+    res = ''
+    while i < n:
+        c = rule[i]
+        i = i + 1
+        if c == '*':
+            if i < n and rule[i] == '*':
+                i = i + 1
+                res = res + '.*'
```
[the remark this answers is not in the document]
jameinel: just because fnmatch didn't use them, but I'm happier with them

charmcraft/jujuignore.py
```diff
@@ -0,0 +1,215 @@
...
+            if j < n and rule[j] == ']':
+                j = j + 1
+            while j < n and rule[j] != ']':
+                j = j + 1
+            if j >= n:
+                res = res + '\\['
+            else:
+                stuff = rule[i:j]
+                if '--' not in stuff:
+                    stuff = stuff.replace('\\', r'\\')
```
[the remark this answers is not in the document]
jameinel: yeah. I know I used r'\' somewhere and it immediately said "no end of string found".
Which seemed to negate the point of r and the handling of '\' but its how it goes.

charmcraft/jujuignore.py
```diff
@@ -0,0 +1,215 @@
...
+                        if k < 0:
+                            break
+                        chunks.append(rule[i:k])
+                        i = k + 1
+                        k = k + 3
+                    chunks.append(rule[i:j])
+                    # Escape backslashes and hyphens for set difference (--).
+                    # Hyphens that create ranges shouldn't be escaped.
+                    stuff = '-'.join(s.replace('\\', r'\\').replace('-', r'\-')
+                                     for s in chunks)
```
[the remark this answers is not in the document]
jameinel: I didn't know about translate. I would have like to have that for unescaping, but it turns out that translate can only handle when the source is a single character.

charmcraft/jujuignore.py
```diff
@@ -0,0 +1,215 @@
...
+class _Matcher:
+    """Couple a regex with other metadata for how we should match a given pattern."""
+
+    def __init__(self, line_num: int, orig_rule: str, invert: bool, only_dirs: bool,
+                 regex: typing.Pattern):
+        self.line_num = line_num
+        self.orig_rule = orig_rule
+        self.invert = invert
+        self.only_dirs = only_dirs
+        self.regex = regex
```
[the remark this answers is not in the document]
jameinel: 👍

charmcraft/jujuignore.py
jameinel: yeah, that is what juju itself is doing: https://github.com/juju/charm/blob/master/jujuignore.go#L176
I'll add a test

charmcraft/jujuignore.py
```diff
@@ -0,0 +1,235 @@
...
+        else:
+            res += re.escape(c)
+    res += r'\Z'
+    return res
+
+
+class _Matcher:
+    """Couple a regex with other metadata for how we should match a given pattern."""
+
+    def __init__(self, line_num: int, orig_rule: str, invert: bool, only_dirs: bool,
```
facundobatista, line 141: We're not using type annotations in the rest of the project (and in other parts of the file)... It bothers me to have these in only some places...

charmcraft/jujuignore.py
```diff
@@ -0,0 +1,235 @@
...
+                invert=invert,
+                only_dirs=only_dirs,
+                regex=regex,
+            )
+            self._matchers.append(m)
+            logger.debug(
+                'Translated .jujuignore %d "%s" => "%s"',
+                line_num,
+                orig_rule,
+                regex)
```
facundobatista, line 205: why not everything in the same line?

charmcraft/jujuignore.py
```diff
@@ -0,0 +1,235 @@
...
+                keep = False
+            elif matchRes == FORCEKEEP:
+                keep = True
+                break
+        return not keep
+
+
+# default_juju_ignore is the initial set of ignores.
+# juju itself always includes these before adding the contents of .jujuignore
+default_juju_ignore = '''
```
facundobatista, line 223: I'm concerned about including PYC files (as we still don't use the `.gitignore` file)

charmcraft/jujuignore.py
```diff
@@ -0,0 +1,235 @@
...
+def _unescape_rule(rule):
+    """Take out escape characters and trailing unescaped whitespace from the rule"""
+    rule = rule.lstrip()
+    rule = _rstrip_unescaped(rule)
+    for old, new in _unescapes.items():
+        rule = rule.replace(old, new)
+    return rule
+
+
+def _rule_to_regex(rule):
```
facundobatista, line 59: It SUPER scares me the complexity of this function.

We need to test this conciusly. I went ahead and removed some lines and the tests kept passing, so we need to improve its coverage.

charmcraft/jujuignore.py
jameinel: These are the defaults from juju. We can diverge, but it means people used to be able to deploy from a directory get different results than deploying a .charm than has been built.
If we don't really intend that you can build in place (which makes me a bit sad) then it matters less to diverge.
Otherwise, that is impetus for them to add a .jujuignore if they don't want the .pyc.

charmcraft/jujuignore.py
```diff
@@ -0,0 +1,235 @@
...
+
+def _rstrip_unescaped(rule):
+    """Remove trailing whitespace that isn't escaped."""
+    i = len(rule) - 1
+    while i >= 0:
+        if rule[i] == '\n' or rule[i] == '\r':
+            rule = rule[:i]
+        elif rule[i] != ' ':
+            break
+        elif i == 0 or rule[i - 1] != '\\':
```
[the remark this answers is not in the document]
jameinel: it can, the while loop is i >= 0. 
If it is 0 the code 'works' because rule[-1] is the last character, but that feels like it 'works' for the wrong reason.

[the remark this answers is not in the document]
facundobatista: ack

charmcraft/jujuignore.py
facundobatista: thanks!

charmcraft/jujuignore.py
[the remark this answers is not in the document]
facundobatista: so, this means that backslash in `.jujuitgnores` is not a scaping char? I'm lost with all these `\`s :p

charmcraft/jujuignore.py
facundobatista: It bothers me the internal inconsistency. But I guess that if they are needed in some points, it doesn't harm. What are you using them for?

charmcraft/jujuignore.py
facundobatista: It's not a matter of including them or not. 

We will NOT include PYCs, as we plan to support `.gitignore` and PYCs are ignored there. So, at the end, we will not have them included. And now, we don't have them included.

My concern is if including them in the brief period between we "get everything except jujuignored" and "get everything except jujuignored and gitignored" is problematic or not.

charmcraft/jujuignore.py
chipaca: I'd rather have them than not. They are an easy way of documenting what's expected, and we should be able to grow them to the point where some static analysis can be done. I'm in no rush to get there, but I'd rather be closer than further away

charmcraft/jujuignore.py
chipaca: It's unclear to me whether we actually want to support `.gitignore` over `.jujuignore`. We want to support at least one of them, for sure, and `.jujuignore` does make some sense because it's what `juju deploy .` does for example and the argument that you can have artifacts you want in the charm but not in git, and viceversa, is a reasonable one.

Supporting both is possible, but as I said yesterday in the standup, supporting one and then circling back to that forum post to see what our users thing is a reasonable step.

Having said all that, supporting `.jujuignore` means having this implicit list. We don't get to have opinions about this list, here; if we need it changed we'd have to address it with juju and then update this one to reflect that. We can add `*.py[cod]` to an explicit `.jujuignore`, and in fact #91 does that, and that's probably good enough.

jameinel:
Thanks for the clarification, will add better coverage and poke here when it is done.

jameinel:
Ok, the testing has now been updating, including running 'git check-ignore' for a bunch of the test patterns and making sure that we match what git does/doesn't match.
Arguably we should be testing Juju, but even there Juju should probably conform to how Git handles the special characters.

PR #91, opened 2020-07-22
First pass at 'charmcraft init'

chipaca:
Some files need more work. Especially README.md.
Some more options needed.
It's not interactive (should it be?).


charmcraft/commands/init.py
```diff
@@ -0,0 +1,112 @@
...
+            optimized=False,             # optimization doesn't make sense for one-offs
+            undefined=StrictUndefined)   # fail on undefined
+
+        # bare minimum:
+        #  🗷 a req.txt w/ops,
+        #  ☐ src/charm.py w/structure & xxx & a minimum hook & stored state
+        #  ☐ a basic fixme test which pulls in the harness and tests that the minimum thing works
+        #  🗷 a run_tests script & flake8
+        #  🗷 a req-dev.txt w/flake8 (etc?)
+        # maybe a templates directory, later :-)
```
chipaca, line 93: whoops forgot to delete my TODO :-)

charmcraft/templates/init/README.md
```diff
@@ -0,0 +1,16 @@
+# {{ name }}
+
+## XXX: fill this out.
+
+## developing
+
+Create a virtualenv in your preferred way, and install `requirements-dev.txt`.
+For example,
+
+    pipenv install --dev
```
chipaca, line 10: I'll make five different readme each with a different way of creating and using a virtualenv :-p

requirements-dev.txt
```diff
@@ -2,3 +2,5 @@
 
 pytest == 5.4.1
 flake8 == 3.7.9
+git+https://github.com/canonical/operator
```
chipaca, line 5: should change this to just `ops` when we release 0.7 (or drop the `harness.cleanup` code in the test until 0.7 is a thing)

charmcraft/commands/_utils.py
```diff
@@ -0,0 +1,26 @@
...
+import os
+from stat import S_IXUSR, S_IXGRP, S_IXOTH
+
+S_IXALL = S_IXUSR | S_IXGRP | S_IXOTH
+
+
+def make_executable(fh):
+    """make open file fh executable"""
+    fileno = fh.fileno()
+    os.fchmod(fileno, os.fstat(fileno).st_mode | S_IXALL)
```
jameinel, line 26: usually you only set X for ones that have read/write something else. I'm ok with this but it does seem odd that you might set executable when that person cannot read the content.
If you wanted that I think it is:
```
mode = os.fstat(fileno).st_mode
mode_r = mode & (S_IRUSR | S_IRGRP | S_IROTH)
mode_x = mode_r >> 2
mode = mode | mode_x
```

charmcraft/commands/init.py
```diff
@@ -0,0 +1,112 @@
...
+
+class InitCommand(BaseCommand):
+    """Initialize a directory to be a charm project."""
+    name = "init"
+    help_msg = "initialize a directory to be a charm project"
+
+    def fill_parser(self, parser):
+        parser.add_argument(
+            "-d", "--directory", type=Path, default=Path("."),
+            help="the directory to initialize. Must not exist or be empty; defaults to '.'")
```
jameinel, line 40: Must be empty or nonexistent ?
If it exists it must be empty?

charmcraft/commands/init.py
```diff
@@ -0,0 +1,112 @@
...
+            logger.debug("Using existing project directory '%s'", args.directory)
+        else:
+            logger.debug("Creating project directory '%s'", args.directory)
+            args.directory.mkdir()
+
+        if args.author is None:
+            gecos = pwd.getpwuid(os.getuid()).pw_gecos.split(',', 1)[0]
+            if not gecos:
+                raise CommandError("Author not given, and nothing in GECOS field")
+            logger.debug("Setting author to %r from GECOS field", gecos)
```
jameinel, line 66: is GECOS useful to report? I know it doesn't mean much to me.

charmcraft/commands/init.py
```diff
@@ -0,0 +1,112 @@
...
+            "name": args.name,
+            "author": args.author,
+            "year": date.today().year,
+            "class_name": "".join(re.split(r"\W+", args.name.title())) + "Charm",
+        }
+
+        env = Environment(
+            loader=PackageLoader('charmcraft', 'templates/init'),
+            autoescape=False,            # no need to escape things here :-)
+            keep_trailing_newline=True,  # they're not text files if they don't end in newline!
```
jameinel, line 83: :)

charmcraft/commands/init.py
```diff
@@ -0,0 +1,112 @@
...
+
+        # bare minimum:
+        #  🗷 a req.txt w/ops,
+        #  ☐ src/charm.py w/structure & xxx & a minimum hook & stored state
+        #  ☐ a basic fixme test which pulls in the harness and tests that the minimum thing works
+        #  🗷 a run_tests script & flake8
+        #  🗷 a req-dev.txt w/flake8 (etc?)
+        # maybe a templates directory, later :-)
+
+        _xxx_rx = re.compile("XXX: (.*)")
```
jameinel, line 95: should we use XXX or should we use TODO: ?

charmcraft/commands/init.py
```diff
@@ -0,0 +1,112 @@
...
+        executables = ["run_tests", "src/charm.py"]
+        for template_name in env.list_templates():
+            if template_name.endswith(".pyc") or template_name.endswith("~"):
+                # would be nice if we could stop from shipping these 🤷
```
jameinel, line 99: Why would we have these? Where is it reading the templates from that we are actually executing the template code? Isn't it illegal code since it hasn't expanded the template variables?

charmcraft/templates/init/LICENSE
```diff
@@ -0,0 +1 @@
+# XXX: replace with the right license file
```
jameinel, line 1: The right stuff
I'm pretty sure I'd start with GPLv3

charmcraft/templates/init/README.md
jameinel: I've always used `python3 -m venv  .venv`
because that comes with python. I support pipenv means you don't have to create and then install it?

charmcraft/templates/init/config.yaml
```diff
@@ -0,0 +1,7 @@
+# Copyright (c) {{ year }} {{ author }}
+# See LICENSE file for licensing details.
+options:
+  thing:
+    default: 🪄
```
jameinel, line 5: I definitely wouldn't default to a magic wand, it doesn't render on Github and many won't actually understand how it should be replaced.

charmcraft/templates/init/run_tests
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh -e
+# Copyright (c) {{ year }} {{ author }}
+# See LICENSE file for licensing details.
+
+flake8 --max-line-length=99
+PYTHONPATH=src${PYTHONPATH:+:}$PYTHONPATH python3 -m unittest -v "$@"
```
jameinel, line 6: behold, the magic of Bash !

charmcraft/templates/init/src/charm.py
```diff
@@ -0,0 +1,30 @@
...
+import logging
+
+from ops.charm import CharmBase
+from ops.main import main
+from ops.framework import StoredState
+
+logger = logging.getLogger(__name__)
+
+
+class {{ class_name }}(CharmBase):
```
jameinel, line 14: Should we default to including a CharmEvents and a no-op event so that people know what that should look like?

charmcraft/templates/init/src/charm.py
```diff
@@ -0,0 +1,30 @@
...
+    _stored = StoredState()
+
+    def __init__(self, *args):
+        super().__init__(*args)
+        self.framework.observe(self.on.config_changed, self._on_config_changed)
+        self._stored.set_default(things=[])
+
+    def _on_config_changed(self, _):
+        current = self.framework.model.config["thing"]
```
jameinel, line 23: I think we generally discourage 'self.framework.*' and just do 'self.model.config'

requirements-dev.txt
jameinel: You don't *need* cleanup unless you're doing resources, but it does feel better to have it in case we add something else in the future.

setup.py
```diff
@@ -62,6 +62,7 @@
         },
         python_requires='>=3',
         install_requires=requirements,
+        include_package_data=True,
```
jameinel, line 65: what does this do?

tests/commands/test_init.py
```diff
@@ -0,0 +1,73 @@
...
+        python_paths=[])
+    pep8_test(paths)
+
+
+def test_init_non_ascii_author(tmp_path):
+    test_init_pep8(tmp_path, author="فلانة الفلانية")
+
+
+def test_init_non_ascii_name(tmp_path):
+    test_init_pep8(tmp_path, name="ჭიჭიკო")
```
jameinel, line 41: fwiw, I know Juju expressly forbids nonascii charm names, I'm not sure if the charm store does/doesn't want to support nonascii elsewhere.

charmcraft/commands/init.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+class InitCommand(BaseCommand):
+    """Initialize a directory to be a charm project."""
+    name = "init"
+    help_msg = "initialize a directory to be a charm project"
+
+    def fill_parser(self, parser):
+        parser.add_argument(
+            "-d", "--directory", type=Path, default=Path("."),
```
facundobatista, line 39: `--directory` is too ambiguous... yes, it's a directory, a directory for what? Maybe `--base-directory`, or `--base`, or `--deploy-directory`?

charmcraft/commands/init.py
```diff
@@ -0,0 +1,112 @@
...
+        parser.add_argument(
+            "--author", type=str,
+            help="the author of the project; defaults to the directory name;"
+            " defaults to the current user's name as present in the GECOS field.")
+
+    def run(self, args):
+        args.directory = args.directory.resolve()
+        if args.directory.exists():
+            if not args.directory.is_dir():
+                raise CommandError("{} is not a directory".format(args.directory))
```
facundobatista, line 54: In other errors, we tend to be more explicit about what is failing or not found, and showing the repr of the string.

IOW, something like:
```
    raise CommandError("the given option to where init the project is not a directory: {!r}".format(str(args.directory)))
```

charmcraft/commands/init.py
```diff
@@ -0,0 +1,112 @@
...
+            for _ in args.directory.iterdir():
```
facundobatista, line 55: it works, but looks weird a for to raise an exception... what about `if next(args.directory.iterdir(), False):`

charmcraft/commands/init.py
facundobatista: I didn't knew about GECOS until now...

charmcraft/commands/init.py
```diff
@@ -0,0 +1,112 @@
...
+                raise CommandError("{} is not empty".format(args.directory))
+            logger.debug("Using existing project directory '%s'", args.directory)
+        else:
+            logger.debug("Creating project directory '%s'", args.directory)
+            args.directory.mkdir()
+
+        if args.author is None:
+            gecos = pwd.getpwuid(os.getuid()).pw_gecos.split(',', 1)[0]
```
facundobatista, line 63: This will not work in windows

charmcraft/commands/init.py
```diff
@@ -0,0 +1,112 @@
...
+                # would be nice if we could stop from shipping these 🤷
+                continue
+            logger.debug("Rendering %s", template_name)
+            template = env.get_template(template_name)
+            path = args.directory / template_name
+            path.parent.mkdir(parents=True, exist_ok=True)
+            with path.open("wt", encoding="utf8") as fh:
+                out = template.render(context)
+                fh.write(out)
+                for xxx in _xxx_rx.findall(out):
```
facundobatista, line 108: Mmm.. I think we need to create the structure and let the user fill it.

As those structures will not be there until user's run `init`, these warnings will always appear, giving the user the impression of that something weird is going on with the `init` command itself.

charmcraft/main.py
```diff
@@ -160,13 +160,16 @@ def _build_argument_parser(self, commands_groups):
...
 
-def main(argv=sys.argv):
+def main(argv=None):
     """Main entry point."""
     # Setup logging, using DEBUG envvar in case dev wants to show info before
     # command parsing.
     mode = message_handler.VERBOSE if 'DEBUG' in os.environ else message_handler.NORMAL
     message_handler.init(mode)
 
+    if argv is None:
```
facundobatista, line 170: thanks for this

charmcraft/templates/init/README.md
facundobatista: Yes, use `python3 -m venv` (as it's always there) or `virtualenv` (which is the canonical tool to create venvs).

charmcraft/templates/init/README.md
```diff
@@ -0,0 +1,16 @@
+# {{ name }}
+
+## XXX: fill this out.
```
facundobatista, line 3: with what?

charmcraft/templates/init/README.md
facundobatista: Also:

https://chriswarrick.com/blog/2018/07/17/pipenv-promises-a-lot-delivers-very-little/
https://www.reddit.com/r/Python/comments/chkah3/is_pipenv_dead_why_has_the_project_stopped/

charmcraft/templates/init/config.yaml
```diff
@@ -0,0 +1,7 @@
+# Copyright (c) {{ year }} {{ author }}
```
facundobatista, line 1: (c) means Copyright, so it's redundant

charmcraft/templates/init/run_tests
facundobatista: Will this work on windows?

Can't we get away without fixing the PATH? What about something like `env/bin/python -m unittest discover tests/`

requirements-dev.txt
facundobatista, line 5: Why we need `ops` in the charmcraft project?

charmcraft/templates/init/run_tests
facundobatista: Also you probably should be able to improve the README saying that just run this file, and here you do:

```
venv/bin/flake8...
venv/bin/python3 -m unittest discover...
```

run_tests
```diff
@@ -1,3 +1,3 @@
 #!/bin/sh
 
-python3 -m pytest -v "$@"
+python3 -m pytest --ignore=charmcraft/templates -v "$@"
```
facundobatista, line 3: revisit this after the `.j2` renaming

tests/commands/test_init.py
facundobatista: The Store is surely opinionated about this (it's probably a subset of ascii, like a-z, A-Z, 0-9, -, and not much more...

tests/test_infra.py
```diff
@@ -29,20 +29,29 @@
...
+    if python_paths is None:
+        python_paths = ['setup.py']
+    if roots is None:
+        roots = ['charmcraft', 'tests']
+    for root in roots:
         for dirpath, dirnames, filenames in os.walk(root):
             for filename in filenames:
                 if filename.endswith(".py"):
                     python_paths.append(os.path.join(dirpath, filename))
+            for i, dirname in enumerate(dirnames):
```
facundobatista, line 43: revisit all this after the `.j2` renaming

charmcraft/templates/init/README.md
chipaca: On Debian-derived `venv` is not always there (you need to install `python3-venv`, which is exactly the same amount of work as installing `virtualenv` except the latter has `command-not-found` support). We might as well go with the 'canonical' one.

charmcraft/commands/init.py
chipaca: if set, it's expected to be the user's name

charmcraft/commands/init.py
chipaca: let's go with TODO

charmcraft/commands/init.py
```diff
@@ -0,0 +1,112 @@
...
+            "-d", "--directory", type=Path, default=Path("."),
+            help="the directory to initialize. Must not exist or be empty; defaults to '.'")
+        parser.add_argument(
+            "--name", type=str,
+            help="the name of the project; defaults to the directory name")
+
+        parser.add_argument(
+            "--author", type=str,
+            help="the author of the project; defaults to the directory name;"
+            " defaults to the current user's name as present in the GECOS field.")
```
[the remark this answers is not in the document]
chipaca: oops

charmcraft/commands/init.py
chipaca: both of those are worse, for different reasons: 'base' doesn't add anything other than length (ie it doesn't _mean_ anything), and 'deploy' means something else entirely :-)
but, yes, directory is not clear.

charmcraft/commands/init.py
chipaca: it's the call to iterdir that'll fail, not looping it, so might as well do that directly

charmcraft/commands/init.py
chipaca: I don't follow, what do you mean?

requirements-dev.txt
chipaca: to run the charm's tests

charmcraft/commands/init.py
chipaca: (discussed in the standup, but: the last step of install is a compileall pass; pytest also aggressively compiles things during discovery but i dissuaded it)

charmcraft/templates/init/run_tests
chipaca: it's not bash but `/bin/sh`, and it'll work on windows as much as charmcraft's and the operator's one will :-)

charmcraft/templates/init/src/charm.py
chipaca: discussed in the standup: no; custom events are an advanced feature really only people writing components should be looking at

charmcraft/commands/_utils.py
```diff
@@ -0,0 +1,26 @@
+# Copyright 2020 Canonical Ltd.
```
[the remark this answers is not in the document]
chipaca: I'll un-private it then

charmcraft/commands/init.py
```diff
@@ -0,0 +1,112 @@
...
+    """Initialize a directory to be a charm project."""
+    name = "init"
+    help_msg = "initialize a directory to be a charm project"
+
+    def fill_parser(self, parser):
+        parser.add_argument(
+            "-d", "--directory", type=Path, default=Path("."),
+            help="the directory to initialize. Must not exist or be empty; defaults to '.'")
+        parser.add_argument(
+            "--name", type=str,
```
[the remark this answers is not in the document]
chipaca: Yes (are you asking that I drop it?)

charmcraft/commands/init.py
chipaca: the whole thing doesn't work on windows (until you run tests there, it's broken) ...
what are you asking that I do?

charmcraft/commands/init.py
chipaca: is `--project-directory` too long? Maybe `--project-dir`?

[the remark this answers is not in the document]
chipaca: > ```
> -f FROM, --from FROM  the directory where the charm project is located, from where the build is done; defaults to '.'
> -e ENTRYPOINT, --entrypoint ENTRYPOINT
>                       the executable script or program which is the entry point to all the charm code; defaults to 'src/charm.py'
> -r REQUIREMENT, --requirement REQUIREMENT
>                       the file(s) with the needed dependencies (this option can be used multiple times); defaults to 'requirements.txt'
> ```
...?

chipaca: I'm liking `--project-dir`, but it does open the question as to whether `build` shouldn't use the same name for the same thing. And if not, maybe `init` should use `--to` (cf. `build`'s `--from`), but that's less than clear for the user imho.

charmcraft/templates/init/run_tests
chipaca: I don't think there's a way for things to work without fixing PYTHONPATH

setup.py
chipaca: without it the templates aren't distributed

tests/commands/test_init.py
chipaca: The [docs] say `^[a-z][a-z0-9-]*[a-z]$`, pretty much

[docs]: https://discourse.juju.is/t/charm-metadata/1043

chipaca: I've added a test for this

charmcraft/commands/init.py
chipaca: gah, i confused this iterdir with os.listdir. Going with your suggestion verbatim

requirements-dev.txt
chipaca: dropped cleanup for now

charmcraft/commands/init.py
chipaca: confirmed the new approach is alright (in a meet)

charmcraft/commands/init.py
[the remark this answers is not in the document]
facundobatista: I'm more wondering why to include it... *I* will not include it, it's ok if you do, but don't know the value it provides.

charmcraft/commands/init.py
facundobatista: You're right.

run_tests
facundobatista: we still need this?

tests/test_infra.py
facundobatista: we still need all these changes?

run_tests
chipaca: huh, i thought i'd dropped that

tests/test_infra.py
chipaca: probably not

chipaca: removed it

setup.py
chipaca: in particular they're not included in the bdist_wheel. I've added a comment to that effect.

run_tests
chipaca: done it now for real

PR #92, opened 2020-07-23
flake8 support

jameinel:
Add .flake8 and tweak the tests so that they use that flake8.
If we want to bikeshed on the excluded files, I'm happy to tweak it.

Essentially by not passing separate arguments to get_style_guide() it finds the .flake8 file, which means manually running 'flake8' (or in my case autopep8) will read the same rules.

This also fixes one PEP8 issue found by autopep8 where we didn't have a blank line between a docstring and the next function.


charmcraft/cmdbase.py
```diff
@@ -21,6 +21,7 @@ class CommandError(Exception):
     It optionally receives a `retcode` parameter that will be the returned code
     by the process on exit.
     """
+
```
facundobatista, line 24: Please revert this change so this file is not included in the change history...

.flake8
```diff
@@ -0,0 +1,12 @@
+[flake8]
+max-line-length = 99
+select: E,W,F,C,N
+exclude:
```
facundobatista, line 4: why do we need this "exclude" list? you're getting flake8 issues from .git dir, for example?

jameinel:
It just means flake8 doesn't have to look at it. I had a branch in the past
called 'foo.py' and flake8 started complaining about that directory.
What I actually found is that flake8 gets *very* slow if you leave things
like venv around because it goes and audits the code of all your
dependencies.

jameinel:
It seemed relevant to include the change to a PEP8 compliance when adding
extra PEP8 checking.

facundobatista:
> What I actually found is that flake8 gets *very* slow if you leave things like venv around because it goes and audits the code of all your

But that will not happen here, as we only check our code:
```
    for root in ['charmcraft', 'tests']:
```

jameinel:
It doesn't matter for the test, it does matter if you run 'flake8' yourself.

PR #94, opened 2020-07-23
Store commands autocomplete

facundobatista:
This depends on #93


completion.bash
```diff
@@ -42,10 +42,28 @@ _charmcraft()
             COMPREPLY=( $(compgen -W "--help" -- "$cur") )
             return
             ;;
+
+        # store related commands
```
chipaca, line 46: misplaced comment is misplaced

completion.bash
```diff
@@ -78,6 +96,13 @@ _charmcraft()
                     ;;
             esac
             ;;
+        "upload")
+            case "$prev" in
+                --charm-file)
+                    _filedir charm
+                    ;;
+            esac
+            ;;
```
chipaca, line 105: also `release`'s `--revision` should autocomplete, n'eh?

completion.bash
```diff
@@ -42,10 +42,28 @@ _charmcraft()
             COMPREPLY=( $(compgen -W "--help" -- "$cur") )
```
chipaca, line 42: way up there ^^^ there's a `cmds=(...)` that lists all the commands the completer knows about...

completion.bash
facundobatista: indeed, fixed, thanks!

completion.bash
facundobatista: As we discussed, let's not hit the Store to autocomplete.

completion.bash
facundobatista: FTR, I removed the "message as a separation/grouping", as I like to keep more simple (grouping everything that doesn't have anything else than --help)

completion.bash
facundobatista: Ah, nice, thanks

chipaca:
Here's the [delta] FWIW

[delta]: https://github.com/facundobatista/charmcraft/compare/store-status...facundobatista:store-commands-autocomplete

completion.bash
```diff
@@ -19,7 +19,7 @@
 _charmcraft()
 {
     local cur prev words cword cmd cmds
-    cmds=(build version login logout whoami)
+    cmds=(build version login logout whoami names upload revisions status release)
```
jameinel, line 22: this is starting to get big enough that it feels like something we'd want to start automatically generating.
Manual is still fine for now, but it does mean having to keep 2 things up to date whenever adding an argument or optional parameter.

PR #96, opened 2020-07-24
Updated snapcraft.yaml for new Py deps.

facundobatista:


snap/snapcraft.yaml
```diff
@@ -128,6 +128,12 @@ parts:
       - lib/google
       - lib/protobuf-*.pth
       - lib/protobuf-*.dist-info
+      - lib/tabulate-*.dist-info
+      - lib/tabulate.py
```
chipaca, line 132: am I very anal if it bothers me that you switched the order of dist-info and the libraries?

chipaca:
thank you for the HOWTO :-)

HOWTO_SNAP.md
```diff
@@ -0,0 +1,61 @@
+Steps:
+
+- First check snapcraft is currently building something fine:
```
jameinel, line 3: First check that snapcraft is able to build something:

HOWTO_SNAP.md
```diff
@@ -0,0 +1,61 @@
...
+
+```
+snapcraft
+```
+
+- Install it and try it, something like the following, replacing 
+  for what you just built:
```
jameinel, line 10: replacing the snap name with the one you just built

HOWTO_SNAP.md
```diff
@@ -0,0 +1,61 @@
...
+sudo snap install --dangerous charmcraft_0.2.0+95.g6f8c4cc_amd64.snap
+```
+
+- If needs to be fixed, see section(s) below, GOTO 10
+
+
+## How to add new Python dependencies
+
+If trying a new snap fails because import errors, you need to include new 
+dependencies. Easiest way to do this is to open the snap, copy manually the
```
jameinel, line 22: manually copy the needed dependency

HOWTO_SNAP.md
facundobatista: thanks

HOWTO_SNAP.md
facundobatista: thanks

HOWTO_SNAP.md
facundobatista: thanks

PR #99, opened 2020-07-29
charmcraft/jujuignore.py: Allow extending the list of patterns.

jameinel:
This adds the ability for JujuIgnore to pull in the default ignores and extend them with .jujuignore contents.

I don't have tests around the walk functionality, but I figured I would leave it in there to show you how I was thinking it would look.

### DIFF
--- a/charmcraft/commands/build.py
+++ b/charmcraft/commands/build.py
@@ -26,6 +26,7 @@ import yaml
 
 from charmcraft.cmdbase import BaseCommand, CommandError
 from .utils import make_executable
+from .jujuignore import JujuIgnore, default_juju_ignore
 
 logger = logging.getLogger(__name__)
 
@@ -87,6 +88,7 @@ class Builder:
         self.requirement_paths = args['requirement']
 
         self.buildpath = self.charmdir / BUILD_DIRNAME
+        self.ignore_rules = self._load_juju_ignore()
 
     def run(self):
         """Main building process."""
@@ -110,11 +112,38 @@ class Builder:
         destpath.symlink_to(srcpath)
         return destpath
 
+    def _load_juju_ignore(self):
+        ignore = JujuIgnore(default_juju_ignore)
+        path = self.charmdir / '.jujuignore'
+        if path.exists():
+            with path.open() as ignores:
+                ignore.extend_patterns(ignores)
+        return ignore
+
+    def _walk_unignored(self, topdir):
+        """Same as os.walk but filter out ignored items."""
+        for top, dirs, nondirs in os.walk(str(topdir)):
+            relpath = pathlib.Path(top).relative_to(topdir)
+            i = 0
+            for d in dirs[:]:
+                if self.ignore_rules.match(str(relpath / d), is_dir=True):
+                    del dirs[i]
+                else:
+                    i++
+            remaining = []
+            for nondir in nondirs:
+                rel = relpath / nondir
+                if not self.ignore_rules.match(rel, is_dir=False):
+                    remaining.append(nondir)
+            yield top, dir, remaining
+
     def handle_code(self):
         """Handle basic files and the charm source code."""
         # basic files
         logger.debug("Linking in basic files and charm code")
         self._link_to_buildpath(self.charmdir / CHARM_METADATA)
+
+
         for fn in CHARM_OPTIONAL:
             path = self.charmdir / fn
             if path.exists():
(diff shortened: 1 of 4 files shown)


charmcraft/commands/build.py
```diff
@@ -109,11 +111,38 @@ def _link_to_buildpath(self, srcpath):
...
 
+    def _load_juju_ignore(self):
+        ignore = JujuIgnore(default_juju_ignore)
+        path = self.charmdir / '.jujuignore'
+        if path.exists():
+            with path.open() as ignores:
+                ignore.extend_patterns(ignores)
+        return ignore
+
+    def _walk_unignored(self, topdir):
```
facundobatista, line 122: Thanks for this.

Anyway, I need to implement the "walk" outside, as it will be ignoring different stuff (a part of which is jujuignored).

So I think this shouldn't be included in this layer.

charmcraft/commands/build.py
```diff
@@ -109,11 +111,38 @@ def _link_to_buildpath(self, srcpath):
...
+                if not self.ignore_rules.match(rel, is_dir=False):
+                    remaining.append(nondir)
+            yield top, dir, remaining
+
     def handle_code(self):
         """Handle basic files and the charm source code."""
         # basic files
         logger.debug("Linking in basic files and charm code")
         self._link_to_buildpath(self.charmdir / CHARM_METADATA)
+
```
facundobatista, line 144: Please reduce this to only one blank line.

charmcraft/jujuignore.py
```diff
@@ -151,7 +151,12 @@ def __init__(self, patterns: typing.Iterable[str]):
         self._matchers = []
         self._compile_from(patterns)
 
-    def _compile_from(self, patterns):
+    def extend_patterns(self, patterns: typing.Iterable[str]) -> None:
+        """Add more patterns to the ignore list.
+        """
```
facundobatista, line 156: Please close the docstring in the same line above.

charmcraft/commands/build.py
```diff
@@ -26,6 +26,7 @@
 
 from charmcraft.cmdbase import BaseCommand, CommandError
 from .utils import make_executable
+from .jujuignore import JujuIgnore, default_juju_ignore
```
facundobatista, line 29: I expect flake8 will fail on these...

jameinel: So it was supposed to be "..jujuignore" since it is in the parent directory, but flake8 doesn't complain at that point.

charmcraft/commands/build.py
```diff
@@ -109,11 +111,20 @@ def _link_to_buildpath(self, srcpath):
         destpath.symlink_to(srcpath)
         return destpath
 
+    def _load_juju_ignore(self):
+        ignore = JujuIgnore(default_juju_ignore)
+        path = self.charmdir / '.jujuignore'
+        if path.exists():
+            with path.open() as ignores:
```
facundobatista, line 118: Shouldn't we specify the encoding here? Which is the expected encoding for the `.jujuignore` file?

charmcraft/commands/build.py
```diff
@@ -109,11 +111,20 @@ def _link_to_buildpath(self, srcpath):
         destpath.symlink_to(srcpath)
         return destpath
 
+    def _load_juju_ignore(self):
```
facundobatista, line 114: This function needs tests.

charmcraft/jujuignore.py
```diff
@@ -151,7 +151,11 @@ def __init__(self, patterns: typing.Iterable[str]):
         self._matchers = []
         self._compile_from(patterns)
 
-    def _compile_from(self, patterns):
+    def extend_patterns(self, patterns: typing.Iterable[str]) -> None:
```
facundobatista, line 154: Thanks! I'll use this also from tests!

charmcraft/commands/build.py
jameinel: .gitignore is officially UTF-8, I don't know that Juju specifically declares an encoding, but there isn't any reason to believe it anything but UTF-8.

jameinel:
I'm happy either way.

charmcraft/commands/build.py
```diff
@@ -26,6 +26,7 @@
 
 from charmcraft.cmdbase import BaseCommand, CommandError
 from .utils import make_executable
+from ..jujuignore import JujuIgnore, default_juju_ignore
```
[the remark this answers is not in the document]
chipaca: It's weird. I like `from . import foo` and `from .foo import bar` and even `from .. import foo`, but not `from ..foo import bar`.

PR #106, opened 2020-08-04
Support old and new field names in Store responses.

facundobatista:


jameinel:
Someone else raised the question "if there is a new version of the Store, how could we be getting either response". Is this more "we should just move to the new names" ?

Ideally we wouldn't have compatibility code before charmhub is actually public if it can leave the code cleaner.

PR #107, opened 2020-08-04
Improve the tests run in the initted project to use the outside virtualenv.

facundobatista:


tests/commands/test_init.py
```diff
@@ -74,6 +76,15 @@ def test_executables(tmp_path):
...
+    env = os.environ.copy()
+    env_paths = [p for p in sys.path if 'env/lib/python' in p]
+    if env_paths:
+        (env_path,) = env_paths
+        env['PYTHONPATH'] += ':' + env_path
+
     cmd = InitCommand('group')
     cmd.run(Namespace(path=tmp_path, name='my-charm', author="だれだれ"))
-    subprocess.run(["./run_tests"], cwd=str(tmp_path), check=True)
+    subprocess.run(["./run_tests"], cwd=str(tmp_path), check=True, env=env)
```
jameinel, line 90: Why are we recursively running all tests in the directory? Is this just that we are testing that './run_tests' works for the charm that we've built?
Isn't our venv already in PYTHONPATH since we're running from there?

facundobatista: This test checks that the "tests run ok for the template project created by init".

The venv path is not in the PYTHONPATH of the child process only if the tests are run after activating our venv. I don't do that (run them calling `venv/bin/python...`).

PR #113, opened 2020-08-07
init: include series in metadata.yaml

chipaca:
fixes: #111


chipaca:
An alternative would be to make series default to kubernetes, instead of parsing os-release

charmcraft/commands/init.py
```diff
@@ -72,11 +76,20 @@ def run(self, args):
         if not re.match(r"[a-z][a-z0-9-]*[a-z0-9]$", args.name):
             raise CommandError("{} is not a valid charm name".format(args.name))
 
+        if not args.series:
+            try:
+                os_release = parse_os_release()
+                # TODO: support non-Ubuntu things
+                args.series = [os_release["UBUNTU_CODENAME"]]
+            except (UnicodeDecodeError, KeyError):
```
facundobatista, line 84: When you can have UnicodeDecodeError?

charmcraft/commands/utils.py
```diff
@@ -29,3 +30,16 @@ def make_executable(fh):
     mode_x = mode_r >> 2
     mode = mode | mode_x
     os.fchmod(fileno, mode)
+
+
+def parse_os_release(filename="/etc/os-release"):
```
facundobatista, line 35: Docstring and meaningful variable names, please.

charmcraft/templates/init/metadata.yaml.j2
```diff
@@ -5,3 +5,6 @@ description: |
   TODO: fill out the charm's description
 summary: |
   TODO: fill out the charm's summary
+{%- if series %}
```
facundobatista, line 8: Shouldn't we **always** have a series?

tests/commands/test_init.py
```diff
@@ -63,17 +65,43 @@ def test_all_the_files(tmp_path):
...
     subprocess.run(["./run_tests"], cwd=str(tmp_path), check=True)
+
+
+def _test_metadata_series(tmp_path, *, os_release=None, args=(), expected):
+    with patch('charmcraft.commands.init.parse_os_release') as osr:
+        osr.return_value = os_release
+
+        cmd = InitCommand('group')
+        cmd.run(Namespace(path=tmp_path, name='my-charm', author="fred", series=args))
+        subprocess.run(["./run_tests"], cwd=str(tmp_path), check=True)
```
facundobatista, line 90: Why are you running all the tests again here?

charmcraft/commands/utils.py
```diff
@@ -29,3 +30,16 @@ def make_executable(fh):
...
+    d = {}
+    with open(filename) as f:
+        x = shlex(instream=f, posix=True)
+        while 1:
+            k, eq, v = (x.get_token(), x.get_token(), x.get_token())
+            if k == '' or eq != '=':
+                # error, or eof
```
facundobatista, line 42: EOF isn't None in posix mode?

jameinel:
The guidance we have had elsewhere has been to focus on making a very smooth onramp for people writing K8s charms (wave of the future) and it has the significant advantage of simplifying the code.
I'd still expose the flag but change the default.

charmcraft/commands/init.py
chipaca: `os-release` could be garbage?
I'm dropping this entirely to default to `"kubernetes"` fwiw

charmcraft/commands/utils.py
chipaca: ack (dropping it entirely instead)

charmcraft/commands/utils.py
chipaca: yep!

charmcraft/templates/init/metadata.yaml.j2
chipaca: now yes

tests/commands/test_init.py
chipaca: by mistake

tests/commands/test_init.py
```diff
@@ -63,17 +64,35 @@ def test_all_the_files(tmp_path):
...
+    cmd.run(Namespace(path=tmp_path, name='my-charm', author="fred", series=None))
+
+    with (tmp_path / "metadata.yaml").open("rt", encoding="utf8") as f:
+        metadata = yaml.safe_load(f)
+    assert metadata.get("series") == ['kubernetes']
+
+
+def test_manual_overrides_defaults(tmp_path):
+    cmd = InitCommand('group')
+    cmd.run(Namespace(path=tmp_path, name='my-charm', author="fred", series=['xenial', 'precise']))
```
jameinel, line 94: One concern for me is that this doesn't expose how a user actually interacts with the tool. Because I think this means
```
charmcraft init my-charm --author=fred --series xenial --series precise
```

Which feels a bit awkward to me. Does it work if you supply:
```
charmcraft init my-charm --author=fred --series xenial,precise
```

(I have the feeling that is treated as the 'xenial,precise' series.)

Do we care to support multiple series in "init" ? I know we want to allow for multiple series in charms, but I wonder if the syntax is such that it is too clumsy to do nicely.

tests/commands/test_init.py
jameinel: Given path=tmp_path this is saying we are running the test suite *for the initialized charm*, not the test suite for charmcraft.

tests/commands/test_init.py
chipaca: yeah, i think i like the comma-separated more than the current thing. I did the current thing because the simplest and apparently recommended is horrible to me,
```python
parser.add_argument('--series', nargs='*')
```
because the way you use that is
```
charmcraft init --series xenial precise --author fred
```

but, yes, the current one is not really better. Comma-separated it is then.

tests/commands/test_init.py
chipaca: yes, this was because i started from the test that runs the unit tests of the template charm :-)

PR #116, opened 2020-08-13
include install info in README

chipaca:


dstathis:
> @chipaca I don't want to be too nit-picky here, but it might be good to be more explicit. Something like:
> 
> Start with `snap info charmcraft` to see which channels are available. For example, if you want to use the `beta` version, install with `snap install --beta charmcraft`.

I would even go as far as to just put `snap install charmcraft --beta`. I think anyone who wants `--edge` will know how to do it anyway.

chipaca:
The problem is that for most but not all people the command would be `sudo snap install --beta charmcraft`, but i guess if you've logged in or want edge you already know to tweak this, so, yeah.

chipaca:
@dstathis @justinmclark please re-review

Issue #105, opened 2020-08-04
charmcraft-log files in /tmp


jameinel:
I was looking in '/tmp' and I noticed:
```
$ ls
charmcraft-log-2ym5flsu  charmcraft-log-j4fnixya  charmcraft-log-uksp6zlp
charmcraft-log-3_a70vuh  charmcraft-log-j5jis5ub  charmcraft-log-uxyytzhv
charmcraft-log-3didu38k  charmcraft-log-k5eeys7n  charmcraft-log-vt2rd08o
charmcraft-log-4quh1ckg  charmcraft-log-l2vlihqk  charmcraft-log-w6xxbie_
charmcraft-log-50lp0y6z  charmcraft-log-lw12v3in  charmcraft-log-wo3iad9f
```

This is likely left over from the test suite, not from running charmcraft itself (I have 48 of them).
I'm guessing something in our test suite isn't triggering the correct cleanup.

facundobatista:
By design, pytest leaves the latest N temporary directories, in case you want to debug anything there.

That said, there are reports of that mechanism [not working exactly ok](https://github.com/pytest-dev/pytest/issues/543), but it seems it was [fixed recently](https://github.com/pytest-dev/pytest/pull/6944).

I think we're ok with this, we just need to wait future pytest releases. I'd close this issue.

chipaca:
every time I run pytest on charmcraft (with both pytest 5.4.1 and pytest 6.0.1) I get three new `charmcraft-log-<hex>` directories.

chipaca:
These are not pytest logs! they're charmcraft logs. Pytest logs will be `/tmp/pytest-of-john/pytest-NN/`

PR #117, opened 2020-08-13
a couple of fixes for tests

chipaca:
fixes: #105

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -15,8 +15,14 @@
 # For further info, check https://github.com/canonical/charmcraft
 
 import pathlib
+import tempfile
 
 import pytest
+
+
+@pytest.fixture(autouse=True, scope="session")
+def tmpdir_under_tmpdir(tmpdir_factory):
+    tempfile.tempdir = str(tmpdir_factory.getbasetemp())
 
 
 @pytest.fixture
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -15,10 +15,16 @@
...
 
 import pathlib
+import tempfile
 
 import pytest
 
 
+@pytest.fixture(autouse=True, scope="session")
+def tmpdir_under_tmpdir(tmpdir_factory):
+    tempfile.tempdir = str(tmpdir_factory.getbasetemp())
```
justinmclark, line 25: I was trying to understand how this works based on `tmpdir_factory` examples -- can all the tests using this fixture (which should be all of them due to `autouse=True`) access `tempfile.tempdir`?

Looking at docs, it seems as though when `tempfile.tempdir` is not `None`, it is only used as the `dir` default for other `tempfile` functions and it was unclear to me if it is useful outside of the context of the `tempfile` module: [1]

[1] https://docs.python.org/3/library/tempfile.html#tempfile.tempdir

chipaca: this change is exactly so charmcraft's calls to `tempfile.mkstemp` (in `logsetup`) use the test's tempdir instead of the default; this is the change that fixes #105.

PR #118, opened 2020-08-13
Some changes in preparation for big "include everything but jujuignored" branch.

facundobatista:
Included here:

- put the 'hooks' dir as a constant

- introduced a 'relativise' helper, that will be used in the next branch 

- added a verification in the build command line: the entry point must be inside the project

### DIFF
--- a/charmcraft/commands/build.py
+++ b/charmcraft/commands/build.py
@@ -15,6 +15,7 @@
 # For further info, check https://github.com/canonical/charmcraft
 
 
+import itertools
 import logging
 import os
 import pathlib
@@ -58,6 +59,7 @@ JUJU_DISPATCH_PATH="${{JUJU_DISPATCH_PATH:-$0}}" PYTHONPATH=lib:venv ./{entrypoi
 
 # The minimum set of hooks to be provided for compatibility with old Juju
 MANDATORY_HOOK_NAMES = {'install', 'start', 'upgrade-charm'}
+HOOKS_DIR = 'hooks'
 
 
 def polite_exec(cmd):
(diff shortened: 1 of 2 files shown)


charmcraft/commands/build.py
```diff
@@ -79,6 +81,21 @@ def polite_exec(cmd):
     return retcode
 
 
+def relativise(src, dst):
```
justinmclark, line 84: Could we just use `os.path.relpath()` here? Something like `os.path.relpath(dst, src.parent)` should return the same thing.

chipaca: `os.path.relpath` falls flat on its face for most of these:

| `src` | `dst` | `relativise(src, dst)` | `os.path.relpath(dst, src.parent)` |
| --- | --- | -- | -- |
| `/tmp/foo/bar/src.txt` | `/tmp/foo/bar/dst.txt` | `dst.txt` | `dst.txt` |
| `/tmp/foo/bar/baz/src.txt` | `/tmp/foo/dst.txt` | `../../dst.txt` | `dst.txt` |
| `/tmp/foo/src.txt` | `/tmp/foo/bar/baz/dst.txt` | `bar/baz/dst.txt` | `dst.txt` |
| `/tmp/foo/bar/src.txt` | `/tmp/foo/baz/dst.txt` | `../baz/dst.txt` | `dst.txt` |
| `/tmp/foo/bar1/bar2/src.txt` | `/tmp/foo/baz1/baz2/baz3/dst.txt` | `../../baz1/baz2/baz3/dst.txt` | `dst.txt` |

charmcraft/commands/build.py
```diff
@@ -79,6 +81,21 @@ def polite_exec(cmd):
...
+    src_rel_parts = []
+    dst_rel_parts = []
+    for src_part, dst_part in itertools.zip_longest(src.parent.parts, dst.parts):
+        if src_part != dst_part:
+            if src_part is not None:
+                src_rel_parts.append('..')
+            if dst_part is not None:
+                dst_rel_parts.append(dst_part)
+
+    link = pathlib.Path(os.path.join(*(src_rel_parts + dst_rel_parts)))
```
chipaca, line 95: the `os.path.join` call seems spurious to me (all tests pass if I remove it)

also `*(a+b)` is the same as `*a, *b`, but that might be an aesthetic thing

charmcraft/commands/build.py
justinmclark: This is interesting to me. I didn't directly run the test cases, but I locally tested these 4 and they return the same result using `os.path.relpath`

e.g. 
```
>>> src = pathlib.Path("/tmp/foo/bar1/bar2/src.txt")
>>> dst = pathlib.Path("/tmp/foo/baz1/baz2/baz3/dst.txt")
>>> os.path.relpath(dst, src.parent)
'../../baz1/baz2/baz3/dst.txt'
```

chipaca: I don't know what I did to get those results now. I can reproduce what you're doing there and get what you got. Sadly the script I used to create the table was in `/tmp` and I had to reboot :-(
But hey, this means less code \o/

PR #122, opened 2020-08-19
set creds to 0600

chipaca:


charmcraft/commands/store/client.py
```diff
@@ -93,7 +93,9 @@ def _save_credentials_if_changed(self):
             dirpath = os.path.dirname(self._cookiejar_filepath)
             os.makedirs(dirpath, exist_ok=True)
 
-            self._cookiejar.save()
+            fd = os.open(self._cookiejar_filepath, os.O_WRONLY | os.O_CREAT | os.O_TRUNC, 0o600)
+            os.fchmod(fd, 0o600)
```
facundobatista, line 97: is this `fchmod` needed? (as permission flags are already indicated to os.open).

justinmclark: I'm curious about this as well. Is it because `os.open` only sets the permissions of the file where is `fchmod` sets the permissions of the file descriptor?

chipaca: it's because `os.open` will only set the permission bits of the file if it actually created it. Try:
```python
import os

FN = "/tmp/xyzzy"

with open(FN, "w") as f:
    print(oct(os.fstat(f.fileno()).st_mode & 0o777))

fd = os.open(FN, os.O_WRONLY | os.O_CREAT | os.O_TRUNC, 0o600)
# os.fchmod(fd, 0o600)

print(oct(os.stat(FN).st_mode & 0o777))
```

Issue #102, opened 2020-07-30
The build command does not work on bionic


camille-rodriguez:
`charmcraft build` fails when ran on bionic. 

The error traceback:
2020-07-30 20:59:03,438  charmcraft.commands.build      DEBUG    :: Collecting PyYAML (from ops->-r /home/ubuntu/charm-iscsi-connector/requirements.txt (line 2))
2020-07-30 20:59:03,442  charmcraft.commands.build      DEBUG    ::   Cache entry deserialization failed, entry ignored
2020-07-30 20:59:03,527  charmcraft.commands.build      DEBUG    :: Installing collected packages: MarkupSafe, jinja2, PyYAML, ops
2020-07-30 20:59:03,580  charmcraft.commands.build      DEBUG    :: Exception:
2020-07-30 20:59:03,580  charmcraft.commands.build      DEBUG    :: Traceback (most recent call last):
2020-07-30 20:59:03,580  charmcraft.commands.build      DEBUG    ::   File "/usr/lib/python3/dist-packages/pip/basecommand.py", line 215, in main
2020-07-30 20:59:03,581  charmcraft.commands.build      DEBUG    ::     status = self.run(options, args)
2020-07-30 20:59:03,581  charmcraft.commands.build      DEBUG    ::   File "/usr/lib/python3/dist-packages/pip/commands/install.py", line 360, in run
2020-07-30 20:59:03,581  charmcraft.commands.build      DEBUG    ::     prefix=options.prefix_path,
2020-07-30 20:59:03,581  charmcraft.commands.build      DEBUG    ::   File "/usr/lib/python3/dist-packages/pip/req/req_set.py", line 784, in install
2020-07-30 20:59:03,581  charmcraft.commands.build      DEBUG    ::     **kwargs
2020-07-30 20:59:03,581  charmcraft.commands.build      DEBUG    ::   File "/usr/lib/python3/dist-packages/pip/req/req_install.py", line 851, in install
2020-07-30 20:59:03,581  charmcraft.commands.build      DEBUG    ::     self.move_wheel_files(self.source_dir, root=root, prefix=prefix)
2020-07-30 20:59:03,581  charmcraft.commands.build      DEBUG    ::   File "/usr/lib/python3/dist-packages/pip/req/req_install.py", line 1064, in move_wheel_files
2020-07-30 20:59:03,581  charmcraft.commands.build      DEBUG    ::     isolated=self.isolated,
[...]

PR #124, opened 2020-08-19
cope with bionic's slightly broken pip3 command

chipaca:
This makes 'charmcraft build' work even when running using the pip
shipped in bionic, which is a bit "off".

fixes: #102


chipaca:
I just brought up a xenial vm and can confirm:
1. it has the same issue, and
2. this fix fixes the fix it fixes in xenial too.

chipaca:
makes me think maybe i should rename the helper function to something less bionic-y :-)

facundobatista:
+0 to renaming the function, but it's not a show-stopper.

PR #130, opened 2020-08-19
Added an "overview" attribute to each command, with the long help descriptions.

facundobatista:
Also I stopped doing a runtime check of the command integrity, in favor of a test-time verification.

The program output is not tested as it's handled automatically by argparse (it will be properly verified by our tests when we move to the custom-generated helps, in the short future).

@chipaca: feel free to push a commit over this PR for changes in the texts (that will easier for both of us than a bunch of changes).


chipaca (approved):
I approve of the bits I didn't write :-)

PR #131, opened 2020-08-19
Respect the symlink even if it's a directory when building.

facundobatista:


chipaca (approved):
phew

charmcraft/commands/build.py
```diff
@@ -110,6 +110,20 @@ def _load_juju_ignore(self):
                 ignore.extend_patterns(ignores)
         return ignore
 
+    def create_symlink(self, src_path, dest_path):
+        """Create a symlink in dest_path pointing relatively like src_path.
```
balbirthomas, line 114: Nit pick: "pointing relatively like" sounds like a colloquialism. Shouldn't it be "pointing to src_path" ?

balbirthomas (approved):

balbirthomas (approved):

Issue #141, opened 2020-08-27
action created with 'charmcraft init' calls nonexistent method set_result on the action event


chipaca:
It should be set_results, and it should pass in a dict.

PR #142, opened 2020-08-27
fix the action generated by 'charmcraft init'

chipaca:
Whooops.

Fixes: #141

Thanks to @narindergupta for bringing this to our attention.

### DIFF
--- a/charmcraft/templates/init/src/charm.py.j2
+++ b/charmcraft/templates/init/src/charm.py.j2
@@ -31,9 +31,12 @@ class {{ class_name }}(CharmBase):
         if fail:
             event.fail(fail)
         else:
-            event.set_result(
-                "Probability is too important to be left to the experts."
-                " -- Richard Hamming")
+            event.set_results(
+                {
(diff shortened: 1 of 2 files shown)


charmcraft/templates/init/src/charm.py.j2
```diff
@@ -31,9 +31,12 @@ class {{ class_name }}(CharmBase):
         if fail:
             event.fail(fail)
         else:
-            event.set_result(
-                "Probability is too important to be left to the experts."
-                " -- Richard Hamming")
+            event.set_results(
+                {
+                    "fortune": "Probability is too important to be left to the experts."
+                    " -- Richard Hamming"
```
facundobatista, line 37: What an ugly indentation, it looks like there are two entries in the dict. May you indent the second line a little more to match the "value" start?

chipaca: better now?

narindergupta:
Looks good and I tested the same

balbirthomas (approved):

PR #147, opened 2020-09-02
fix PYTHONPATH correctly

dstathis:
The code "env_path,) = env_paths" assumes that env_paths always has exactly one value in it. This breaks if your virtualenv adds more than one path.


facundobatista (approved):
We talked about this, missed the approval! sorry...

justinmclark (approved):
LGTM

Issue #127, opened 2020-08-19
The "release" command must not guess the revision


facundobatista:
Currently the "release" command does not requires a revision, for simplicity as the user just pushed something to the store, and most probably will want to release that revision.

This is dangerous, as there's the race condition of a CI or anything pushing *another* revision to the Store in the middle.

So we need to remove this automatic behaviour, and the revision will need to be always specified (so probably this will migrate from being an option to a mandatory parameter:

```
charmcraft release 17 edge
```

OTOH, if the user want to do this in a simpler way, she could just release at the moment of uploading, and charmcraft will take care automatically, see #126 .

PR #149, opened 2020-09-03
make revision a positional argument to release

dstathis:
fixes: #127

### DIFF
--- a/charmcraft/commands/store/__init__.py
+++ b/charmcraft/commands/store/__init__.py
@@ -346,20 +346,10 @@ class ReleaseCommand(BaseCommand):
                     "be executed in a valid project's directory, or indicate the charm name with "
                     "the --name option.")
 
-        if parsed_args.revision:
-            revision = parsed_args.revision
-        else:
-            # find out which is the latest revision for the charm
-            revisions = store.list_revisions(charm_name)
-            if not revisions:
-                raise CommandError(
-                    "The charm {!r} doesn't have any uploaded revisions.".format(charm_name))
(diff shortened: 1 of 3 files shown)


facundobatista (changes requested):
This is not just a matter of making the parameter mandatory. You also need to remove the code that deals for this arg not being present, and adjust tests.

Also, add a comment to remove the corresponding case in https://docs.google.com/document/d/1Pe-0ED6db53SmrUGIAgVMzxeCOGJQwsZJWf7jzFSagQ/

And also fix `completion.bash` file

charmcraft/commands/store/__init__.py
```diff
@@ -330,13 +330,13 @@ class ReleaseCommand(BaseCommand):
 
     def fill_parser(self, parser):
         """Add own parameters to the general parser."""
+        parser.add_argument(
+            'revision', type=int, help='The revision to release.')
         parser.add_argument(
             'channels', metavar='channel', nargs='+',
+
```
facundobatista, line 337: Please remove this empty line.

facundobatista (changes requested):
Almost done, great! Please check the inline comment, and remember to fix the `completion.bash` file.

charmcraft/commands/store/__init__.py
dstathis: Done.

facundobatista (approved):
Great, thanks! FTR, I already updated the doc after your suggestion there.

chipaca (approved):

PR #150, opened 2020-09-04
when failing to create hard link attempt to use shutil.copy2

matuskosut:
When using e.g. vagrant build fails because using hard links is not allowed.
Could catching the permission exception and attempting to copy the file instead be a way to manage builds in vagrant?
https://github.com/canonical/charmcraft/issues/151

### DIFF
--- a/tests/commands/test_build.py
+++ b/tests/commands/test_build.py
@@ -14,6 +14,7 @@
 #
 # For further info, check https://github.com/canonical/charmcraft
 
+import filecmp
 import logging
 import pathlib
 import os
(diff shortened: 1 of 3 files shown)


facundobatista (changes requested):
Hello @matuskosut , thanks for your work here!

The patch looks solid, but we need a test for this case to let it in.

Thanks again!

chipaca:
@matuskosut thanks for this! are you ok if i pick up writing the test for it, or are you already doing so?

matuskosut:
@chipaca sure go ahead, I haven't got back to it yet. Thanks :)

tests/commands/test_build.py
```diff
@@ -596,6 +595,29 @@ def test_build_generics_tree(tmp_path, caplog):
...
+    if expect_hardlinks:
+        # they're hard links
+        assert (build_dir / 'crazycharm.py').samefile(entrypoint)
+        assert (build_dir / 'file1.txt').samefile(file1)
+        assert (build_dir / 'dir2' / 'file2.txt').samefile(file2)
+    else:
+        # they're *not* hard links
+        assert not (build_dir / 'crazycharm.py').samefile(entrypoint)
+        assert not (build_dir / 'file1.txt').samefile(file1)
+        assert not (build_dir / 'dir2' / 'file2.txt').samefile(file2)
```
facundobatista, line 607: Please check that the files has the same content, permissions, etc... thanks!

facundobatista (changes requested):
If we copy the files, we need to verify they are properly copied.

tests/commands/test_build.py
chipaca: picky, picky...! :-D

facundobatista (approved):
Wonderful, thanks

balbirthomas (approved):

PR #153, opened 2020-09-09
Adjust help messages in all cases to UX indications

facundobatista:
Integrated the rest of the help functionality.

Specifically:

- have a (special) "help" command to provide general, detailed, and per command help

- stop ArgumentParser to automatic handling help texts, as we're doing that now

- hack ArgumentParser to provide better usage and general messages on error

- generate the structures from Argparse internals to build help texts correctly

These "internal hacking" details will be heavily refactored when we go to the "lazy module import and manually support global options" model, in the short future.

Also note that I moved several tests to `test_help.py`, where I implemented them in a more "execution POV", to closer reproduce/test desired behaviour.


charmcraft/main.py
```diff
@@ -58,41 +86,89 @@ def _get_option(args, name):
...
+
+def _get_subparser(parser, namespace):
+    """Return a subparser for that command, if any."""
+    command = getattr(namespace, '_command', None)
+    if command is not None:
+        subactions = [
+            action for action in parser._actions
+            if isinstance(action, argparse._SubParsersAction)]
+        if subactions:
+            (subparseraction,) = subactions
```
chipaca, line 132: this kind of thing has bitten us before, where it explodes with an obscure error. Better to check the length and be explicit about what went wrong if it's not exactly 1, than this.

chipaca (changes requested):
*almost* a +1 :-)

charmcraft/main.py
facundobatista: Added a explicit check for that situation (but didn't add a test for that, as it would imply messing with argparse in a different way that we do now, kind of "breaking" the other part of the code)

chipaca (approved):

justinmclark (approved):
+1

PR #154, opened 2020-09-10
Improved short help messages.

facundobatista:
This includes the command ones and command's parameters ones.

Also fixed some output messages as per UX dept. suggestion.

This is part of the [big adjust help process](https://github.com/canonical/charmcraft/pull/153).

PR #155, opened 2020-09-10
Produce the default help text.

facundobatista:
For this I had to add a 'common' attribute to commands, to flag those are of "common use" (the "common" term was UX used in the docs).

Not integrated yet, this is part of the [big adjust help process](https://github.com/canonical/charmcraft/pull/153).

This PR includes [this previous one](https://github.com/canonical/charmcraft/pull/154).

### DIFF
--- a/charmcraft/cmdbase.py
+++ b/charmcraft/cmdbase.py
@@ -34,6 +34,7 @@ class BaseCommand:
 
     - name: the identifier in the command line
     - help_msg: a one line help for user documentation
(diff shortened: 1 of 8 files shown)


chipaca (approved):
This seems reasonable to me. If it got any more complicated I'd eschew the common flag and build the help summary by hand, but as long as it's this complicated and not (significantly) more, it's ok :-)

balbirthomas (approved):

PR #156, opened 2020-09-14
Text builders for simple usage, detailed help, and command help.

facundobatista:
Not integrated yet, this is part of the [big adjust help process](https://github.com/canonical/charmcraft/pull/153).

### DIFF
--- a/charmcraft/helptexts.py
+++ b/charmcraft/helptexts.py
@@ -32,6 +32,7 @@ GENERAL_SUMMARY = """
 """
 # XXX Facundo 2020-09-10: we should add an extra (separated) line to the summary with:
 #   See <url> for additional documentation.
+# Related issue: https://github.com/canonical/charmcraft/issues/161
 
 # generic intro and outro texts
 HEADER = """
@@ -48,7 +49,11 @@ Error: {error_message}
 
 
 def get_usage_message(fullcommand, error_message):
-    """Build a usage and error message."""
+    """Build a usage and error message.
+
+    The fullcommand is the command used by the user (`charmcraft`, `charmcraft build`, etc),
+    and the error message is the specific problem in the given parameters.
+    """
     return USAGE.format(fullcommand=fullcommand, error_message=error_message)
 
 
@@ -75,7 +80,13 @@ def _build_item(title, text, title_space):
 def get_full_help(command_groups, global_options):
     """Produce the text for the default help.
 
-    It has the following structure:
+    - command_groups: list of grouped commands, as it's defined in the main
+      module
+
+    - global_options: options defined at charmcraft level (not in the commands),
+      with the (options, description) structure
+
+    The help text has the following structure:
 
     - usage
     - summary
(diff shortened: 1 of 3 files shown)


charmcraft/helptexts.py
```diff
@@ -33,6 +33,24 @@
...
+Try '{fullcommand} -h' for help.
+
+Error: {error_message}
+"""
+
+
+def get_usage_message(fullcommand, error_message):
+    """Build a usage and error message."""
+    return USAGE.format(fullcommand=fullcommand, error_message=error_message)
+
```
dstathis, line 53: It seems a bit weird to me to have usage only be usable for errors. I would do something like:
```suggestion
USAGE = """\
￼Usage: charmcraft [OPTIONS] COMMAND [ARGS]...
￼Try '{fullcommand} -h' for help.
￼"""

ERROR_MSG = """\
{usage}

Error: {error_message}
"""
￼
def get_usage_message(fullcommand):
    """Build a usage message"""
    return USAGE.format(fullcommand=fullcommand)￼

￼def get_error_message(fullcommand, error_message):
￼    """Build an error message."""
￼    return ERROR_MSG.format(usage=get_usage_message(fullcommand=fullcommand), error_message=error_message)
```

dstathis, line 53: It seems a bit weird to me to have usage only be usable for errors. I would do something like:
```suggestion
USAGE = """\
￼Usage: charmcraft [OPTIONS] COMMAND [ARGS]...
￼Try '{fullcommand} -h' for help.
￼"""

ERROR_MSG = """\
{usage}

Error: {error_message}
"""
￼
def get_usage_message(fullcommand):
    """Build a usage message"""
    return USAGE.format(fullcommand=fullcommand)￼

￼def get_error_message(fullcommand, error_message):
￼    """Build an error message."""
￼    return ERROR_MSG.format(usage=get_usage_message(fullcommand=fullcommand), error_message=error_message)
```

facundobatista: This is a good idea, if we would have separate usage for those. Otherwise it's convoluted code with the same result.

So, if in the future we have a use for this, we can separate it, but for now I'll keep it simple.

dstathis (approved):

charmcraft/helptexts.py
```diff
@@ -118,3 +133,116 @@ def get_full_help(command_groups, global_options):
...
+    """Produce the text for each command's help.
+
+    It has the following structure:
+
+    - usage
+    - summary
+    - options
+    - other related commands
+    - footer
+    """
```
balbirthomas, line 200: May be worth putting a line for each function parameter, describing its structure. For example "command_group is a list of tuples ..."

charmcraft/helptexts.py
```diff
@@ -118,3 +133,116 @@ def get_full_help(command_groups, global_options):
...
+def get_detailed_help(command_groups, global_options):
+    """Produce the text for the detailed help.
+
+    It has the following structure:
+
+    - usage
+    - summary
+    - global options
+    - all commands shown with description, grouped
+    - more help
```
balbirthomas, line 147: Same comment as below. It would be nice to have the expected type/structure of function parameters described, since it makes the code easier to understand with them in mind, even though the structure can be inferred from the code and tests.

balbirthomas (approved):

charmcraft/helptexts.py
facundobatista: Thanks, I improved the docstring with the suggestion.

charmcraft/helptexts.py
facundobatista: Thanks, I improved the docstring with the suggestion.

PR #162, opened 2020-09-21
use pytest.approx to compare times

chipaca:


facundobatista (approved):
Thanks!

tests/commands/test_build.py
```diff
@@ -612,9 +612,8 @@ def _test_build_generics_tree(tmp_path, caplog, *, expect_hardlinks):
             assert filecmp.cmp(str(p1), str(p2), shallow=False)
             assert p1.stat().st_mode == p2.stat().st_mode
             assert p1.stat().st_size == p2.stat().st_size
-            # checking st_atime seems to be a bit racy
-            # assert p1.stat().st_atime == p2.stat().st_atime
-            assert p1.stat().st_mtime == p2.stat().st_mtime
+            assert pytest.approx(p1.stat().st_atime, p2.stat().st_atime)
+            assert pytest.approx(p1.stat().st_mtime, p2.stat().st_mtime)
```
dstathis, line 616: I think the syntax here is not correct and will always pass. I believe you would need:

`p1.stat().st_mtime == pytest.approx(p2.stat().st_mtime)`

dstathis (changes requested):

tests/commands/test_build.py
facundobatista: You're right, Dylan!

dstathis (approved):

chipaca:
@dstathis welcome back! thanks for the review

PR #163, opened 2020-09-21
Two small details before integrating help messages to main.

facundobatista:
Specifically:

- support "argsparsing" option in CommandError to just present the message to the user (this will be refactored heavily in the short future when we change how all messages are sent to the user, but we need it now to just show the help)

- Added a small helper to check for options in the arguments (which will be more used in the next branch).

### DIFF
--- a/charmcraft/main.py
+++ b/charmcraft/main.py
@@ -46,6 +46,15 @@ COMMAND_GROUPS = [
 ]
 
 
+def _get_option(args, name):
+    """Get a general option, merging with OR both global and command versions.
+
+    This is done manually because otherwise Argparse would overwrite the
+    command ont into thee global one.
+    """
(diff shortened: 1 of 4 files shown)


charmcraft/main.py
```diff
@@ -46,6 +46,15 @@
 ]
 
 
+def _get_option(args, name):
+    """Get a general option, merging with OR both global and command versions.
```
balbirthomas, line 50: "merging with OR" seems to be redundant. I suppose it meant "Get the global or command option, including both if present'.

balbirthomas (approved):

charmcraft/main.py
facundobatista: Improved the docstring, thanks!

dstathis (approved):

PR #166, opened 2020-09-23
Don't ignore important files (like version) by default.

facundobatista:


jameinel (commented):
The actual Juju ignore list is here:
https://github.com/juju/charm/blob/d7e8e94ba938a277283201e693f699a9d09a1f72/charmdir.go#L30

Which *does* include version and revision as ignored, because those are filled out by Juju during deploy.
If we would rather have some other process populating them, and then charmcraft doesn't fill them out itself, we can go with pulling them from source. However, 'external process' tends to mean they get stale. (eg, if you have to 'make version' for the file to be up to date, then changing a source file and publishing without doing 'make version' will have the wrong content, and give confusing information when trying to debug what you are actually running.)

facundobatista:
**facubatista:** jam, but charmcraft should fill it itself in the case it's missing, otherwise we're overruling the author, and that's probably always a bad idea
**facubatista:** if it's there, why not respect it?
**jam**: facubatista, so 'revision' can also be there, but is almost always wrong/stale. Same for 'version'. If you unpack a .zip of a charm, you would expect it to be there, but if you then reuploaded that, you would not want to take the stale one IMO
**facubatista:** jam, revision and version are handled diferently; the revision is always indicated by the store, it's an integer, always growing; the version can be any string that the user wants, to represent the proper version of the charm
**facubatista:** the revision will be incremented on each binary uploaded to the store, the version is author responsibility
**facubatista:** we're ignoring a posible "revision" file too (don't know why), but in any case the store will not use that for anything; OTOH the store parses and keeps what's in "version"
**jam**: I'm not opposed to having it, but I'm much more in favor of populating it by the build command.
**facubatista:** jam, so you say that if the author indicates that version should be X, we go ahead and push to the store something that will end up in version Y? that's super surprising
**jam**: facubatista, you can warn if you want, but I trust a static file on disk much less than the build process
**facubatista:** for me it's matter of trusting the author, not the file; if the author would want automatic versioning, she can always remove the file and let charmcraft take charge

dstathis (approved):

chipaca (approved):

PR #173, opened 2020-10-16
Load only the indicated command (if any), not everything in advance.

facundobatista:
For this I pre-parsed the sysargs, which moved ourselves to a MUCH SIMPLER use of ArgumentParser, as there are no more "commands" involved at that level.

Of course we're still using ArgumentParser, but only for each command.

Everything got MUCH MORE SIMPLE now, as there is quite less fiddling with ArgumentParser internals.

Sorry fo the length of the branch, but I improved several tests, added corner cases to others, and created a couple of new ones.

### DIFF
--- a/charmcraft/main.py
+++ b/charmcraft/main.py
@@ -208,7 +208,6 @@ class Dispatcher:
 
         # control and use quiet/verbose options
         if flags['quiet'] and flags['verbose']:
-            print("Error, mutex!")
             raise CommandError("The 'verbose' and 'quiet' options are mutually exclusive.")
         if flags['quiet']:
             message_handler.set_mode(message_handler.QUIET)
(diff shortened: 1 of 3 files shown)


chipaca (approved):
noice

charmcraft/main.py
```diff
@@ -181,91 +151,96 @@ class Dispatcher:
...
+                filtered_sysargs.append(arg)
 
-    def _handle_global_params(self):
-        """Set up and process global parameters."""
-        if _get_option(self.parsed_args, 'verbose'):
-            message_handler.set_mode(message_handler.VERBOSE)
-        if _get_option(self.parsed_args, 'quiet'):
+        # control and use quiet/verbose options
+        if flags['quiet'] and flags['verbose']:
+            print("Error, mutex!")
```
justinmclark, line 211: Should this print still be here?

justinmclark (approved):
+1

Just one comment/question

PR #175, opened 2020-10-22
Tweek template

chipaca:
ops 0.8 is kinda already out so update the template to reflect that.

Also, why was charmcraft pinning it to 0.7? :-/


facundobatista (approved):
Thanks

justinmclark (approved):

PR #177, opened 2020-10-31
Make use of the self.config alias

erik78se:
This PR changes the template for new charms to make use of the more user-friendly "self.config" alias discussed here:

https://discourse.juju.is/t/first-steps-with-the-operator-framework/3006/20?u=erik-lonroth

It makes the charming experience somewhat more casual-programmer-friendly


facundobatista (approved):
Great, thanks!

chipaca (approved):
thank you!

PR #179, opened 2020-11-06
Added function to load libraries information.

facundobatista:
This function is not used yet (but adding first code that uses it makes the branch too large).

This is part of [all libraries related functionality branch](https://github.com/facundobatista/charmcraft/tree/charmlibs-draft).

### DIFF
--- a/charmcraft/commands/store/__init__.py
+++ b/charmcraft/commands/store/__init__.py
@@ -469,34 +469,52 @@ class StatusCommand(BaseCommand):
             logger.info(line)
 
 
+class _BadLibraryPathError(CommandError):
+    """Subclass to provide a specific error for a bad library path."""
+    def __init__(self, path):
+        super().__init__(
+            "Library path {} must conform to the lib/charms/<charm>/v<API>/<libname>.py "
+            "structure.".format(path))
+
+
+class _BadLibraryNameError(CommandError):
+    """Subclass to provide a specific error for a bad library name."""
+    def __init__(self, name):
+        super().__init__(
+            "Library full name {!r} must conform to the charms.<charm>.v<API>.<libname> structure."
+            .format(name))
+
+
+def _get_positive_int(raw_value):
+    """Convert the raw value for api/patch into a positive integer."""
+    value = int(raw_value)
+    if value < 0:
+        raise ValueError('negative')
+    return value
+
+
 def _get_lib_info(*, full_name=None, lib_path=None):
     """Get the whole lib info from the path/file."""
     if full_name is None:
         # get it from the lib_path
-        bad_structure_msg = (
-            "Library path {} must conform to the lib/charms/<charm>/v<API>/<libname>.py "
-            "structure.".format(lib_path))
         try:
             libsdir, charmsdir, charm_name, v_api = lib_path.parts[:-1]
         except ValueError:
-            raise CommandError(bad_structure_msg)
+            raise _BadLibraryPathError(lib_path)
         if libsdir != 'lib' or charmsdir != 'charms' or lib_path.suffix != '.py':
-            raise CommandError(bad_structure_msg)
+            raise _BadLibraryPathError(lib_path)
         full_name = '.'.join((charmsdir, charm_name, v_api, lib_path.stem))
 
     else:
         # build the path! convert a lib name with dots to the full path, including lib
         # dir and Python extension.
         #    e.g.: charms.mycharm.v4.foo -> lib/charms/mycharm/v4/foo.py
-        bad_structure_msg = (
-            "Library full name {!r} must conform to the charms.<charm>.v<API>.<libname> structure."
-            .format(full_name))
         try:
             charmsdir, charm_name, v_api, libfile = full_name.split('.')
         except ValueError:
-            raise CommandError(bad_structure_msg)
+            raise _BadLibraryNameError(full_name)
         if charmsdir != 'charms':
-            raise CommandError(bad_structure_msg)
+            raise _BadLibraryNameError(full_name)
         lib_path = pathlib.Path('lib') / charmsdir / charm_name / v_api / (libfile + '.py')
 
     if v_api[0] != 'v' or not v_api[1:].isdigit():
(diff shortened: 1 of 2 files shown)


charmcraft/commands/store/__init__.py
```diff
@@ -461,3 +467,109 @@ def run(self, parsed_args):
         table = tabulate(data, headers=headers, tablefmt='plain', numalign='left')
         for line in table.splitlines():
             logger.info(line)
+
+
+def _get_lib_info(*, full_name=None, lib_path=None):
+    """Get the whole lib info from the path/file."""
+    if full_name is None:
+        # get it from the lib_path
+        bad_structure_msg = (
```
chipaca, line 476: wouldn't it be better to have a couple of `BadLib{Path,Name}(CommandError)` classes, rather than building these strings on the fast path?

charmcraft/commands/store/__init__.py
```diff
@@ -461,3 +467,109 @@ def run(self, parsed_args):
...
+    with lib_path.open('rb') as fh:
+        for line in fh:
+            if line.startswith(metadata_fields):
+                try:
+                    field, value = [x.strip() for x in line.split(b'=')]
+                except ValueError:
+                    raise CommandError("Bad metadata line in {}: {!r}".format(lib_path, line))
+                metadata[field] = value
+            else:
+                hasher.update(line)
```
chipaca, line 526: why only hash the non-metadata lines?

charmcraft/commands/store/__init__.py
```diff
@@ -461,3 +467,109 @@ def run(self, parsed_args):
...
+
+    missing = [k.decode('ascii') for k, v in metadata.items() if v is None]
+    if missing:
+        raise CommandError(
+            "Library {} is missing the mandatory metadata fields: {}."
+            .format(lib_path, ', '.join(sorted(missing))))
+
+    def _get_positive_int(key):
+        """Convert the raw value for api/patch into a positive integer."""
+        value = metadata[key].decode('ascii')
```
chipaca, line 536: why the decode?
and why define this function on every call to _get_lib_info?

charmcraft/commands/store/__init__.py
```diff
@@ -461,3 +467,109 @@ def run(self, parsed_args):
...
+    except Exception:
+        raise CommandError(bad_api_patch_msg.format(lib_path, 'LIBAPI'))
+    try:
+        libpatch = _get_positive_int(b'LIBPATCH')
+    except Exception:
+        raise CommandError(bad_api_patch_msg.format(lib_path, 'LIBPATCH'))
+
+    if libapi == 0 and libpatch == 0:
+        raise CommandError(
+            "Library {} metadata fields LIBAPI and LIBPATCH can not be both zero."
```
chipaca, line 554:
```suggestion
            "Library {} metadata fields LIBAPI and LIBPATCH cannot both be zero."
```

charmcraft/commands/store/__init__.py
```diff
@@ -461,3 +467,109 @@ def run(self, parsed_args):
...
+            .format(lib_path))
+
+    if libapi != api_from_path:
+        raise CommandError(
+            "Library {} metadata field LIBAPI is different than the version in the path."
```
chipaca, line 559:
```suggestion
            "Library {} metadata field LIBAPI is different from the version in the path."
```

charmcraft/commands/store/__init__.py
facundobatista: Great idea! Did that now.

charmcraft/commands/store/__init__.py
facundobatista: Because we need to know if the file *real content* changed or not, with *real content* meaning the file without it's metadata. This way we can support for cases like for example the user updating the patch version but not touching the rest of the file (it's a duplicate, and we should not have a new revision for it).

charmcraft/commands/store/__init__.py
facundobatista: Right, no need to decode!

And moved the function outside.

charmcraft/commands/store/__init__.py
facundobatista: Improved.

charmcraft/commands/store/__init__.py
facundobatista: Oh, you british! Improved.

chipaca (approved):

tests/commands/test_store_commands.py
```diff
@@ -939,3 +941,257 @@ def test_status_with_multiple_branches(caplog, store_mock):
...
+    if metadata_api is None:
+        metadata_api = "LIBAPI = 3"
+    if metadata_patch is None:
+        metadata_patch = "LIBPATCH = 14"
+
+    fields = [metadata_id, metadata_api, metadata_patch]
+    with lib_file.open('wt', encoding='utf8') as fh:
+        for f in fields:
+            if f:
+                fh.write(f + '\n')
```
sergiusens, line 972: side comment: interesting that @chipaca did not encourage `print(f, file=fh)` :-)

sergiusens (commented):
Nice work. Some comments.

This does trigger some thoughts about how to converge the tooling in the future.

charmcraft/commands/store/__init__.py
```diff
@@ -461,3 +467,119 @@ def run(self, parsed_args):
...
+        #    e.g.: charms.mycharm.v4.foo -> lib/charms/mycharm/v4/foo.py
+        try:
+            charmsdir, charm_name, v_api, libfile = full_name.split('.')
+        except ValueError:
+            raise _BadLibraryNameError(full_name)
+        if charmsdir != 'charms':
+            raise _BadLibraryNameError(full_name)
+        lib_path = pathlib.Path('lib') / charmsdir / charm_name / v_api / (libfile + '.py')
+
+    if v_api[0] != 'v' or not v_api[1:].isdigit():
```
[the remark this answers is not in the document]
facundobatista: Exactly! 

That said, I think I'm missing your point here. A `v-1` value is invalid, and there's already a test case for it.

charmcraft/commands/store/__init__.py
```diff
@@ -461,3 +467,119 @@ def run(self, parsed_args):
...
+    metadata_fields = (b'LIBAPI', b'LIBPATCH', b'LIBID')
+    metadata = dict.fromkeys(metadata_fields)
+    hasher = hashlib.sha256()
+    with lib_path.open('rb') as fh:
+        for line in fh:
+            if line.startswith(metadata_fields):
+                try:
+                    field, value = [x.strip() for x in line.split(b'=')]
+                except ValueError:
+                    raise CommandError("Bad metadata line in {}: {!r}".format(lib_path, line))
```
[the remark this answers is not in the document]
facundobatista: There's no simple way to express valid metadata fields in less than a couple of paragraphs... what about pointing to some documentation?

tests/commands/test_store_commands.py
```diff
@@ -939,3 +941,257 @@ def test_status_with_multiple_branches(caplog, store_mock):
         "         beta/branch-2  15.0.0     15          2020-07-03T20:30:40+00:00",
     ]
     assert expected == [rec.message for rec in caplog.records]
+
+
+# -- tests for _get_lib_info helper
+
+def _create_lib(extra_content=None, metadata_id=None, metadata_api=None, metadata_patch=None):
+    """Helper to create the structures in disk for a given lib.
```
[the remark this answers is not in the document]
facundobatista: Fixed, thanks!

tests/commands/test_store_commands.py
```diff
@@ -939,3 +941,257 @@ def test_status_with_multiple_branches(caplog, store_mock):
...
+
+    WARNING: this function has the capability of creating WRONG structures in disk.
```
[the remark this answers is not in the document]
facundobatista: Fixed, thanks!

tests/commands/test_store_commands.py
```diff
@@ -939,3 +941,257 @@ def test_status_with_multiple_branches(caplog, store_mock):
...
+
+    This is specific for the _get_lib_info tests below, other tests should use the safe
+    similar one from factory.
```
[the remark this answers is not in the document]
facundobatista: Improved, thanks.

tests/commands/test_store_commands.py
facundobatista: I don't like to use `print` to write into files :)

charmcraft/commands/store/__init__.py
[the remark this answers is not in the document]
sergiusens: my bad!

charmcraft/commands/store/__init__.py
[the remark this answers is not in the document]
sergiusens: Linking to docs is what we do in Snapcraft, that could work here. We also do multiline errors: What went wrong, why it went wrong, how to fix it, further documentation from past design recommendations

tests/commands/test_store_commands.py
sergiusens: but everything is a file :sweat_smile:

sergiusens (approved):
errors I guess need general consensus on how to manage, so we can ignore that for now.

PR #180, opened 2020-11-11
Support to create a library, first step in all the lifecycle.

facundobatista:
This is the second cut of the big [charmlibs draft](https://github.com/facundobatista/charmcraft/tree/charmlibs-draft).

### DIFF
--- a/charmcraft/commands/store/__init__.py
+++ b/charmcraft/commands/store/__init__.py
@@ -21,6 +21,7 @@ import hashlib
 import logging
 import os
 import pathlib
+import string
 import textwrap
 from collections import namedtuple
 from operator import attrgetter
@@ -36,6 +37,30 @@ logger = logging.getLogger('charmcraft.commands.store')
 
 LibData = namedtuple(
     'LibData', 'lib_id api patch content content_hash full_name path lib_name charm_name')
+
+LIBRARY_TEMPLATE = """
+\"""TEMPLATE FIXME: Add a proper docstring here.
+
+This is the main documentation of the library, will be exposed by Charmhub after
+the lib is published.
+
+Markdown is supported.
+\"""
+
+# Never change this field, it's the unique identifier to track the library in
+# all systems
+LIBID = "{lib_id}"
+
+# Update this API version when introducing backwards incompatible
+# changes in the library.
+LIBAPI = 0
+
+# Update this version for every change in the library before (re)publishing it
+# (except for the initial content).
+LIBPATCH = 1
+
+# TEMPLATE FIXME: add your code here! Happy coding!
+"""
 
 
 def get_name_from_metadata():
@@ -583,3 +608,54 @@ def _get_lib_info(*, full_name=None, lib_path=None):
     return LibData(
         lib_id=libid, api=libapi, patch=libpatch, content_hash=content_hash, content=content,
         full_name=full_name, path=lib_path, lib_name=lib_name, charm_name=charm_name)
+
+
+class CreateLibCommand(BaseCommand):
+    """Create a charm library."""
+    name = 'create-lib'
+    help_msg = "Create a charm library."
+    overview = textwrap.dedent("""
+        Create a charm library.
+
+        It will request a unique ID from Charmhub and bootstrap a
+        template file in the proper local directory.
+
+        It will automatically take you through the login process if
+        your credentials are missing or too old.
+    """)
+
+    def fill_parser(self, parser):
+        """Add own parameters to the general parser."""
+        parser.add_argument(
+            'lib_name', metavar='lib-name',
+            help="The name of the library file (e.g. 'db').")
+
+    def run(self, parsed_args):
+        """Run the command."""
+        lib_name = parsed_args.lib_name
+        valid_chars = set(string.ascii_lowercase + string.digits + '_')
+        if set(lib_name) - valid_chars or lib_name[0] not in string.ascii_lowercase:
+            raise CommandError(
+                "Invalid library name (can be only lowercase alphanumeric "
+                "characters and underscore, starting with alpha).")
+
+        charm_name = get_name_from_metadata()
+        if charm_name is None:
+            raise CommandError(
+                "Can't access name in 'metadata.yaml' file. The 'create-lib' command needs to "
+                "be executed in a valid project's directory.")
+
+        full_name = 'charms.{}.v0.{}'.format(charm_name, lib_name)
+        lib_data = _get_lib_info(full_name=full_name)
+        lib_path = lib_data.path
+        if lib_path.exists():
+            raise CommandError('The indicated library already exists on {}'.format(lib_path))
+
+        store = Store()
+        lib_id = store.create_library_id(charm_name, lib_name)
+
+        lib_path.parent.mkdir(parents=True, exist_ok=True)
+        lib_path.write_text(LIBRARY_TEMPLATE.format(lib_id=lib_id))
+
+        logger.info("Library %s created with id %s.", full_name, lib_id)
+        logger.info("Make sure to add the library file to your project: %s", lib_path)
--- a/charmcraft/commands/store/__init__.py
+++ b/charmcraft/commands/store/__init__.py
@@ -620,14 +620,15 @@ class CreateLibCommand(BaseCommand):
         charm_name = get_name_from_metadata()
         if charm_name is None:
             raise CommandError(
-                "Can't access name in 'metadata.yaml' file. The 'create-lib' command needs to "
+                "Cannot access name in 'metadata.yaml' file. The 'create-lib' command needs to "
                 "be executed in a valid project's directory.")
 
+        # all libraries born with API version in 0
         full_name = 'charms.{}.v0.{}'.format(charm_name, lib_name)
         lib_data = _get_lib_info(full_name=full_name)
         lib_path = lib_data.path
         if lib_path.exists():
-            raise CommandError('The indicated library already exists on {}'.format(lib_path))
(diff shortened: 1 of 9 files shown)


charmcraft/commands/store/__init__.py
```diff
@@ -583,3 +608,54 @@ def _get_lib_info(*, full_name=None, lib_path=None):
...
+            raise CommandError('The indicated library already exists on {}'.format(lib_path))
+
+        store = Store()
+        lib_id = store.create_library_id(charm_name, lib_name)
+
+        lib_path.parent.mkdir(parents=True, exist_ok=True)
+        lib_path.write_text(LIBRARY_TEMPLATE.format(lib_id=lib_id))
+
+        logger.info("Library %s created with id %s.", full_name, lib_id)
+        logger.info("Make sure to add the library file to your project: %s", lib_path)
```
chipaca, line 661: How about
```suggestion
        logger.info("Make sure to add the library file to your project: for example, 'git add %s'", lib_path)
```

chipaca (approved):
+1 but see notes

charmcraft/commands/store/__init__.py
```diff
@@ -583,3 +608,54 @@ def _get_lib_info(*, full_name=None, lib_path=None):
...
+    name = 'create-lib'
+    help_msg = "Create a charm library."
+    overview = textwrap.dedent("""
+        Create a charm library.
+
+        It will request a unique ID from Charmhub and bootstrap a
+        template file in the proper local directory.
+
+        It will automatically take you through the login process if
+        your credentials are missing or too old.
```
sergiusens, line 624:
```suggestion
        You will be interactively queried for your login credentials if they
        are missing or if they have expired.
```

charmcraft/commands/store/__init__.py
```diff
@@ -583,3 +608,54 @@ def _get_lib_info(*, full_name=None, lib_path=None):
...
+        valid_chars = set(string.ascii_lowercase + string.digits + '_')
+        if set(lib_name) - valid_chars or lib_name[0] not in string.ascii_lowercase:
+            raise CommandError(
+                "Invalid library name (can be only lowercase alphanumeric "
+                "characters and underscore, starting with alpha).")
+
+        charm_name = get_name_from_metadata()
+        if charm_name is None:
+            raise CommandError(
+                "Can't access name in 'metadata.yaml' file. The 'create-lib' command needs to "
```
sergiusens, line 645:
```suggestion
                "Cannot access name in 'metadata.yaml' file. The 'create-lib' command needs to "
```

charmcraft/commands/store/__init__.py
```diff
@@ -583,3 +608,54 @@ def _get_lib_info(*, full_name=None, lib_path=None):
...
+                "be executed in a valid project's directory.")
+
+        full_name = 'charms.{}.v0.{}'.format(charm_name, lib_name)
```
sergiusens, line 648: Can you add a code comment for v0?

charmcraft/commands/store/__init__.py
```diff
@@ -583,3 +608,54 @@ def _get_lib_info(*, full_name=None, lib_path=None):
...
+        lib_data = _get_lib_info(full_name=full_name)
+        lib_path = lib_data.path
+        if lib_path.exists():
+            raise CommandError('The indicated library already exists on {}'.format(lib_path))
```
sergiusens, line 652:
```suggestion
            raise CommandError('This library already exists: {}'.format(lib_path))
```

charmcraft/commands/store/__init__.py
sergiusens: Is the library already part of the project by virtue of its creation? If so, instead of adding to the project, should that word not be replaced with "version control"?

tests/factory.py
```diff
@@ -36,3 +40,27 @@ def run(self, parsed_args):
             pass
 
     return MyCommand
+
+
+def create_lib_filepath(charm_name, lib_name, api=0, patch=1, lib_id='test-lib-id'):
+    """Helper to create the structures in disk for a given lib."""
```
sergiusens, line 46:
```suggestion
    """Helper to create the structures on disk for a given lib."""
```

charmcraft/commands/store/__init__.py
```diff
@@ -37,6 +38,30 @@
...
+
+Markdown is supported.
+\"""
+
+# Never change this field, it's the unique identifier to track the library in
+# all systems
+LIBID = "{lib_id}"
+
+# Update this API version when introducing backwards incompatible
+# changes in the library.
```
sergiusens, line 55: maybe, just maybe mention that this should be incremented by 1 when introducing incompatible changes?

sergiusens (approved):
Looks good, I just have questions and cosmetics

charmcraft/commands/store/__init__.py
```diff
@@ -37,6 +38,30 @@
 LibData = namedtuple(
     'LibData', 'lib_id api patch content content_hash full_name path lib_name charm_name')
 
+LIBRARY_TEMPLATE = """
+\"""TEMPLATE FIXME: Add a proper docstring here.
```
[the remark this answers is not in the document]
facundobatista: Improved the FIXME.

And used the jinja template for this, it was a great idea!!!

charmcraft/commands/store/__init__.py
```diff
@@ -583,3 +608,54 @@ def _get_lib_info(*, full_name=None, lib_path=None):
...
+        template file in the proper local directory.
+
+        It will automatically take you through the login process if
+        your credentials are missing or too old.
+    """)
+
+    def fill_parser(self, parser):
+        """Add own parameters to the general parser."""
+        parser.add_argument(
+            'lib_name', metavar='lib-name',
```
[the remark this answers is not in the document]
facundobatista: This is because the `list-lib` command also needs a name, but the charm name! So I used `--charm-name` there, and `--lib-name` here. But if you think it's enough difference to use `--charm-name` there, and when just using `--name` is obviously the library name, I'm fine to change it.

charmcraft/commands/store/__init__.py
chipaca: FWIW also as written it sounds like you need to increase PATCH every time, even when you just reset API, and I don't think that's the case

charmcraft/commands/store/__init__.py
```diff
@@ -583,3 +608,54 @@ def _get_lib_info(*, full_name=None, lib_path=None):
...
+            help="The name of the library file (e.g. 'db').")
+
+    def run(self, parsed_args):
+        """Run the command."""
+        lib_name = parsed_args.lib_name
+        valid_chars = set(string.ascii_lowercase + string.digits + '_')
+        if set(lib_name) - valid_chars or lib_name[0] not in string.ascii_lowercase:
+            raise CommandError(
+                "Invalid library name (can be only lowercase alphanumeric "
+                "characters and underscore, starting with alpha).")
```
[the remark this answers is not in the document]
facundobatista: I added a check for the `""` name, which is obviously incorrect, but not for a "too long" situation. However, I started handling (with its test!) a more generic problem of not being able to write the library (because of name too long, or permissions, disk full, or whatever may happen there).

charmcraft/commands/store/__init__.py
facundobatista: Nice, improved.

charmcraft/commands/store/__init__.py
facundobatista: I think this improvement is nice, but won't apply it here for consistency (this message appears in a lot of commands, so we would better change it in all at the same time). If you think that's worthwhile, please open an issue for it, thanks!

charmcraft/commands/store/__init__.py
facundobatista: Improved.

charmcraft/commands/store/__init__.py
facundobatista: Added!

charmcraft/commands/store/__init__.py
facundobatista: Nice, improved.

tests/factory.py
facundobatista: Improved!

charmcraft/commands/store/__init__.py
facundobatista: Nice, I changed both "update" for "increment by 1" (not exactly, better rephrasing).

charmcraft/commands/store/__init__.py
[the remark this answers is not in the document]
facundobatista: *`<Chipaca> for create-lib's argument, 'name' makes the most sense, as what else would it be`*

Changing it.

charmcraft/commands/store/__init__.py
facundobatista: And improved the explanation to show what should happen with PATCH when changing API.

charmcraft/templates/charmlibs/new_library.py.j2
```diff
@@ -0,0 +1,23 @@
...
+# changes in the library.
+LIBAPI = 0
+
+# Increment this PATCH version by 1 for every change in the library before 
+# (re)publishing it (except for the initial content), or reset it to 0 if you
+# are changing the API version.
+LIBPATCH = 1
+
+# FIXME: add your code here! Happy coding!
+"""
```
chipaca, line 23: that's a syntax error (EOF while scanning triple-quoted string literal)

PR #182, opened 2020-11-13
Added the pack command, initially to work only with bundles.

facundobatista:


charmcraft/commands/pack.py
```diff
@@ -0,0 +1,130 @@
...
+                "Missing or invalid main bundle file: '{}'.".format(bundle_filepath))
+        bundle_name = bundle_config.get('name')
+        if not bundle_name:
+            raise CommandError(
+                "Invalid bundle config; missing a 'name' field indicating the bundle's name in "
+                "file '{}'.".format(bundle_filepath))
+
+        charmcraft_filepath = dirpath / 'charmcraft.yaml'
+        charmcraft_config = load_yaml(charmcraft_filepath)
+        if charmcraft_config is None:
```
cmatsuoka, line 118: At some point we'll probably want to validate the schema of the config files.

cmatsuoka (approved):
Thanks, this lgtm assuming that the implemented parsing matches the specs. Like in the previous PR it would be interesting to add a description to provide context to the changes so the reviewer can see if the code is actually doing what it's supposed to do.

charmcraft/commands/pack.py
```diff
@@ -0,0 +1,130 @@
...
+logger = logging.getLogger(__name__)
+
+# the minimum set of files in a bundle
+MANDATORY_FILES = {'bundle.yaml'}
+
+
+def build_zip(zippath, basedir, fpaths):
+    """Build the final file.
+
+    Note we convert all paths to str to support Python 3.5.
```
cjp256, line 34: If shipping `charmcraft` a snap, why is python 3.5 required?

charmcraft/commands/pack.py
```diff
@@ -0,0 +1,130 @@
...
+    # it's verified before)
+    for fname in MANDATORY_FILES:
+        allpaths.add(dirpath / fname)
+
+    # the extra files, which must be relative
+    config = load_yaml(dirpath / 'charmcraft.yaml') or {}
+    prime_specs = config.get('parts', {}).get('bundle', {}).get('prime', [])
+
+    for spec in prime_specs:
+        if spec[0] == os.path.sep:
```
cjp256, line 56: I don't know if it's bad practice or not... but if the yaml spec requires posix, i'd use "/" instead of os.sep.  Any chance this will ever run native on windows?

charmcraft/commands/pack.py
```diff
@@ -0,0 +1,130 @@
...
+
+        fpaths = sorted(fpath for fpath in dirpath.glob(spec) if fpath.is_file())
+        logger.debug("Including per prime config %r: %s.", spec, fpaths)
+        allpaths.update(fpaths)
+
+    return sorted(allpaths)
+
+
+_overview = """
+Build the bundle, leaving a .zip file as the result of the process.
```
cjp256, line 68: Maybe something like:
```suggestion
Build the bundle and package it as a zip archive.
```

charmcraft/commands/pack.py
```diff
@@ -0,0 +1,130 @@
...
+
+You can `juju deploy` directly from the resulting .zip file, or upload it to
```
cjp256, line 70:
```suggestion
You can `juju deploy` the resulting .zip file, or upload it to
```

charmcraft/commands/pack.py
```diff
@@ -0,0 +1,130 @@
...
+        """Run the command."""
+        if parsed_args.from_dir is None:
+            dirpath = pathlib.Path.cwd()
+        else:
+            dirpath = parsed_args.from_dir.expanduser()
+            if not dirpath.exists():
+                raise CommandError("Bundle project directory was not found: '{}'.".format(dirpath))
+            if not dirpath.is_dir():
+                raise CommandError(
+                    "Bundle project directory is not really a directory: '{}'.".format(dirpath))
```
cjp256, line 102:
```suggestion
                    "Bundle project directory is not a directory: {!r}.".format(str(dirpath)))
```

charmcraft/commands/utils.py
```diff
@@ -33,6 +39,20 @@ def make_executable(fh):
     os.fchmod(fileno, mode)
 
 
+def load_yaml(fpath):
+    """Return the content of a YAML file."""
+    if not fpath.exists():
```
cjp256, line 44:
```suggestion
    if not fpath.is_file():
```

cjp256:
Unrelated, but have you by chance considered using `black` for automatic code formatting?

charmcraft/commands/pack.py
```diff
@@ -0,0 +1,130 @@
...
+        if charmcraft_config.get('type') != 'bundle':
+            raise CommandError(
+                "Invalid charmcraft config; 'type' must be 'bundle' in file '{}'."
+                .format(charmcraft_filepath))
+
+        # pack everything
+        paths = get_paths_to_include(dirpath)
+        zipname = dirpath / (bundle_name + '.zip')
+        build_zip(zipname, dirpath, paths)
+        logger.info("Done, bundle left in '%s'.", zipname)
```
cjp256, line 130:
```suggestion
        logger.info("Bundled %s", zipname)
```

would be consistent with Snapcraft's output, if desired

tests/commands/test_pack.py
```diff
@@ -0,0 +1,418 @@
...
+    build_zip,
+    get_paths_to_include,
+)
+
+
+@pytest.fixture
+def bundle_yaml(tmp_path):
+    """Create an empty bundle.yaml, with the option to set values to it."""
+    bundle_path = tmp_path / 'bundle.yaml'
+    bundle_path.write_text("{}")
```
cjp256, line 39: redundant
```suggestion
    bundle_path.write_text("{}")
```

charmcraft/commands/pack.py
facundobatista: Good idea!

charmcraft/commands/pack.py
facundobatista: Because it's also provided as something installable from PyPI (and you can run it from the project, anyway). In those cases, if you're in a Xenial machine (because you're building for Xenial), it needs to work with 3.5.

charmcraft/commands/pack.py
facundobatista: Yes, we'll run this in Windows at some point.

charmcraft/commands/pack.py
cjp256: Once the lifecycle is integrated, you will undoubtedly have a higher python requirement. It is worth considering the option of taking the position that the snap must be used if the environment does not support the minimum required versions.

charmcraft/commands/pack.py
facundobatista: Oh, very nice! Improved, thanks!

charmcraft/commands/pack.py
facundobatista: nice, thanks! Improved.

charmcraft/commands/pack.py
facundobatista: Removed the "really". Regarding how to show dirpath, this is for consistency with the rest of the project.

I'm +1 to revisit that decision (if you want) and propose a branch to change them all (if needed).

charmcraft/commands/utils.py
facundobatista: Nice, improved (and added a test for it).

charmcraft/commands/pack.py
facundobatista: It was internal consistency with the result of "charm build". That said, Mark change that output, so I'm just adapting this to what Mark used there (which is: `logger.info("Created '%s'.", zipname)`).

tests/commands/test_pack.py
facundobatista: I like the added simplicity of just calling it. However I prefer the explicitness of passing the value you want there. By default it's empty.

Changed it to do all this.

charmcraft/commands/pack.py
cjp256: Yeah, that's the right call. I thought about that afterwards and forgot to come back.  `!r` is the approach made throughout most of snapcraft.

charmcraft/commands/pack.py
cjp256: "/foo" would not fail on Windows in this case.

If the user should always use POSIX pathing in the yaml, we should replace `os.path.sep` with `/`.

Or would the user be using Windows style pathing for windows series charms?

tests/commands/test_pack.py
cjp256: looks great, thanks!

charmcraft/commands/pack.py
facundobatista: Ah, I see your point now!!! So, I changed to `/`, and also added a comment explaining why.

charmcraft/commands/pack.py
facundobatista: Opened https://github.com/canonical/charmcraft/issues/208 for this

chipaca (approved):

PR #200, opened 2020-11-16
Add force option to charmcraft init

dstathis:


charmcraft/commands/init.py
```diff
@@ -79,13 +79,16 @@ def fill_parser(self, parser):
             "--series", default="kubernetes",
             help="A comma-separated list of supported platform series;"
             " defaults to 'kubernetes'")
+        parser.add_argument(
+            "-f", "--force", type="store_true",
+            help="Initialize even if the project dir is not empty. Will not overwrite files")
```
facundobatista, line 84: As we're not finishing with dots, it looks weird to have two sentences. What about "Initialize even if the directory is not empty (will not overwrite files)"

charmcraft/commands/init.py
```diff
@@ -126,6 +129,8 @@ def run(self, args):
             template_name = template_name[:-3]
             logger.debug("Rendering %s", template_name)
             path = args.path / template_name
+            if path.exists():
```
facundobatista, line 132: What if it's a directory?

If I have an empty "src" directory, this code will not create the "charm.py" in it. Should it?

facundobatista (changes requested):
Added a couple of details.

Notably, we need to define how it will behave with directories.

Thanks!!!

charmcraft/commands/init.py
dstathis: I don't understand why this would not work for a directory. path.exists would be true and we would continue. Then when we got to the file it would write it

dstathis: @facundobatista

facundobatista: Ah, now I understand the code: it **cannot** be a directory, as `path` is always the full path of the template file. So we're ok.

facundobatista (approved):

chipaca (approved):

Issue #55, opened 2020-06-24
`charmcraft build` breaks when pip packages come from github


jamesbeedy:
`charmcraft build` errors when pip packages located in github are included in the requirements.txt. 

This issue can be reproduced if you build a charm where the requirements.txt contains pip packages that live in github.

Reproducer:
```
git clone https://github.com/ducks23/charm-slurmdbd -b bdx_charmcraft

cd charm-slurmdbd

charmcraft build
```

Here is our example of the breakage below:

```
$ cat requirements.txt
ops
#git+https://github.com/omnivector-solutions/interface-mysql.git@master
#git+https://github.com/omnivector-solutions/slurm-snap-manager.git@master

$ charmcraft build
Done, charm left in 'slurmdbd.charm'

$ vim requirements.txt

$ cat requirements.txt
ops
git+https://github.com/omnivector-solutions/interface-mysql.git@master
git+https://github.com/omnivector-solutions/slurm-snap-manager.git@master

$ charmcraft build
Executing ['pip3', 'install', '--target=/home/bdx/allcode/github/ducks23/charm-slurmdbd/build/venv', '--requirement=/home/bdx/allcode/github/ducks23/charm-slurmdbd/requirements.txt'] failed with return code 1
problems installing dependencies
```
I can install the requirements.txt by hand successfully
```
$ pip install -r requirements.txt
Collecting git+https://github.com/omnivector-solutions/interface-mysql.git@master (from -r requirements.txt (line 2))
  Cloning https://github.com/omnivector-solutions/interface-mysql.git (to revision master) to /tmp/pip-req-build-ztlxwhvf
  Running command git clone -q https://github.com/omnivector-solutions/interface-mysql.git /tmp/pip-req-build-ztlxwhvf
Collecting git+https://github.com/omnivector-solutions/slurm-snap-manager.git@master (from -r requirements.txt (line 3))
  Cloning https://github.com/omnivector-solutions/slurm-snap-manager.git (to revision master) to /tmp/pip-req-build-pwosnebt
  Running command git clone -q https://github.com/omnivector-solutions/slurm-snap-manager.git /tmp/pip-req-build-pwosnebt
Requirement already satisfied: ops in /home/bdx/allcode/github/venv/lib/python3.8/site-packages (from -r requirements.txt (line 1)) (0.6.1)
[...]

PR #202, opened 2020-11-20
Prod backend configurable

facundobatista:
This draft branch proposes two changes at once. After we settle some design decisions, I will cut both changes in different PRs, but for now I'm putting this here so it serves as example code.

Changes:

- Make the server backend urls configurable

- Have a very simple configuration layer to hold (still optional) charmcraft.yaml


facundobatista:
Related issue: #55

charmcraft/config.py
```diff
@@ -0,0 +1,54 @@
...
+from charmcraft.utils import load_yaml
+
+from collections import UserDict
+
+
+class _Config(UserDict):
+    """Hold the config from charmcraft.yaml."""
+
+    def __init__(self):
+        self.data = {}
```
cjp256, line 31: I've been pushing Snapcraft away from (ab)using dictionaries and instead making this type of object into first-class python object to make it easier to use/maintain. Something like:

```
class Config:
  def __init__(self, *, name: str) -> None:
    self.name = name

  def validate(self) -> None:
    """Validate all properties."""
    ...

  @classmethod
  def unmarshal(self, raw_data) -> "Config":
    """Validate the schema & mypy typing assumptions and return Config."""
    if not isinstance(raw_data, dict):
            raise CommandError("Invalid charmcraft.yaml structure: must be a dictionary.")
    name = raw_data.get("name")
    if not isinstance(name, str):
            raise CommandError(f"Charm name {name!r} is not a valid string.")
    config = Config(name=name)
    config.validate()
    return config
````

File: charmcraft/config.py [hunk 2]
```diff
@@ -0,0 +1,54 @@
...
+        content = load_yaml(project_directory / 'charmcraft.yaml')
+        if content is None:
+            # so far charmcraft.yaml is optional
+            return
+
+        self._validate(content)
+        self.data = content
+
+
+config = _Config()
```
cjp256, line 54: I would reconsider this global variable.  I expect it will cause unnecessary coupling.  Maybe instead just pass `config` as required, or limit to the properties relevant the user's need?

charmcraft/utils.py
```diff
@@ -33,6 +39,20 @@ def make_executable(fh):
...
+def load_yaml(fpath):
+    """Return the content of a YAML file."""
+    if not fpath.is_file():
+        logger.debug("Couldn't find config file %s", fpath)
+        return
+    try:
+        with fpath.open('rb') as fh:
+            content = yaml.safe_load(fh)
+    except (yaml.error.YAMLError, OSError) as err:
+        logger.error("Failed to read/parse config file %s (got %r)", fpath, err)
```
cmatsuoka, line 51: Perhaps simply displaying the original error message after a colon would be easier to read ("Failed to do something: this bad thing happened").

charmcraft/main.py
```diff
@@ -192,32 +198,56 @@ def _pre_parse_args(self, sysargs):
...
+        for sysarg in sysargs:
+            if sysarg in arg_per_option:
+                arg = arg_per_option[sysarg]
+                if arg.type == 'flag':
+                    value = True
+                else:
+                    try:
+                        value = next(sysargs)
+                    except StopIteration:
+                        raise CommandError("The 'from' option expected one argument.")
```
cmatsuoka, line 229: Maybe "expects" instead of "expected"? Or "requires an argument"?

charmcraft/config.py
cmatsuoka: I second that. This will make a difference if for some reason you need to refactor a subsystem into a reusable library in the future.

charmcraft/config.py
facundobatista: As we discussed, I'll go to a schema oriented complex object for this config.

charmcraft/config.py
facundobatista: As we discussed, I'll go to a model where just the minimum config is passed around.

charmcraft/utils.py
facundobatista: nice, improved

charmcraft/main.py
facundobatista: much better, thanks!

facundobatista:
@cjp256 @cmatsuoka current code applies the two concepts we talked about: a hard-defined schema for the config, hold by a frozen complex object, and then passing around the minimum of what's needed, not having everything available everywhere.

I also transformed the just-landed Pack command for bundles to use this config (it previously had to deal directly with charmcraft.yaml).

Of course, this branch is not to be landed as is (that's why Ieft it in Draft), but the idea is to have approval of the general direction, and then I'll start cutting branches from here. First one will be just the config infrastructure with tests (without using it in that branch).

charmcraft/commands/pack.py
```diff
@@ -114,18 +88,14 @@ def run(self, parsed_args):
...
-        if charmcraft_config is None:
-            raise CommandError(
-                "Missing or invalid charmcraft file: '{}'.".format(charmcraft_filepath))
-        if charmcraft_config.get('type') != 'bundle':
+        # so far 'pack' works for bundles only (later this will operate also on charms)
+        if self.config.type != 'bundle':
             raise CommandError(
-                "Invalid charmcraft config; 'type' must be 'bundle' in file '{}'."
-                .format(charmcraft_filepath))
+                "Bad config; charmcraft.yaml missing or with an invalid 'type' field: "
```
cjp256, line 94: Consider splitting up missing vs. unsupported type.

charmcraft/config.py
```diff
@@ -0,0 +1,152 @@
...
+    charmhub = attr.ib(default={}, converter=_CharmhubConfig.from_dict)
+    parts = attr.ib(default={}, converter=_BasicPrime.from_dict)
+    type = attr.ib(default=None, validator=[_check_typefield])
+
+    # this value is provided by the code itself, not the user, as convenience for the
+    # rest of the code
+    project_dirpath = attr.ib(default=None)
+
+    @classmethod
+    def from_file(cls, project_directory):
```
cjp256, line 138: deconflict "from_file" vs directory argument. perhaps use the chamcraft file path as arg?

charmcraft/config.py
```diff
@@ -0,0 +1,152 @@
...
+        else:
+            project_directory = pathlib.Path(project_directory)
+
+        content = load_yaml(project_directory / 'charmcraft.yaml')
+        if content is None:
+            content = {}
+        elif not isinstance(content, dict):
+            raise CommandError("Invalid charmcraft.yaml structure: must be a dictionary.")
+
+        content['project_dirpath'] = project_directory
```
cjp256, line 151: consider decoupling this from the Config (maybe there's a case either for testing or future cases where you need to create configs without a project path readily available).  I'd perhaps consider a 'Project' class that encompasses these aspects of the project that aren't specifically detailed in the yaml.

charmcraft/main.py
```diff
@@ -95,11 +97,14 @@ def run(self, parsed_args, all_commands):
 ]
 
 
-# global options: the flag used internally, short and long parameters, and the help text
-GLOBAL_FLAGS = [
-    ('help', '-h', '--help', "Show this help message and exit"),
-    ('verbose', '-v', '--verbose', "Show debug information and be more verbose"),
-    ('quiet', '-q', '--quiet', "Only show warnings and errors, not progress"),
+# global options: the name used internally, its type, short and long parameters, and help text
+_Global = namedtuple('Global', 'name type short_option long_option help_message')
```
cjp256, line 101: Not important, but dataclass (or attr) would be a better choice than namedtuple imo (type annotations, etc.)

charmcraft/config.py
```diff
@@ -0,0 +1,152 @@
...
+    if not url.scheme or not url.netloc:
+        key = _get_key(config, field)
+        raise CommandError(
+            "Bad charmcraft.yaml content; field '{}' must be a full URL "
+            "(e.g. 'https://some.server.com'): got {!r}.".format(key, value))
+
+
+def _check_typefield(config, field, value):
+    """Check that the value of 'type' field is valid."""
+    # None is allowed until we make mandatory this configuration field
```
sergiusens, line 59:
```suggestion
    # None is allowed until we make this configuration field mandatory
```

charmcraft/config.py
```diff
@@ -0,0 +1,152 @@
...
+
+class _BasicPrime(tuple):
+    """Hold the list of files to include, specified under parts/bundle/prime configs.
+
+    This is a intermediate structure until we have the full Lifecycle in place.
+    """
+    @classmethod
+    def from_dict(cls, parts):
+        """Build from a dicts sequence."""
+        if not isinstance(parts, dict):
```
sergiusens, line 99: one advantage of using jsonschema here is that you eventually get autocomplete on IDEs (VS Code, PyCharms, Emacs, etc)

charmcraft/config.py
```diff
@@ -0,0 +1,152 @@
...
+                "Bad charmcraft.yaml content; the 'parts.bundle' field must be a dict: got {!r}."
+                .format(type(bundle).__name__))
+
+        prime = bundle.get('prime', [])
+        if not isinstance(prime, list):
+            raise CommandError(
+                "Bad charmcraft.yaml content; the 'parts.bundle.prime' field must be a list: "
+                "got {!r}.".format(type(prime).__name__))
+
+        # validate that all are relative
```
sergiusens, line 116: Should this not handle:

```
$ ls prime
foo/ bar foobar
```

- Only Include foo and everything inside it
```yaml
prime:
    - foo
```

- Exclude foo and everything inside it (keeping bar and foobar)
```yaml
prime:
    - -foo
```

sergiusens: If so, spec[0] would need some extra logic

charmcraft/config.py
sergiusens: I am +1 on decoupling file finding (and opening) from config creation

charmcraft/commands/pack.py
facundobatista: Ok, will make this a more specific requirement.

charmcraft/config.py
facundobatista: Will decouple it even more. I will put a "load" function at module level.

charmcraft/config.py
facundobatista: IIUC you're talking about different things here.

On one side, the config will hold a generic `project` option that would hold all project related details, like `directory`, and could be extended in the future.

On the other hand, I'll decouple the config loading from the config itself.

charmcraft/main.py
facundobatista: It's simpler like this and good enough for now, we can improve it in the future.

charmcraft/config.py
facundobatista: thanks, improved.

charmcraft/config.py
facundobatista: Will give it a try to `jsonschema`, I think all could be simpler and more declarative when using this.

charmcraft/config.py
facundobatista: I don't like this "everything but this" rule, it suffers the same problem than "including everything by default": today you exclude `foo` and get `bar` and `foobar`, tomorrow you also add some confidential keys file and it will be included in the charm.

That said, note that this "basic prime" is a simple intermediate thing that will lead us towards the "full lifecycle capabilities", it doesn't need to cover all the cases.

facundobatista:
@cmatsuoka @sergiusens @cjp256 

As per the last round of reviews, I did the following changes:

- The Pack command now separates the cases of "missing a charmcraft.yaml config" (which is mandatory for this command) and the `type` field not being `bundle`. Note that it doesn't check that the field is there (this is handled in the config, as that field is mandatory).

- Removed `project_dirpath` in favor of a first class `project` entry that currently just holds `dirpath` but it may grow in the future.

- Refactored the whole config infrastructure to now use `jsonschema`; some details about this:

    - had to adapt the validation/format errors produced by that lib; surprisingly the adaptor function is small and without magic, and the error messages produced are the same than before using the library (this could grow/change in the future, though, if we need different messages -- probably not, but looking at how the jsonschema errors are massaged in snapcraft, we will see).

    - there's lot less manual checks, as `jsonschema` handled them; the only function in that regard that survived the refactor is the one validation that urls are as we need.

    - the config classes are smaller and neater; I even thought if we should keep these "attr classes" or we could go towards something simpler, and took the decission of leaving them as probably we will replace them by "dataclasses" when using a modern enough Python.

    - sure, now we have a splicit schema that we need to maintain, but this is **better**, as it's easier to understand than fields scattered all over the place; note that all the schema is unittested, specially to validate that the messages we're producing to the users are the one that we want.

    - decoupled the config creation from the config classes, now it's done through module function.

charmcraft/commands/store/client.py
```diff
@@ -164,8 +159,14 @@ def _storage_push(monitor):
 class Client:
     """Lightweight layer above _AuthHolder to present a more network oriented interface."""
 
-    def __init__(self):
+    # default production URLs
```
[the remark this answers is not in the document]
cjp256: Also, you may find it nicer to set the parameters with their respective default rather than write as a class/instance var.  Then when creating a client, the parameters are optional (nice for testing etc).

tests/conftest.py
```diff
@@ -60,3 +62,27 @@ def __getattribute__(self, name):
...
+
+
+@pytest.fixture
+def config(tmp_path):
+    """Provide a config class with an extra set method for the test to change it."""
+
+    class TestConfig(config_module._Config):
+        """The Config, but with a method to set test values."""
+
+        def set(self, **kwargs):
```
cjp256, line 74: IMO, now that Config is a proper object, this fixture should no longer be used.  The tests should instantiate their own Config() with the necessary parameters required for that test rather than jump through this hoop.

charmcraft/commands/store/client.py
```diff
@@ -139,9 +134,9 @@ def request(self, method, url, body):
         return resp
 
 
-def _storage_push(monitor):
+def _storage_push(monitor, storage_base_url):
```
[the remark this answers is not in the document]
facundobatista: I agree only when the multiple parameters are options.

charmcraft/commands/store/client.py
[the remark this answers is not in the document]
facundobatista: As discussed, we'll defer using docstrings for sphinx-generated documentation for a future when charmcraft is a library (we're not generating docs from docstrings as it's only a CLI tool).

charmcraft/config.py
```diff
@@ -0,0 +1,170 @@
...
+format_checker = jsonschema.FormatChecker()
+
+# translator between "json" and pythonic yaml names
+TYPE_TRANSLATOR = {
+    'object': 'dict',
+    'array': 'list',
+}
+
+
+def adapt_validation_error(error):
```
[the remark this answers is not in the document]
facundobatista: We'll go "full type hinting" in the future. After that, all code landed should have type hints, yes.

charmcraft/config.py
```diff
@@ -0,0 +1,170 @@
...
+                'bundle': {
+                    'type': 'object',
+                    'properties': {
+                        'prime': {'type': 'array'},
+                    },
+                },
+            },
+        },
+    },
+    'required': ['type'],
```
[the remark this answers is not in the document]
sergiusens: this will help with the top level entries, we should not do this for parts though

sergiusens (commented):
this should move forward into its smaller PRs, general structure looks good

charmcraft/config.py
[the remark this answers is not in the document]
facundobatista: Good idea. Will do.

charmcraft/config.py
```diff
@@ -0,0 +1,170 @@
...
+    if dirpath is None:
+        dirpath = pathlib.Path.cwd()
+    else:
+        dirpath = pathlib.Path(dirpath)
+
+    content = load_yaml(dirpath / 'charmcraft.yaml')
+    # XXX Facundo 2021-01-04: we will make this configuration mandatory in the future, but
+    # so far is ok to not have it
+    if content is None:
+        return
```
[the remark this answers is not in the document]
facundobatista: I don't like explicit None in empty returns :/

tests/conftest.py
facundobatista: As the config is not omnipresent anymore, the config needs to be used in ~130 tests; having it as a fixture, easy to use in most of the cases where a special value is not needed reduces a lot the code needed in all the tests.

facundobatista:
All this went in (except #229, but already proposed).

PR #204, opened 2020-11-28
Update snapcraft.yaml

markshuttle:
Hi folks - some text changes for snapcraft.io and snap users


facundobatista (approved):
Thanks.

chipaca (changes requested):
the description in snapcraft.yaml does not support links

chipaca (approved):

Issue #70, opened 2020-07-02
Verify types in the store layer for stuff returned to the commands


facundobatista:
We could something like this (after we forget Py3.5):

```python
import typing
from collections import namedtuple


class TypedNamedTuple:
    """A namedtuple that verifies hinted types."""

    def __new__(cls, **kwargs):
        annotations = typing.get_type_hints(cls)
        for attr, hint in annotations.items():
            assert isinstance(kwargs[attr], hint)
        nt = namedtuple(cls.__name__, list(annotations))
        return nt(**kwargs)


class User(TypedNamedTuple):
    """The authenticated user information."""
    name: str
    username: str
    userid: str
```

Extra benefit: we get to have docstrings for each type :)

PR #206, opened 2020-12-02
(charmlibs) List libraries for a given charm

facundobatista:
This adds the command itself, and the supporting method in the store layer.

For ListLibCommand:

- it can run standalone, or with a given charm name

- when standalone, it will try to access metadata.yaml, and if there it will list all libraries for that charm

- if a charm name is indicated, will just list all libraries for that charm

In the store layer:

- a new multi-functional `get_libraries_tips` is added

- it's bulk, and supports all the flexibility in the store for each query:

    - just the library id (will get the tip for that library)

    - just the charm name (will get the tip for all libraries of that charm)

    - charm and library names (alternative identification of a specific library)

    - for any of those cases, `api` is supported (it will give the tip for that API version)

- note that this is current more flexible than needed for ListLibCommand itself, but more of this will be used in the FetchLibCommand (in the next branch)

### DIFF
--- a/charmcraft/commands/store/__init__.py
+++ b/charmcraft/commands/store/__init__.py
@@ -825,3 +825,47 @@ class PublishLibCommand(BaseCommand):
             logger.info(
                 "Library %s sent to the store with version %d.%d",
                 lib_data.full_name, lib_data.api, lib_data.patch)
+
+
+class ListLibCommand(BaseCommand):
+    """List all libraries belonging to a charm."""
+    name = 'list-lib'
+    help_msg = "List all libraries from a charm"
+    overview = textwrap.dedent("""
+        List all libraries from a charm.
+
+        For each library, it will show the name and the api and patch versions
+        for its tip.
+    """)
+
+    def fill_parser(self, parser):
+        """Add own parameters to the general parser."""
+        parser.add_argument('--charm-name', help="The name of the charm")
+
+    def run(self, parsed_args):
+        """Run the command."""
+        if parsed_args.charm_name:
+            charm_name = parsed_args.charm_name
+        else:
+            charm_name = get_name_from_metadata()
+            if charm_name is None:
+                raise CommandError(
+                    "Can't access name in 'metadata.yaml' file. The 'list-lib' command needs to "
+                    "be executed in a valid project's directory, or indicate the charm name with "
+                    "the --charm-name option.")
+
+        # get tips from the Store
+        store = Store()
+        to_query = [{'charm_name': charm_name}]
+        libs_tips = store.get_libraries_tips(to_query)
+
+        if not libs_tips:
+            logger.info("Nothing found.")
+            return
+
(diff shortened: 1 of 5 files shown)


chipaca (approved):

charmcraft/commands/store/store.py
```diff
@@ -210,3 +210,26 @@ def create_library_revision(self, charm_name, lib_id, api, patch, content, conte
...
         result = _build_library(response)
         return result
+
+    def get_libraries_tips(self, libraries):
+        """Get the tip details for several libraries at once."""
+        endpoint = '/v1/charm/libraries/bulk'
+        payload = []
+        for lib in libraries:
+            if 'lib_id' in lib:
+                d = {
```
cjp256, line 220: nit: `d` could use a better name

charmcraft/commands/store/__init__.py
```diff
@@ -825,3 +825,50 @@ def run(self, parsed_args):
...
+        """Run the command."""
+        if parsed_args.name:
+            charm_name = parsed_args.name
+        else:
+            charm_name = get_name_from_metadata()
+            if charm_name is None:
+                raise CommandError(
+                    "Can't access name in 'metadata.yaml' file. The 'list-lib' command needs to "
+                    "be executed in a valid project's directory, or indicate the charm name with "
+                    "the --charm-name option.")
```
cmatsuoka, line 858: This error message is sounding a bit strange to me, I wonder if we could use something like "... command must either be executed from a valid project directory, or specify a charm name using the --charm-name option"? @cjp256 may have a better suggestion here.

charmcraft/commands/store/store.py
```diff
@@ -210,3 +210,26 @@ def create_library_revision(self, charm_name, lib_id, api, patch, content, conte
         response = self._client.post(endpoint, payload)
         result = _build_library(response)
         return result
+
+    def get_libraries_tips(self, libraries):
+        """Get the tip details for several libraries at once."""
```
cjp256, line 215: This could use a bit more description on what it's doing, input, outputs, etc. It took me some time to parse what's happening.  

Seems like this queries a set of libraries referenced by ID, or charm name [+ library name]. It would help if these inputs could be encapsulated into better defined parameters / class(es) such as `Library`

cmatsuoka (approved):
Thanks Facundo, this looks good. I only added a minor comment about the wording in an error message.

charmcraft/commands/store/store.py
facundobatista: Improved.

charmcraft/commands/store/store.py
facundobatista: Improved the docstring to describe better what's going on.

Regarding interfaces, it's better to keep each item as a dictionary, because if the key is present we need to send that info in the payload (otherwise we would need a specific token to flag that the key should not be sent). That said, we can improve this in the future if needed; #70 is related.

PR #207, opened 2020-12-02
Avoid a duplicated -h/--help option when getting help for a command.

facundobatista:
This was because we had two `-h/--help` options, one which is the global one (that works for the tool in general and for each command, as with any global option), and other specific for the command added by `ArgumentParser`. The fix is to tell `ArgumentParser` to not automatically add a help option.

Regarding tests, no new one was needed, just improved the current ones to stop hiding duplicates.


cjp256 (approved):
LGTM

chipaca (approved):

PR #209, opened 2020-12-04
Add coverage to unit tests by default in 'charmcraft init'

mthaddon:
I think it'd be nice to include a coverage report in the default unit test output from what's generated by 'charmcraft init'. This PR adds that.

### DIFF
--- a/charmcraft/templates/init/requirements-dev.txt.j2
+++ b/charmcraft/templates/init/requirements-dev.txt.j2
@@ -1,2 +1,3 @@
 -r requirements.txt
+coverage
 flake8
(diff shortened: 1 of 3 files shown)


chipaca:
should we switch the template to use tox?

mthaddon:
> should we switch the template to use tox?

I'd be happy to see that, as my team tends to favour tox for our charms. Worth doing as a separate PR, or would you like to see that here? And is that something you'd like me to have a go at or would you prefer to do it?

chipaca:
I think it's something for a separate PR, at least, and we should first decide if it's the right thing.
The current approach works, and the resulting tests are faster to get up and running, but as we see more people use tox, I wonder if we should just bite the bullet across the board.
It's no longer my decision to make, though.

chipaca (approved):

facundobatista (approved):
Great, thanks!

cjp256 (approved):
LGTM

PR #211, opened 2020-12-08
Exercise command parsers.

chipaca:
Before this change, asking for help would always return with error.
This made testing actually asking for help hard, so we didn't, so we
never exercised the actual command parser, so a bug snuck in. This
changes things so that help returns with error only when actually in
error, and adds a test that asks for help over all commands, meaning
we exercise all the parsers, so we would actually catch this bug.

Fixes: #210


facundobatista (approved):
Wonderful, thanks!

dstathis (approved):

PR #215, opened 2020-12-17
(charmlibs) Fetch the specified library

facundobatista:
This adds the command itself, and the supporting method in the store layer.

The FetchLibCommand can run standalone, or with a given library full name:

- when standalone, it will update all the libraries already locally in disk

- if a full library name is indicated, will just get/update that library

In any case, verifications are done that the library can be downloaded for the first time or updated, and the user will be informed of the final action.

Regarding the store layer, a simple method was added to support the GET library.

NOTE: this is the final cut of the original `charmlibs-draft` branch.


charmcraft/commands/store/__init__.py
```diff
@@ -827,6 +827,93 @@ def run(self, parsed_args):
...
 
 
+class FetchLibCommand(BaseCommand):
+    """Fetch one or more charm libraries."""
+    name = 'fetch-lib'
+    help_msg = "Fetch one or more charm libraries"
+    overview = textwrap.dedent("""
+        Fetch charm libraries.
+
+        It will download the library the first time, and update it the next times.
```
sergiusens, line 837: Questions/comments:

- maybe something like "subsequent fetches of the same libraries will update the local copies"
- the libraries are always downloaded, not just the first time, right?
- if wanting to keep the current message then s/times/time/

charmcraft/commands/store/__init__.py
```diff
@@ -827,6 +827,93 @@ def run(self, parsed_args):
...
+                item = dict(lib_id=lib.lib_id)
+            item['api'] = lib.api
+            to_query.append(item)
+        libs_tips = store.get_libraries_tips(to_query)
+
+        # check if something needs to be done
+        to_fetch = []
+        for lib_data in local_libs_data:
+            logger.debug("Verifying local lib %s", lib_data)
+            # if locally we didn't have the lib id, let's fix it from the Store info
```
sergiusens, line 869:
```suggestion
            # fix any missing lib id using Store info.
```

charmcraft/commands/store/__init__.py
```diff
@@ -827,6 +827,93 @@ def run(self, parsed_args):
...
+            logger.debug("Store tip: %s", tip)
+            if tip is None:
+                logger.info("Library %s not found in Charmhub.", lib_data.full_name)
+                continue
+
+            if tip.patch > lib_data.patch:
+                # the store is more advanced than local
+                to_fetch.append(lib_data)
+            elif tip.patch < lib_data.patch:
+                # the store has smaller version numbers than local
```
sergiusens, line 886:
```suggestion
                # the store has a lower version numbers than local
```

charmcraft/commands/store/__init__.py
```diff
@@ -827,6 +827,93 @@ def run(self, parsed_args):
...
+                        break
+
+            tip = libs_tips.get((lib_data.lib_id, lib_data.api))
+            logger.debug("Store tip: %s", tip)
+            if tip is None:
+                logger.info("Library %s not found in Charmhub.", lib_data.full_name)
+                continue
+
+            if tip.patch > lib_data.patch:
+                # the store is more advanced than local
```
sergiusens, line 883:
```suggestion
                # the store has a higher version than local
```

charmcraft/commands/store/__init__.py
sergiusens: Reading the code below, maybe clarify when local gets updated when fetching

charmcraft/commands/store/store.py
```diff
@@ -211,6 +211,13 @@ def create_library_revision(self, charm_name, lib_id, api, patch, content, conte
         result = _build_library(response)
         return result
 
+    def get_library(self, charm_name, lib_id, api):
+        """Get the library tip by id for a given api version."""
+        endpoint = '/v1/charm/libraries/{}/{}?api={}'.format(charm_name, lib_id, api)
```
sergiusens, line 216: question, is there any charm_name or lib_id that would ever need to be url encoded? I suspect not

charmcraft/commands/store/__init__.py
facundobatista: Rewrote it: *The first time a library is downloaded the command will create the needed directories to place it, subsequent fetches will just update the local copy.*

charmcraft/commands/store/__init__.py
facundobatista: improved, thanks

charmcraft/commands/store/__init__.py
facundobatista: thanks

charmcraft/commands/store/__init__.py
facundobatista: fixed

charmcraft/commands/store/store.py
facundobatista: Right, both names are validated to be a small subset of ASCII.

charmcraft/commands/store/__init__.py
```diff
@@ -827,6 +827,94 @@ def run(self, parsed_args):
                 lib_data.full_name, lib_data.api, lib_data.patch)
 
 
+class FetchLibCommand(BaseCommand):
+    """Fetch one or more charm libraries."""
+    name = 'fetch-lib'
```
sergiusens, line 832: This is bikeshedding territory, but I find a disconnect between the command name and what it does (on lib by name versus N in tree)

tests/commands/test_store_commands.py
```diff
@@ -1591,6 +1592,205 @@ def test_getlibinfo_libid_empty(tmp_path, monkeypatch):
...
+def test_fetchlib_simple_updated(caplog, store_mock, tmp_path, monkeypatch):
+    """Happy path fetching the lib for Nth time (updating it)."""
+    caplog.set_level(logging.INFO, logger="charmcraft.commands")
+    monkeypatch.chdir(tmp_path)
+
+    lib_id = 'test-example-lib-id'
+    content, content_hash = factory.create_lib_filepath(
+        'testcharm', 'testlib', api=0, patch=1, lib_id=lib_id)
+
+    new_lib_content = 'some test content with uñicode ;)'
```
sergiusens, line 1635: uñicode :laughing:

sergiusens (commented):
Minor comments, general approve though

charmcraft/commands/store/__init__.py
facundobatista: It already went through a couple of design meetings with Mark and Gustavo, so probably it's fine to leave the first version like this, yes.

tests/commands/test_store_commands.py
```diff
@@ -1591,6 +1592,205 @@ def test_getlibinfo_libid_empty(tmp_path, monkeypatch):
         "Library {} metadata field LIBID must be a non-empty ASCII string.".format(test_path))
 
 
+# -- tests for fetch libraries command
+
+def test_fetchlib_simple_downloaded(caplog, store_mock, tmp_path, monkeypatch):
+    """Happy path fetching the lib for first time (downloading it)."""
```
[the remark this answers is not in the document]
facundobatista: Fixed, thanks.

tests/commands/test_store_commands.py
```diff
@@ -1591,6 +1592,205 @@ def test_getlibinfo_libid_empty(tmp_path, monkeypatch):
...
+        call.get_library('testcharm', lib_id, 0),
+    ]
+    expected = "Library charms.testcharm.v0.testlib updated to version 0.2."
+    assert [expected] == [rec.message for rec in caplog.records]
+    saved_file = tmp_path / 'lib' / 'charms' / 'testcharm' / 'v0' / 'testlib.py'
+    assert saved_file.read_text() == new_lib_content
+
+
+def test_fetchlib_all(caplog, store_mock, tmp_path, monkeypatch):
+    """Publish all the libraries found in disk."""
```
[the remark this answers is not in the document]
facundobatista: Silly C&P mistake. Fixed, thanks.

sergiusens (approved):

PR #218, opened 2021-01-13
Promoted the utils module, and a simple test enhancement.

facundobatista:
Promoted the `utils` module out of `commands` because the configuration module will use it.

Also included a fix in the test that runs the tests from the created project by `charmcraft init` (now it uses `coverage`, so depending on how the project tests are run, we also need to fix the path).


tests/commands/test_init.py
```diff
@@ -95,16 +95,19 @@ def test_executables(tmp_path):
 
 
 def test_tests(tmp_path):
-    # fix the PYTHONPATH so the tests in the initted environment use our own
-    # virtualenv (if any), as they need one, but we're not creating one for them; note
-    # that for CI this normally doesn't run under a venv, so this may fix nothing
+    # fix the PYTHONPATH and PATH so the tests in the initted environment use our own
+    # virtualenv libs and bins (if any), as they need them, but we're not creating a
+    # venv for the tests (note that for CI doesn't use a venv, so this is for local tests)
```
cjp256, line 100:
```suggestion
    # venv for local tests (note that CI doesn't use a venv)
```

tests/commands/test_init.py
```diff
@@ -95,16 +95,19 @@ def test_executables(tmp_path):
...
     env = os.environ.copy()
     env_paths = [p for p in sys.path if 'env/lib/python' in p]
     if env_paths:
         if 'PYTHONPATH' in env:
             env['PYTHONPATH'] += ':' + ':'.join(env_paths)
         else:
             env['PYTHONPATH'] = ':'.join(env_paths)
+        for path in env_paths:
+            bin_path = path[:path.index('env/lib/python')] + 'env/bin'
+            env['PATH'] = bin_path + ':' + env['PATH']
```
cjp256, line 110: I'm wondering why the environment needs to be instrumented at all?

cjp256 (approved):

tests/commands/test_init.py
facundobatista: improved, thanks

tests/commands/test_init.py
facundobatista: The user (as is, the charmcraft developer) has the choice. I, for example, don't do that. I run the tests doing `env/bin/python3 -m pytest`

cjp256 (approved):

PR #220, opened 2021-01-14
improve the user-agent string sent to servers

### LINKED ISSUE #74: Improve UserAgent string sent to the servers
Currently it's `"charmcraft/{}".format(__version__)`

Probably include the platform, architecture, Python version...

jbpratt:
Extend the current user-agent to include the current system, machine
arch and python version, using conventions from the MDN docs. The
user-agent test has been updated to mock these values though it is a bit
ugly to do so, I don't exactly like the patching of `platform` so there may
be a better/cleaner way to handle that.

Also update copyright dates in files modified.

`459 passed, 8 warnings in 5.12s`

fixes #74

Happy to rework this however is desired!

### DIFF
--- a/charmcraft/commands/store/client.py
+++ b/charmcraft/commands/store/client.py
@@ -1,4 +1,4 @@
-# Copyright 2020 Canonical Ltd.
+# Copyright 2020-2021 Canonical Ltd.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -18,6 +18,7 @@
 
 import logging
 import os
+import platform
 import webbrowser
 from http.cookiejar import MozillaCookieJar
 
@@ -46,9 +47,10 @@ STORAGE_BASE_URL = 'https://storage.staging.snapcraftcontent.com'
 
 def build_user_agent():
     """Build the charmcraft's user agent."""
-    # XXX Facundo 2020-06-29: we need to include here at least the platform, maybe
-    # architecture, etc. Related: issue #74.
-    return "charmcraft/{}".format(__version__)
+    return "charmcraft/{} ({}; {}) python/{}".format(__version__,
+                                                     platform.system(),
+                                                     platform.machine(),
+                                                     platform.python_version())
 
 
 def visit_page_with_browser(visit_url):
--- a/charmcraft/commands/store/client.py
+++ b/charmcraft/commands/store/client.py
@@ -18,6 +18,7 @@
 
 import logging
 import os
+import pathlib
 import platform
 import webbrowser
 from http.cookiejar import MozillaCookieJar
@@ -44,43 +45,42 @@ logger = logging.getLogger('charmcraft.commands.store')
 API_BASE_URL = 'https://api.staging.charmhub.io'
 STORAGE_BASE_URL = 'https://storage.staging.snapcraftcontent.com'
 
-
-def _is_ci_env() -> bool:
-    """Utilize standard env variables to determine if the environment is for testing."""
-    env_prefixes = ["TRAVIS", "AUTOPKGTEST_TMP"]
-    for prefix in env_prefixes:
-        for key in os.environ.keys():
-            if key.startswith(prefix):
-                return True
-    return False
+TESTING_ENV_PREFIXES = ["TRAVIS", "AUTOPKGTEST_TMP"]
 
 
-def _get_os_platform(filepath: str = "/etc/os-release") -> str:
+def _get_os_platform(filepath=pathlib.Path("/etc/os-release")):
     """Determine a system/release combo for an OS using /etc/os-release if available."""
-    if platform.system() == "Linux":
-        os_release = dict()
-        try:
-            with open(filepath) as f:
-                for line in f:
-                    entry = line.rstrip().split("=")
-                    if len(entry) == 2:
-                        os_release[entry[0]] = entry[1].strip('"')
-        except FileNotFoundError:
-            logger.warning("Unable to locate 'os-release' file, using default values")
-        finally:
-            return "{}/{} ({})".format(os_release.get("NAME", "Unknown"),
-                                       os_release.get("VERSION_ID", "Unknown Version"),
-                                       platform.machine())
+    system = platform.system()
+    release = platform.release()
+    machine = platform.machine()
 
-    return "{}/{} ({})".format(platform.system(), platform.release(), platform.machine())
+    if system == "Linux":
+        os_release = {}
+        try:
+            with filepath.open("r", encoding='utf-8') as f:
+                for line in f:
+                    if "=" in line:
+                        key, value = line.rstrip().split("=", 1)
+                        os_release[key] = value.strip('"')
+        except FileNotFoundError:
+            logger.debug("Unable to locate 'os-release' file, using default values")
+        finally:
+            system = os_release.get("NAME", "Unknown")
+            release = os_release.get("VERSION_ID", "Unknown Version")
+
+    return "{}/{} ({})".format(system, release, machine)
 
 
 def build_user_agent():
     """Build the charmcraft's user agent."""
-    return "charmcraft/{} {} {} python/{}".format(__version__,
-                                                  "(testing)" if _is_ci_env() else "",
-                                                  _get_os_platform(),
-                                                  platform.python_version())
+    if any(key.startswith(prefix) for prefix in TESTING_ENV_PREFIXES for key in os.environ.keys()):
+        testing = " (testing) "
+    else:
+        testing = " "
+    return "charmcraft/{}{}{} python/{}".format(__version__,
+                                                testing,
+                                                _get_os_platform(),
+                                                platform.python_version())
 
 
 def visit_page_with_browser(visit_url):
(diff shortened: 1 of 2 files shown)


jbpratt:
Hi @facundobatista, this was just some low hanging fruit that I saw when looking over the issues. Please let me know if I can update this in any way!

facundobatista:
Yes, thanks for this! I'm trying to find a reviewer that know more about useful user agents than me. We'll review this super soon, thanks again!

charmcraft/commands/store/client.py
```diff
@@ -46,9 +47,10 @@
 
 def build_user_agent():
     """Build the charmcraft's user agent."""
-    # XXX Facundo 2020-06-29: we need to include here at least the platform, maybe
-    # architecture, etc. Related: issue #74.
-    return "charmcraft/{}".format(__version__)
+    return "charmcraft/{} ({}; {}) python/{}".format(__version__,
```
sergiusens, line 50: Just for reference, this is the UA we agreed to for Snapcraft
`"snapcraft/{snapcraft.__version__} {testing}{os_platform} ({arch})"`

Which can lead to:

- snapcraft/4.4 Unknown/Unknown Version (armhf)
- snapcraft/4.4 Ubuntu/18.04 (armhf)
- snapcraft/4.4 (testing) Ubuntu/18.04 (armhf)
- snapcraft/4.5 Windows (amd64)
- snapcraft/4.5 Darwin (amd64)

jbpratt: Ah, I see, much more detailed than my simple implementation. I am more than fine with fleshing this out, I should have started a discussion around this anyway but wanted to jump into something ( :smile: ). Standardizing with Snapcraft seems like the best idea to me since it has been discussed and agreed on there.

If anyone has any thoughts please let me know! cc @facundobatista @cjp256

facundobatista: Thanks @jbpratt for working on this!

jbpratt:
I looked over Snapcraft, the information is coming from `ProjectOptions` and `OsRelease`, the former doing some additional mapping of build info and the latter being a wrapper around `/etc/os-release`. Reading in `/etc/os-release` can hopefully give us a bit more detail, I believe my UA would be near equivalent between Snapcraft and Charmcraft on my local machine since I am relying on the same fields (`NAME`, and `VERSION_ID`).

Testing this required quite a bit of patching that is ugly, maybe I can clean that up. Let me know if this way off, or any changes/mistakes I made and I can rework it.

charmcraft/commands/store/client.py
```diff
@@ -44,11 +45,42 @@
...
 
 
+def _is_ci_env() -> bool:
+    """Utilize standard env variables to determine if the environment is for testing."""
+    env_prefixes = ["TRAVIS", "AUTOPKGTEST_TMP"]
+    for prefix in env_prefixes:
+        for key in os.environ.keys():
+            if key.startswith(prefix):
+                return True
+    return False
```
facundobatista, line 55: No need for this function.

The `env_prefixes` can be a global constant (`TESTING_ENV_PREFIXES`?), and the rest of the code can be written like:

     if any(key.startswith(TESTING_ENV_PREFIXES) for key in os.environ.keys()):

charmcraft/commands/store/client.py
```diff
@@ -44,11 +45,42 @@
...
+
+
+def _get_os_platform(filepath: str = "/etc/os-release") -> str:
+    """Determine a system/release combo for an OS using /etc/os-release if available."""
+    if platform.system() == "Linux":
+        os_release = dict()
```
facundobatista, line 61: No need to call a function, can create a dict syntactically: `{}`.

charmcraft/commands/store/client.py
```diff
@@ -44,11 +45,42 @@
...
+                    if len(entry) == 2:
+                        os_release[entry[0]] = entry[1].strip('"')
+        except FileNotFoundError:
+            logger.warning("Unable to locate 'os-release' file, using default values")
+        finally:
+            return "{}/{} ({})".format(os_release.get("NAME", "Unknown"),
+                                       os_release.get("VERSION_ID", "Unknown Version"),
+                                       platform.machine())
+
+    return "{}/{} ({})".format(platform.system(), platform.release(), platform.machine())
```
facundobatista, line 75: Don't format the platform twice (as this could lead to inconsistencies in the future because changing only one).

So, just do `"{}/{} ({})".format(system, release, machine)` at the end, and arrive there with the proper values for each of those.

charmcraft/commands/store/client.py
```diff
@@ -44,11 +45,42 @@
...
+    if platform.system() == "Linux":
+        os_release = dict()
+        try:
+            with open(filepath) as f:
+                for line in f:
+                    entry = line.rstrip().split("=")
+                    if len(entry) == 2:
+                        os_release[entry[0]] = entry[1].strip('"')
+        except FileNotFoundError:
+            logger.warning("Unable to locate 'os-release' file, using default values")
```
facundobatista, line 69: This should be in DEBUG mode (as there's no really a situation that the users need to be aware if they don't have that file).

charmcraft/commands/store/client.py
```diff
@@ -44,11 +45,42 @@
...
+
+
+def _get_os_platform(filepath: str = "/etc/os-release") -> str:
+    """Determine a system/release combo for an OS using /etc/os-release if available."""
+    if platform.system() == "Linux":
+        os_release = dict()
+        try:
+            with open(filepath) as f:
+                for line in f:
+                    entry = line.rstrip().split("=")
```
facundobatista, line 65: Just do `key, value = line.strip().split("=", 1)`, no need for subsequent `if` (which will make all this work better if the value has an equal sign).

tests/commands/test_store_client.py
```diff
@@ -36,15 +37,77 @@
...
+
+def test_useragent_windows():
+    with patch('charmcraft.commands.store.client.__version__', '1.2.3'), \
+            patch('charmcraft.commands.store.client._is_ci_env', return_value=False), \
+            patch('platform.system', return_value='Windows'), \
+            patch('platform.release', return_value='10'), \
+            patch('platform.machine', return_value='AMD64'), \
+            patch('platform.python_version', return_value='3.9.1'):
+        ua = build_user_agent()
+    assert ua == "charmcraft/1.2.3  Windows/10 (AMD64) python/3.9.1"
```
facundobatista, line 68: Please remove the double space here (between `1.2.3` and `Windows`).

charmcraft/commands/store/client.py
```diff
@@ -44,11 +45,42 @@
...
+                return True
+    return False
+
+
+def _get_os_platform(filepath: str = "/etc/os-release") -> str:
+    """Determine a system/release combo for an OS using /etc/os-release if available."""
+    if platform.system() == "Linux":
+        os_release = dict()
+        try:
+            with open(filepath) as f:
```
facundobatista, line 63: If the file is open in text mode, please specify the encoding (if we leave it platform dependent may lead to obscure errors).

facundobatista (changes requested):
Great work, thanks for your time and effort here! I annotated a couple of points to fix/improve.

Thanks again!!!

charmcraft/commands/store/client.py
```diff
@@ -43,12 +45,42 @@
...
+    system = platform.system()
+    release = platform.release()
+    machine = platform.machine()
+
+    if system == "Linux":
+        os_release = {}
+        try:
+            with filepath.open("r", encoding='utf-8') as f:
+                for line in f:
+                    if "=" in line:
```
jbpratt, line 62: I left a check here as my test case `/etc/os-release` file was causing this to split and unpack an empty line which would panic. Rather than fixing the test case, I figured it is better safe than to cause a crash because someone/thing modified a file they shouldn't have.

jbpratt:
Thanks for the review! I have made the requested changes.

charmcraft/commands/store/client.py
```diff
@@ -43,12 +45,42 @@
...
+                        key, value = line.rstrip().split("=", 1)
+                        os_release[key] = value.strip('"')
+        except FileNotFoundError:
+            logger.debug("Unable to locate 'os-release' file, using default values")
+        finally:
+            system = os_release.get("NAME", "Unknown")
+            release = os_release.get("VERSION_ID", "Unknown Version")
```
facundobatista, line 69: I think it's better, in case of not having `NAME` or `VERSION_ID`, to just leave original `system` and `release` (from `platform`, thank "Unknown"s.

IOW, what do you think about doing?

            system = os_release.get("NAME", system)
            release = os_release.get("VERSION_ID", release)

facundobatista (changes requested):
Great, thanks! I added another small request for your consideration (sorry I didn't spot this before). Thanks again for your work here!

charmcraft/commands/store/client.py
jbpratt: No worries, happy to make changes. I emulated what Snapcraft was doing (https://github.com/snapcore/snapcraft/blob/b5138e03e53727347f65b0b1499a02fccd1ef9b4/snapcraft/storeapi/_agent.py#L35) but I agree that this is more clear. I have pushed an update!

facundobatista (approved):
Wonderful, thanks for the work! 

Please resolve the conflict, and we're ready to go!  Thanks again.

jbpratt:
Resolved! Thanks for the review. If you know of any issues I can try to tackle please feel free to send it my way, though I don't want to retract from your development time.

jbpratt:
Oh! I should have squashed this up, I have made quite the mess. FWIW it is also possible to squash it when merging, the drop down allows for that. 

Very sorry on my part.

PR #221, opened 2021-01-14
Configuration schema and support classes.

facundobatista:
Not used from the rest of the system, yet, to keep the PR reviewable.

### DIFF
--- a/charmcraft/config.py
+++ b/charmcraft/config.py
@@ -132,6 +132,7 @@ CONFIG_SCHEMA = {
                 'api_url': {'type': 'string', 'format': 'url'},
                 'storage_url': {'type': 'string', 'format': 'url'},
             },
+            'additionalProperties': False,
         },
         'parts': {
             'type': 'object',
--- a/charmcraft/config.py
+++ b/charmcraft/config.py
@@ -58,7 +58,7 @@ def adapt_validation_error(error):
 
 
 @format_checker.checks('url', raises=ValueError)
-def _check_url(value):
+def check_url(value):
     """Check that the URL has at least scheme and net location."""
     if isinstance(value, str):
         url = urlparse(value)
@@ -68,7 +68,7 @@ def _check_url(value):
 
 
 @attr.s(kw_only=True, frozen=True)
-class _CharmhubConfig:
+class CharmhubConfig:
     """Configuration for all Charmhub related options."""
 
     api_url = attr.ib(default='https://api.staging.charmhub.io')
@@ -80,7 +80,7 @@ class _CharmhubConfig:
         return cls(**source)
 
 
-class _BasicPrime(tuple):
+class BasicPrime(tuple):
     """Hold the list of files to include, specified under parts/bundle/prime configs.
 
     This is a intermediate structure until we have the full Lifecycle in place.
@@ -103,18 +103,18 @@ class _BasicPrime(tuple):
 
 
 @attr.s(kw_only=True, frozen=True)
-class _Project:
+class Project:
     """Configuration for all project-related options, used internally."""
 
     dirpath = attr.ib(default=None)
 
 
 @attr.s(kw_only=True, frozen=True)
-class _Config:
+class Config:
     """Root of all the configuration."""
 
-    charmhub = attr.ib(default={}, converter=_CharmhubConfig.from_dict)
-    parts = attr.ib(default={}, converter=_BasicPrime.from_dict)
+    charmhub = attr.ib(default={}, converter=CharmhubConfig.from_dict)
+    parts = attr.ib(default={}, converter=BasicPrime.from_dict)
     type = attr.ib()
 
     # this item is provided by the code itself, not the user, as convenience for the
@@ -170,6 +170,6 @@ def load(dirpath):
         adapt_validation_error(exc)
 
     # inject project's config
-    content['project'] = _Project(dirpath=dirpath)
+    content['project'] = Project(dirpath=dirpath)
 
-    return _Config(**content)
+    return Config(**content)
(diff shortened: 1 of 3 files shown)


charmcraft/config.py
```diff
@@ -0,0 +1,175 @@
...
+
+@attr.s(kw_only=True, frozen=True)
+class _Project:
+    """Configuration for all project-related options, used internally."""
+
+    dirpath = attr.ib(default=None)
+
+
+@attr.s(kw_only=True, frozen=True)
+class _Config:
```
cjp256, line 113: out of curiosity why lead with '_'?

Snapcraft is all over the place with its naming convention, but generally I'd expect if a class or function is meant to be used external to the module, it should not be labeled as such.  And I'd expect there to be other usage of Config, Project, etc. but maybe not :D

tests/test_config.py
```diff
@@ -0,0 +1,299 @@
...
+        test_file.write_text(text)
+        return tmp_path
+    return create_config
+
+
+# -- tests for the config loading
+
+def test_load_current_directory(create_config, monkeypatch):
+    """Init the config using charmcraft.yaml in current directory."""
+    tmp_path = create_config("""
```
cjp256, line 48: FWIW Over on Snapcraft, we heavily make use textwrap.dedent() to make sure the text blocks have correct alignment.

tests/test_config.py
```diff
@@ -0,0 +1,299 @@
...
+
+def test_schema_type_mandatory(create_config, check_schema_error):
+    """Schema validation, type is mandatory."""
+    create_config("""
+        someconfig: None
+    """)
+    if is_py35:
+        check_schema_error(
+            "Bad charmcraft.yaml content; missing fields: type.",
+            "Bad charmcraft.yaml content; the 'type' field must be one of: 'charm', 'bundle'.",
```
cjp256, line 108: So do this `the 'type' field must be one of: 'charm', 'bundle'` not surface on 3.6+?

tests/test_config.py
```diff
@@ -0,0 +1,299 @@
...
+from charmcraft.config import (
+    _BasicPrime,
+    _CharmhubConfig,
+    _check_url,
+    load,
+)
+
+# Decide if we're using a Python 3.5 or older to support a jsonschema detail that uses
+# randomly ordered dictionaries
+is_py35 = (sys.version_info.major, sys.version_info.minor) < (3, 6)
```
cjp256, line 31: Just a nit, but I intuitively expected this to be `python == 3.5` or a compat level (py >= 3.5).

charmcraft/config.py
facundobatista: My idea was to indicate the possible developers that they are not expected to instantiate that class directly, but using the module's `load` function (which receives the directory where the config is, and stores that in the config itself, a value that is critical for some commands).

tests/test_config.py
facundobatista: Yes, used it in other places where text MUST start in column 0 (like help texts, init code templates, etc). But for YAMLs there is no need of that, so code gets cleaner.

tests/test_config.py
facundobatista: No. In 3.6+, if you don't include the `type` field it errors out indicating (consistently, not randomly like in 3.5) that the field is missing. The problem is that in 3.5 it fails sometimes with one error, sometimes with the other one. Note that this is internal to jsonschema.

tests/test_config.py
facundobatista: I tried to code it generically like "3.5 or older", but then the name `is_py35` was more clear in the rest of the code. I can change it, I guess this will never run in py3.4 or 3.3, etc....

charmcraft/config.py
```diff
@@ -0,0 +1,175 @@
...
+CONFIG_SCHEMA = {
+    'type': 'object',
+    'properties': {
+        'type': {'type': 'string', 'enum': ['charm', 'bundle']},
+        'charmhub': {
+            'type': 'object',
+            'properties': {
+                'api_url': {'type': 'string', 'format': 'url'},
+                'storage_url': {'type': 'string', 'format': 'url'},
+            },
```
cjp256, line 134: You may want `'additionalProperties': False,`  for each json object to ensure it is supported.

charmcraft/config.py
cjp256: If anyone outside the module is going to use an interface, it should not be marked as internal.

Since load() is returning a Config object, presumably for external use, the Config object itself should be public.

FWIW load() may be encapsulated as a classmethod in Config if it's strictly to init the Config.

Also, consider the case where you may want to create a Config() object without a backing yaml (helpful in tests).

facundobatista: It was originally a classmethod, I was requested to decouple it from the class in the draft branch. In any case, it's not a problem for the tests.

I'll make the classes public, though.

charmcraft/config.py
facundobatista: Good catch! Added.

charmcraft/config.py
```diff
@@ -0,0 +1,176 @@
...
+            'additionalProperties': False,
+        },
+        'parts': {
+            'type': 'object',
+            'properties': {
+                'bundle': {
+                    'type': 'object',
+                    'properties': {
+                        'prime': {'type': 'array'},
```
sergiusens, line 143: shouldn't there be some checks for minItems and types for those items (string)?

cjp256 (dismissed):
LGTM

charmcraft/config.py
facundobatista: Added control for these URLs! not only type, but also format (which is guarded now by the schema, and not by the class).

Regarding minItems, I don't see a problem of having an empty list there.

Anyway, this detail can be improved/fixed later if we decide to.

cjp256 (approved):
LGTM

PR #222, opened 2021-01-14
Modernize the Bash completion definition.

facundobatista:
The sad thing is that there's no possible to add a test that checks that we forgot to update this file after adding a new command...


sergiusens (approved):
one TODO here is to shellcheck this file

cjp256 (approved):
LGTM

PR #225, opened 2021-01-19
Better init template outputs: correct license and not real actions/config contents

facundobatista:


charmcraft/templates/init/src/charm.py.j2
```diff
@@ -25,12 +25,16 @@ class {{ class_name }}(CharmBase):
         self._stored.set_default(things=[])
 
     def _on_config_changed(self, _):
+        # Note: you need to uncomment the example in the config.yaml file for this to work (ensure
```
sergiusens, line 28: This is a tough call, perhaps it might be better to comment the code block entirely so it is a noop?

facundobatista: As we just talked, I'll leave this so project tests work fine with just the YAML files "restored".

tests/commands/test_init.py
```diff
@@ -98,6 +98,21 @@ def test_executables(tmp_path, config):
     assert (tmp_path / "src/charm.py").stat().st_mode & S_IXALL == S_IXALL
 
 
+def uncomment(filepath, token):
+    """Uncomment a file after and including the line with the given token."""
+    new_content = []
+    with filepath.open('rt', encoding='utf8') as fh:
```
sergiusens, line 104: not that it matter, but I like using fileinput for these inplace replacements.

sergiusens (approved):

PR #226, opened 2021-01-21
Pass the config to all commands.

facundobatista:
Now the main module passes the config to all the commands.

The branch is a little too big, but most of the diff are mechanical changes (as most of the commands don't use the config, at least so far).

But there are two commands that are impacted by this change:

- `pack`: which previously loaded/validated the `charmcraft.yaml` file itself, now it just uses the config.

- `init`: which had a parameter to identify the project dir, and now this parameter is global to all the system.

### DIFF
--- a/tests/commands/test_pack.py
+++ b/tests/commands/test_pack.py
@@ -1,4 +1,4 @@
-# Copyright 2020 Canonical Ltd.
+# Copyright 2020-2021 Canonical Ltd.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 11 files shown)


cmatsuoka (dismissed):
Thanks, it looks good.

tests/commands/test_pack.py
```diff
@@ -31,6 +31,9 @@
     get_paths_to_include,
 )
 
+# empty namespace
+noargs = Namespace()
```
cjp256, line 35: I'm not a huge fan of doing this to save a couple lines, but it is harmless.

cjp256 (approved):
LGTM

cmatsuoka (approved):

PR #227, opened 2021-01-22
Pass the charmhub config to the Store API layer.

facundobatista:
Almost all are mechanic changes, the functional impact is super low (the config still doesn't reach the store client, which is the one that changes).

Also moved a test file (which I forgot to move when moved the real `utils.py` file).


cjp256 (approved):
LGTM

cmatsuoka (approved):
Thanks, this seems good.

PR #229, opened 2021-01-25
Make the Store client to use the Config (with production default URLs now).

facundobatista:
Note: I had to change the behaviour when the `charmcraft.yaml` file is missing (so far is optional, in the future will be mandatory): previously in that case the whole config was None, but this way the Store URLs wouldn't get to the Store client, or the project directory wouldn't get to the `init` command.

Now the config object will always be there and useful, but with `type` in None. This is fine because the only ones using `type` are the bundle related commands (only `pack` so far, the next coming in the following days), for which the config is also mandatory. 

So if there is no charmcraft.yaml file at all the  config is created and passed around for all objects just fine and `pack` detects that config.type is None, so it complains that config is mandatory (like before).

In the future (following weeks), when we make the config mandatory, we'll put back that `type` is mandatory in the schema.

### DIFF
--- a/charmcraft/commands/pack.py
+++ b/charmcraft/commands/pack.py
@@ -79,7 +79,7 @@ class PackCommand(BaseCommand):
 
     def run(self, parsed_args):
         """Run the command."""
-        if self.config is None:
+        if self.config.type is None:
             raise CommandError(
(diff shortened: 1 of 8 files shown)


cmatsuoka (approved):
Thanks Facundo, lgtm.

PR #236, opened 2021-02-05
Fix infrastructure around the concept that the config is not always mandatory.

facundobatista:
This is because it will be mandatory for the more building-related commands, but not for the more store-related ones.

So, now:

- each command indicates if config is mandatory for it, the infrastructure will check that, so the for the command is safe to assume it's always there (see change in `pack.py`).

- if the config was provided by the user is now an attribute of the Project in config, so the infrastructure (at least) can check it.

- if the config is provided, it's verified with the schema; otherwise everything is bootstrapped by default.

- `type` is back to mandatory if the config is provided.

### DIFF
--- a/charmcraft/cmdbase.py
+++ b/charmcraft/cmdbase.py
@@ -42,6 +42,7 @@ class BaseCommand:
     - name: the identifier in the command line
     - help_msg: a one line help for user documentation
     - common: if it's a common/starter command, which are prioritized in the help
+    - needs_config: will ensure a config is provided when executing the command
 
     It also must/can override some methods for the proper command behaviour (see each
     method's docstring).
@@ -55,6 +56,7 @@ class BaseCommand:
     help_msg = None
     overview = None
     common = False
+    needs_config = False
 
     def __init__(self, group, config):
         self.group = group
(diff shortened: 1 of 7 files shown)


cjp256 (commented):
Out of curiosity, what's the difference if config_provided is True or False, if the config will have the same defaults either way?

facundobatista:
> Out of curiosity, what's the difference if config_provided is True or False, if the config will have the same defaults either way?

The `config_provided` flag is used by the "framework" to avoid executing a command if that commands needs the config and it wasn't provided. In the future it may have more uses, but so far just that.

The defaults are the same in one case or the other, but if the config is provided, the schema is verified. So, see how `pack` command operates, then: it declares that needs the config, so it knows that whenever it's executed, the config will be there; furthermore the config schema indicates that `type` is mandatory (restricting its value, even), so for the command the attribute always be there.

In short, all this mechanism makes the config robust and well validated the moment it gets to the command, so there's a LOT of less checking on command side.

cmatsuoka (approved):
Thanks Facundo, lgtm.

cjp256 (approved):

PR #240, opened 2021-02-10
Handling names for bundles: register (new command) and list (generalized previous one).

facundobatista:
Also took the opportunity for a couple of small improvements:

- empty line after commands help text is superfluous (the formatter takes care of that, so I changed those that were inconsistent)

- a `meta` was redundantly used in a argparse option

### DIFF
--- a/charmcraft/commands/store/__init__.py
+++ b/charmcraft/commands/store/__init__.py
@@ -36,6 +36,10 @@ from .store import Store
 
 logger = logging.getLogger('charmcraft.commands.store')
 
+# entity types
+CHARM = 'charm'
+BUNDLE = 'bundle'
+
 LibData = namedtuple(
     'LibData', 'lib_id api patch content content_hash full_name path lib_name charm_name')
 
@@ -67,7 +71,6 @@ class LoginCommand(BaseCommand):
         from your local system, especially in a shared environment.
 
         See also `charmcraft whoami` to verify that you are logged in.
-
     """)
 
     def run(self, parsed_args):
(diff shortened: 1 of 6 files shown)


charmcraft/commands/store/__init__.py
```diff
@@ -36,6 +36,10 @@
 
 logger = logging.getLogger('charmcraft.commands.store')
 
+# entity types
+CHARM = 'charm'
```
sergiusens, line 40: Random suggestion (might need a better name that TYPES)

```python
TYPES = namedtuple("Types", "CHARM BUNDLE")(CHARM="charm", BUNDLE="bundle")
```

sergiusens:
There are also conflicts

charmcraft/commands/store/__init__.py
```diff
@@ -162,28 +163,64 @@ def fill_parser(self, parser):
...
+        release them for wider consumption.
+
+        Charmhub operates on the 'principle of least surprise' with regard
+        to naming. A bundle with a well-known name should provide the best
+        system for the service most people associate with that name.  Bundles
+        can be renamed in the Charmhub, but we would nonetheless ask
+        you to use a qualified name, such as `yourname-bundlename` if you are
+        in any doubt about your ability to meet that standard.
+
+        We discuss registrations in Charmhub Discourse:
```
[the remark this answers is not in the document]
facundobatista: Fixed here and in the `register`'s help text too.

charmcraft/commands/store/__init__.py
facundobatista: I considered using an Enum, but it was overkilling (and didn't fit ok). Using a namedtuple here looks like a hack, but fits perfectly with the usage.

I'd drop the uppercases as this is not longer a "const", but more a class, same with attributes.

Improved! Thanks.

sergiusens (approved):

sergiusens (approved):

Issue #128, opened 2020-08-19
Handle bundles when interacting with the store


facundobatista:
We need to handle bundles properly in all the operations related to the store.

Most of that should be ok already, but for sure we need to improve some details:

- in the `register` command, give the option to register the name for a bundle. The idea is to provide a `--type` option that will allow the `charm` and `bundle` values (and possibly more in the future, like `stack`); this `--type` can be avoided if the project from where the command is issued can be identified as one or the other, otherwise it's mandatory.

- in the `names` command, add a column with the type of the package.

- in the `build` command, we should support a `metadata.yaml` not being there but a `bundle.yaml`: in that case just build a zip named `.bundle` with everything that is there (pretty much like we do for charms) but without building dependencies, or creating hooks/dispatch, etc. So in reality build will have two modes ("charm" and "bundle"), decided automatically after what is found in the project. Note also that some options are valid for the different modes (like `--from`) but others are not (like `--entrypoint`): in the latter cases we need to raise an error.

- in the `upload` command, we need to rename current `--charm-file` to be `--file` (supporting both charm and bundle files), and improve what is automatically detected from the project where the command is issued (we can guess the name from a possible `metadata.yaml`, for example, but not from a `bundle.yaml`).

facundobatista:
We went to a more explicit model regarding the commands...

These are the specific ones:

- `register-bundle`
- `upload-bundle`
- `release-bundle`

These are still under study (if to be shared, or new ones be created):

- `revisions`
- `status`

OTOH, `names` if for "everything registered by the user", so it makes totally sense to work for "everything".

hmlanigan:
Can we have a bundle init like we have a charm init?

facundobatista:
Everything landed. Still to work on some docs.

facundobatista:
> Can we have a bundle init like we have a charm init?

No really needed, as normally we ask Juju to create the bundles, [see here](https://juju.is/docs/charm-bundles).

facundobatista:
We got [docs](https://charmhub.io/tutorials/how-to-work-with-bundles-in-charmcraft#1-introduction).

PR #246, opened 2021-02-18
Support for bundles in the Upload command.

### LINKED ISSUE #77: Deduce the charm's name from the file's included metadata, not its name
Currently in the upload command, if a charm binary is pointed out to upload, we deduce the charm's name from the pointed out binary name.

We should open that binary (unzip it), get its included `metadata.yaml`, parse it, and get the name from there.

facundobatista:
Part of #128.

Fixes #77.

### DIFF
--- a/charmcraft/commands/store/__init__.py
+++ b/charmcraft/commands/store/__init__.py
@@ -250,7 +250,7 @@ class ListNamesCommand(BaseCommand):
 def get_name_from_zip(filepath):
     """Get the charm/bundle name from a zip file."""
     try:
-        zf = zipfile.ZipFile(filepath)
+        zf = zipfile.ZipFile(str(filepath))
     except zipfile.BadZipFile:
         raise CommandError("Cannot open {!r} (bad zip file).".format(str(filepath)))
 
--- a/charmcraft/commands/store/__init__.py
+++ b/charmcraft/commands/store/__init__.py
@@ -19,7 +19,6 @@
 import ast
 import hashlib
 import logging
-import os
 import pathlib
 import string
 import textwrap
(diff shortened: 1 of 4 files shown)


charmcraft/commands/store/__init__.py
```diff
@@ -247,75 +248,66 @@ def run(self, parsed_args):
...
     overview = textwrap.dedent("""
-        Upload a charm to Charmhub.
+        Upload a charm or bundle to Charmhub.
 
-        Push a charm to Charmhub where it will be verified for conformance
-        to the packaging standard. This command will finish successfully
-        once the package is approved by Charmhub.
+        Push a charm or bundle to Charmhub where it will be verified for
+        conformance to the packaging standard. This command will finish
+        successfully once the package is approved by Charmhub.
```
sergiusens, line 292:
```suggestion
        Push a charm or bundle to Charmhub where it will be verified.
        This command will finish successfully once the package is
        approved by Charmhub.
```

charmcraft/utils.py
```diff
@@ -64,3 +67,17 @@ def get_templates_environment(templates_dir):
...
         undefined=StrictUndefined)   # fail on undefined
     return env
+
+
+def useful_filepath(filepath):
+    """Convert the string to Path and verify that is a useful file path.
+
+    It checks that the file exists and it's readable, and that it's actually a
+    file. Also the `~` is expanded to the user's home.
+    """
```
sergiusens, line 77:
```suggestion
    """Return a valid Path with user name expansion for filepath.

    CommandError is raised if filepath is not a valid file or is not readable.
    """
```

sergiusens (dismissed):
Looks good, just a couple of docstring suggestions

charmcraft/commands/store/__init__.py
facundobatista: improved

charmcraft/utils.py
facundobatista: better, thanks!

sergiusens (approved):

sergiusens (approved):

PR #248, opened 2021-02-22
Final commands changes for Bundles support.

facundobatista:
Explained changes (UX changes accoording to [this doc](https://docs.google.com/document/d/1dqzoFLEaB_1N_jJJdmsoCxeLUtTZRakveXF8aHx6lFA/)):

- `revisions` and `status`: now `name` is mandatory (no longer guessed from metadata)

- `release` now gets `name` as first mandatory parameter, then `--revision` as a mandatory option (only one allowed!) and `--channel` as mandatory option (multiple allowed).

### DIFF
--- a/charmcraft/commands/store/__init__.py
+++ b/charmcraft/commands/store/__init__.py
@@ -30,7 +30,7 @@ import yaml
 from tabulate import tabulate
 
 from charmcraft.cmdbase import BaseCommand, CommandError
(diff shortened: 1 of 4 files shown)


cjp256 (approved):
LGTM, though I'm curious what `cm` means in the context of pytest.raises() (i.e. ExceptionInfo)?

cmatsuoka (approved):
Thanks Facundo, lgtm.

PR #252, opened 2021-03-02
Command to list resources

facundobatista:


tests/commands/test_store_api.py
```diff
@@ -764,3 +764,75 @@ def test_get_tips_query_combinations(client_mock, config):
...
+        }
+    ]}
+
+    store = Store(config.charmhub)
+    result = store.list_resources('some-name')
+
+    (item1, item2) = result
+
+    assert item1.name == 'testresource1'
+    assert item1.optional
```
cjp256, line 831: Perhaps `assert item1.optional is True` ?

cjp256 (approved):

tests/commands/test_store_api.py
facundobatista: It's what `assert` itself is doing, validating that `item1.optional` is true...

PR #253, opened 2021-03-02
Support for resource uploading

facundobatista:


charmcraft/commands/store/store.py
```diff
@@ -144,11 +144,9 @@ def list_registered_names(self):
                 entity_type=item['type']))
         return result
 
-    def upload(self, name, filepath):
-        """Upload the content of filepath to the indicated charm."""
+    def _upload(self, endpoint, filepath):
+        """Uploadfor all charms, bundles, resources (generic process)."""
```
cmatsuoka, line 148: A typo here? Perhaps this should be "Upload routine for..." (or even "Common upload routine for..."?)

tests/commands/test_store_api.py
```diff
@@ -282,6 +280,31 @@ def test_upload_error(client_mock, config):
     assert error2.code == "error-code-2"
 
 
+def test_upload_charmbundles_endpoint(config):
+    """Upload for charm or bundles prepares ok the endpoint and call generic _upload."""
```
cmatsuoka, line 284: I'm a bit confused with this description, perhaps we could use something like "Test charm or bundles upload using a mock endpoint"?

tests/commands/test_store_api.py
```diff
@@ -282,6 +280,31 @@ def test_upload_error(client_mock, config):
...
+
+    with patch.object(store, '_upload') as mock:
+        mock.return_value = test_results
+        result = store.upload('test-charm', 'test-filepath')
+    mock.assert_called_once_with('/v1/charm/test-charm/revisions', 'test-filepath')
+    assert result == test_results
+
+
+def test_upload_resources_endpoint(config):
+    """Upload for resources prepares ok the endpoint and call generic _upload."""
```
cmatsuoka, line 296: Same thing here.

cmatsuoka (approved):
Thanks Facundo. It looks good, only added a few comments on docstring texts.

charmcraft/commands/store/store.py
facundobatista: I actually wrote "Generic upload..." initially, but PEP257 complained that it was not a verb in imperative...

Improved it a little, though.

tests/commands/test_store_api.py
facundobatista: There's a redundancy when starting the tests docstrings with "Test" or "Check"... it's obvious that the test function is to test something, so I start these docstrings with what *the tested code* should do.

Improved the docstring anyway.

tests/commands/test_store_api.py
facundobatista: Same.

charmcraft/commands/store/store.py
cmatsuoka: I think the PEP257 checker will complain a lot about my docstrings, I'll add it to my lint workflow.

tests/commands/test_store_api.py
cmatsuoka: Ah I see, that's a good guideline to follow. The part I had problems to parse was the "prepares ok the endpoint", but it's now clear what it's doing there. Perhaps the actual phrasing could be improved in a later change if you feel like it.

cmatsuoka (approved):

PR #254, opened 2021-03-03
Support for listing resources revisions.

facundobatista:


charmcraft/commands/store/__init__.py
```diff
@@ -1051,3 +1051,51 @@ def run(self, parsed_args):
...
+    name = 'resource-revisions'
+    help_msg = "List revisions for a resource associated to a charm in Charmhub"
+    overview = textwrap.dedent("""
+        Show size and date for each resource revision in Charmhub.
+
+        For example:
+
+           $ charmcraft resource-revisions
+           Revision    Created at     Size
+           1           2020-11-15   183151
```
cjp256, line 1068: Is size in bytes or kb?  Would be nice to indicate the unit somewhere.

cjp256 (approved):

charmcraft/commands/store/__init__.py
facundobatista: As discussed, will show these values using the `humanize` lib.

cjp256: LGTM, but the doc is out of date now :D

cjp256 (approved):

sergiusens (approved):

cjp256 (approved):

PR #259, opened 2021-03-05
Support for attaching resources when releasing charms.

facundobatista:


charmcraft/commands/store/__init__.py
```diff
@@ -420,14 +433,25 @@ def fill_parser(self, parser):
         parser.add_argument(
             '-c', '--channel', action='append', required=True,
             help="The channel(s) to release to (this option can be indicated multiple times)")
+        parser.add_argument(
+            '--resource', action='append', type=ResourceOption(), default=[],
+            help=(
+                "The resource(s)s to attach to to the release, in the <name>:<revision> format "
```
cmatsuoka, line 439: There's an extra "s" here in resource(s)s.

cmatsuoka (commented):
Thanks Facundo, lgtm. Added a note about a typo in a help message.

charmcraft/commands/store/__init__.py
facundobatista: Thanks! will fix it in the next PR.

cjp256 (approved):
lgtm

cmatsuoka (approved):
Thank you.

PR #262, opened 2021-03-09
Support resources when showing charm status.

### LINKED ISSUE #176: Support for charms' resources
We need to support the resources of the charms, interacting with the Store properly for all these related operations.

In detail:

- there's no "resource registration", as resources are declared in the `metadata.yaml` file (after a charm is uploaded indicating that has some resources in that file, the Store will allow the author to manage those resources).

- list the resources for the charm (using store's [`list_resources`](https://api.staging.charmhub.io/docs/default.html#list_resources) endpoint).

- upload a new resource (using store's [`push_resource`](https://api.staging.charmhub.io/docs/default.html#push_resource) endpoint. 

- to release resources and list those released resources, we need to improve the `release` command (as resources are incorporated in the [release](https://api.staging.charmhub.io/docs/default.html#release) endpoint) and the `status` command (because now the [list_releases](https://api.staging.charmhub.io/docs/default.html#list_releases) endpoint includes the resources information in the channelmap field).

facundobatista:
This is the final PR of the sequence. Fixes #176.

### DIFF
--- a/tests/commands/test_store_commands.py
+++ b/tests/commands/test_store_commands.py
@@ -1087,8 +1087,8 @@ def test_status_with_resources(caplog, store_mock, config):
     expected = [
         "Track    Channel    Version    Revision    Resources",
         "latest   stable     -          -           -",
-        "         candidate  5.1        5           resource1 r1, resource2 r54",
-        "         beta       5.1        5           resource1 r1",
+        "         candidate  5.1        5           resource1 (r1), resource2 (r54)",
+        "         beta       5.1        5           resource1 (r1)",
         "         edge       ↑          ↑           ↑",
     ]
     assert expected == [rec.message for rec in caplog.records]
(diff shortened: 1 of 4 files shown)


cjp256 (approved):

tests/commands/test_store_commands.py
```diff
@@ -1063,6 +1065,73 @@ def test_status_with_multiple_branches(caplog, store_mock, config):
...
+    revisions = [
+        _build_revision(revno=5, version='5.1'),
+    ]
+    store_mock.list_releases.return_value = (channel_map, channels, revisions)
+
+    args = Namespace(name='testcharm')
+    StatusCommand('group', config).run(args)
+
+    expected = [
+        "Track    Channel    Version    Revision    Resources",
```
sergiusens, line 1088: not related speficically two this, but the separation between headers/cloumns is supposed to be 2 spaces (https://discourse.ubuntu.com/t/tabular-data/18886)

tests/commands/test_store_commands.py
```diff
@@ -1063,6 +1065,73 @@ def test_status_with_multiple_branches(caplog, store_mock, config):
...
+        "latest   stable     -          -           -",
+        "         candidate  5.1        5           resource1 r1, resource2 r54",
```
sergiusens, line 1090: resource listing can get incredibly long here, there is a supposed limitation of 80 chars per table.

Without addressing the long list, does this look better?
```suggestion
        "         candidate  5.1        5           resource1 (r1), resource2 (r54)",
```

tests/commands/test_store_commands.py
facundobatista: Opened #264 to fix all tables at once... we may need to stop using `tabulate` for this.

tests/commands/test_store_commands.py
facundobatista: Improved it using the parenthesis, let's talk about line length.

PR #266, opened 2021-03-10
Mutate - to _ to make importable names

### LINKED ISSUE #265: Creating a charm library can create an invalid import
If you have a charm with '-' in the name, and run `charmcraft create-lib`, the path that is created will contain a '-', which is an invalid name in a python import.

Example:
Charm called 'harbor-registry' in it's metadata.yaml
`charmcraft create-lib harborregistrylib`

Path created is: `lib/charms/harbor-registry/v0/harborregistrylib.py`.

An import of `import charms.harbor-registry.v0.harborregistrylib` is not valid python.

tomwardill:
Using '-' in a name means that the charmlib will have an un-importable path.

Mutate that to '_', and use that in places where we need to. This can be reversed from the filepath as '_' is not valid in a charm name (via https://juju.is/docs/charm-metadata )

Closes #265

### DIFF
--- a/charmcraft/commands/store/__init__.py
+++ b/charmcraft/commands/store/__init__.py
@@ -719,6 +719,8 @@ def _get_libs_from_tree(charm_name=None):
 
     It only follows/uses the the directories/files for a correct charmlibs
     disk structure.
+
+    This can take charm_name as both importable and normal form.
     """
     local_libs_data = []
 
@@ -726,8 +728,8 @@ def _get_libs_from_tree(charm_name=None):
         base_dir = pathlib.Path('lib') / 'charms'
         charm_dirs = sorted(base_dir.iterdir()) if base_dir.is_dir() else []
     else:
-        charm_name = create_importable_name(charm_name)
-        base_dir = pathlib.Path('lib') / 'charms' / charm_name
+        importable_charm_name = create_importable_name(charm_name)
+        base_dir = pathlib.Path('lib') / 'charms' / importable_charm_name
         charm_dirs = [base_dir] if base_dir.is_dir() else []
 
     for charm_dir in charm_dirs:
--- a/charmcraft/commands/store/__init__.py
+++ b/charmcraft/commands/store/__init__.py
@@ -881,7 +881,7 @@ class PublishLibCommand(BaseCommand):
 
         for lib_data in to_publish:
             store.create_library_revision(
-                charm_name, lib_data.lib_id, lib_data.api, lib_data.patch,
+                lib_data.charm_name, lib_data.lib_id, lib_data.api, lib_data.patch,
                 lib_data.content, lib_data.content_hash)
             logger.info(
                 "Library %s sent to the store with version %d.%d",
--- a/charmcraft/commands/store/__init__.py
+++ b/charmcraft/commands/store/__init__.py
@@ -598,32 +598,40 @@ def _get_positive_int(raw_value):
 
 
 def _get_lib_info(*, full_name=None, lib_path=None):
-    """Get the whole lib info from the path/file."""
+    """Get the whole lib info from the path/file.
+
+    This will perform mutation of the charm name to create importable paths.
+    * `charm_name` and `libdata.charm_name`: `foo-bar`
+    * `full_name` and `libdata.full_name`: `charms.foo_bar.v0.somelib`
+    * paths, including `libdata.path`: `lib/charms/foo_bar/v0/somelib`
+
+    """
     if full_name is None:
         # get it from the lib_path
         try:
-            libsdir, charmsdir, charm_name, v_api = lib_path.parts[:-1]
+            libsdir, charmsdir, importable_charm_name, v_api = lib_path.parts[:-1]
         except ValueError:
             raise _BadLibraryPathError(lib_path)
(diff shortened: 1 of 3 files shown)


tomwardill:
> Would it be possible to forbid "-" in the name instead to avoid potential aliasing?

https://juju.is/docs/charm-metadata says that '-' is valid in a charm name. We could drop it from the library name, but then we have the same problem of not being able to reverse it.

charmcraft/commands/store/__init__.py
```diff
@@ -861,7 +881,7 @@ def run(self, parsed_args):
 
         for lib_data in to_publish:
             store.create_library_revision(
-                lib_data.charm_name, lib_data.lib_id, lib_data.api, lib_data.patch,
+                charm_name, lib_data.lib_id, lib_data.api, lib_data.patch,
```
facundobatista, line 884: But `lib_data.charm_name` should be equal to `charm_name`, right? this is even validated above..

tests/commands/test_store_commands.py
```diff
@@ -1726,6 +1777,35 @@ def test_fetchlib_simple_downloaded(caplog, store_mock, tmp_path, monkeypatch, c
     assert saved_file.read_text() == lib_content
 
 
+def test_fetchlib_simple_dash_in_name(caplog, store_mock, tmp_path, monkeypatch, config):
```
facundobatista, line 1780: Please add a test for "fetchlib" using a path already in disk with underscore (NOT passing the `library` as a parameter, so it needs to pickup what's on disk to know what to fetch)

facundobatista (changes requested):
Awesome work, thank you very much!

I pointed a couple of details in the code, any doubt ping me in Mattermost or we can discuss them here.

Thanks!!!

charmcraft/commands/store/__init__.py
```diff
@@ -610,6 +621,10 @@ def _get_lib_info(*, full_name=None, lib_path=None):
             raise _BadLibraryNameError(full_name)
         lib_path = pathlib.Path('lib') / charmsdir / charm_name / v_api / (libfile + '.py')
 
+    # charm names in the path can contain '_' to be importable
+    # these should be '-', so change them back
+    charm_name = create_charm_name_from_importable(charm_name)
```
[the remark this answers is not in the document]
tomwardill: My assumption here was that the fully qualified name would also be mutated to '_'. This would mean that full name would look like 'charms.test_charm.v0.testlib', and therefore needs transforming also.
Otherwise the fully qualified name is also an invalid import.

charmcraft/commands/store/__init__.py
tomwardill: Ah yes, this is a mistake from an earlier iteration. Removed.

charmcraft/commands/store/__init__.py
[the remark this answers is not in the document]
facundobatista: Ah, you're right!!!!

It gets a little confusing.

Let me ask you to enhance the code with two things.

First, put the "rules" in `_get_lib_info`'s docstring, something like:

- `charm_name` (and `libdata.charm_name`): `foo-bar`
- `full_name` (and `libdata.full_name`): `charms.foo_bar.v0.somelib`
- any path (specially `libdata.path`): `lib/charms/foo_bar/v0/somelib`

Second, please rename `charm_name` to `importable_charm_name` where it corresponds. So it will be explicit that in the "if blocks" we're handling `importable_charm_name`, it won't be mixed afterwards, and even your conversion line is more straightforward:

    charm_name = create_charm_name_from_importable(importable_charm_name)

charmcraft/commands/store/__init__.py
```diff
@@ -703,6 +726,7 @@ def _get_libs_from_tree(charm_name=None):
         base_dir = pathlib.Path('lib') / 'charms'
         charm_dirs = sorted(base_dir.iterdir()) if base_dir.is_dir() else []
     else:
+        charm_name = create_importable_name(charm_name)
```
facundobatista, line 729: Use `importable_charm_name` as the result here and in the rest of the function (for example, what it receives is importable_charm_name, too...

facundobatista (changes requested):
Wonderfult. We're almost there, it's missing an extra conversion to "importable_charm_name", thanks!!!

charmcraft/commands/store/__init__.py
tomwardill: I've changed the result of this to be `importable_charm_name`, and where it is used to build the directory path. 
The argument to the function I've left as `charm_name`, as theoretically this function could be used with either form, since the result of transforming an already transformed name is the same result.

facundobatista (approved):
Wonderful, thanks!!

charmcraft/commands/store/__init__.py
```diff
@@ -60,6 +60,17 @@ def get_name_from_metadata():
...
 
 
+def create_importable_name(charm_name):
+    """Convert a charm name to something that is importable in python."""
+    return charm_name.replace('-', '_')
+
+
+def create_charm_name_from_importable(charm_name):
+    """Convert a charm name from the importable form to the real form."""
+    # _ is invalid in charm names, so we know it's intended to be '-'
```
sergiusens, line 70: :+1:

sergiusens (approved):

Issue #267, opened 2021-03-10
init doesn't set a charm status


chris-sanders:
The charm that's produced doesn't set charm status, which is something all charms should be doing. It would be nice if we include it in the init to get developers in the mindset of reporting charm status.

PR #268, opened 2021-03-10
Set unit status in default charm

chris-sanders:
This addresses https://github.com/canonical/charmcraft/issues/267 by adding a call to set the status in the init template for the charm.


sergiusens (approved):

facundobatista (approved):
I like it, thanks!

PR #272, opened 2021-03-16
Add lib path to run_tests template

tomwardill:
If you added a library to your charm, the tests would not run as it is not in the PYTHONPATH provided by the `run_tests` script.

Add it, so that tests will run.


facundobatista (approved):
Awesome, thanks!

sergiusens (approved):

PR #279, opened 2021-03-29
Produce a manifest when building charms and bundles.

facundobatista:


charmcraft/utils.py
```diff
@@ -136,3 +140,52 @@ def useful_filepath(filepath):
...
+    """
+    os_platform = get_os_platform()
+
+    # XXX Facundo 2021-03-29: the architectures list will be provided by the caller when
+    # we integrate lifecycle lib in future branches
+    architectures = [os_platform.machine]
+
+    content = {
+        'charmcraft-version': __version__,
+        'charmcraft-started-at': started_at.isoformat() + "Z",
```
cmatsuoka, line 181: It seems that we're converting from ISO 8601 to RFC3339 date formats here, is the original `started_at` time UTC+0 so we can safely say it's Zulu time?

charmcraft/utils.py
```diff
@@ -136,3 +140,52 @@ def useful_filepath(filepath):
...
+        'charmcraft-os-release-name': os_platform.system,
+        'charmcraft-os-release-version-id': os_platform.release,
+        'architectures': architectures,
+    }
+    filepath = basedir / 'manifest.yaml'
+    if filepath.exists():
+        raise CommandError(
+            "Cannot write the manifest as there is already a 'manifest.yaml' in disk.")
```
cmatsuoka, line 189: Another common way to express this is the classic `cannot write 'foo.txt': file exists"`. (But if the rest of the error messages is in a more verbose style, it's better to stay consistent)

PR #280, opened 2021-03-29
Added a helper to create a manifest

facundobatista:
This will be used later from the charm and bundle builders.

This PR is part of [this draft](https://github.com/canonical/charmcraft/pull/279).

### DIFF
--- a/charmcraft/utils.py
+++ b/charmcraft/utils.py
@@ -155,7 +155,9 @@ def get_os_platform(filepath=pathlib.Path("/etc/os-release")):
                 for line in f:
                     if "=" in line:
                         key, value = line.rstrip().split("=", 1)
-                        os_release[key] = value.strip('"')
+                        if value[0] == value[-1] == '"':
+                            value = value[1:-1]
+                        os_release[key] = value
         except FileNotFoundError:
             logger.debug("Unable to locate 'os-release' file, using default values")
         finally:
--- a/charmcraft/utils.py
+++ b/charmcraft/utils.py
@@ -19,16 +19,20 @@
 import logging
 import os
 import pathlib
+import platform
+from collections import namedtuple
 from stat import S_IXUSR, S_IXGRP, S_IXOTH, S_IRUSR, S_IRGRP, S_IROTH
 
 import attr
 import yaml
 from jinja2 import Environment, PackageLoader, StrictUndefined
 
+from charmcraft import __version__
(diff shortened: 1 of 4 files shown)


charmcraft/utils.py
```diff
@@ -136,3 +140,52 @@ def useful_filepath(filepath):
...
+    release = platform.release()
+    machine = platform.machine()
+
+    if system == "Linux":
+        os_release = {}
+        try:
+            with filepath.open("r", encoding='utf-8') as f:
+                for line in f:
+                    if "=" in line:
+                        key, value = line.rstrip().split("=", 1)
```
sergiusens, line 157: maybe catch ValueError (for missing =) and debug level log?

tests/test_utils.py
```diff
@@ -192,3 +200,73 @@ def test_usefulfilepath_not_a_file(tmp_path):
...
+        with patch('platform.system', return_value='Linux'):
+            os_platform = get_os_platform(filepath)
+    assert os_platform.system == "Ubuntu"
+    assert os_platform.release == "20.04"
+    assert os_platform.machine == "x86_64"
+
+
+def test_get_os_platform_windows():
+    """Get platform from a patched Windows machine"""
+    with patch('platform.system', return_value='Windows'):
```
sergiusens, line 236: curiosity, is there a preference to this over monkeypatch from pytest?

sergiusens (approved):

charmcraft/utils.py
sergiusens: An now I see the previous if

charmcraft/utils.py
```diff
@@ -136,3 +140,52 @@ def useful_filepath(filepath):
...
+                        os_release[key] = value.strip('"')
```
cjp256, line 158: Seems unlikely, but if `"` were to show up in a string, this would break it.

```suggestion
                        value = value.strip()
                        if value.startswith('"') and value.endswith('"'):
                            value = value[1:-1]
                        os_release[key] = value
```

cjp256 (approved):
LGTM, optional suggestion.  And maybe breaking the reading/parsing of os_release into its own function would be good.  I have a similar one in craft providers...  we could at least sync them :D

charmcraft/utils.py
facundobatista: Nice, I added several tests for this case and similar.

PR #281, opened 2021-04-05
Include a manifest in the built result when packing charms and bundles.

### LINKED ISSUE #273: Create a manifest at build time with building information
After packing charms or bundles, we need to leave a manifest.yaml in place (and include it in the packed file) with the following fields:

- charmcraft-version: the version of the executed charmcraft
- charmcraft-started-at: timestamp in ISO format for when the packing started
- charmcraft-os-release-id: OS name
- charmcraft-os-release-version-id: OS version
- architectures: the list of archs where the charm/bundle can run on

Example: 

```
charmcraft-version: 4.5.4
charmcraft-started-at: '2021-03-13T10:36:59.250861Z'
charmcraft-os-release-id: ubuntu
charmcraft-os-release-version-id: '18.04'
architectures:
- amd64
```

facundobatista:
This is a little too big branch, but a lot of it are repetitive test changes (because `Builder` now receives the config).

Fixes #273.


cjp256 (approved):
LGTM

tests/commands/test_pack.py
```diff
@@ -72,6 +73,14 @@ def test_simple_succesful_build(tmp_path, caplog, bundle_yaml, config):
     expected = "Created '{}'.".format(zipname)
     assert [expected] == [rec.message for rec in caplog.records]
 
+    # check the created manifest for these particular values that depend on given info
+    manifest = yaml.safe_load(zf.read('manifest.yaml'))
+    assert manifest['architectures'] == [platform.machine()]
```
sergiusens, line 78: I missed this in the previous PR, we want this to be the equivalent architecture entries for deb and snaps.

facundobatista: This was addressed in the previous branch.

cmatsuoka (approved):

PR #282, opened 2021-04-07
Have 'bases' in the manifest, with the translated architecture.

facundobatista:
This PR should land *before* #281 (which I just converted to "draft").

### DIFF
--- a/charmcraft/utils.py
+++ b/charmcraft/utils.py
@@ -172,7 +172,7 @@ def get_os_platform(filepath=pathlib.Path("/etc/os-release")):
                 if not line or line.startswith('#') or '=' not in line:
                     continue
                 key, value = line.rstrip().split("=", 1)
-                if value[0] == value[-1] == '"':
+                if value[0] == value[-1] and value[0] in ('"', "'"):
                     value = value[1:-1]
                 os_release[key] = value
             system = os_release.get("NAME", system)
--- a/charmcraft/utils.py
+++ b/charmcraft/utils.py
@@ -20,7 +20,6 @@ import logging
 import os
 import pathlib
 import platform
-from urllib.request import parse_keqv_list
 from collections import namedtuple
 from stat import S_IXUSR, S_IXGRP, S_IXOTH, S_IRUSR, S_IRGRP, S_IROTH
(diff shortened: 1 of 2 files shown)


sergiusens (approved):

charmcraft/utils.py
```diff
@@ -151,16 +163,12 @@ def get_os_platform(filepath=pathlib.Path("/etc/os-release")):
...
-                        key, value = line.rstrip().split("=", 1)
-                        if value[0] == value[-1] == '"':
-                            value = value[1:-1]
-                        os_release[key] = value
+            with filepath.open("rt", encoding='utf-8') as fh:
+                lines = [x.strip() for x in fh.readlines() if x.strip()]
         except FileNotFoundError:
             logger.debug("Unable to locate 'os-release' file, using default values")
         finally:
+            os_release = parse_keqv_list(lines)
```
cjp256, line 171: Just know that this would break if a commented line were in the os-release file.  Not that I'm aware of any cases where it happens, but it's allowed in the spec.

cjp256 (commented):
LGTM

charmcraft/utils.py
facundobatista: Indeed. I added that and other cases to the tests, and improved the code.

charmcraft/utils.py
```diff
@@ -149,18 +160,21 @@ def get_os_platform(filepath=pathlib.Path("/etc/os-release")):
...
             logger.debug("Unable to locate 'os-release' file, using default values")
-        finally:
+        else:
+            os_release = {}
+            for line in lines:
+                line = line.strip()
+                if not line or line.startswith('#') or '=' not in line:
+                    continue
+                key, value = line.rstrip().split("=", 1)
+                if value[0] == value[-1] == '"':
```
cjp256, line 175: single quotes are also allowed by spec.

cjp256 (approved):
LGTM

charmcraft/utils.py
facundobatista: Added a bunch of new tests for this case and a slight improvement here.

Issue #247, opened 2021-02-19
init and build: a complete and sane project bootstrap but not buildable


facundobatista:
`charmcraft init` should provide a complete and sane project: it should work in theory, and tests should pass.

At the same time, `charmcraft build` (or `pack` in the future) should not build a project which is just bootstrapped from `init` and not paid attention nor care from the developer.

To get to that situation, we will:

- fix templates left by `init`, so tests pass ok
- mark with a comment using an unusual token (e.g.: `TEMPLATE-TODO`) all those corners where the developer needs to work to get started a real project.
- check if any of those tokens are leftovers when `build`ing, and refuse to build if any of those are present (indicating which one(s))

mthaddon:
I think this may be an issue again. I just tried running `charmcraft init` and then deploying on rc11 and got:
```
mthaddon@tenaya:~/repos/k8s-charms$ mkdir charm-test
mthaddon@tenaya:~/repos/k8s-charms$ cd charm-test
mthaddon@tenaya:~/repos/k8s-charms/charm-test$ charmcraft init
Charm operator package file and directory tree initialized.
TODO: 

      README.md: Describe your charm in a few paragraphs of Markdown
      README.md: Provide high-level usage, such as required config or relations
   actions.yaml: change this example to suit your needs.
    config.yaml: change this example to suit your needs.
  metadata.yaml: fill out the charm's description
  metadata.yaml: fill out the charm's summary
  metadata.yaml: change to an Ubuntu series if not using k8s
   src/charm.py: change this example to suit your needs.
   src/charm.py: change this example to suit your needs.
mthaddon@tenaya:~/repos/k8s-charms/charm-test$ charmcraft build
Created 'charm-test.charm'.
mthaddon@tenaya:~/repos/k8s-charms/charm-test$ juju add-model charm-test
Added 'charm-test' model on microk8s/localhost with credential 'microk8s' for user 'admin'
mthaddon@tenaya:~/repos/k8s-charms/charm-test$ juju status
Model       Controller          Cloud/Region        Version   SLA          Timestamp
charm-test  microk8s-localhost  microk8s/localhost  2.9-rc11  unsupported  09:40:26+02:00

Model "admin/charm-test" is empty.
mthaddon@tenaya:~/repos/k8s-charms/charm-test$ juju deploy ./charm-test.charm
ERROR series with metadata v2 not valid
```

manadart:
This is the issue as it stands from Juju's perspective.

[...]

Issue #243, opened 2021-02-12
Initial tests from `charmcraft init` do not pass


tomwardill:
If you create a new charm with `charmcraft init`, the tests will not pass out of the box.

This appears to be because the `config.yaml` file exists, but does not contain any valid yaml (just comments).

```
tomwardill@microk8s-juju:~/work/harbor-charms/harbor-core$ ./run_tests 
test_action (tests.test_charm.TestCharm) ... ERROR
test_action_fail (tests.test_charm.TestCharm) ... ERROR
test_config_changed (tests.test_charm.TestCharm) ... ERROR

======================================================================
ERROR: test_action (tests.test_charm.TestCharm)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/tomwardill/work/harbor-charms/harbor-core/tests/test_charm.py", line 21, in test_action
    harness = Harness(HarborCoreCharm)
  File "/home/tomwardill/work/harbor-charms/harbor-core/venv/lib/python3.8/site-packages/ops/testing.py", line 92, in __init__
    self._update_config(key_values=self._load_config_defaults(config))
  File "/home/tomwardill/work/harbor-charms/harbor-core/venv/lib/python3.8/site-packages/ops/testing.py", line 277, in _load_config_defaults
    charm_config = charm_config.get('options', {})
AttributeError: 'NoneType' object has no attribute 'get'

======================================================================
ERROR: test_action_fail (tests.test_charm.TestCharm)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/tomwardill/work/harbor-charms/harbor-core/tests/test_charm.py", line 30, in test_action_fail
    harness = Harness(HarborCoreCharm)
  File "/home/tomwardill/work/harbor-charms/harbor-core/venv/lib/python3.8/site-packages/ops/testing.py", line 92, in __init__
    self._update_config(key_values=self._load_config_defaults(config))
  File "/home/tomwardill/work/harbor-charms/harbor-core/venv/lib/python3.8/site-packages/ops/testing.py", line 277, in _load_config_defaults
    charm_config = charm_config.get('options', {})
AttributeError: 'NoneType' object has no attribute 'get'

======================================================================
ERROR: test_config_changed (tests.test_charm.TestCharm)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/tomwardill/work/harbor-charms/harbor-core/tests/test_charm.py", line 13, in test_config_changed
[...]

Issue #238, opened 2021-02-09
`charmcraft init` creates broken charm


hatched:
It looks like the install hook expects an action to exist? 

To reproduce:
- run `charmcraft init`
- change the series in `metadata.yaml` to `focal`
- `charmcraft build`
- deploy the charm

```
juju-dashboard-charm-0: 16:21:17 INFO juju.worker.uniter awaiting error resolution for "install" hook
juju-dashboard-charm-0: 16:21:17 INFO unit.juju-dashboard-charm/0.juju-log Running legacy hooks/install.
juju-dashboard-charm-0: 16:21:17 ERROR unit.juju-dashboard-charm/0.juju-log Uncaught exception while in charm code:
Traceback (most recent call last):
  File "./src/charm.py", line 46, in <module>
    main(JujuDashboardCharmCharm)
  File "/var/lib/juju/agents/juju-dashboard-charm-0/charm/venv/ops/main.py", line 390, in main
    charm = charm_class(framework)
  File "./src/charm.py", line 24, in __init__
    self.framework.observe(self.on.fortune_action, self._on_fortune_action)
AttributeError: 'CharmEvents' object has no attribute 'fortune_action'
juju-dashboard-charm-0: 16:21:17 ERROR juju.worker.uniter.operation hook "install" (via hook dispatching script: dispatch) failed: exit status 1
juju-dashboard-charm-0: 16:21:17 INFO juju.worker.uniter awaiting error resolution for "install" hook
juju-dashboard-charm-0: 16:22:36 INFO juju.worker.uniter awaiting error resolution for "install" hook
juju-dashboard-charm-0: 16:22:36 INFO unit.juju-dashboard-charm/0.juju-log Running legacy hooks/install.
juju-dashboard-charm-0: 16:22:36 ERROR unit.juju-dashboard-charm/0.juju-log Uncaught exception while in charm code:
Traceback (most recent call last):
  File "./src/charm.py", line 46, in <module>
    main(JujuDashboardCharmCharm)
  File "/var/lib/juju/agents/juju-dashboard-charm-0/charm/venv/ops/main.py", line 390, in main
    charm = charm_class(framework)
  File "./src/charm.py", line 24, in __init__
    self.framework.observe(self.on.fortune_action, self._on_fortune_action)
AttributeError: 'CharmEvents' object has no attribute 'fortune_action'
[...]

Issue #197, opened 2020-11-15
`charmcraft init` should not populate actions.yaml with real data


markshuttle:
After `charmcraft init` I notice that there is real YAML in `config.yaml` and `actions.yaml`.

While it would be OK to have examples in comments, having real YAML in there is likely to lead to someone uploading it to Charmhub without realising it.

PR #283, opened 2021-04-08
Smoother and safer template project from init command.

### LINKED ISSUE #237: Built charm defaults to kubernetes cloud
When you run `charmcraft init` it builds a charm with a default series being `kubernetes` but a default beginner Juju bootstrap is usually not.

A simple fix might be to expand the TODO items to include modifying the series array in the `metadata.yaml`. To hopefully avoid this error:

`ERROR cannot add application "my-charm": series "kubernetes" in a non container model not valid`

In the linked issue it suggests a few options for commands to build charms specifically for k8s but maybe the init command should be interactive instead.

Related: https://github.com/canonical/charmcraft/issues/111

facundobatista:
This involves several details:

- Do NOT build the charm if the project has leftover TODOs from when it was bootstrapped with `init` .
- For that, started to use `TEMPLATE-TODO` to make it very distinct.
- Added a couple of those to different places where the developer also needs to pay attention.
- Left the templates in an state that will create a "solid" charm project.

Fixes #237, #238, #243 and #247.


cjp256 (commented):
What about just warning on TODOs rather than trying to hard error?

It does seem a little odd that the example is uncommented now in the template.  Was there an issue with the comment-only file?

jnsgruk:
I agree that warning about TEMPLATE-TODO items would be nicer than failing the build (thinking about first run, Hello, World style experiences with charmcraft)

facundobatista:
@cjp256 @jnsgruk: we want a hard error because the developers could totally ignore the warnings and just build and push a very low quality charm to Charmhub. We want to lower the barrier as much as possible for people creating charms, but at the same time avoid as much as possible to saturate Charmhub with those just-bootstrapped charms. See issue #197 from Mark, for example.

Regarding the un-commented example: that is how it born, we commented it out at some point in time (see same issue I mentioned above) so people doesn't publish a charm with real (but bogus) actions and config without realizing it. In this new model, after `init` bootstrapping, you have a real charm, that has that functionality, and even passing tests for those examples, but you HAVE to pay attention to all that (at least to consciously remove the TODO tokens or the whole examples).

jnsgruk:
Is it possible to restrict this at the point of upload/publish, rather than build?

facundobatista:
> Is it possible to restrict this at the point of upload/publish, rather than build? It seems for the purposes of getting people started with development, you should be able to quite simply follow something like:
> 
> ```shell
> mkdir my-charm && cd my-charm
> charmcraft init
> charmcraft buiild
> juju deploy ./my-charm.charm
> ```

This is a great idea. I moved the detection from build to upload. I also made it so it flags *all* files that needs to be fixed (not the first one), as the loop of changing the file and detecting the "problem" is bigger.

charmcraft/templates/init/src/charm.py.j2
```diff
@@ -33,20 +33,30 @@ class {{ class_name }}(CharmBase):
         self._stored.set_default(things=[])
 
     def _on_install(self, _):
-        # Perform any install steps and report status
+        """Perform any install steps and report status."""
         self.unit.status = ActiveStatus("Active and running")
```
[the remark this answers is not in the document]
facundobatista: I like it showing that a message can be given, as it's the only example in the file showing how to set the unit status.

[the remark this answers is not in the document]
jnsgruk: Maybe less of a problem now there are actually some developer docs? :rofl:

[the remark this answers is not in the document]
jnsgruk: We could remove the message for now, but put a comment that links to the docs?

cjp256:
I'd also suggest reconsidering how "init" works.  If it were more interactive, charmcraft could help the user create a more tailored template as needed.

cjp256 (approved):

charmcraft/commands/build.py
```diff
@@ -87,6 +90,19 @@ def relativise(src, dst):
     return pathlib.Path(os.path.relpath(str(dst), str(src.parent)))
 
 
+def validate_template_is_handled(filepath):
+    """Check that the file has not a mark of coming from the template and wasn't handled.
```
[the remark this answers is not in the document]
facundobatista: Improved the docstring in this function (moved to the upload phase).

charmcraft/commands/build.py
```diff
@@ -87,6 +90,19 @@ def relativise(src, dst):
...
+
+    This is important to eliminate the risk of getting built charms that are just created
+    with the `init` commands and not paid attentiont.
```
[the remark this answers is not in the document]
facundobatista: Same, thanks!

facundobatista:
> I'd also suggest reconsidering how "init" works. If it were more interactive, charmcraft could help the user create a more tailored template as needed.

Not a bad idea, for sure not for this PR.

charmcraft/commands/init.py
```diff
@@ -105,12 +106,17 @@ def run(self, args):
         if not re.match(r"[a-z][a-z0-9-]*[a-z0-9]$", args.name):
             raise CommandError("{} is not a valid charm name".format(args.name))
 
+        if args.series is None:
+            series = "[kubernetes]  # TEMPLATE-TODO: fix to an Ubuntu series if not using k8s"
```
cmatsuoka, line 110: Maybe "change to"?

charmcraft/commands/store/__init__.py
```diff
@@ -322,9 +325,31 @@ def fill_parser(self, parser):
...
+        This is important to avoid uploading low-quality charms that are just
+        bootstrapped and not corrected.
+        """
+        # we're already sure we can open it ok
+        zf = zipfile.ZipFile(str(filepath))
+
+        tainted_filenames = []
+        for name in zf.namelist():
+            content = zf.read(name)
+            if INIT_TEMPLATE_TOKEN in content:
```
cmatsuoka, line 340: Is it safe to assume that all files in the zipfile will be small enough to read to memory? Can we filter the possible candidates to have `TEMPLATE-TODO` by filename, or could it really be any file?

charmcraft/templates/init/config.yaml.j2
```diff
@@ -1,11 +1,12 @@
...
-#     default: 🎁
-#     description: A thing used by the charm.
-#     type: string
+# TEMPLATE-TODO: change this example to suit your needs.
+# If you don't need a config, you can remove the file entirely.
+# It ties in to the example _on_config_changed handler in src/charm.py
+
+options:
+  thing:
+    default: 🎁
```
cmatsuoka, line 10: This is a nice default.

charmcraft/commands/init.py
facundobatista: I like it! improved.

charmcraft/commands/store/__init__.py
facundobatista: Charm files tend to be very small, yes. We *could* filter by files (e.g., *.py, *.yaml, *.md, etc), but I wanted to keep this as decoupled as possible from the other part of the code.

charmcraft/templates/init/config.yaml.j2
facundobatista: :)

facundobatista:
functionality landed!

cmatsuoka (approved):
Thanks Facundo, it looks good. Maybe I'm not 100% comfortable with the possibility of reading large files to memory but this can be handled later if needed.

PR #288, opened 2021-04-15
snap: strict -> classic confinement for Charmcraft snap

cjp256:
Classic snaps on core20 require python3.  Add this collection
of stage packages as its own part to separate stage filtering
used in the charmcraft part.

There is an apparent bug in Snacpraft where it will set the
bin/python3 to an invalid symlink in this scenario.  Hardcode
the correct symlink until we have a fix in place for Snapcraft.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -50,7 +50,7 @@ architectures:
 
 apps:
   charmcraft:
-    command: bin/charmcraft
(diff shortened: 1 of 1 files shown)


facundobatista (approved):
thanks

facundobatista:
Let's talk about holding this change for 0.11 (0.10 will be used in Kubecon)

PR #289, opened 2021-04-15
Specify the resource type when uploading one.

facundobatista:
This is because the API changed, specifying the type for files is optional (for now), but we will need it soon to indicate oci-image types.

Also fixed/improved two help-related details in other commands.

### DIFF
--- a/charmcraft/commands/store/store.py
+++ b/charmcraft/commands/store/store.py
@@ -155,10 +155,13 @@ class Store:
                 entity_type=item['type']))
         return result
 
-    def _upload(self, endpoint, filepath):
+    def _upload(self, endpoint, filepath, *, extra_fields=None):
         """Upload for all charms, bundles and resources (generic process)."""
         upload_id = self._client.push(filepath)
-        response = self._client.post(endpoint, {'upload-id': upload_id})
+        payload = {'upload-id': upload_id}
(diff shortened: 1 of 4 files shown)


cjp256 (approved):
LGTM

charmcraft/commands/store/store.py
```diff
@@ -184,10 +187,10 @@ def upload(self, name, filepath):
         endpoint = '/v1/charm/{}/revisions'.format(name)
         return self._upload(endpoint, filepath)
 
-    def upload_resource(self, charm_name, resource_name, filepath):
+    def upload_resource(self, charm_name, resource_name, resource_type, filepath):
         """Upload the content of filepath to the indicated resource."""
         endpoint = '/v1/charm/{}/resources/{}/revisions'.format(charm_name, resource_name)
-        return self._upload(endpoint, filepath)
+        return self._upload(endpoint, filepath, extra_fields={'type': resource_type})
```
sergiusens, line 193: what happened to reading the value for metadata?

sergiusens (approved):
Looks good. Just a question to keep track of changes

charmcraft/commands/store/store.py
facundobatista: Don't know, will find out,

PR #291, opened 2021-04-15
Initial code for the OCI Registry functionality: auth and hit the server.

facundobatista:


charmcraft/commands/store/registry.py
```diff
@@ -0,0 +1,117 @@
...
+from urllib.request import parse_http_list, parse_keqv_list
+
+import requests
+
+from charmcraft.cmdbase import CommandError
+
+logger = logging.getLogger(__name__)
+
+# some mimetypes
+MANIFEST_LISTS = 'application/vnd.docker.distribution.manifest.list.v2+json'
```
cjp256, line 29: Googled this and found surprisingly few hits. Not a friendly search term I guess :rofl: Found some docs @ https://docs.docker.com/registry/spec/manifest-v2-2/

cjp256 (approved):
LGTM

charmcraft/commands/store/registry.py
facundobatista: yes, those are the ONLY docs...

cmatsuoka (approved):
Looks good.

PR #292, opened 2021-04-16
Added methods to check if the manifest is uploaded and get it

facundobatista:


cmatsuoka (approved):
Implementation looks correct, and matches the design and prototype explained in a previous meeting.

sergiusens (approved):

PR #294, opened 2021-04-16
The ImageHandler with the external get_destination_url funtionality.

facundobatista:


charmcraft/commands/store/registry.py
```diff
@@ -181,3 +181,28 @@ def get_manifest(self, reference):
             logger.debug("Got the v2 manifest ok")
             digest = response.headers['Docker-Content-Digest']
         return (None, digest, response.text)
+
+
+class PublicDockerhubRegistry(OCIRegistry):
+    """Hardcoded Dockerhub registry without special credentials."""
```
sergiusens, line 187:
```suggestion
    """Dockerhub registry without special credentials."""
```

sergiusens (approved):

charmcraft/commands/store/registry.py
facundobatista: thanks

cjp256 (approved):

PR #295, opened 2021-04-19
Update charm library template for correct command to publish or update libraries

mthaddon:
`charmcraft push-lib` doesn't seem to exist currently, update the template to use the correct command.


facundobatista (approved):
Awesome, thanks!!

sergiusens (approved):

PR #297, opened 2021-04-19
Better manifest in case of strict snap.

facundobatista:
This is temporary, these new translations will be removed when we get into a classic snap.

Also changed the name to com from the 'id' from os_release file.

### DIFF
--- a/charmcraft/utils.py
+++ b/charmcraft/utils.py
@@ -175,7 +175,7 @@ def get_os_platform(filepath=pathlib.Path("/etc/os-release")):
                 if value[0] == value[-1] and value[0] in ('"', "'"):
                     value = value[1:-1]
                 os_release[key] = value
-            system = os_release.get("NAME", system)
+            system = os_release.get("ID", system)
             release = os_release.get("VERSION_ID", release)
 
     return OSPlatform(system=system, release=release, machine=machine)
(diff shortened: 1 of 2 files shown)


charmcraft/utils.py
```diff
@@ -192,13 +192,21 @@ def create_manifest(basedir, started_at):
     # we integrate lifecycle lib in future branches
     architectures = [ARCH_TRANSLATIONS.get(os_platform.machine, os_platform.machine)]
 
+    # XXX Facundo 2021-04-19: these are temporary translations until charmcraft
+    # changes to be a "classic" snap
+    name_translation = {'ubuntu core': 'ubuntu'}
```
cmatsuoka, line 197: Shouldn't this be `ubuntu-core` instead since it translates the value in the os release ID?

tests/test_utils.py
```diff
@@ -235,7 +235,7 @@ def test_get_os_platform_linux(tmp_path):
     with patch('platform.machine', return_value='x86_64'):
         with patch('platform.system', return_value='Linux'):
             os_platform = get_os_platform(filepath)
-    assert os_platform.system == "Ubuntu"
+    assert os_platform.system == "ubuntu"
```
cmatsuoka, line 238: We could also add a test using core 20 os-release values.

charmcraft/utils.py
facundobatista: Fixed, thanks.

tests/test_utils.py
facundobatista: Unsure of provided value, but added anyway.

cmatsuoka (approved):
Thanks!

sergiusens (approved):

PR #298, opened 2021-04-19
Mandatorily include a README.md file when packing bundles.

facundobatista:


sergiusens (commented):
should there not be a failing pack test when README.md is missing?

sergiusens (approved):

cmatsuoka (approved):

PR #299, opened 2021-04-19
[WIP] charmcraft config: introduce support for bases keyword

cjp256:
- Support short-form syntax where a base without 'build-on'
  or 'run-on' implies build_on=base coupled with run_on=base.

- Support long-form syntax where explicit build-on and run-on
  lists are used to support any configuration combination.

- Support combined lists, where a short-form base item will
  be expanded into the long-form equivalent.

Note that this does not include unit testing for errors.  There
is a fair bit of other work to be done to improve error handling.

Note that this does not automatically set any implied architecture
in the Config, though an empty list of architectures would imply
the default (host's) is to be used.

### DIFF
--- a/charmcraft/config.py
+++ b/charmcraft/config.py
@@ -71,6 +71,15 @@ def adapt_validation_error(error):
     raise CommandError(msg)
 
 
+@format_checker.checks('architecture', raises=ValueError)
+def check_architectures(value):
+    """Check the base architectures are valid charm-supported architectures."""
+    if isinstance(value, str):
+        if value in ["amd64", "arm64"]:
+            return True
+    raise ValueError("must be a valid architecture (e.g. amd64, arm64)")
+
+
 @format_checker.checks('url', raises=ValueError)
 def check_url(value):
     """Check that the URL has at least scheme and net location."""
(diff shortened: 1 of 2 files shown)


charmcraft/config.py
```diff
@@ -105,6 +114,73 @@ def from_dict(cls, source):
...
+    name = attr.ib(validator=attr.validators.instance_of(str))
+    channel = attr.ib(validator=attr.validators.instance_of(str))
+    architectures = attr.ib(default=attr.Factory(list), validator=attr.validators.deep_iterable(
+        member_validator=attr.validators.instance_of(str),
+        iterable_validator=attr.validators.instance_of(list)))
+
+    @classmethod
+    def from_dict(cls, source):
+        """Build from a raw dict."""
+        # Ensure channel is treated as string if interpreted as number, e.g. 20.04.
```
facundobatista, line 130: But

```
>>> str(20.10)
'20.1'
```

I propose to just error out if it's a float.

charmcraft/config.py
```diff
@@ -105,6 +114,73 @@ def from_dict(cls, source):
...
+        build_on = list()
+        run_on = list()
+
+        if "build-on" in source or "run-on" in source:
+            # Long-form syntax for explicit build-on/run-on scenarios.
+            for build_base in source.get("build-on", list()):
+                build_on.append(BaseConfig.from_dict(build_base))
+
+            for run_base in source.get("run-on", list()):
+                run_on.append(BaseConfig.from_dict(run_base))
```
facundobatista, line 155: Two details I think are not handled in this logic:

- if the config says "run-on" but not "build-on" it should be an error.
- if the config says "build-on" but not "run-on", the later should default to the former (not being empty).

Another detail: it's faster and simpler to initialize an empty list using `[]` than calling the function.

charmcraft/config.py
```diff
@@ -105,6 +114,73 @@ def from_dict(cls, source):
...
+class BasesConfig:
+    """Configures for bases and build-on/run-on pairs."""
+
+    buildon_runons = attr.ib(default=attr.Factory(list))
+
+    @classmethod
+    def unmarshal(cls, sources):
+        """Convert short-form / long-form bases into long form."""
+        if not isinstance(sources, list):
+            raise ValueError("bases must be a list")
```
facundobatista, line 175: Shouldn't this be caught by the schema?

charmcraft/config.py
```diff
@@ -105,6 +114,73 @@ def from_dict(cls, source):
...
+
+        buildon_runons = list()
```
facundobatista, line 177: why not `[]`?

charmcraft/config.py
```diff
@@ -142,6 +219,24 @@ class Config:
 
 CONFIG_SCHEMA = {
     'type': 'object',
+    'definitions': {
+        'base': {
+            'type': 'object',
+            'properties': {
+                'name': {'type': 'string'},
+                'channel': {"oneOf": [{'type': 'string'}, {'type': 'number'}]},
```
facundobatista, line 227: IMO channel must be always a string, not a number. Better explicit than implicit.

charmcraft/config.py
```diff
@@ -71,6 +71,15 @@ def adapt_validation_error(error):
     raise CommandError(msg)
 
 
+@format_checker.checks('architecture', raises=ValueError)
+def check_architectures(value):
+    """Check the base architectures are valid charm-supported architectures."""
+    if isinstance(value, str):
+        if value in ["amd64", "arm64"]:
```
facundobatista, line 78: Why other archs are not supported in the config? 

Charmcraft may not be able to build in other archs, but this is a decision to be made by the building command, the config does not know enough about the system to apply the restriction here.

tests/test_config.py
```diff
@@ -397,3 +400,79 @@ def test_basicprime_empty():
...
+                BaseConfig(name="runtime-base", channel="runtime-channel", architectures=["arm64"])
+            ]),
+        BuildonRunonBaseConfig(
+            build_on=[
+                BaseConfig(name="short-base", channel="short-channel", architectures=["arm64"])
+            ],
+            run_on=[
+                BaseConfig(name="short-base", channel="short-channel", architectures=["arm64"])
+            ]),
+    ])
```
facundobatista, line 478: These unittests are fine, but please provide "tests from the human-like configuration", using the `create_config` fixture (see current tests like `test_schema_charmhub_api_url_bad_format`) specially for the errors, to ensure we're providing good error messages for those cases.

facundobatista (changes requested):
Great, thanks! I annotated some details inline.

charmcraft/config.py
cjp256: Yikes, good point.  I was wondering if it's worth moving to strictyaml.load() to get some better behavior out of the box and avoid this kind of issue?

tests/test_config.py
cjp256: Planning on it :D

charmcraft/config.py
facundobatista: `strictyaml` looks nice, but then everything will be read as strings, unless you give it a schema, and I don't like having the schema twice.

But IMO there's no need for that! Our schema already handles this detail, in the sense that it will error out with a clear message: `the 'channel' field must be a string: got 'float'.`. And the person writing the YAML can easily fix it. 

Remember our users are developers...

charmcraft/config.py
cjp256: Just a starting point for the draft. Would you prefer an enumerated list (matching what's in ARCH_TRANSLATIONS), or just leave it unchecked?

facundobatista: IMO there's no option, we must leave it unchecked. Think for example in this situation: two bases, one with `arch=[amd64]`, the other with `arch=[something-bizarre-that-will-work-in-a-private-builder]`, the charm config is totally fine to run on an `amd64` environment as "one base is valid".

PR #307, opened 2021-04-21
Improved the message for the case of Dockerhub not having v2 manifests.

facundobatista:
The previous error was a mess in the CLI; now the info is just logged and the message is nicer.

### DIFF
--- a/charmcraft/commands/store/registry.py
+++ b/charmcraft/commands/store/registry.py
@@ -176,8 +176,8 @@ class OCIRegistry:
             response = self._hit('GET', url, headers=headers)
             result = assert_response_ok(response)
             if result.get('schemaVersion') != 2:
-                raise CommandError(
-                    "Manifest v2 requested but got something else: {}".format(result))
+                logger.debug("Got something else when asking for a v2 manifest: %s", result)
+                raise CommandError("Manifest v2 not found for the indicated reference.")
             logger.debug("Got the v2 manifest ok")
             digest = response.headers['Docker-Content-Digest']
         return (None, digest, response.text)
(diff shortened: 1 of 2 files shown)


charmcraft/commands/store/registry.py
```diff
@@ -176,8 +176,8 @@ def get_manifest(self, reference):
             response = self._hit('GET', url, headers=headers)
             result = assert_response_ok(response)
             if result.get('schemaVersion') != 2:
-                raise CommandError(
-                    "Manifest v2 requested but got something else: {}".format(result))
+                logger.debug("Got something else when asking for a v2 manifest: %s", result)
+                raise CommandError("Manifest v2 not found for the indicated reference.")
             logger.debug("Got the v2 manifest ok")
```
jameinel, line 181: would it be clearer to just have "not found for {}".format(ref)

rather than "indicated reference"?

jameinel (commented):
thanks for the quick polish. I have not QAd this directly.

jameinel (approved):

charmcraft/commands/store/registry.py
```diff
@@ -176,8 +176,8 @@ def get_manifest(self, reference):
             response = self._hit('GET', url, headers=headers)
             result = assert_response_ok(response)
             if result.get('schemaVersion') != 2:
-                raise CommandError(
-                    "Manifest v2 requested but got something else: {}".format(result))
+                logger.debug("Got something else when asking for a v2 manifest: %s", result)
+                raise CommandError("Manifest v2 not found for the indicated reference.")
```
sergiusens, line 180:
```suggestion
                raise CommandError("Manifest v2 not found for {reference!r}.")
```

sergiusens (approved):

cjp256 (approved):

charmcraft/commands/store/registry.py
facundobatista: Indeed, thanks!

charmcraft/commands/store/registry.py
facundobatista: Yes,

PR #315, opened 2021-04-26
style: use black for consistent and automated formatting

cjp256:
- Add test case to check formatting.

- Add a couple of ignores to flake8 that conflict with black, including
  E203 and W503. For further reading, the rationale is provided @
  https://black.readthedocs.io/en/stable/the_black_code_style.html

They also recommend to disable E501 (line-too-long), but for this PR
I simply bumped the minimum required line length after using black. We
need to revisit our approach if we want to be stricter than that.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -10,3 +10,8 @@ exclude:
     *.egg_info
     dist
     snap
+extend-ignore:
+    # E203 whitespace before ‘:’ (Not PEP8 compliant, Python Black)
+    E203
+    # W503 line break before binary operator (Not PEP8 compliant, Python Black)
+    W503
--- a/tests/test_infra.py
+++ b/tests/test_infra.py
@@ -21,6 +21,8 @@ import re
 import subprocess
 from unittest.mock import patch
 
+import black
+import click.testing
 import pydocstyle
 import pytest
 from flake8.api.legacy import get_style_guide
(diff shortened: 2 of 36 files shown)


.flake8
```diff
@@ -1,5 +1,5 @@
 [flake8]
-max-line-length = 99
+max-line-length = 107
```
facundobatista, line 2: Let's fix those weird lines and keep this at 99, so we don't go to a "broken master" then to fix it.

(And I need to see that this is doable with black ;) )

tests/test_infra.py
```diff
@@ -132,3 +138,12 @@ def test_bashcompletion_all_commands():
         real_command_names.update(cmd.name for cmd in cmds)
 
     assert completed_commands == real_command_names
+
+
+def test_black():
+    """Verify all files are nicely styled."""
+    proc = subprocess.run(
+        ["black", "--check", "--diff", "."], check=False, capture_output=True, text=True
```
facundobatista, line 146: Can we use black programatically (and not as a subscript)? this is breaking my development setup (I don't run the tests from inside the VM, so "black" is not in the path)

facundobatista (changes requested):
Two details, thanks!

sergiusens (approved):

tests/test_infra.py
cjp256: I can give it a shot.

cjp256: Done!

.flake8
cjp256: Addressed https://github.com/canonical/charmcraft/pull/315/commits/c43e8a7a0834a15d992455a7f36c7d09ab7b26ec

facundobatista (approved):
I will hate you forever, but thanks! :)

sergiusens:
black has cookies

PR #316, opened 2021-04-26
snap: various fixes for classic

cjp256:


.github/workflows/tests.yaml
```diff
@@ -33,3 +33,32 @@ jobs:
...
+        with:
+          name: snap
+          path: ${{ steps.snapcraft.outputs.snap }}
+      - name: Install snap and test dependencies
+        run: |
+          sudo snap install --classic --dangerous ${{ steps.snapcraft.outputs.snap }}
+          sudo apt update
+          sudo apt install -y python3-pip
+      - name: Run smoke tests
+        run: |
```
jnsgruk, line 59: I think this will need `sudo`. From memory the GitHub Actions runners normally run as a user setup with sudo/NOPASSWD

jnsgruk (commented):
Added comment on integration test

.github/workflows/tests.yaml
cjp256: Yeppers, it does :D thx

.github/workflows/tests.yaml
```diff
@@ -33,3 +33,32 @@ jobs:
...
+      - name: Upload snap artifact
+        uses: actions/upload-artifact@v2
+        with:
+          name: snap
+          path: ${{ steps.snapcraft.outputs.snap }}
+      - name: Install snap and test dependencies
+        run: |
+          sudo snap install --classic --dangerous ${{ steps.snapcraft.outputs.snap }}
+          sudo apt update
+          sudo apt install -y python3-pip
```
sergiusens, line 57: python3-pip is not installed by default on most systems, specially mutlipass.

This would ideally be handled by the charm plugin, but that is not integrated yet, so this would need to be manually handled until then

.github/workflows/tests.yaml
```diff
@@ -33,3 +33,32 @@ jobs:
...
+      - name: Run smoke tests
```
sergiusens, line 58: can we also add 18.04 for the Run smoke tests part?

snap-extras/sitecustomize.py
```diff
@@ -0,0 +1,8 @@
+import site
```
facundobatista, line 1: Can this new file or the whole dir to be put inside current `snap` directory? just to not pollute the project's root .

facundobatista (changes requested):
Looks good, a small change requested (if possible, thanks!)

snap-extras/sitecustomize.py
```diff
@@ -0,0 +1,8 @@
...
+import os
+
+snap_dir = os.getenv("SNAP")
+if snap_dir:
```
sergiusens, line 5: This might make it a bit less error prone (running a terminal that is a SNAP like the VS Code shell) and setting the wrong thing up if this ever becomes more that snap-extras (which I suppose is why you added the conditional) :-)

```suggestion
if snap_dir and os.getenv("SNAP_NAME") == "charmcraft":
```

sergiusens (approved):

facundobatista (approved):
Thanks!

PR #323, opened 2021-04-29
config: migrate to use pydantic BaseModel

cjp256:
- Drop json schema in favor of BaseModel's validations

- Add unmarshal() interface to Config to handle unmarshaling the
  yaml object.

- Since `type` is not allowed to be None, set to "no-charmcraft-yaml"
when there is no YAML.

- Add a minimal Parts() and Part() implementation to allow for
  parts.bundle.prime definition.  This should ultimately be replaced
  with a reference to the implementation in craft-parts.

- Minor logging message changes:

  - must be a full URL -> must be a fully qualified URL

  - must be a valid relative URL -> must be a valid relative path

- Errors are now output as a list, with fields represented with
  python-like notation.  Lists items are indexed with [n] and
  dictionary items are reduced to the key. For example:

```
Bad charmcraft.yaml content:
- field: <broken-field>
  reason: <some reason>
- field: <parts.bundle>
  reason: <some reason>
- field: <broken.field_list[3]>
  reason: <some reason>
- ...
```

...where:

- "bundle" is the object key under parts

- "field_list[3]" is indicating the third element in field_list

### DIFF
--- a/charmcraft/config.py
+++ b/charmcraft/config.py
@@ -18,167 +18,195 @@
 
 import datetime
 import pathlib
+from typing import Any, Dict, List
 from urllib.parse import urlparse
 
-import attr
-import jsonschema
+import pydantic
 
 from charmcraft.cmdbase import CommandError
 from charmcraft.utils import load_yaml
 
-format_checker = jsonschema.FormatChecker()
 
-# translator between "json" and pythonic yaml names
-TYPE_TRANSLATOR = {
-    "object": "dict",
-    "array": "list",
-}
-
-
-def get_field_reference(path):
-    """Get a field indicator from the received path."""
-    if isinstance(path[-1], int):
-        field = ".".join(list(path)[:-1])
-        ref = "item {} in '{}' field".format(path[-1], field)
-    else:
-        field = ".".join(path)
-        ref = "'{}' field".format(field)
-    return ref
-
-
-def adapt_validation_error(error):
-    """Take a jsonschema.ValidationError and create a proper CommandError."""
-    if error.validator == "required":
-        msg = "Bad charmcraft.yaml content; missing fields: {}.".format(
-            ", ".join(error.validator_value)
-        )
-    elif error.validator == "type":
-        expected_type = TYPE_TRANSLATOR.get(
-            error.validator_value, error.validator_value
-        )
-        field_ref = get_field_reference(error.absolute_path)
-        msg = "Bad charmcraft.yaml content; the {} must be a {}: got '{}'.".format(
-            field_ref, expected_type, error.instance.__class__.__name__
-        )
-    elif error.validator == "enum":
-        field_ref = get_field_reference(error.absolute_path)
-        msg = "Bad charmcraft.yaml content; the {} must be one of: {}.".format(
-            field_ref, ", ".join(map(repr, error.validator_value))
-        )
-    elif error.validator == "format":
-        field_ref = get_field_reference(error.absolute_path)
-        msg = "Bad charmcraft.yaml content; the {} {}: got {!r}.".format(
-            field_ref, error.cause, error.instance
-        )
-    else:
-        # safe fallback
-        msg = error.message
-
-    raise CommandError(msg)
-
-
-@format_checker.checks("url", raises=ValueError)
 def check_url(value):
     """Check that the URL has at least scheme and net location."""
     if isinstance(value, str):
         url = urlparse(value)
         if url.scheme and url.netloc:
-            return True
-    raise ValueError("must be a full URL (e.g. 'https://some.server.com')")
+            return value
+    raise ValueError(
+        "must be a fully qualified URL (such as 'https://some.server.com')"
+    )
 
 
-@format_checker.checks("relative_path", raises=ValueError)
 def check_relative_paths(value):
     """Check that the received paths are all valid relative ones."""
     if isinstance(value, str):
         # check if it's an absolute path using POSIX's '/' (not os.path.sep, as the charm's
         # config is independent of the platform where charmcraft is running)
         if value and value[0] != "/":
-            return True
-    raise ValueError("must be a valid relative URL")
+            return value
+    raise ValueError("must be a valid relative path")
 
 
-@attr.s(kw_only=True, frozen=True)
-class CharmhubConfig:
-    """Configuration for all Charmhub related options."""
+def format_pydantic_error_location(loc):
+    """Format location."""
+    loc_parts = []
+    for loc_part in loc:
+        if isinstance(loc_part, str):
+            loc_parts.append(loc_part)
+        elif isinstance(loc_part, int):
+            # Integer indicates an index. Go
+            # back and fix up previous part.
+            previous_part = loc_parts.pop()
+            previous_part += f"[{loc_part}]"
+            loc_parts.append(previous_part)
+        else:
+            raise RuntimeError(f"unhandled loc: {loc_part}")
 
-    api_url = attr.ib(default="https://api.charmhub.io")
-    storage_url = attr.ib(default="https://storage.snapcraftcontent.com")
+    loc = ".".join(loc_parts)
 
-    @classmethod
-    def from_dict(cls, source):
-        """Build from a raw dict."""
-        return cls(**source)
+    # Filter out internal __root__ detail.
+    loc = loc.replace(".__root__", "")
+    return loc
 
 
-class BasicPrime(tuple):
-    """Hold the list of files to include, specified under parts/bundle/prime configs.
+def format_pydantic_error_msg(msg):
+    """Format pydantic's error message field."""
+    # Replace shorthand "str" with "string".
+    msg = msg.replace("str type expected", "string type expected")
+    return msg
 
-    This is a intermediate structure until we have the full Lifecycle in place.
+
+def format_pydantic_errors(errors):
+    """Format errors.
+
+    Example 1: Single error.
+
+    Bad charmcraft.yaml content:
+    - field: <some field>
+      reason: <some reason>
+
+    Example 2: Multiple errors.
+
+    Bad charmcraft.yaml content:
+    - field: <some field>
+      reason: <some reason>
+    - field: <some field 2>
+      reason: <some reason 2>
     """
+    combined = ["Bad charmcraft.yaml content:"]
+    for error in errors:
+        loc = "- field: " + format_pydantic_error_location(error["loc"])
+        combined.append(loc)
 
-    @classmethod
-    def from_dict(cls, parts):
-        """Build from a dicts sequence."""
-        prime = parts.get("bundle", {}).get("prime", [])
-        return cls(prime)
+        msg = "  reason: " + format_pydantic_error_msg(error["msg"])
+        combined.append(msg)
+
+    return "\n".join(combined)
 
 
-@attr.s(kw_only=True, frozen=True)
-class Project:
+class Part(
+    pydantic.BaseModel, extra=pydantic.Extra.forbid, frozen=True, validate_all=True
+):
+    """Placeholder representation for part.  To be replaced by craft-parts."""
+
+    prime: List[pydantic.StrictStr] = []
+
+    @pydantic.validator("prime", each_item=True)
+    def validate_relative_paths(cls, prime):
+        """Verify relative paths are used in prime."""
+        return check_relative_paths(prime)
+
+
+class Parts(
+    pydantic.BaseModel, extra=pydantic.Extra.forbid, frozen=True, validate_all=True
+):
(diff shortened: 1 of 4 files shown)


charmcraft/config.py
```diff
@@ -18,167 +18,195 @@
...
+            return None
+
+
+class CharmhubConfig(
+    pydantic.BaseModel, extra=pydantic.Extra.forbid, frozen=True, validate_all=True
+):
+    """Represents top-level charmhub object."""
+
+    api_url: pydantic.StrictStr = "https://api.charmhub.io"
+    storage_url: pydantic.StrictStr = "https://storage.snapcraftcontent.com"
```
cmatsuoka, line 145: Maybe those could be `pydantic.HttpUrl` so they can be automatically validated by pydantic instead of using `validate_urls()`.

charmcraft/config.py
```diff
@@ -18,167 +18,195 @@
...
+
+    @pydantic.validator("api_url", "storage_url")
+    def validate_urls(cls, url):
```
cmatsuoka, line 148: Can be removed if validating using `HttpUrl`.

charmcraft/config.py
```diff
@@ -18,167 +18,195 @@
...
 
+def format_pydantic_errors(errors):
+    """Format errors.
 
-class BasicPrime(tuple):
-    """Hold the list of files to include, specified under parts/bundle/prime configs.
+    Example 1: Single error.
 
-    This is a intermediate structure until we have the full Lifecycle in place.
+    Bad charmcraft.yaml content:
```
sergiusens, line 85: do we want to use doctests?

charmcraft/config.py
```diff
@@ -18,167 +18,195 @@
...
-    "required": ["type"],
-    "additionalProperties": False,
-}
+    dirpath: pydantic.DirectoryPath
+    content: Dict[str, Any] = {}
+    config_provided: bool = False
+    started_at: datetime.datetime
+
+    @pydantic.validator("dirpath")
+    def validate_path(cls, path):
```
cmatsuoka, line 164: Is this necessary? Validation of `DirectoryPath` should handle this automatically.

charmcraft/config.py
```diff
@@ -18,167 +18,195 @@
...
+            raise ValueError("invalid path")
+        return path
+
+
+class Config(
+    pydantic.BaseModel, extra=pydantic.Extra.forbid, frozen=True, validate_all=True
+):
+    """Charmcraft config.
+
+    :ivar type: If none, no charmcraft.yaml is present.
```
cmatsuoka, line 176: It could be worth to list the valid types.

charmcraft/config.py
```diff
@@ -18,167 +18,195 @@
...
+
+    type: pydantic.StrictStr
+    charmhub: CharmhubConfig = CharmhubConfig()
+    parts: Parts = Parts()
+    project: Project
+
+    @pydantic.validator("type")
+    def validate_charm_type(cls, charm_type):
+        """Verify charm type is valid with exception when instantiated without YAML."""
+        if charm_type not in ["bundle", "charm", "no-charmcraft-yaml"]:
```
sergiusens, line 190: I would rather use undefined here.

```suggestion
        if charm_type not in ["bundle", "charm", "undefined"]:
```

charmcraft/config.py
cjp256: good point :D

charmcraft/config.py
cjp256: Sounds good, i know it's ugly, but temporary :D

charmcraft/config.py
cjp256: Nice, didn't notice that one.

charmcraft/config.py
cjp256: Nice, didn't notice that one, thanks!

cjp256: Done!

charmcraft/config.py
cjp256: Done!

charmcraft/config.py
cjp256: done!

charmcraft/config.py
cjp256: Simplified the documentation here so we can call schema().

charmcraft/config.py
cjp256: Done!

charmcraft/config.py
cjp256: I did add one doctest to document the schema in the module because @facundobatista asked for the schema to be documented.  But I'll leave it up to @facundobatista if he wants more pervasive doctests (and a more generalized enablement across the codebase).

cmatsuoka (approved):
Thanks Chris, lgtm.

charmcraft/config.py
```diff
@@ -14,171 +14,280 @@
...
+                    "maxLength": 2083,
+                    "format": "uri",
+                    "type": "string"
+                }
+            },
+            "additionalProperties": false
+        },
+        "Part": {
+            "title": "Part",
+            "description": "Defintiion of part to build.",
```
sergiusens, line 87:
```suggestion
            "description": "Defintion of part to build.",
```

charmcraft/config.py
```diff
@@ -14,171 +14,282 @@
...
-class Project:
-    """Configuration for all project-related options, used internally."""
 
-    dirpath = attr.ib(default=None)
-    config_provided = attr.ib(default=None)
-    started_at = attr.ib(default=None)
+class Parts(pydantic.BaseModel, frozen=True, validate_all=True):
+    """Definition of parts to build."""
 
+    __root__: Dict[pydantic.StrictStr, Part] = {}
```
facundobatista, line 208: We're relaxing too much the config here. Let's try to not change the rules just because we're changing the implementation/lib used.

Currently the only thing that be used here is "bundle", whatever different should be an error, because it's something we're not supporting.

IOW, if the user puts the following, it should fail:

```
parts:
   bumdle:
      prime: []
```

charmcraft/config.py
```diff
@@ -14,171 +14,282 @@
...
+            "title": "Parts",
+            "description": "Definition of parts to build.",
+            "default": {},
+            "type": "object",
+            "additionalProperties": {
+                "$ref": "#/definitions/Part"
+            }
+        }
+    }
+}
```
facundobatista, line 111: All this definition is precise and exact, but IMO does not help the user to understand the schema.

For "precise and exact", the code below can be used. I think we should provide a **much more human** info, like (beware, needs some adjustments):

```
Definition of charmcraft.yaml configuration:

type: [string] one of "charm" or "bundle"

charmhub:
    api_url: [HttpUrl] optional, defaults to "https://api.charmhub.io"
    storage_url: [HttpUrl] optional, defaults to "https://storage.snapcraftcontent.com"

parts: 
    bundle:
        prime: [list of strings] 
```

charmcraft/config.py
facundobatista: I don't like doctests :/. -1 to start using them.

charmcraft/config.py
```diff
@@ -14,171 +14,282 @@
...
-    """Configuration for all Charmhub related options."""
+    Example 2: Multiple errors.
 
-    api_url = attr.ib(default="https://api.charmhub.io")
-    storage_url = attr.ib(default="https://storage.snapcraftcontent.com")
+    Bad charmcraft.yaml content:
+    - field: <some field>
+      reason: <some reason>
+    - field: <some field 2>
+      reason: <some reason 2>
```
facundobatista, line 179: This format of "list of keys", used in yaml, may work but it's not too helpful to understand what's going on.

I'd stick to use one line per error. For example, instead of the following:

```
Bad charmcraft.yaml content:
- field: parts.bundle
  reason: value is not a valid dict
```

I prefer:

```
Bad charmcraft.yaml content:
- value is not a valid dict in field 'parts.bundle'
```

facundobatista: Extra points to specify *what* is the program interpreting that is not a dict. Why not? the developer may be yelling at the screen because for they **it is a dict**.

See in opposition the current error in this situation: `Bad charmcraft.yaml content; the 'parts' field must be a dict: got 'list'.`

We really should try to get the extra mile here, otherwise we're losing functionality with this change...

charmcraft/config.py
```diff
@@ -14,171 +14,282 @@
...
-    def from_dict(cls, parts):
-        """Build from a dicts sequence."""
-        prime = parts.get("bundle", {}).get("prime", [])
-        return cls(prime)
+    prime: List[pydantic.StrictStr] = []
 
+    @pydantic.validator("prime", each_item=True)
+    def validate_relative_paths(cls, prime):
+        """Verify relative paths are used in prime."""
+        return check_relative_paths(prime)
```
facundobatista, line 202: Question: instead of declaring that the type is `StrictStr` and then apply a validation, can't we just specify that the type is something like `RelativePath` or something? Of course, `RelativePath` will be something we're coding above, that will validate it's a string and a relative path...

charmcraft/config.py
```diff
@@ -14,171 +14,282 @@
...
+    storage_url: pydantic.HttpUrl = "https://storage.snapcraftcontent.com"
+
+
+class Project(
+    pydantic.BaseModel, extra=pydantic.Extra.forbid, frozen=True, validate_all=True
+):
+    """Internal-only project configuration."""
+
+    dirpath: pydantic.DirectoryPath
+    content: Dict[str, Any] = {}
```
facundobatista, line 236: Why we need this `content` here?

charmcraft/config.py
```diff
@@ -14,171 +14,282 @@
...
+
+class Config(
+    pydantic.BaseModel,
+    extra=pydantic.Extra.forbid,
+    frozen=True,
+    validate_all=True,
+):
+    """Definition of charmcraft.yaml configuration."""
+
+    type: pydantic.StrictStr
```
facundobatista, line 251: Can we put something here like `StrictStr | None` or `Optional(StrictStr)`, I totally hate this new `undefined`.

charmcraft/config.py
```diff
@@ -14,171 +14,282 @@
...
+    pydantic.BaseModel, extra=pydantic.Extra.forbid, frozen=True, validate_all=True
+):
+    """Internal-only project configuration."""
+
+    dirpath: pydantic.DirectoryPath
+    content: Dict[str, Any] = {}
+    config_provided: bool = False
+
+    # this timestamp will be used in several places, even sent to Charmhub: needs to be UTC
+    started_at: datetime.datetime = datetime.datetime.utcnow()
```
facundobatista, line 240: Why this default is needed here, if always is passed below?

facundobatista (changes requested):
Awesome work! I left some comments and questions inline, thanks!

charmcraft/config.py
cjp256: Good point, will fix

charmcraft/config.py
cjp256: :+1: will do

charmcraft/config.py
cjp256: It was already there https://github.com/canonical/charmcraft/pull/323/files#diff-0657ee63e8720cf85a4fc089151f357cd8309bf35f2213c20f2845f9cd67683aL211  - looks unused so I removed it

facundobatista: FTR, there is no "content" attribute in the current (as in `master` branch) config.

charmcraft/config.py
cjp256: I hate it too but it was a bit painful to deal with the model being optionally None but only in certain cases. I figured since it's only short-lived, who cares.  I worked around it though.

charmcraft/config.py
cjp256: At least one test fixture in pack wasn't passing it, so I put a default in for sanity unless passed. I can remove the one passed in, make it optional, or address the missing values - your choice.

charmcraft/config.py
cjp256: Done.

charmcraft/config.py
cjp256: done

charmcraft/config.py
cjp256: Unfortunately that information is not provided in pydantic exception and I don't see a straight forward way to capture the item that was the source of the error in pydantic.  I expect we could certainly improve some pydantic validators in the future if desired to improve context...

charmcraft/config.py
cjp256: Removed

facundobatista (approved):
Wonderful, thanks!

PR #324, opened 2021-04-29
Provide the building charms functionality through the pack command.

facundobatista:


cjp256 (approved):

charmcraft/commands/pack.py
```diff
@@ -58,27 +65,93 @@ def get_paths_to_include(config):
...
+file directly, or upload it to Charmhub with `charmcraft upload`.
+
+For the charm you must be inside a charm directory with a valid
+`metadata.yaml`, `requirements.txt` including the `ops` package
+for the Python operator framework, and an operator entrypoint,
+usually `src/charm.py`.  See `charmcraft init` to create a
+template charm directory structure.
+
+For the bundle you must already have the `bundle.yaml` (can be
+generated by Juju) and the README.md file.
```
sergiusens, line 80: if not, charms are generic and bundles are specific :-)

```suggestion
For the bundle you must already have a `bundle.yaml` (can be
generated by Juju) and a README.md file.
```

charmcraft/commands/pack.py
```diff
@@ -58,27 +65,93 @@ def get_paths_to_include(config):
...
 """
 
 
 class PackCommand(BaseCommand):
     """Build the bundle or the charm.
 
-    Eventually this command will also support charms, but for now it will work only
-    on bundles.
+    If charmcraft.yaml missing or indicates a charm, use the "build"
```
sergiusens, line 87:
```suggestion
    If charmcraft.yaml is missing or its type is charm, use the "build"
```

charmcraft/commands/pack.py
```diff
@@ -58,27 +65,93 @@ def get_paths_to_include(config):
...
+    infrastructure to build the charm.
```
sergiusens, line 88:
```suggestion
    infrastructure to create the charm.
```

tests/commands/test_pack.py
```diff
@@ -51,10 +53,66 @@ def func(*, name):
     return func
 
 
-# -- tests for main building process
+# -- tests for the project type decissor
 
 
-def test_simple_succesful_build(tmp_path, caplog, bundle_yaml, config):
+def test_decissor_charm_type(config):
```
sergiusens, line 59:
```suggestion
def test_resolve_charm_type(config):
```

Noun. decisor (plural decisors) (obsolete) a person who resolves or settles matters or controversies

tests/commands/test_pack.py
```diff
@@ -51,10 +53,66 @@ def func(*, name):
...
+    """The config indicates the project is a charm."""
+    config.set(type="charm")
+    cmd = PackCommand("group", config)
+
+    with patch.object(cmd, "_pack_charm") as mock:
+        cmd.run(noargs)
+    mock.assert_called_with(noargs)
+
+
+def test_decissor_bundle_type(config):
```
sergiusens, line 69:
```suggestion
def test_resolve_bundle_type(config):
```

tests/commands/test_pack.py
```diff
@@ -51,10 +53,66 @@ def func(*, name):
...
+    """The config indicates the project is a bundle."""
+    config.set(type="bundle")
+    cmd = PackCommand("group", config)
+
+    with patch.object(cmd, "_pack_bundle") as mock:
+        cmd.run(noargs)
+    mock.assert_called_with()
+
+
+def test_decissor_no_given_config(config):
```
sergiusens, line 79:
```suggestion
def test_resolve_no_config_packs_charm(config):
```

tests/commands/test_pack.py
```diff
@@ -51,10 +53,66 @@ def func(*, name):
...
+
+    with pytest.raises(CommandError) as cm:
+        PackCommand("group", config).run(args)
+    assert (
+        str(cm.value)
+        == "The -r/--requirement option is valid only when packing a charm"
+    )
+
+
+def test_decissor_bundle_with_entrypoint(config):
```
sergiusens, line 102:
```suggestion
def test_resolve_bundle_with_entrypoint(config):
```

sergiusens (approved):
2 lines they said

charmcraft/commands/pack.py
```diff
@@ -58,27 +65,93 @@ def get_paths_to_include(config):
...
+    help_msg = "Build the charm or bundle"
     overview = _overview
-    needs_config = True
+    needs_config = False  # optional until we fully support charms here
+
+    def fill_parser(self, parser):
+        """Add own parameters to the general parser."""
+        parser.add_argument(
+            "-e",
+            "--entrypoint",
```
cmatsuoka, line 102: Is `-- entrypoint` already agreed upon, or should it be `--entry-point`?

cmatsuoka (approved):
Looks good, +1 with changes suggested by @sergiusens. Added a question about `entrypoint` vs `entry-point` (this affects the user interface so it will be a bit annoying to change it later).

charmcraft/commands/pack.py
facundobatista: Thanks.,

charmcraft/commands/pack.py
facundobatista: Adapted, thanks.

charmcraft/commands/pack.py
facundobatista: Yes.

tests/commands/test_pack.py
facundobatista: So the idea was right. Test the "decisor piece of code", which decides if it's a charm or a bundle.

Taking your suggestions in, anyway.

charmcraft/commands/pack.py
facundobatista: I'd stick with the same exact parameter that comes from the "build" command. In the medium term we need to discuss if we should even keep these two options, they probably should be options for the "Charm Plugin". Let's talk about this.

sergiusens: This should probably go away and become a charmcraft.yaml proper field equivalent to apps or something the charm plugin deals with

PR #325, opened 2021-04-30
build: shell out to python to check host pip rather than environment's

cjp256:
Importing pip directly is checking the environment in which charmcraft
is running, not the version of pip that will be used by host's pip3.

Without changing the check behavior, simply call out to python3 and
perform the check.


cjp256:
This is in #316's queue as a requirement to pass integration tests on Bionic with classic snap.

sergiusens (approved):
Looks sane, we cannot rely on the pip bundled with charmcraft as it will be tied to the interpreter charmcraft is using if directly imported.

charmcraft/commands/build.py
```diff
@@ -55,13 +55,16 @@
...
-        return False
+    cmd = [
+        "python3",
+        "-c",
+        (
+            "from pip.commands.install import InstallCommand; "
+            'assert InstallCommand().cmd_opts.get_option("--system") is not None'
+        ),
+    ]
+    proc = subprocess.run(cmd, capture_output=True)
```
facundobatista, line 66: `capture_output` does not exist in Python 3.6. Why we need it anyway?

facundobatista (changes requested):

charmcraft/commands/build.py
sergiusens: For context, subprocess.run is processed using charmcraft's python interpreter (3.8), not the one on the system, the python in the arguments to subprocess.run is the one on the system

facundobatista: Note that charmcraft may still be installed in Bionic from PyPI.

sergiusens: I thought we agreed to baseline on 3.8, was that not the case? python3.8 is available on 18.04.

facundobatista: Let's talk about this. It's not clear to me where/how we want charmcraft to run.

facundobatista (approved):
Wonderful, thanks!

PR #333, opened 2021-05-05
utils: introduce get_host_architecture()

cjp256:


sergiusens (approved):

facundobatista:
What's the idea of this? How/where you will use it?

cjp256:
Pretty much the same as its current use.

When `bases` is specified in charmcraft.yaml, determining the host arch as the default if `architectures:` is not set.

facundobatista (approved):
Thanks!

Issue #329, opened 2021-05-02
charmcraft upload of "bash"/"hooks only" charm doesn't work


erik78se:
I'm trying to upload a "hooks only charm"  ............... that fails.

1. The documentation on publishing does not cover this scenario at all https://juju.is/docs/sdk/publishing

2. The video covering the process is now outdated and contains stuff that is invalid today. Such as referencing "--charm-file"- 

3. charmcraft upload fails when a simple "hooks only" charm has been zip:ed into a file.
```
charmcraft upload gitlab-runner.charm
The indicated zip file 'gitlab-runner.charm' is not a charm ('metadata.yaml' not found) nor a bundle ('bundle.yaml' not found). (full execution logs in /home/erik/snap/charmcraft/common/charmcraft-log-afu1wu5p)
```

erik78se:
In the description of charmcraft (https://discourse.charmhub.io/t/release-1-0-0/4537) it say:

_We recommend the Charmed Operator Framework for charm development, but Charmcraft will help you publish charms written in any language. ..._

So, charms written in bash "should" work?

jameinel:
charms written in bash might not be supported by charmcraft but should be
usable by both the store and juju.

The actual error was no 'metadata.yaml' which is definitely needed by all
parties involved.  But it may be that charmcraft pack failed to assemble
the contents because of other incompatibilities.

erik78se:
... but there is a metadata.yaml in my charm.

https://github.com/erik78se/charm-gitlab-runner

I don't know how to get my charm uploaded atm.

Should I use "charm push" instead?



Den sön 2 maj 2021 14:22John Arbash Meinel ***@***.***> skrev:

> charms written in bash might not be supported by charmcraft but should be
> usable by both the store and juju.
>
> The actual error was no 'metadata.yaml' which is definitely needed by all
> parties involved. But it may be that charmcraft pack failed to assemble
> the contents because of other incompatibilities.
>

facundobatista:
Hello erik78se,

Would you please attach here the building logs? You can see those by doing `charmcraft build --verbose`, thanks.

jameinel:
The existing tooling (charm push) should still be working publishing to Charmstore. If you are using Juju 2.9, you can reference them as `cs:~user/charm-name`.

erik78se:
I really appreciate you taking the time to help me. But I headed out to
refactoring to operator for the pieces.

I kind of agree on the line of charmcraft and operators which I think will
[...]

PR #335, opened 2021-05-06
Ability to build a hooks only charm

### LINKED ISSUE #334: Charmcraft cannot build a non-operator framework charm
The current version of charmcraft cannot build a "hooks only" charm. An example of such a charm is [tiny-bash](https://github.com/erik78se/tiny-bash).

If you try to build that charm, you get the following output:

```
❯ git clone https://github.com/erik78se/tiny-bash.git
Cloning into 'tiny-bash'...
remote: Enumerating objects: 52, done.
remote: Counting objects: 100% (17/17), done.
remote: Compressing objects: 100% (16/16), done.
remote: Total 52 (delta 1), reused 5 (delta 0), pack-reused 35
Receiving objects: 100% (52/52), 22.74 KiB | 2.27 MiB/s, done.
Resolving deltas: 100% (13/13), done.
❯ cd tiny-bash
❯ charmcraft -v pack
2021-05-06 12:18:15,939  charmcraft                     DEBUG    Starting charmcraft version 0.10.0+58.ge8d67aa
2021-05-06 12:18:15,939  charmcraft.main                DEBUG    Raw pre-parsed sysargs: args={'help': False, 'verbose': True, 'quiet': False, 'project_dir': None} filtered=['pack']
2021-05-06 12:18:15,939  charmcraft.commands            DEBUG    Couldn't find config file /home/jon/temp/tiny-bash/charmcraft.yaml
2021-05-06 12:18:15,939  charmcraft.main                DEBUG    General parsed sysargs: command='pack' args=[]
2021-05-06 12:18:15,940  charmcraft.main                DEBUG    Command parsed sysargs: Namespace(entrypoint=None, requirement=None)
2021-05-06 12:18:15,940  charmcraft                     ERROR    Charm entry point was not found: '/home/jon/temp/tiny-bash/src/charm.py' (full execution logs in /home/jon/snap/charmcraft/common/charmcraft-log-tdb6s3cj)
```

There isn't an obvious `entrypoint` to specify here, and even if we did, we'd need to not include the `dispatch` and virtualenv in the built charm, which would require some changes to the `build`/`pack` command.

Related to #329

jnsgruk:
A very simple hack on the current charmcraft code to support building hooks only charms, by adding a `--bare/-b` option to `build`/`pack`. 

This will prevent `charmcraft` from requiring an `entrypoint`, prevent it from creating `dispatch` when building a bare charm. A `venv` directory will still be created and populated in the charm *if* there is a `requirements.txt` file

Fixes #334 
Related #329

### DIFF
--- a/charmcraft/commands/build.py
+++ b/charmcraft/commands/build.py
@@ -116,11 +116,12 @@ class Builder:
 
         linked_entrypoint = self.handle_generic_paths()
 
-        # Do not create dispatch or venv in built charm if we're building a
+        # Do not create dispatch in built charm if we're building a
         # bare charm
         if not self.bare:
             self.handle_dispatcher(linked_entrypoint)
-            self.handle_dependencies()
+
+        self.handle_dependencies()
         zipname = self.handle_package()
 
         logger.info("Created '%s'.", zipname)
@@ -313,7 +314,7 @@ class Validator:
 
     _options = [
         "from",  # this needs to be processed first, as it's a base dir to find other files
-        "bare",
+        "bare",  # we check this next, as it obviates the need for the following two
         "entrypoint",
         "requirement",
     ]
@@ -356,7 +357,9 @@ class Validator:
 
     def validate_entrypoint(self, filepath):
         """Validate that the entrypoint exists and is executable."""
-        if self.bare:
+        if filepath is not None and self.bare:
+            raise CommandError("Cannot specify 'bare' and 'entrypoint'")
+        elif self.bare:
             return ""
 
         if filepath is None:
@@ -385,9 +388,6 @@ class Validator:
 
         If not specified, default to requirements.txt if there.
         """
-        if self.bare:
-            return []
-
         if filepaths is None:
             req = self.basedir / "requirements.txt"
             if req.exists() and os.access(req, os.R_OK):
--- a/charmcraft/commands/build.py
+++ b/charmcraft/commands/build.py
@@ -331,7 +331,7 @@ class Validator:
         return result
 
     def validate_bare(self, bare):
-        """Validate that the charm dir is there and yes, a directory."""
+        """Check if we're building a bare (non-operator framework) charm."""
         self.bare = bare
         return bare
 
--- a/charmcraft/commands/build.py
+++ b/charmcraft/commands/build.py
@@ -25,10 +25,9 @@ import subprocess
 import zipfile
 
 import yaml
-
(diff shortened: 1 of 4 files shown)


jnsgruk:
Tested with [tiny-bash](https://github.com/erik78se/tiny-bash):

```
❯ git clone https://github.com/erik78se/tiny-bash.git
Cloning into 'tiny-bash'...
remote: Enumerating objects: 52, done.
remote: Counting objects: 100% (17/17), done.
remote: Compressing objects: 100% (16/16), done.
remote: Total 52 (delta 1), reused 5 (delta 0), pack-reused 35
Receiving objects: 100% (52/52), 22.74 KiB | 1.08 MiB/s, done.
Resolving deltas: 100% (13/13), done.
❯ cd tiny-bash
❯ python3 -m charmcraft build -b
Created 'tiny-bash.charm'.
```

jnsgruk:
Latest commit should fix all existing tests, though no tests exist specifically for the new option yet.

There are more formatting changes than I might have anticipated, the code was formatted with `black`, though...

jnsgruk:
Updated to add some tests for the 'bare' option and also allow a virtualenv to be created and populated in a 'bare' charm if there is a 'requirements.txt' in the working directory. This enables people who are creating non-operator framework Python charms to still have a clean packaged venv at runtime

jameinel:
I agree that putting this as some sort of config in charmcraft.yaml is a better path forward. You shouldn't have to remember how each charm is built and pass it on the CLI each time.

I'm not sure about the specific parts/plugins for how that would be done. There is a lot of bikeshedding that could be done to get the naming/structure right, but its probably better to have a deliberate approach to that.

jnsgruk:
It would be quite simple to remove the extra command line option from this so that a 'bare' charm is built by default if `pack` is called, and the validator detects there is no `src/charm.py`, but there is `hooks/` directory?

jameinel:
for backwards compatibility (with juju 2.7 which didn't support dispatch) the guidline was that you should always create hooks/install and the Operator Framework would check if it was invoked via `dispatch` and if not, would create the ~15 symlinks in hooks/ necessary to catch all the possible events.
I don't know if backwards compatibility is as necessary now that 2.8 and then 2.9 is out, but the presence of 'hooks' is not strictly an indicator whether or not 'dispatch' should be created. (There exist operator framework charms that want dispatch, but also have hooks/install)

facundobatista:
As we discussed, we'll solve this through the new building mechanism soon to be incorporated...

PR #336, opened 2021-05-07
Include branch coverage measurement in unit tests

mthaddon:
Add branch coverage measurement to unit tests.

See https://coverage.readthedocs.io/en/coverage-5.5/branch.html?highlight=cov-branch#branch-coverage-measurement for details, and https://github.com/mthaddon/plan-services-test for an example of a side effect of this.


sergiusens (approved):

jameinel:
Just to clarify, the issue is that code coverage was reporting that we had covered everything,but what we hadn't covered is both a 'true' evaluation of an if and a 'false' evaluation of an if (because there was no explicit 'else' clause, there weren't any lines to be covered in the case of false)
I'm curious why --branch isn't the default, is there significant overhead in using it?

PR #339, opened 2021-05-10
config: introduce long-form bases support

cjp256:
Add support `bases:` keyword, targetting only the long-form.

Short-form support will be added later.

### DIFF
--- a/charmcraft/config.py
+++ b/charmcraft/config.py
@@ -19,7 +19,8 @@
 Using pydantic's BaseModel, this module supports the translation of the
 charmcraft.yaml to a python object.
 
-Configuration schema:
+Configuration Schema
+====================
 
 type: [string] one of "charm" or "bundle"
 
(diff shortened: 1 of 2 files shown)


charmcraft/config.py
```diff
@@ -164,6 +185,29 @@ class CharmhubConfig(
     storage_url: pydantic.HttpUrl = "https://storage.snapcraftcontent.com"
 
 
+class Base(
+    pydantic.BaseModel, extra=pydantic.Extra.forbid, frozen=True, validate_all=True
```
facundobatista, line 189: I wonder if repeating this again and again, we could have something like...

````
class ConfigBaseModel(pydantic.BaseModel, extra=pydantic.Extra.forbid, frozen=True, validate_all=True):
    pass
```

And then just use that to inherit all our classes.

facundobatista (approved):
I like it! Check an inline comment/question, though.

cmatsuoka (approved):

PR #340, opened 2021-05-11
config: improve error messages for extra & missing fields

cjp256:

### DIFF
--- a/charmcraft/config.py
+++ b/charmcraft/config.py
@@ -174,12 +174,12 @@ def format_pydantic_errors(errors):
         if formatted_msg == "field required":
             field_name, location = printable_field_location_split(formatted_loc)
             combined.append(
-                f"- field {field_name} required for {location} configuration"
+                f"- field {field_name} required in {location} configuration"
             )
         elif formatted_msg == "extra fields not permitted":
             field_name, location = printable_field_location_split(formatted_loc)
             combined.append(
-                f"- extra field {field_name} not permitted for {location} configuration"
+                f"- extra field {field_name} not permitted in {location} configuration"
             )
         else:
             combined.append(f"- {formatted_msg} in field {formatted_loc!r}")
--- a/charmcraft/config.py
+++ b/charmcraft/config.py
@@ -56,7 +56,7 @@ Object with the following properties:
 
 import datetime
 import pathlib
-from typing import Any, Dict, List, Optional
+from typing import Any, Dict, List, Optional, Tuple
 
 import pydantic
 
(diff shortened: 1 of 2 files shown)


charmcraft/config.py
```diff
@@ -150,7 +171,18 @@ def format_pydantic_errors(errors):
...
-        combined.append(f"- {formatted_msg} in field {formatted_loc!r}")
+        if formatted_msg == "field required":
+            field_name, location = printable_field_location_split(formatted_loc)
+            combined.append(
+                f"- field {field_name} required for {location} configuration"
+            )
+        elif formatted_msg == "extra fields not permitted":
+            field_name, location = printable_field_location_split(formatted_loc)
+            combined.append(
+                f"- extra field {field_name} not permitted for {location} configuration"
```
facundobatista, line 182: "required for" and "permitted for", or (taking in consideration that the problematic field happens inside a config key, "required in" and "permitted in"?

facundobatista (approved):
+1, but please take a look to the question inside. Thanks!

charmcraft/config.py
cjp256: I actually did it both ways and kinda flip-flopped.  Take your pick and I'll update it if desired :)

charmcraft/config.py
```diff
@@ -150,7 +171,18 @@ def format_pydantic_errors(errors):
         formatted_loc = format_pydantic_error_location(error["loc"])
         formatted_msg = format_pydantic_error_message(error["msg"])
 
-        combined.append(f"- {formatted_msg} in field {formatted_loc!r}")
+        if formatted_msg == "field required":
```
cmatsuoka, line 174: Could this be affected by locale?

charmcraft/config.py
```diff
@@ -150,7 +171,18 @@ def format_pydantic_errors(errors):
...
+            combined.append(
+                f"- field {field_name} required in {location} configuration"
+            )
+        elif formatted_msg == "extra fields not permitted":
+            field_name, location = printable_field_location_split(formatted_loc)
+            combined.append(
+                f"- extra field {field_name} not permitted in {location} configuration"
+            )
+        else:
+            combined.append(f"- {formatted_msg} in field {formatted_loc!r}")
```
cmatsuoka, line 185: This kind of handling would be useful in craft-parts too. Perhaps we could move it to an external common toolbox later.

cmatsuoka (approved):

charmcraft/config.py
cjp256: negative, but I wonder if we can catch / determine the error class directly, that may be better in the future

charmcraft/config.py
cjp256: Sounds like a good idea

PR #341, opened 2021-05-13
config: add short-form bases support (CRAFT-11)

cjp256:
Support short-form bases, where a base implies the long-form
base configuration equivalent of:

```
- build-on: [<base>]
  run-on: [<base>]
```

### DIFF
--- a/charmcraft/config.py
+++ b/charmcraft/config.py
@@ -263,6 +263,35 @@ class Config(ModelConfigDefaults, validate_all=False):
         return charm_type
 
     @classmethod
+    def expand_short_form_bases(cls, bases: List[Dict[str, Any]]) -> None:
+        """Expand short-form base configuration into long-form in-place."""
+        for index, base in enumerate(bases):
+            # Skip if already long-form. Account for common typos in case user
+            # intends to use long-form, but did so incorrectly (for better
+            # error message handling).
+            if (
+                "run-on" in base
+                or "run_on" in base
+                or "build-on" in base
+                or "build_on" in base
+            ):
+                continue
+
+            try:
+                converted_base = Base(**base)
+            except pydantic.error_wrappers.ValidationError as error:
+                # Rewrite location to assist user.
+                pydantic_errors = error.errors()
+                for pydantic_error in pydantic_errors:
+                    pydantic_error["loc"] = ("bases", index, pydantic_error["loc"][0])
+
+                raise CommandError(format_pydantic_errors(pydantic_errors))
+
+            base.clear()
+            base["build-on"] = [converted_base.dict()]
+            base["run-on"] = [converted_base.dict()]
+
+    @classmethod
     def unmarshal(cls, obj: Dict[str, Any], project: Project):
         """Unmarshal object with necessary translations and error handling.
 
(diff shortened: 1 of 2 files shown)


charmcraft/config.py
```diff
@@ -262,6 +262,35 @@ def validate_charm_type(cls, charm_type, values):
...
+        for index, base in enumerate(bases):
+            # Skip if already long-form. Account for common typos in case user
+            # intends to use long-form, but did so incorrectly (for better
+            # error message handling).
+            if (
+                "run-on" in base
+                or "run_on" in base
+                or "build-on" in base
+                or "build_on" in base
+            ):
```
facundobatista, line 277: We also need to support this case:

```
bases:
  - build-on:
    - name: ubuntu
      channel: 20.04
      architectures: [<machine arch>]
```

facundobatista (changes requested):
Code looks nice, but there's an extra case we need to support (see comment)

charmcraft/config.py
cjp256: As in the 'run-on' is implied to match build-on?

facundobatista: Yeap.

cjp256: Seems odd to support configuring the same thing three different ways out of the gate, but can do.

To clarify, the `run-on` implication is squarely the matching the *chosen* build-on that was used to create the build environment (as `build-on` is a list of options).  It may make sense to allow run-on to be None, in which case it will always be set to match the build environment.

How about we add this in a follow up PR?

cjp256: We've agreed that we will avoid this extra form for the time being as it provides no value.

charmcraft/config.py
```diff
@@ -262,6 +262,35 @@ def validate_charm_type(cls, charm_type, values):
...
+    def expand_short_form_bases(cls, bases: List[Dict[str, Any]]) -> None:
+        """Expand short-form base configuration into long-form in-place."""
+        for index, base in enumerate(bases):
+            # Skip if already long-form. Account for common typos in case user
+            # intends to use long-form, but did so incorrectly (for better
+            # error message handling).
+            if (
+                "run-on" in base
+                or "run_on" in base
+                or "build-on" in base
```
facundobatista, line 275: Alternative way of writing this, if you like:

```
if {"run-on", "run_on", "build-on", "build_on"} & set(base):
    continue
```

facundobatista (approved):
Great! Thanks! Just check a suggestion in the comment.

charmcraft/config.py
```diff
@@ -262,6 +262,35 @@ def validate_charm_type(cls, charm_type, values):
...
+    @classmethod
+    def expand_short_form_bases(cls, bases: List[Dict[str, Any]]) -> None:
+        """Expand short-form base configuration into long-form in-place."""
+        for index, base in enumerate(bases):
+            # Skip if already long-form. Account for common typos in case user
+            # intends to use long-form, but did so incorrectly (for better
+            # error message handling).
+            if (
+                "run-on" in base
+                or "run_on" in base
```
cmatsuoka, line 274: Why do we check both `run-on` and `run_on`? Shouldn't the underscore be translated to dash on unmarshal?

cmatsuoka (approved):
+1

charmcraft/config.py
cjp256: This check happens before the unmarshal, so I want to make sure we capture the mispelled case.

charmcraft/config.py
cjp256: interesting, though I will keep it as-is for readability.

another succinct option could have been:
`any(k in base for k in ["build-on", "build_on", "run-on", "run_on"])`

PR #362, opened 2021-05-31
New config, and deprecation infrastructure

facundobatista:
Added `registry-url` to the configuration, in preparation for the resources refactor.

Also changed the other `charmhub` keys to use dash, instead of underscore. But to do that change, I included some deprecation infrastructure.

I thought about preparing a branch with the deprecation infra alone, but I took the opportunity of being able to include it here (not too many lines) as it shows also how that infra is used.

### DIFF
--- a/charmcraft/deprecations.py
+++ b/charmcraft/deprecations.py
@@ -29,16 +29,26 @@ import logging
 logger = logging.getLogger(__name__)
 
 
+# the message to show for each deprecation ID (this needs to be in sync with the
+# documentation)
 _DEPRECATION_MESSAGES = {
     "dn01": "Configuration keywords are now separated using dashes.",
 }
 
+# the URL to point to the deprecation entry in the documentation
 _DEPRECATION_URL_FMT = "https://discourse.charmhub.io/t/4652#heading--{deprecation_id}"
+
+# already-notified deprecations will be stored here to not log them twice
+_ALREADY_NOTIFIED = set()
 
 
 def notify_deprecation(deprecation_id):
     """Present proper messages to the user for the indicated deprecation id."""
+    if deprecation_id in _ALREADY_NOTIFIED:
+        return
+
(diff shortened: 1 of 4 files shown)


charmcraft/deprecations.py
```diff
@@ -0,0 +1,44 @@
...
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+_DEPRECATION_MESSAGES = {
+    "dn01": "Configuration keywords are now separated using dashes.",
+}
+
+_DEPRECATION_URL_FMT = "https://discourse.charmhub.io/t/4652#heading--{deprecation_id}"
```
sergiusens, line 36: Do you think we can get a docs endpoint for this?

tests/test_config.py
```diff
@@ -437,6 +474,29 @@ def test_charmhub_frozen():
...
+    """
+    )
+    cfg = load(tmp_path)
+    assert cfg.charmhub.storage_url == "https://server1.com"
+    assert cfg.charmhub.api_url == "https://server2.com"
+    assert cfg.charmhub.registry_url == "https://server3.com"
+    deprecation_msg = (
+        "DEPRECATED: Configuration keywords are now separated using dashes."
+    )
+    assert deprecation_msg in [rec.message for rec in caplog.records]
```
sergiusens, line 497: One thing we have as tech debt in Snapcraft is to log the deprecation notice only once. It would be nice, but not mandatory to have that here from the get go.

sergiusens (approved):
Nice, looks good, added an itsy bitsy comment

tests/test_config.py
facundobatista: Good idea, added this detail.

charmcraft/deprecations.py
facundobatista: For sure in the future, we can fix this URL later.

cmatsuoka (approved):
A very good solution to handle the field name change problem.

PR #364, opened 2021-06-02
Refactor how OCIRegistry and ImageHandler are used (CRAFT-53)

facundobatista:
This is to align the classes' signatures for the big refactor ahead.

Specifically, in OCIRegistry:

- receive just the image name (whatever it could be) and not the two "orga" and "name" parameters which were too specific for image names with only two "tokens"

- also receive username and password, and store the ready-to-use encoded credentials

- changed slightly a couple of url-massaging functions to reflect the fact that the server has now the schema in it (not faking it anymore, more robust)

And in ImageHandler:

- receive now just the registry and use it

- it will hold only one registry now, so now need for the "dst_" prefix

Also in the upload-command command, changed from passing the components to just pass the registry itself.

Note I minimized the changes in the tests; some of those may be simplified further, but they will disappear in the refactor, so no need for the noise now.

### DIFF
--- a/charmcraft/commands/store/registry.py
+++ b/charmcraft/commands/store/registry.py
@@ -16,7 +16,9 @@
 
 """Module to work with OCI registries."""
 
+import base64
 import logging
+import urllib.parse
 from urllib.request import parse_http_list, parse_keqv_list
 
 import requests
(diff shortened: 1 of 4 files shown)


cmatsuoka (approved):
Thanks for the detailed PR description. Couldn't see any problems, implementation seems good and the refactoring improves code clarity and organization.

charmcraft/commands/store/registry.py
```diff
@@ -64,13 +66,25 @@ def assert_response_ok(response, expected_status=200):
...
         self.server = server
-        self.orga = organization
-        self.name = image_name
-
+        self.image_name = image_name
         self.auth_token = None
-        self.auth_encoded_credentials = None
+
+        if username:
+            _u_p = "{}:{}".format(username, password)
```
sergiusens, line 75: this assumes password is set but not checked for, what are the implecations of setting "<username>:" ?

sergiusens (changes requested):
There's a test missing for username and no password set which would document if that is allowed or not

charmcraft/commands/store/registry.py
facundobatista: This makes no assumptions on password. If the username is set, this means there are credentials, so `auth_encoded_credentials` is calculed, otherwise is set to None (later will be used by the `_authenticate` method.

If password is "foo", "asjkldhakl", or "", it's just fine.

facundobatista:
> There's a test missing for username and no password set which would document if that is allowed or not

Test added.

sergiusens (approved):

PR #367, opened 2021-06-04
Tests improvements to minimize future merges from resources draft branch

facundobatista:
Some small changes, everything just in tests, no affected functionality:

- simplified some OCIRegistry calls, the "test-orga" makes no more sense (left it as-is in other tests that will disappear in the future)

- improved some test names

- added schema to "service" key in auth tests, it's more real now

- improved a test in test_main.py to isolate it from the running environment


sergiusens (approved):
Looks good and straightforward.

I personally would have used a commit message that made more sense when looking in historical context (specifically the "draft branch" part) and described *what* is being introduced (the longer description could say *why*)

cjp256 (approved):

PR #369, opened 2021-06-07
Support the case of no info for the user at all in init command (CRAFT-249)

facundobatista:


cjp256 (approved):

charmcraft/commands/init.py
```diff
@@ -92,7 +92,11 @@ def run(self, args):
         logger.debug("Using project directory '%s'", self.config.project.dirpath)
 
         if args.author is None:
-            gecos = pwd.getpwuid(os.getuid()).pw_gecos.split(",", 1)[0]
+            try:
+                gecos = pwd.getpwuid(os.getuid()).pw_gecos.split(",", 1)[0]
+            except KeyError:
+                # no info for the user
+                gecos = None
```
sergiusens, line 99: if gecos is only ever assigned in this try block, why not directly raise the CommandError here?

sergiusens (approved):

PR #370, opened 2021-06-07
Added the blob upload functionality to OCIRegistry (CRAFT-234)

facundobatista:


charmcraft/commands/store/registry.py
```diff
@@ -37,6 +37,10 @@
     MANIFEST_LISTS,
     MANIFEST_V2_MIMETYPE,
 }
+OCTET_STREAM_MIMETYPE = "application/octet-stream"
+
+# downloads and uploads happen in chunks
```
sergiusens, line 42: ah, maybe the code comment needs to happen here :-)

charmcraft/commands/store/registry.py
```diff
@@ -193,6 +198,63 @@ def upload_manifest(self, manifest_data, reference):
...
+        logger.debug(
+            "Got upload URL ok with range %s-%s", range_from, range_to_inclusive
+        )
+        if range_from != 0:
+            raise CommandError("Server error: bad range received")
+        if range_to_inclusive == 0:
+            range_to_inclusive = -1
+
+        # start the chunked upload
+        from_position = range_to_inclusive + 1
```
cmatsuoka, line 221: This looks a bit suspicious, is it correct that we'll never start at position 1? Maybe positions are always a multiple of `CHUNK_SIZE` and therefore always greater than 1? If so, I think it's worth to add a comment explaining the situation.

cmatsuoka (approved):
Looks good, just added a small note asking for a clarification.

charmcraft/commands/store/registry.py
cmatsuoka: Oh, it seems that @sergiusens found the comment in line 42.

charmcraft/commands/store/registry.py
facundobatista: Improved this in the other PR

charmcraft/commands/store/registry.py
facundobatista: It's a side effect of the range being inclusive.

So, if the server tells us that it already has "0-80", means that it has 81 bytes (from 0 to 80 inclusive), we set `from_position` in 81 and read from there. Going down, if the server says it has "0-10" it means it has 11 bytes. "0-2" would mean it has 3 bytes (from positions 0, 1 and 2). "0-1" would mean it has bytes 0 and 1. "0-0" would mean... WAIT! this is what the store returns when it does not have ANY bytes at all.

So we comply with Range parameter, but addressing this special case, I don't see many alternatives here. The worst think it could happen, IIUC, is that we start from 0 when the server has 1 byte already, which is not a problem.

sergiusens: this could be a code comment itself. I remember this comment from a past discussion and quick demo you did.

charmcraft/commands/store/registry.py
```diff
@@ -203,6 +208,63 @@ def upload_manifest(self, manifest_data, reference):
...
+                    break
+
+                end_position = from_position + len(chunk)
+                headers = {
+                    "Content-Length": str(len(chunk)),
+                    "Content-Range": "{}-{}".format(from_position, end_position),
+                    "Content-Type": OCTET_STREAM_MIMETYPE,
+                }
+                progress = 100 * end_position / size
+                print("Uploading.. {:.2f}%\r".format(progress), end="", flush=True)
```
sergiusens, line 246: this makes me a bit uncomfortable if we intend to have an eventual common library (the printing from a package/library may not be as nice) and taking that into account or to start thinking about this would be nice

sergiusens (approved):
Looks good, a comment as Claudio asked again for the start "+1" even if the var name gives a hint would be nice (as it was asked even though we had the var name). I have another comment which would be nice to start floating around as things progress.

charmcraft/commands/store/registry.py
facundobatista: This will be replaced when we have the library, of course.

I added a comment here in he code, and opened [an issue about it](https://github.com/canonical/charmcraft/issues/381).

charmcraft/commands/store/registry.py
facundobatista: Added a comment explaining this...

PR #376, opened 2021-06-09
Added a helper in ImageHandler to extract a particular file from a tar archive. (CRAFT-236)

facundobatista:


charmcraft/commands/store/registry.py
```diff
@@ -38,6 +43,9 @@
     MANIFEST_V2_MIMETYPE,
 }
 
+# downloads and uploads happen in chunks
+CHUNK_SIZE = 2 ** 20
```
sergiusens, line 47: why this size?

charmcraft/commands/store/registry.py
```diff
@@ -249,3 +279,34 @@ def get_destination_url(self, reference):
...
+                mtime=mtime,
+            )
+        else:
+            dst_filehandler = hashing_temp_file
+        while True:
+            chunk = src_fh.read(CHUNK_SIZE)
+            if not chunk:
+                break
+            dst_filehandler.write(chunk)
+        dst_filehandler.close()
```
sergiusens, line 307: do you intend to catch any error produced by this and retry? If so, this may warrant a try and finally to ensure the close takes place for this and the temp file.

Also, aren't both of these handles? The naming is different for each

charmcraft/commands/store/registry.py
facundobatista: This is mostly driven by the usage in the "upload blob" part, where I found that the cost in time was actually doing the dance to push bytes, not because of the quantity of bytes sent.

IOW (numbers are to exemplify), sending 64k takes 200ms, sending 1M takes 250ms.

charmcraft/commands/store/registry.py
facundobatista: I don't intend to catch errors here, this is a local communication that actually ends quite fast, I don't think we won't see many failures here.

Improved the naming.

charmcraft/commands/store/registry.py
sergiusens: Sorry for being vague, I was hoping the reason for the size made it into the code comment :-)

sergiusens (approved):
Looks good.

If possible, I would appreciate a code comment on why you chose that chunk size.

charmcraft/commands/store/registry.py
cmatsuoka: Ideally files should be closed on write error, but I don't know the usage context to see if it could lead to a descriptor leak or something similar. From your previous comment I assume that there will be no major consequences of leaving them open, but adding a try-finally around the while loop shouldn't cause any problems.

charmcraft/commands/store/registry.py
facundobatista: Added.

charmcraft/commands/store/registry.py
facundobatista: Improved, thanks.

cmatsuoka (approved):
Great, thanks!

PR #379, opened 2021-06-11
main: verify runtime environment is supported (CRAFT-65)

cjp256:
Add env module with helpers for querying/validating environment:
- is_charmcraft_running_from_snap()
- is_charmcraft_running_in_developer_mode()
- is_charmcraft_running_in_supported_environment()
- ensure_environment_is_supported()

Call ensure_environment_is_supported() from main to verify that
the runtime environment is supported.  If not supported,
CHARMCRAFT_DEVELOPER=1 would be required to be set in environment
for development purposes.

Update main tests to bypass the "unsupported" configuration error
with an autouse fixture.


charmcraft/env.py
```diff
@@ -0,0 +1,53 @@
...
+    return os.getenv("SNAP_NAME") == "charmcraft" and os.getenv("SNAP") is not None
+
+
+def is_charmcraft_running_in_developer_mode():
+    """Check if Charmcraft is running under developer mode."""
+    return os.getenv("CHARMCRAFT_DEVELOPER") == "1"
+
+
+def is_charmcraft_running_in_supported_environment():
+    """Check if Charmcraft is running under developer mode."""
```
facundobatista, line 37: Bad docstring.

facundobatista (commented):
Added a small detail in the code.

Beyond that, what will be the plan for people actually NOT using the snap? When they update, `charmcraft` will just start crashing under their noses... they will complain, we will tell them to use the snap or set the environment variable... maybe we could make this more auto-discoverable?

charmcraft/env.py
cjp256, line 37:
```suggestion
    """Check if Charmcraft is running in a supported environment."""
```

facundobatista (approved):
+1 if we change the link in the future to the proper documentation on which are the supported platforms and how to run charmcraft elsewhere.

charmcraft/env.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+def is_charmcraft_running_from_snap():
+    """Check if charmcraft is running from the snap."""
+    return os.getenv("SNAP_NAME") == "charmcraft" and os.getenv("SNAP") is not None
+
+
+def is_charmcraft_running_in_developer_mode():
+    """Check if Charmcraft is running under developer mode."""
+    return os.getenv("CHARMCRAFT_DEVELOPER") == "1"
```
sergiusens, line 33: if we want to support more "true" type values in this field, maybe
```suggestion
    return distutils.util.strtobool(os.getenv("CHARMCRAFT_DEVELOPER", "n")) == 1
```

sergiusens (approved):
Looks good, just a suggestion brought in from @chipaca's past recommendation of best practices instilled in me :-)

charmcraft/env.py
cjp256: I considered that, but thought it'd be better to make it more restrictive (at least initially).  I'm OK either way, I'll defer to @facundobatista.

cjp256: Done

PR #384, opened 2021-06-14
manifest: add bases support to create_manifest() (CRAFT-63)

cjp256:
- Move create_manifest() out of utils into its own module.  This will
prevent a potential import loop due to dependencies on util in config.

- Add optional bases_config parameter to use the run-on bases
configuration.

Note that runtime usage still only invokes the "legacy" non-bases
configuration.  Future work will enable the usage of bases and this
just adds test coverage for the future case.

### DIFF
--- a/charmcraft/manifest.py
+++ b/charmcraft/manifest.py
@@ -64,7 +64,14 @@ def create_manifest(
             }
         ]
     else:
-        bases = [r.dict() for r in bases_config.run_on]
+        bases = [
+            {
+                "name": r.name,
+                "channel": r.channel,
+                "architectures": r.architectures,
+            }
+            for r in bases_config.run_on
+        ]
 
     content = {
         "charmcraft-version": __version__,
--- a/charmcraft/manifest.py
+++ b/charmcraft/manifest.py
@@ -1,0 +1,80 @@
+# Copyright 2020-2021 Canonical Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# For further info, check https://github.com/canonical/charmcraft
+
+"""Charmcraft manifest.yaml related functionality."""
+
+import datetime
+import logging
+import pathlib
+from typing import Optional
+
+import yaml
+
+from charmcraft import __version__, config, utils
+from charmcraft.cmdbase import CommandError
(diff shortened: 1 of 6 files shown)


cmatsuoka (approved):
Looks good, implementation matches what's in the PR description.

charmcraft/manifest.py
```diff
@@ -0,0 +1,80 @@
...
+        channel = channel_translation.get(os_platform.release, os_platform.release)
+        bases = [
+            {
+                "name": name,
+                "channel": channel,
+                "architectures": architectures,
+            }
+        ]
+    else:
+        bases = [r.dict() for r in bases_config.run_on]
```
facundobatista, line 67: I will decouple the config from the manifest here. IOW, don't throw in whatever is in the config, but explicitly take the fields that should be included.

This is because two reasons:
- legibility/discoverability/learnability: if you're trying to debug something related to the manifest or learn about it, it's better to have everything explicitly stated here than trying to chase other code to see how fits in here
- future proof: maybe later we want to include stuff in `run-on` in the config that we don't want to leak into the manifest, or modify something, and we may forget to fix it here

facundobatista, line 67: And a question regarding to use just what's in the config: there is no translation needed?

Does the user (and charmcraft to build) use the same terms than we need to send to Charmhub and Juju?

charmcraft/manifest.py
```diff
@@ -0,0 +1,80 @@
...
+    """
+    if bases_config is None:
+        os_platform = utils.get_os_platform()
+
+        # XXX Facundo 2021-03-29: the architectures list will be provided by the caller when
+        # we integrate lifecycle lib in future branches
+        architectures = [utils.get_host_architecture()]
+
+        # XXX Facundo 2021-04-19: these are temporary translations until charmcraft
+        # changes to be a "classic" snap
```
facundobatista, line 53: Maybe this is an opportunity to clean this? we're not *strict* anymore, right?

facundobatista (changes requested):
Great, I like the manifest creation in its own module.

Annotated something to change and a couple of questions. Thanks!

charmcraft/manifest.py
cjp256: Sure, I can do that np.

cjp256: I would _expect_ it to match Charmhub/Juju by the very nature of `run-on`.  Its only relevant for this purpose and should never contain extraneous data.  However, perhaps the short-form could possibly prove problematic someday if something is introduced into `base` that is somehow only relevant to `build-on` bases configuration.

I updated it to be explicit in construction.

There should be an explicit spec for this somewhere though.  https://juju.is/docs/sdk/charm-anatomy just refers us back to here.

charmcraft/manifest.py
cjp256: I was avoiding unnecessary refactoring in the migration, but may as well rip that bandaid...

Done!

charmcraft/manifest.py
cjp256: Done!

facundobatista (approved):
wonderful, thanks!

PR #385, opened 2021-06-15
Added an lightweight interface to communicate with local docker daemon.

facundobatista:


charmcraft/commands/store/registry.py
```diff
@@ -331,6 +332,45 @@ def write(self, data):
...
+            response = self.session.get(url)
+        except requests.exceptions.ConnectionError:
+            logger.debug(
+                "Cannot connect to /var/run/docker.sock , please ensure dockerd is running.",
+            )
+            return
+
+        if response.status_code == 200:
+            # image is there, we're fine
+            pass
```
cjp256, line 357:
```suggestion
            return response.json()
```

charmcraft/commands/store/registry.py
```diff
@@ -331,6 +332,45 @@ def write(self, data):
...
+        elif response.status_code == 404:
+            # image not found (known error)
+            return
+        else:
+            logger.debug(
+                "Bad response when validation local image: %s", response.status_code
+            )
+            return
```
cjp256, line 365: Personally I prefer to explicitly "return None" if the method is returning a value.

charmcraft/commands/store/registry.py
```diff
@@ -331,6 +332,45 @@ def write(self, data):
...
+class LocalDockerdInterface:
+    """Functionality to interact with a local Docker daemon."""
+
+    # the address of the dockerd socket
+    dockerd_socket_baseurl = "http+unix://%2Fvar%2Frun%2Fdocker.sock"
+
+    def __init__(self):
+        self.session = requests_unixsocket.Session()
+
+    def get_image_info(self, digest):
```
cjp256, line 344: Consider adding type information with new code?

cjp256 (approved):

charmcraft/commands/store/registry.py
facundobatista: Refactored a little.

charmcraft/commands/store/registry.py
facundobatista: I don't like it, for me it's redundant.

charmcraft/commands/store/registry.py
facundobatista: Added.

cmatsuoka (approved):
Looks nice!

PR #387, opened 2021-06-15
Added the upload_from_local functionality to ImageHandler (CRAFT-239)

facundobatista:
Also improve the `assert_response_ok` function to better detect if the response is JSON, and added a bunch of type hints.

### DIFF
--- a/charmcraft/commands/store/registry.py
+++ b/charmcraft/commands/store/registry.py
@@ -463,18 +463,18 @@ class ImageHandler:
         logger.debug("Getting the image from the local repo; size=%s", local_image_size)
         response = dockerd.get_streamed_image_content(digest)
 
-        tmp_exported = tempfile.NamedTemporaryFile(mode="wb")
+        tmp_exported = tempfile.NamedTemporaryFile(mode="wb", delete=False)
         extracted_total = 0
-        for chunk in response.iter_content(2 ** 20):
+        for chunk in response.iter_content(CHUNK_SIZE):
             extracted_total += len(chunk)
             progress = 100 * extracted_total / local_image_size
             print("Extracting... {:.2f}%\r".format(progress), end="", flush=True)
             tmp_exported.file.write(chunk)
+        tmp_exported.close()
 
-        # open the image tar and inspect it to get the config and layers for the manifest
-        tmp_exported.file.flush()
+        # open the image tar and inspect it to get the config and layers from the only one
+        # manifest inside (as it's a list of one)
         image_tar = tarfile.open(tmp_exported.name)
-        tmp_exported.close()  # closing implies deletion, but the tar lib already grabbed it ok
         local_manifest = json.load(image_tar.extractfile("manifest.json"))
         (local_manifest,) = local_manifest
         config_name = local_manifest.get("Config")
@@ -514,6 +514,9 @@ class ImageHandler:
                     "size": size,
                 }
             )
+
+        # remove the temp tar file
+        os.unlink(tmp_exported.name)
 
         # upload the manifest
         manifest_data = json.dumps(manifest)
--- a/charmcraft/commands/store/registry.py
+++ b/charmcraft/commands/store/registry.py
@@ -26,7 +26,7 @@ import os
 import tarfile
 import tempfile
 import urllib.parse
-from typing import Union
+from typing import Union, Dict, Any
 from urllib.request import parse_http_list, parse_keqv_list
 
 import requests
@@ -57,7 +57,7 @@ CHUNK_SIZE = 2 ** 20
 
 def assert_response_ok(
     response: requests.Response, expected_status: int = 200
-) -> Union[dict, None]:
+) -> Union[Dict[str, Any], None]:
     """Assert the response is ok."""
     if response.status_code != expected_status:
         ct = response.headers.get("Content-Type", "")
@@ -449,8 +449,11 @@ class ImageHandler:
         # finally remove the temp filepath
         os.unlink(filepath)
 
-    def upload_from_local(self, digest: str) -> str:
-        """Upload the image from the local registry."""
+    def upload_from_local(self, digest: str) -> Union[str, None]:
+        """Upload the image from the local registry.
+
+        Returns the new remote digest, or None if the image was not found locally.
+        """
         dockerd = LocalDockerdInterface()
 
         # validate the image is present locally
--- a/charmcraft/commands/store/registry.py
+++ b/charmcraft/commands/store/registry.py
@@ -20,8 +20,10 @@ import base64
 import gzip
 import hashlib
 import io
+import json
 import logging
 import os
+import tarfile
 import tempfile
 import urllib.parse
 from typing import Union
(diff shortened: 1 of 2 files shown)


charmcraft/commands/store/registry.py
```diff
@@ -440,3 +448,77 @@ def _upload_blob(self, filepath, size, digest):
...
+        if image_info is None:
+            return
+        local_image_size = image_info["Size"]
+
+        logger.debug("Getting the image from the local repo; size=%s", local_image_size)
+        response = dockerd.get_streamed_image_content(digest)
+
+        tmp_exported = tempfile.NamedTemporaryFile(mode="wb")
+        extracted_total = 0
+        for chunk in response.iter_content(2 ** 20):
```
cmatsuoka, line 468: Any special reason for not using `CHUNK_SIZE` (defined in line 52) here?

charmcraft/commands/store/registry.py
```diff
@@ -440,3 +448,77 @@ def _upload_blob(self, filepath, size, digest):
...
+            extracted_total += len(chunk)
+            progress = 100 * extracted_total / local_image_size
+            print("Extracting... {:.2f}%\r".format(progress), end="", flush=True)
+            tmp_exported.file.write(chunk)
+
+        # open the image tar and inspect it to get the config and layers for the manifest
+        tmp_exported.file.flush()
+        image_tar = tarfile.open(tmp_exported.name)
+        tmp_exported.close()  # closing implies deletion, but the tar lib already grabbed it ok
```
cmatsuoka, line 477: Will this work on Windows?  IIRC it doesn't allow open files to be deleted.

charmcraft/commands/store/registry.py
```diff
@@ -440,3 +448,77 @@ def _upload_blob(self, filepath, size, digest):
...
+        local_manifest = json.load(image_tar.extractfile("manifest.json"))
+        (local_manifest,) = local_manifest
```
cmatsuoka, line 479: I'm curious about this line, why do we need this transformation? What is the original format of `local_manifest`?

charmcraft/commands/store/registry.py
facundobatista: Good catch! Fixed.

charmcraft/commands/store/registry.py
facundobatista: Furthermore, it does not allow an open file to be opened! So it's not that I just can close it later. 

So, I had to tell the temp file to not delete on close, and handle that manually later.

The equivalent to this code (which is the one I wrote to try it on Windows):

```
import os
import tempfile

tmp_exported = tempfile.NamedTemporaryFile(mode="wb", delete=False)
tmp_exported.file.write(b"1234567\n")
tmp_exported.close()

tmp_path = tmp_exported.name

image_tar = open(tmp_path, 'rb')

print("Content:", image_tar.read())
print("Deleted?", not os.path.exists(tmp_path))
```

Anyway, fixed.

charmcraft/commands/store/registry.py
facundobatista: It's a list of one. Real life example:

```
[{'Config': '4616ab87e79ca4a4475211cf3f69eb40267c3b041f45b5bcecade8191fe2d97e.json',
  'Layers': ['fe44bcd0154eabf7a5ffdf2e270de700ea90c6155a7ccbd26bd8a791ce7e8e42/layer.tar',
             'bceaa585b938da41b8dfa56f497ece3a251cc7a70d9dde885499795b492f21eb/layer.tar',
             '7c46cbc5d4f29498503431edb7a151b4059a5e2ceaf1faf27a5d936198eb0213/layer.tar',
             'f615336d566d20f7bd51f01e28ad41bc7dc745e468eede82524a0508a8caf20f/layer.tar',
             '1e60b1df8ed245cbb286dd9e04f769a583534c0c396c583080e52c41dbad49b5/layer.tar',
             '0588cf2dc6457432fc6c2623f32f48fa82c4f6fd528d6296ad4dc74b8196a6bf/layer.tar',
             'e949573175f54ca50a024487a2126202e1e0217a8a88c8d027da17c9d10660b9/layer.tar',
             'db065a320c5e41441dfd3d58ea54ac1ba97a353becd925a7e30988854c18974a/layer.tar',
             '5728c11f035ae937880b6775817122890e0496bd60ad42de7967cbc08c58b2e8/layer.tar',
             '968d68636c32452e8ab9b2a401742fd65000eb91e689b879e4b128ccc20cc5be/layer.tar',
             'bfd82e46ae0b90aaadae7ba2a7dd72899fd2e63207f3165ff0cd45dfb49de0a7/layer.tar',
             '5e23da1d3ad0f4de734c74dd1fc615fc6ed88407638c9cc866582bfd289610c3/layer.tar'],
  'RepoTags': None}]
```

Improved the comment to reflect this.

cmatsuoka (approved):
+1,  thanks for the clarification! (added a note about removing a temporary file, this may require some refactoring so it can be done in a follow-up if needed)

charmcraft/commands/store/registry.py
```diff
@@ -440,3 +448,80 @@ def _upload_blob(self, filepath, size, digest):
...
+            manifest_layers.append(
+                {
+                    "digest": digest,
+                    "mediaType": LAYER_MIMETYPE,
+                    "size": size,
+                }
+            )
+
+        # remove the temp tar file
+        os.unlink(tmp_exported.name)
```
cmatsuoka, line 519: Do we need to guarantee that this file will be removed if the upload fails?

facundobatista: As we just discussed, this is good enough. May be improved in the future if really necessary, when we see the value of the added complexity.

sergiusens (approved):

sergiusens:
Heh, my approval is with comments :-)

charmcraft/commands/store/registry.py
```diff
@@ -440,3 +448,80 @@ def _upload_blob(self, filepath, size, digest):
...
+
+    def upload_from_local(self, digest: str) -> str:
+        """Upload the image from the local registry."""
+        dockerd = LocalDockerdInterface()
+
+        # validate the image is present locally
+        logger.debug("Checking image is present locally")
+        image_info = dockerd.get_image_info(digest)
+        if image_info is None:
+            return
```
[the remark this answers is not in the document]
facundobatista: Oops! Fixed, and improved the docstring.

charmcraft/commands/store/registry.py
```diff
@@ -52,10 +55,13 @@
 CHUNK_SIZE = 2 ** 20
 
 
-def assert_response_ok(response, expected_status=200):
+def assert_response_ok(
+    response: requests.Response, expected_status: int = 200
+) -> Union[dict, None]:
```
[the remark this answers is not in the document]
facundobatista: I prefer `Optional[SomeType]` for when a receiving parameter is optional, but to express the function output, I think that `Union[SomeType, None]` is better, as there is nothing really optional.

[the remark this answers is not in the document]
facundobatista: that said, improved to show the dict content

charmcraft/commands/store/registry.py
```diff
@@ -440,3 +448,80 @@ def _upload_blob(self, filepath, size, digest):
...
+
+        # validate the image is present locally
+        logger.debug("Checking image is present locally")
+        image_info = dockerd.get_image_info(digest)
+        if image_info is None:
+            return
+        local_image_size = image_info["Size"]
+
+        logger.debug("Getting the image from the local repo; size=%s", local_image_size)
+        response = dockerd.get_streamed_image_content(digest)
```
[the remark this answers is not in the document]
facundobatista: Let's talk about this. Can be improved later, anyway.

PR #389, opened 2021-06-16
build & pack: add and use default ubuntu 20.04 bases configuration (CRAFT-266)

cjp256:
- Add a default base of Ubuntu 20.04 to match the behavior of the
strict snap which always reports as Ubuntu 20.04.  Upcoming work
will enable the use of `bases:` configuration in the charmcraft.yaml
to override this.

- Update pack command for bundles to use default bases configuration
as well.

- Remove legacy charm name format as it will no longer be used.

- Remove legacy manifest bases support as it will no longer be used.
Update associated unit tests.

Note that there is no deprecation warning for the use of default
   bases, this will be added in a follow up PR.

### DIFF
--- a/charmcraft/commands/build.py
+++ b/charmcraft/commands/build.py
@@ -23,7 +23,6 @@ import pathlib
 import shutil
 import subprocess
 import zipfile
-from typing import Optional
 
 import yaml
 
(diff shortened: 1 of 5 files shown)


facundobatista (changes requested):
Let's talk about this. If you add this "default bases configuration" temporarily hardcoded in the config (before really using what the user specifies, which will be in the next branches), all code here will be almost final (or final), there will be no need of creating new methods or using constants from build in pack. 

And then, all the changes remaining will be confined to config itself.

cjp256:
I'm going to close this for the moment and rework this.

PR #390, opened 2021-06-16
build/pack: add support for bases in managed mode (CRAFT-131)

cjp256:
- Intoduce bases module with `get_host_as_base()` and
  `check_if_base_matches_host()`.

- Extend build/pack to support `bases` configuration if present
  and managed-mode is enabled.  Error if unable to build for any of
  the configured bases.  Warn for each bases config that cannot be
  built for.  Provide additional debug logging on matching behavior
  to help diagnose matching issues when they occur.

- Update run() to return a list of charms built rather than just
  the charm file name.

Note bases are currently required with the use of managed mode until
we introduce the default bases configuration.

### DIFF
--- a/charmcraft/commands/build.py
+++ b/charmcraft/commands/build.py
@@ -168,7 +168,7 @@ class Builder:
 
         In managed-mode (and eventually destructive-mode), build for each bases
         configuration which has a matching build-on to the host we are executing
-        on.  Warn for each bases configuration that is unbuildable.  Error if
+        on.  Warn for each base configuration that is incompatible.  Error if
         unable to produce any builds for any bases configuration.
 
         Otherwise, build the charm in the legacy mode until we have complete
--- a/charmcraft/commands/build.py
+++ b/charmcraft/commands/build.py
@@ -23,12 +23,14 @@ import pathlib
 import shutil
 import subprocess
 import zipfile
-from typing import Optional
+from typing import List, Optional
 
 import yaml
 
+from charmcraft.bases import check_if_base_matches_host
 from charmcraft.config import Base, BasesConfiguration
 from charmcraft.cmdbase import BaseCommand, CommandError
+from charmcraft.env import is_charmcraft_running_in_managed_mode
 from charmcraft.jujuignore import JujuIgnore, default_juju_ignore
 from charmcraft.manifest import create_manifest
 from charmcraft.utils import make_executable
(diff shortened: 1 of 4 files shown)


charmcraft/commands/build.py
```diff
@@ -138,24 +140,79 @@ def __init__(self, args, config):
...
-        zipname = self.handle_package()
+        zipname = self.handle_package(bases_config)
 
         logger.info("Created '%s'.", zipname)
         return zipname
 
+    def run(self) -> List[str]:
+        """Run build process.
+
+        In managed-mode (and eventually destructive-mode), build for each bases
```
sergiusens, line 169:
```suggestion
        In managed-mode (and eventually destructive-mode), build for each base
```

charmcraft/commands/build.py
```diff
@@ -138,24 +140,79 @@ def __init__(self, args, config):
...
+        configuration which has a matching build-on to the host we are executing
+        on.  Warn for each bases configuration that is unbuildable.  Error if
```
sergiusens, line 171:
```suggestion
        on.  Warn for each base configuration that is incompatible.  Error if
```

sergiusens (approved):

charmcraft/commands/build.py
```diff
@@ -138,24 +140,79 @@ def __init__(self, args, config):
...
+        if is_charmcraft_running_in_managed_mode():
+            if self.config.bases is None:
+                # XXX: 2021-06-16 Patterson temporary until we have the default base
+                raise CommandError("Bases are currently required in managed-mode.")
+
+            for i, bases_config in enumerate(self.config.bases):
+                for j, build_on in enumerate(bases_config.build_on):
+                    matches, reason = check_if_base_matches_host(build_on)
+                    if matches:
+                        logger.debug(
```
facundobatista, line 190: This and other log lines: do NOT compose the string yourself, but pass the template string as first parameter and the arguments as next parameters...

This is for performance and robustness.

For performance because if the logger decides that the message will not be sent, the string is not built at all. 

For robustness because in the event of introducing a bug in the line, the logging infrastructure will not crash, but report the problem. For example, in this case: `logger.info("Foo %s: %d", self.quantity)`.

charmcraft/commands/build.py
```diff
@@ -138,24 +140,79 @@ def __init__(self, args, config):
...
+            if not charms:
+                raise CommandError(
+                    "No suitable 'build-on' environment found in any 'bases' configuration."
+                )
+
+        else:
+            charm_name = self.build_charm(None)
+            charms.append(charm_name)
+
+        return charms
```
facundobatista, line 214: Currently this list will always have one entry only, but you're using a list here because in the future (multiple bases in the config, using managed environments) you actually will get multiple entries, right?

facundobatista (changes requested):
Great work! A detail to fix, and a question. Thanks!!

charmcraft/commands/build.py
cjp256: Indeed, pylint usually reminds me of that ;) will fix

charmcraft/commands/build.py
cjp256: It can return multiple now, see https://github.com/canonical/charmcraft/pull/390/files#diff-cdf2ae508434a0f8c34dfa438ccb1e70776b6a499ad71e4dea2f20fbf2e15358R555

The next step will be to add `--bases-index` which will be used to constrain it to one result in real-world usage.  But for now, the behavior matches the anticipated destructive-mode.

charmcraft/commands/build.py
cjp256: I have been uses `bases configuration` to imply a unit/entry of `bases:` vs `base configuration` when specifically discussing the content of a `base` object.

charmcraft/commands/build.py
cjp256: Done!

charmcraft/commands/build.py
facundobatista: ah, right! Great

facundobatista (approved):
Great, thanks! BTW, "black" is awful :p

PR #391, opened 2021-06-16
Refactored the upload-resource command for it's new behaviour (CRAFT-240)

facundobatista:
Now it will stop using Dockerhub, and the process will be:

- check if the indicated digest is in Canonical's registry; if yes, use it

- else try to upload it; the ImageHandler function will upload it and will return the new remote digest, in which case it will be used, or None, in which case an error will be presented to the user

Once the command has the image's digest from the Canonical's Registry (because it was already there, or just uploaded), will request the formal JSON to Charmhub, which will be uploaded back to Charmhub as any other resource.

### DIFF
--- a/charmcraft/commands/store/__init__.py
+++ b/charmcraft/commands/store/__init__.py
@@ -1353,7 +1353,9 @@ class UploadResourceCommand(BaseCommand):
             resource_filepath = parsed_args.filepath
             resource_filepath_is_temp = False
             resource_type = ResourceType.file
-            logger.debug("Uploading resource directly from file %s", resource_filepath)
+            logger.debug(
+                "Uploading resource directly from file %r.", str(resource_filepath)
+            )
         elif parsed_args.image:
             image_digest = parsed_args.image
             credentials = store.get_oci_registry_credentials(
@@ -1361,7 +1363,7 @@ class UploadResourceCommand(BaseCommand):
             )
             image_name = credentials.image_name.split("/", 1)[1]
             logger.debug(
-                "Uploading resource from image %s @ %s", image_name, image_digest
+                "Uploading resource from image %s @ %s.", image_name, image_digest
             )
 
             # build the image handler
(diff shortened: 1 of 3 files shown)


charmcraft/commands/store/__init__.py
```diff
@@ -1396,26 +1355,50 @@ def run(self, parsed_args):
             resource_type = ResourceType.file
             logger.debug("Uploading resource directly from file %s", resource_filepath)
         elif parsed_args.image:
+            image_digest = parsed_args.image
+            credentials = store.get_oci_registry_credentials(
+                parsed_args.charm_name, parsed_args.resource_name
+            )
+            image_name = credentials.image_name.split("/", 1)[1]
```
sergiusens, line 1362: store.get_oci_registry_credentials' .image_name always returned a / (or validated), right?

sergiusens: from this operation it looks like the image_name is not really the image_name, can you add a code comment on why this is?

charmcraft/commands/store/__init__.py
```diff
@@ -1396,26 +1355,50 @@ def run(self, parsed_args):
...
-                "ImageName": final_resource_url,
-            }
+
+            # check if the specific image is already in Canonical's registry
+            already_uploaded = ih.check_in_registry(image_digest)
+            if already_uploaded:
+                logger.info("Using OCI image from Canonical's registry")
+            else:
+                # upload it from local registry
+                logger.info("Remote image not found, uploading from local registry")
```
sergiusens, line 1382: https://discourse.ubuntu.com/t/punctuation/18884

sergiusens: and in other places

charmcraft/commands/store/__init__.py
facundobatista: The credentials.image_name is a standard OCI registry construct, example: `'registry.jujucharms.com/charm/45kk8smbsqp2f90ssgtygsp8atuzup4biyn2e/redis-image'`.

The "image name" from the credentials is this full image name used by OCI Registries. But the "image name" that we handled internally is without the server part.

Maybe we could improve this by changing `credentials` to not hold "image_name" internally, but "full_image_name" or something...

charmcraft/commands/store/__init__.py
facundobatista: Improved here and in other parts.

charmcraft/commands/store/__init__.py
sergiusens: I suspect this is obvious as it is fresh in your head, which is why I thought a code comment for a variable/attribute that had the same name but with different comments for future self would be nice (or a different name as suggested)

sergiusens (approved):

charmcraft/commands/store/__init__.py
cmatsuoka: Using a different variable name if the nature of the content changes would be a good idea IMHO.

cmatsuoka (approved):
It seems good, but consider using a different variable name for the different types of `image_name`.

PR #392, opened 2021-06-17
pack: introduce support for --bases-index to specify targets (CRAFT-148)

cjp256:
Allow the user to specify --bases-index <index> to limit the targets
being built.  Pass this through pack's run() so we don't have to
further extend the deprecated pack's arguments.

These indices are validated at runtime because the current Validator
does not have the config available to validate against.

### DIFF
--- a/charmcraft/commands/build.py
+++ b/charmcraft/commands/build.py
@@ -437,8 +437,7 @@ class Validator:
         """Process the received options."""
         result = {}
         for opt in self._options:
-            print(parsed_args)
-            meth = getattr(self, "validate_" + opt.replace("-", "_"))
+            meth = getattr(self, "validate_" + opt)
             result[opt] = meth(getattr(parsed_args, opt, None))
         return result
 
--- a/charmcraft/commands/build.py
+++ b/charmcraft/commands/build.py
@@ -163,6 +163,19 @@ class Builder:
         logger.info("Created '%s'.", zipname)
         return zipname
 
+    def validate_bases_indices(self, bases_indices: List[int]) -> None:
+        """Validate the indices for bases.
+
+        :param bases_indeces: List of indices.
+
+        :raises CommandError: if invalid.
+        """
+        for base_index in bases_indices:
+            if base_index >= len(self.config.bases):
+                raise CommandError(
+                    f"No bases configuration found for specified index '{base_index}'."
+                )
+
     def run(self, bases_indices: Optional[List[int]] = None) -> List[str]:
         """Run build process.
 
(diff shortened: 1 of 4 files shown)


charmcraft/commands/build.py
```diff
@@ -178,12 +178,29 @@ def run(self) -> List[str]:
         """
         charms: List[str] = []
 
+        # Validate base indices, if any.
+        if bases_indices:
```
facundobatista, line 182: If possible, I'd try to validate command line parameters as soon as possible. In this case, where it can not really be done with argparse (we could validate it's >=0, but at parsing time we don't know the bases length), I'd validate this in the first lines of the Command's `run` method.

charmcraft/commands/build.py
```diff
@@ -178,12 +178,29 @@ def run(self) -> List[str]:
...
+            invalid_indices = [
+                i for i in bases_indices if i < 0 or i >= len(self.config.bases)
+            ]
+            if invalid_indices:
+                raise CommandError(
+                    f"Specified base index '{invalid_indices[0]}' is out of range."
```
facundobatista, line 188: If one element will be just shown, I'd do a simple "for and raise":
```
for i in bases_indices:
    if i < 0 or i >= len(self.config.bases):
        raise CommandError(...)
```

facundobatista (changes requested):
Asked for a little rearrangement, thanks!

charmcraft/commands/build.py
cjp256: Done!

charmcraft/commands/build.py
cjp256: Done (and moved it its own method)!

charmcraft/commands/build.py
```diff
@@ -412,10 +437,22 @@ def process(self, parsed_args):
         """Process the received options."""
         result = {}
         for opt in self._options:
-            meth = getattr(self, "validate_" + opt)
+            print(parsed_args)
```
facundobatista, line 440: leftover print

charmcraft/commands/build.py
```diff
@@ -412,10 +437,22 @@ def process(self, parsed_args):
...
+            meth = getattr(self, "validate_" + opt.replace("-", "_"))
```
facundobatista, line 441: why this replace is needed?

charmcraft/commands/build.py
```diff
@@ -412,10 +437,22 @@ def process(self, parsed_args):
...
             result[opt] = meth(getattr(parsed_args, opt, None))
         return result
 
+    def validate_bases_indices(self, bases_indices):
+        """Validate that bases index is valid."""
+        if not bases_indices:
+            return
+
+        for bases_index in bases_indices:
+            if bases_index < 0:
```
facundobatista, line 451: At this point the config is already loaded. We can totally validate here that the value is not bigger than the confgured bases...

facundobatista (changes requested):

charmcraft/commands/build.py
cjp256: Whoops, accidentally included this bit. Removed

charmcraft/commands/build.py
cjp256: Whoops, accidentally included this bit. Removed

charmcraft/commands/build.py
cjp256: Good point. I added config to Validator and combined the tests.

facundobatista (approved):
Great, thanks!

cmatsuoka (approved):
Looks good.

PR #395, opened 2021-06-17
Update the default bundle from `charmcraft init`

rbarry82:
Include examples of more relation possibilities in `metadata.yaml`, and avoid using hyphens for them in favor of underscores. Also add `display_name`

Split developer contributing instructions to a separate document so the published README doesn't have them.

Also, update the README for charmcraft itself so contributors don't have to read the source to find out why `python -m charmcraft` doesn't actually work when cloned from git

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,28 @@
+## Development environment
+
+To set up an initial development environment:
+    
+    git clone https://github.com/canonical/charmcraft.git
+    cd charmcraft
+    virtualenv venv
+    . venv/bin/activate
+    pip install -r requirements-dev.txt
+
+
+## Developing against Charmcraft source
+
+Make changes as appropriate. Some existing ideas are in the
+[Github Issues](https://github.com/canonical/charmcraft/issues)
+
+To test locally:
+
+    CHARMCRAFT_DEVELOPER=1 python -m charmcraft
+
+When you're done, make sure you run the tests.
+
+You can do so with
+
+    pip install -r requirements-dev.txt
+    ./run_tests
+
+Contributions welcome!
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,7 +6,7 @@ To set up an initial development environment:
     cd charmcraft
     virtualenv venv
     . venv/bin/activate
-    pip install -r requirements-dev.txt
+    pip install -r requirements-dev.txt -e .
 
 
 ## Developing against Charmcraft source
--- a/README.md
+++ b/README.md
@@ -144,21 +144,3 @@ latest/edge
 Use `charmcraft upload` to get a new revision number for your freshly built
 charmed operator, and `charmcraft release` to release a revision into any
 particular channel for your users.
-
-# Charmcraft source
-
-Get the source from github with:
-
-    git clone https://github.com/canonical/charmcraft.git
-    cd charmcraft
-    virtualenv venv
-    . venv/bin/activate
-    pip install -r requirements.txt
-    CHARMCRAFT_DEVELOPER=1 python -m charmcraft
-
-If you would like to run the tests you can do so with
-
-    pip install -r requirements-dev.txt
-    ./run_tests
-
-Contributions welcome!
--- a/README.md
+++ b/README.md
@@ -154,7 +154,7 @@ Get the source from github with:
     virtualenv venv
     . venv/bin/activate
     pip install -r requirements.txt
-    python -m charmcraft
+    CHARMCRAFT_DEVLOPER=1 python -m charmcraft
 
 If you would like to run the tests you can do so with
 
--- a/README.md
+++ b/README.md
@@ -154,7 +154,7 @@ Get the source from github with:
     virtualenv venv
     . venv/bin/activate
     pip install -r requirements.txt
-    CHARMCRAFT_DEVLOPER=1 python -m charmcraft
+    CHARMCRAFT_DEVELOPER=1 python -m charmcraft
 
 If you would like to run the tests you can do so with
 
--- a/charmcraft/commands/init.py
+++ b/charmcraft/commands/init.py
@@ -37,6 +37,7 @@ It will setup the following tree of files and directories:
 .
 ├── metadata.yaml        - Charm operator package description
 ├── README.md            - Frontpage for your charmhub.io/charm/
+├── CONTRIBUTING.md      - Instructions for how to build and develop your charm
 ├── actions.yaml         - Day-2 action declarations, e.g. backup, restore
 ├── config.yaml          - Config schema for your operator
 ├── LICENSE              - Your charm license, we recommend Apache 2
--- a/charmcraft/templates/init/CONTRIBUTING.md.j2
+++ b/charmcraft/templates/init/CONTRIBUTING.md.j2
@@ -8,6 +8,24 @@ Create and activate a virtualenv with the development requirements:
     source venv/bin/activate
     pip install -r requirements-dev.txt
 
+## Code overview
+
+TEMPLATE-TODO: 
+One of the most important things a consumer of your charm (or library)
+needs to know is what set of functionality it provides. Which categories
+does it fit into? Which events do you listen to? Which libraries do you
+consume? Which ones do you export and how are they used?
+
+## Intended use case
+
+TEMPLATE-TODO:
+Why were these decisions made? What's the scope of your charm?
+
+## Roadmap
+
+If this Charm doesn't fulfill all of the initial functionality you were
+hoping for or planning on, please add a Roadmap or TODO here
+
 ## Testing
 
 The Python operator framework includes a very nice harness for testing
--- a/charmcraft/templates/init/CONTRIBUTING.md.j2
+++ b/charmcraft/templates/init/CONTRIBUTING.md.j2
@@ -1,0 +1,16 @@
+# {{ name }}
+
+## Developing
+
+Create and activate a virtualenv with the development requirements:
+
+    virtualenv -p python3 venv
+    source venv/bin/activate
+    pip install -r requirements-dev.txt
+
+## Testing
+
+The Python operator framework includes a very nice harness for testing
+operator behaviour without full deployment. Just `run_tests`:
+
+    ./run_tests
--- a/charmcraft/templates/init/README.md.j2
+++ b/charmcraft/templates/init/README.md.j2
@@ -11,7 +11,7 @@ TODO: Provide high-level usage, such as required config or relations
 
 ## Relations
 
-TODO: Provide any relations which are provides or required by your charm
+TODO: Provide any relations which are provided or required by your charm
 
 ## OCI Images
 
@@ -20,4 +20,5 @@ TODO: Include a link to the default image your charm uses
 ## Contributing
 
 Please see the [Juju SDK docs](https://juju.is/docs/sdk) for guidelines 
-on enhancements to this charm following best practice guidelines.
+on enhancements to this charm following best practice guidelines, and
+`CONTRIBUTING.md` for developer guidance.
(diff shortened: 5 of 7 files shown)


charmcraft/templates/init/CONTRIBUTING.md.j2
```diff
@@ -0,0 +1,16 @@
+# {{ name }}
+
```
balbirthomas, line 2: May I make a couple of suggestions here keeping in view the following rationale -- In order to effectively contribute to a charm that has already reached a certain stage of maturity any developer will need to understand the following types of things

1. How is the code in this charm structured (a high level overview) ? For example the constructor of the charm makes clear what event are handled but it is not immediately obvious why only that particular subset of events were choosen to be handled and others ignored. This is but one example intended to highlight what is ment by code overview. Other obvious examples are -- What are the objects involved in this charm ? What are the responsibilities of these objects ? How do these objects collaborate ? Such understanding is essential for a new developer to understand the intent of the present design of the charm in order to be able to effectively extend it in the future.
2. What assumptions does it make on the use cases for this charm ? For example MongoDB charm only supports replicated databases. It does not support sharding for instance at least as yet. A developer or user should not have to read the entire charm code to understand what use cases are supported and what are not.
3. What were some key design decisions, specific to this charm ? There is a overlap here with point 2 above but this is focused on more the implementation than use cases. 

I think we will be missing a very important aspect of the documentation of charms if we do not encourage charm authors to document such "high level" things so that future contributors can easily understand this.

Whew! Hopefully I have made a case to request that the `CONTRIBUTING.md` file as a few additional headings such as

- Code overview
- Supported Use cases
- Design choices

etc.

charmcraft/templates/init/README.md.j2
```diff
@@ -9,17 +9,15 @@ TODO: Describe your charm in a few paragraphs of Markdown
...
-    source venv/bin/activate
-    pip install -r requirements-dev.txt
+## OCI Images
 
-## Testing
+TODO: Include a link to the default image your charm uses
 
-The Python operator framework includes a very nice harness for testing
-operator behaviour without full deployment. Just `run_tests`:
+## Contributing
```
balbirthomas, line 20: Perhaps also a link to `CONTRIBUTING.md` here ?

charmcraft/templates/init/README.md.j2
```diff
@@ -9,17 +9,15 @@ TODO: Describe your charm in a few paragraphs of Markdown
 TODO: Provide high-level usage, such as required config or relations
 
 
-## Developing
+## Relations
```
balbirthomas, line 12: How about just a list of supported relations here (names of charms or relation names) and a boilerplate linked suggestion to see `INTEGRATION.md` for details on relations, their usage and interfaces (data that is exchanged over the relations) ? If this is agreeable then maybe we need a template for `INTEGRATION.md` too.

charmcraft/templates/init/CONTRIBUTING.md.j2
```diff
@@ -0,0 +1,16 @@
...
+## Developing
```
balbirthomas, line 3: How about a additional heading with links to coding standards, best practice guidelines etc. ? Hopefully some day Juju SDK docs will have this ?

balbirthomas (approved):
A great start but may I request your reflections on a few additional points of consideration.

README.md
```diff
@@ -154,7 +154,7 @@ Get the source from github with:
     virtualenv venv
     . venv/bin/activate
     pip install -r requirements.txt
-    python -m charmcraft
+    CHARMCRAFT_DEVLOPER=1 python -m charmcraft
```
dstathis, line 157: DEVLOPER should be DEVELOPER

dstathis: also, what does this do?

charmcraft/templates/init/README.md.j2
```diff
@@ -9,17 +9,15 @@ TODO: Describe your charm in a few paragraphs of Markdown
...
 
-Create and activate a virtualenv with the development requirements:
+TODO: Provide any relations which are provides or required by your charm
```
dstathis, line 14: provided not provides

dstathis (dismissed):

charmcraft/commands/init.py
```diff
@@ -37,6 +37,7 @@
 .
 ├── metadata.yaml        - Charm operator package description
 ├── README.md            - Frontpage for your charmhub.io/charm/
```
mmanciop, line 39: We should specify it should be written from the PoV of the charm user

charmcraft/templates/init/README.md.j2
mmanciop: List of relations and the rationale of what they accomplish. We should not ask an end user to go read dev docu to find out what `prometheus_scrape` is for.

charmcraft/templates/init/README.md.j2
```diff
@@ -9,17 +9,15 @@ TODO: Describe your charm in a few paragraphs of Markdown
...
 
-## Testing
+TODO: Include a link to the default image your charm uses
 
-The Python operator framework includes a very nice harness for testing
-operator behaviour without full deployment. Just `run_tests`:
+## Contributing
 
-    ./run_tests
+Please see the [Juju SDK docs](https://juju.is/docs/sdk) for guidelines
```
mmanciop, line 22: “ following best practices.”

charmcraft/templates/init/README.md.j2
balbirthomas: Agreed. Perhaps then link to INTEGRATION.md from CONTRIBUTING.md ?

charmcraft/templates/init/metadata.yaml.j2
```diff
@@ -1,10 +1,25 @@
 # Copyright {{ year }} {{ author }}
 # See LICENSE file for licensing details.
 name: {{ name }}
+display-name: |
+  TEMPLATE-TODO: fill out a display name for the Charmcraft store
 description: |
   TEMPLATE-TODO: fill out the charm's description
 summary: |
   TEMPLATE-TODO: fill out the charm's summary
+tags: |
```
jnsgruk, line 10: This may not last much longer as a feature, it doesn't make a lot of sense for a charm developer to specify tags in this way imo. I'd rather not point people at this.

charmcraft/templates/init/metadata.yaml.j2
```diff
@@ -1,10 +1,25 @@
...
+  TEMPLATE-TODO: fill out the tags your charm will present
+
+# TEMPLATE-TODO: replace with any services your charm provides to other charms and the interfaces it presents them on
+provides:
```
jnsgruk, line 14: This seems silly unless the underlying charm does something with it - I think we'd be better off with a comment at the top of the metadata.yaml template that points at the latest spec in the docs

I drafted a spec in a nice template format here: https://discourse.charmhub.io/t/charm-metadata-v2/3674/15?u=jnsgruk

charmcraft/templates/init/metadata.yaml.j2
```diff
@@ -1,10 +1,25 @@
...
+  some_functionality:
+    interface: some_interface
+
+# TEMPLATE-TODO: replace with any services your charm requires from other charms and the interfaces it finds them on
+requires:
```
jnsgruk, line 19: As above

charmcraft/templates/init/README.md.j2
sergiusens: are the underscores in requires and provides mandatory or do they fall under best practices?

README.md
rbarry82: Without this, checking out the git repo and following `python -m charmcraft` immediately dies with an error log (see https://github.com/canonical/charmcraft/blob/342a2ec36c90aa6c7f199c9470256c6e18e760e7/charmcraft/env.py#L54). Seeing it permanently causes the testsuite to fail, though.

charmcraft/templates/init/README.md.j2
rbarry82: "Following best practice guidelines" is in the next line here.

charmcraft/templates/init/metadata.yaml.j2
rbarry82: This is really great. We'll link to this (on discourse) until it makes it into the official docs

charmcraft/commands/init.py
rbarry82: I don't think I'm following you here. `charmcraft init` will only be used by Charm authors.

charmcraft/templates/init/CONTRIBUTING.md.j2
rbarry82: I agree. I'll see if I can find some. For now, it's a little bit of a chicken-and-egg problem. If the Juju SDK docs don't currently have coding standards, best practice guidelines, etc, and we're actively discussing what they should be, it's hard to include them.

There can always be a follow-up patch which adds them once they're hashed out.

charmcraft/templates/init/README.md.j2
rbarry82: As with the other comment, I agree that this would be great. There's the old layer index (https://github.com/juju/layer-index), but I haven't seen a list of relation names anywhere, and Google didn't find one. This would be ideal to have once it exists, to be sure.

The same for `INTEGRATION.md`. Once we have a solid template, make it generic and add it here.

charmcraft/templates/init/metadata.yaml.j2
jnsgruk: It was my intention to make a page for it on the docs following review, I just hadn't yet got around to it ;) I thought in the 'Reference' section of the doc nav

charmcraft/templates/init/CONTRIBUTING.md.j2
jnsgruk: I'd also suggest including a Roadmap/TODO heading so people can identify if there are planned features they could contribute

README.md
```diff
@@ -154,7 +154,7 @@ Get the source from github with:
     virtualenv venv
     . venv/bin/activate
     pip install -r requirements.txt
-    python -m charmcraft
+    CHARMCRAFT_DEVELOPER=1 python -m charmcraft
```
balbirthomas, line 157: If I am not mistaken this should be in CONTRIBUTING.md.

facundobatista (approved):
I like these changes.

facundobatista:
Thanks for working on this to improve the templates!

mthaddon:
My main concern here is the interaction of charmhub with this. CONTRIBUTING.md won't get rendered on the charmhub page currently. Are there plans to address this somehow?

All the guidance we've had so far has been to keep the README.md minimal, and include a link at the bottom of it to the discourse-based documentation on charmhub's "docs" tab.

rbarry82:
You're right that this won't be rendered on the charmhub page, but the plans to address this are more or less "update the developer documentation so it's easier to get started charming". The developer instructions aren't required to download/use charms, only to build them.

The guidance so far (keep the `README.md` minimal and include a link at the bottom) is still good from my POV, just that it will be more focused now

CONTRIBUTING.md
```diff
@@ -0,0 +1,28 @@
+## Development environment
+
+To set up an initial development environment:
+    
+    git clone https://github.com/canonical/charmcraft.git
+    cd charmcraft
+    virtualenv venv
+    . venv/bin/activate
+    pip install -r requirements-dev.txt
```
cjp256, line 9: Don't forget to install Charmcraft itself (neded for user to change directory):

```suggestion
    pip install -r requirements-dev.txt -e .
```

cjp256 (approved):

facundobatista:
@rbarry82 hello! Thanks for this work! Would you solve the conflict and consider Chris' suggestion, so we can land this improvement? Thanks again!

rbarry82:
Done. Just waiting for CI

PR #399, opened 2021-06-21
Implement support for Bases in Store layer (CRAFT-298)

facundobatista:
This is to later be used by the `status` command.


cmatsuoka (approved):
Looks good.

sergiusens (approved):

PR #400, opened 2021-06-21
config: default path to /root/project when in managed mode (CRAFT-300)

cjp256:
Add get_managed_environment_project_path() to get managed directory
path.  When in managed-mode, default to it.  Note that it could still
be overwritten by using --project-dir.

### DIFF
--- a/charmcraft/config.py
+++ b/charmcraft/config.py
@@ -63,6 +63,7 @@ import pydantic
 
 from charmcraft.cmdbase import CommandError
 from charmcraft.deprecations import notify_deprecation
+from charmcraft.env import is_charmcraft_running_in_managed_mode, get_managed_environment_project_path
 from charmcraft.utils import get_host_architecture, load_yaml
 
 
@@ -349,7 +350,10 @@ class Config(ModelConfigDefaults, validate_all=False):
 def load(dirpath):
     """Load the config from charmcraft.yaml in the indicated directory."""
     if dirpath is None:
-        dirpath = pathlib.Path.cwd()
+        if is_charmcraft_running_in_managed_mode():
+            dirpath = get_managed_environment_project_path()
+        else:
+            dirpath = pathlib.Path.cwd()
     else:
         dirpath = pathlib.Path(dirpath).expanduser().resolve()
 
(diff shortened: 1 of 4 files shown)


cmatsuoka (approved):

charmcraft/env.py
```diff
@@ -19,11 +19,17 @@
...
 import os
+import pathlib
 import sys
 
 from charmcraft.cmdbase import CommandError
 
 
+def get_managed_environment_project_path():
+    """Path for project when running in managed environment."""
+    return pathlib.Path("/root/project")
```
facundobatista, line 30: I wonder if this "constant" shouldn't be provided by the `craft_providers` lib.

charmcraft/config.py
```diff
@@ -349,7 +353,10 @@ def schema(cls, **kwargs) -> Dict[str, Any]:
 def load(dirpath):
     """Load the config from charmcraft.yaml in the indicated directory."""
     if dirpath is None:
-        dirpath = pathlib.Path.cwd()
+        if is_charmcraft_running_in_managed_mode():
+            dirpath = get_managed_environment_project_path()
+        else:
+            dirpath = pathlib.Path.cwd()
```
facundobatista, line 359: We need two tests for both these cases.

facundobatista (changes requested):
Nice! A question and two tests needed. Thanks!

charmcraft/env.py
cjp256: The existence of this directory is due to application logic, even if consistent with Snapcraft, so I'd rather not introduce it to craft-providers.  However the use of `/root` is an assumption that craft-providers could perhaps break.  It could be moved into charmcraft.providers.CharmcraftBuilddBaseConfiguration as part of the contract, or a runtime call to query the home directory - but doesn't really help much here.

charmcraft/config.py
cjp256: Ah, forgot to include those, will do.

cjp256: Done!

charmcraft/env.py
facundobatista: I now understood how this is used.

Approving, but consider: maybe this should be a module constant?

```
MANAGED_PROJECT_PATH = pathlib.Path("/root/project")
```

What's the point of a function whose only purpose return that?

facundobatista (approved):

cjp256:
FWIW i made this a function in case we want to make it more dynamic or something in the future. Granted there'd probably be other refactoring required anyhow.

PR #401, opened 2021-06-21
build: require metadata.yaml and parse on Builder init (CRAFT-301)

cjp256:
metadata.yaml is read during handle_package().  Instead, read and
parse the metadata on Builder's init, allowing for re-use across
the class. Add explicit error if metadata.yaml is missing and add
test case (Charmcraft currently fails with FileNotFoundError).


cmatsuoka (approved):
Looks good.

PR #402, opened 2021-06-22
providers: introduce charmcraft base configuration (CRAFT-261)

cjp256:
- Add craft-providers dependency (version will be updated in
  future work) to project & snap.

- Add CharmcraftBuilddBaseConfiguration which extends the buildd
  image configuration with the setup code to:

  - install python3-pip

  - install python3-setuptools (required explicitly for 18.04)

  - install the Charmcraft snap by injecting the host's, which
    should be OK as the snap is currently required.  Future work
    could extend this to add a fallback to install from store using
    snap_installer.install_from_store().  Note that this injection
    is unconditional, future work will add a check to see if the
    injection is necessary prior to doing it (that is, don't inject
    if reusing an instance that has the correct version already).

### DIFF
--- a/charmcraft/providers.py
+++ b/charmcraft/providers.py
@@ -1,0 +1,122 @@
+# Copyright 2021 Canonical Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# For further info, check https://github.com/canonical/charmcraft
+
+"""Build environment provider support for charmcraft."""
+
+import logging
+import subprocess
+from typing import Optional
+
+from craft_providers import Executor, bases
+from craft_providers.actions import snap_installer
+
+logger = logging.getLogger(__name__)
+
+
+class CharmcraftBuilddBaseConfiguration(bases.BuilddBase):
+    """Base configuration for Charmcraft.
+
+    :cvar compatibility_tag: Tag/Version for variant of build configuration and
+        setup.  Any change to this version would indicate that prior [versioned]
+        instances are incompatible and must be cleaned.  As such, any new value
+        should be unique to old values (e.g. incrementing).
+    :cvar instance_config_path: Path to persistent environment configuration
+        used for compatibility checks (or other data).  Set to
+        /etc/craft-instance.conf, but may be overridden for application-specific
+        reasons.
+    :cvar instance_config_class: Class defining instance configuration.  May be
+        overridden with an application-specific subclass of InstanceConfiguration
+        to enable application-specific extensions.
+
+    :param alias: Base alias / version.
+    :param environment: Environment to set in /etc/environment.
+    :param hostname: Hostname to configure.
+    """
+
+    compatibility_tag: str = f"charmcraft-{bases.BuilddBase.compatibility_tag}.0"
+
+    def setup(
+        self,
+        *,
+        executor: Executor,
+        retry_wait: float = 0.25,
+        timeout: Optional[float] = None,
+    ) -> None:
+        """Prepare base instance for use by the application.
+
+        Wait for environment to become ready and configure it.  At completion of
+        setup, the executor environment should have networking up and have all
+        of the installed dependencies required for subsequent use by the
+        application.
+
+        Setup may be called more than once in a given instance to refresh/update
+        the environment.
+
+        If timeout is specified, abort operation if time has been exceeded.
+
+        Guarantees provided by buildd's setup:
+
+            - configured /etc/environment
+
+            - configured hostname
+
+            - networking available (IP & DNS resolution)
+
+            - apt cache up-to-date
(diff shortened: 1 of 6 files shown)


requirements-dev.txt
```diff
@@ -6,6 +6,7 @@ cffi==1.14.5
 chardet==4.0.0
 click==8.0.1
 coverage==5.5
+craft-providers==0.0.1
```
cmatsuoka, line 9: We mentioned this before and I don't remember what was decided back then. Craft-providers is GPLv3 and that should contaminate Charmcraft so it would need to be relicensed under GPLv3 as well -- or is there a different way to address that?

cjp256: For those curious:
```
$ pip-licenses 
 Name                 Version                    License                                             
 Jinja2               3.0.1                      BSD License                                         
 MarkupSafe           2.0.1                      BSD License                                         
 PyNaCl               1.4.0                      Apache License 2.0                                  
 PyYAML               5.4.1                      MIT License                                         
 appdirs              1.4.4                      MIT License                                         
 attrs                21.2.0                     MIT License                                         
 black                21.6b0                     MIT License                                         
 certifi              2021.5.30                  Mozilla Public License 2.0 (MPL 2.0)                
 cffi                 1.14.5                     MIT License                                         
 chardet              4.0.0                      GNU Library or Lesser General Public License (LGPL) 
 charmcraft           0.10.0+224.gb8b8ed2.dirty  Apache Software License                             
 click                8.0.1                      BSD License                                         
 coverage             5.5                        Apache Software License                             
 craft-providers      0.0.1                      GNU General Public License v3 (GPLv3)               
 flake8               3.9.2                      MIT License                                         
 humanize             3.9.0                      MIT License                                         
 idna                 2.10                       BSD License                                         
 iniconfig            1.1.1                      MIT License                                         
 jsonschema           3.2.0                      MIT License                                         
 macaroonbakery       1.3.1                      GNU Lesser General Public License v3 (LGPLv3)       
 mccabe               0.6.1                      MIT License                                         
 mypy-extensions      0.4.3                      MIT License                                         
 ops                  1.2.0                      Apache Software License                             
 packaging            20.9                       Apache Software License, BSD License                
 pathspec             0.8.1                      Mozilla Public License 2.0 (MPL 2.0)                
 pkg-resources        0.0.0                      UNKNOWN                                             
 pluggy               0.13.1                     MIT License                                         
 protobuf             3.17.3                     3-Clause BSD License                                
 py                   1.10.0                     MIT License                                         
 pyRFC3339            1.1                        MIT License                                         
 pycodestyle          2.7.0                      MIT License                                         
 pycparser            2.20                       BSD License                                         
 pydantic             1.8.2                      MIT License                                         
 pydocstyle           6.1.1                      MIT License                                         
 pyflakes             2.3.1                      MIT License                                         
 pymacaroons          0.13.0                     MIT License                                         
 pyparsing            2.4.7                      MIT License                                         
 pyrsistent           0.17.3                     MIT License                                         
 pytest               6.2.4                      MIT License                                         
 python-dateutil      2.8.1                      BSD License, Apache Software License                
 pytz                 2021.1                     MIT License                                         
 regex                2021.4.4                   Apache Software License                             
 requests             2.25.1                     Apache Software License                             
 requests-toolbelt    0.9.1                      Apache Software License                             
 requests-unixsocket  0.2.0                      Apache Software License                             
 responses            0.13.3                     Apache 2.0                                          
 six                  1.16.0                     MIT License                                         
 snowballstemmer      2.1.0                      BSD License                                         
 tabulate             0.8.9                      MIT License                                         
 toml                 0.10.2                     MIT License                                         
 typing-extensions    3.10.0.0                   Python Software Foundation License                  
 urllib3              1.26.5                     MIT License  
 ```

Looks like the two projects using GPL are actually LGPL.

File: charmcraft/providers.py [hunk 2]
```diff
@@ -0,0 +1,124 @@
...
+        /etc/craft-instance.conf, but may be overridden for application-specific
+        reasons.
+    :cvar instance_config_class: Class defining instance configuration.  May be
+        overridden with an application-specific subclass of InstanceConfiguration
+        to enable application-specific extensions.
+
+    :param alias: Base alias / version.
+    :param environment: Environment to set in /etc/environment.
+    :param hostname: Hostname to configure.
+    """
```
facundobatista, line 47: I find weird to have all this documentation about variables and __init__ parameters when they are not really used here.

I understand that because of inheriting `bases.BuilddBase`, the _init__ params are the same... but the doc is super decoupled from the code! If the code in `bases.BuilddBase.__init__` is changed in the future nobody will remember to update the code here, for example...

requirements-dev.txt
facundobatista: But @cmatsuoka 's point is still valid, right?

facundobatista (changes requested):
Asked a couple of questions...

requirements-dev.txt
cmatsuoka: LGPL is fine, but GPL looks problematic to me.

charmcraft/providers.py
cjp256: Done! Updated the docs to remove things not specific to Charmcraft.

requirements-dev.txt
cjp256: I'll defer to @sergiusens, I don't think we had a final resolution on the previous discussion.  We can change the license for craft-providers if desired.

cjp256: I bumped the version to 0.0.2 with the corrected license.

cmatsuoka (approved):

facundobatista (approved):
Looks great! Thanks!

PR #403, opened 2021-06-23
Added bases column in the `status` command output (CRAFT-299).

### LINKED ISSUE #129: Handle multiple platforms for channel map in status
If a charm support multiple series (platforms), there could be different released revisions for each of them (eg. revision 1 released to (`stable`, `xenial`) and revision 2 to (`stable`, `focal`)). FWIW, this is similar for snaps, snapcraft and multiple architectures

`charmcraft` is currently rendering one released value for each channel, ignoring the platform.

facundobatista:
Note about some `all_tracks`-related code deletion: we're now Python 3.6+, so we can rely on dict's order.

Fixes #129.

### DIFF
--- a/tests/commands/test_store_commands.py
+++ b/tests/commands/test_store_commands.py
@@ -958,11 +958,11 @@ def test_status_simple_ok(caplog, store_mock, config):
     ]
 
     expected = [
-        "Track    Channel    Version       Revision",
-        "latest   stable     v7            7",
-        "         candidate  v7            7",
-        "         beta       2.0           80",
-        "         edge       git-0db35ea1  156",
+        "Track    Base                  Channel    Version       Revision",
+        "latest   ubuntu 20.04 (amd64)  stable     v7            7",
+        "                               candidate  v7            7",
+        "                               beta       2.0           80",
+        "                               edge       git-0db35ea1  156",
     ]
     assert expected == [rec.message for rec in caplog.records]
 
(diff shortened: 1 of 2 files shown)


sergiusens (approved):

cmatsuoka (approved):

tests/commands/test_store_commands.py
```diff
@@ -1331,12 +1331,205 @@ def test_status_with_resources_and_branches(caplog, store_mock, config):
...
+    args = Namespace(name="testcharm")
+    StatusCommand("group", config).run(args)
+
+    assert store_mock.mock_calls == [
+        call.list_releases("testcharm"),
+    ]
+
+    expected = [
+        "Track    Base                  Channel        Version       Revision    Resources     Expires at",  # NOQA
+        "latest   ubuntu 20.04 (amd64)  stable         -             -           -",
```
cjp256, line 1514: If a charm supports multiple arches, does it get its own entry even though its the same revision?  Or can "Base" support multiple arch?

cjp256 (approved):

tests/commands/test_store_commands.py
facundobatista: It gets its own entry, you end up getting everything alphabetically sorted and easy to find. In the future we may want to change it, if real usage proves better to "combine" it.

PR #406, opened 2021-06-24
providers: introduce get_command_environment() to build env (CRAFT-262)

cjp256:
Add get_command_environment() method to build and return the command
environment to use when executing commands in the provided container.

Starting with the default buildd environment, add:

- CHARMCRAFT_MANAGED_MODE=1

- [http|https|no]_proxy settings if found in current Charmcraft process.

### DIFF
--- a/charmcraft/providers.py
+++ b/charmcraft/providers.py
@@ -17,13 +17,27 @@
 """Build environment provider support for charmcraft."""
 
 import logging
+import os
 import subprocess
-from typing import Optional
+from typing import Dict, Optional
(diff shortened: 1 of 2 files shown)


facundobatista (approved):
Like it!

sergiusens (approved):
Nice, a test to see that "random" env vars are not captured would be nice :-)

cjp256:
> Nice, a test to see that "random" env vars are not captured would be nice :-)

Good point! Done!

PR #408, opened 2021-06-24
metadata: refactor parsing into own module (CRAFT-311)

cjp256:
Move metadata logic into its own module, migrating to pydantic
and re-using bits found for charmcraft.yaml config parser.

This will allow for re-use of metadata for commands other than
build/pack.


facundobatista (approved):
Nice!

cmatsuoka (approved):

PR #409, opened 2021-06-25
providers: introduce launch_environment() (CRAFT-312)

cjp256:
Launch lxc container and configure for use with charmcraft,
adding/configuring the buildd remote as necessary.

### DIFF
--- a/charmcraft/providers.py
+++ b/charmcraft/providers.py
@@ -110,7 +110,7 @@ def get_command_environment() -> Dict[str, str]:
 
 
 @contextlib.contextmanager
-def launch_environment(
+def launched_environment(
     *,
     charm_name: str,
     project_path: pathlib.Path,
--- a/charmcraft/providers.py
+++ b/charmcraft/providers.py
@@ -16,15 +16,19 @@
 
 """Build environment provider support for charmcraft."""
 
+import contextlib
 import logging
 import os
+import pathlib
 import subprocess
 from typing import Dict, Optional, Tuple, Union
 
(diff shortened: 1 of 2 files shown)


charmcraft/providers.py
```diff
@@ -105,6 +109,66 @@ def get_command_environment() -> Dict[str, str]:
...
+
+    # Mount project.
+    instance.mount(
+        host_source=project_path, target=get_managed_environment_project_path()
+    )
+
+    yield instance
+
+    # Unmount everything.
+    instance.unmount_all()
```
sergiusens, line 167: if instance raises, this won't be run, neither will stop, maybe add a test for that.

cjp256: Ohh, good point.  I'll put a finally clause here and add a test.

charmcraft/providers.py
```diff
@@ -105,6 +109,66 @@ def get_command_environment() -> Dict[str, str]:
     return env
 
 
+@contextlib.contextmanager
+def launch_environment(
```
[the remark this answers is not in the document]
cjp256: Agreed, will do.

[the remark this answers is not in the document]
cjp256: Done

charmcraft/providers.py
cjp256: Done!

sergiusens (approved):

cmatsuoka (approved):
+1

PR #410, opened 2021-06-25
providers: introduce clean_project_environments() (CRAFT-314)

cjp256:
Provide interface for upcoming 'clean' command to clean all
environments created for a given project.

### DIFF
--- a/charmcraft/providers.py
+++ b/charmcraft/providers.py
@@ -18,10 +18,11 @@
 
 import logging
 import os
+import re
 import subprocess
-from typing import Dict, Optional, Tuple, Union
+from typing import Dict, List, Optional, Tuple, Union
 
-from craft_providers import Executor, bases
+from craft_providers import Executor, bases, lxd
 from craft_providers.actions import snap_installer
 
 from charmcraft.config import Base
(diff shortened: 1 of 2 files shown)


charmcraft/providers.py
```diff
@@ -105,6 +106,37 @@ def get_command_environment() -> Dict[str, str]:
...
+    :param lxd_project: Name of LXD project.
+    :param lxd_remote: Name of LXD remote.
+
+    :returns: List of containers deleted.
+    """
+    deleted: List[str] = []
+    lxc = lxd.LXC()
+
+    for name in lxc.list_names(project=lxd_project, remote=lxd_remote):
+        match_regex = f"^charmcraft-{charm_name}-.*-.*-.*$"
```
cmatsuoka, line 127: I wonder if the `.*` between dashes wouldn't be better matched using `[^-]+` (unless `charmcraft-name---` is a valid name).

cjp256: Fair point, I updated it to `r"^charmcraft-{charm_name}-.+-.+.+*$" ` which should address this case. Thanks!

sergiusens (approved):

cmatsuoka (approved):

PR #411, opened 2021-06-25
build: pack charm in provided environment when using bases (CRAFT-313)

cjp256:
For each bases configuration, check if a build-on base can be
provided.  If so, launch environment and execute pack command
inside each to build charm for each viable bases config index.

Future work should extend the pack command to include the parent's
verbose or quiet mode flags for consistency.

### DIFF
--- a/charmcraft/commands/build.py
+++ b/charmcraft/commands/build.py
@@ -25,13 +25,19 @@ import subprocess
 import zipfile
 from typing import List, Optional
 
+from craft_providers import Executor
+
 from charmcraft.bases import check_if_base_matches_host
 from charmcraft.cmdbase import BaseCommand, CommandError
 from charmcraft.config import Base, BasesConfiguration, Config
-from charmcraft.env import is_charmcraft_running_in_managed_mode
+from charmcraft.env import (
+    get_managed_environment_project_path,
+    is_charmcraft_running_in_managed_mode,
+)
 from charmcraft.jujuignore import JujuIgnore, default_juju_ignore
 from charmcraft.manifest import create_manifest
 from charmcraft.metadata import parse_metadata_yaml
+from charmcraft.providers import is_base_providable, launched_environment
 from charmcraft.utils import make_executable
 
 logger = logging.getLogger(__name__)
@@ -177,11 +183,7 @@ class Builder:
         """
         charms: List[str] = []
 
-        if is_charmcraft_running_in_managed_mode():
-            if self.config.bases is None:
-                # XXX: 2021-06-16 Patterson temporary until we have the default base
-                raise CommandError("Bases are currently required in managed-mode.")
-
+        if self.config.bases:
             for i, bases_config in enumerate(self.config.bases):
                 if bases_indices and i not in bases_indices:
                     logger.debug(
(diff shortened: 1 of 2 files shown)


charmcraft/commands/build.py
```diff
@@ -191,14 +193,29 @@ def run(self, bases_indices: Optional[List[int]] = None) -> List[str]:
...
                         )
-                        charm_name = self.build_charm(bases_config)
+                        if is_charmcraft_running_in_managed_mode():
+                            charm_name = self.build_charm(bases_config)
+                        else:
+                            with launched_environment(
+                                charm_name=self.metadata.name,
+                                project_path=self.charmdir,
+                                base=build_on,
+                                bases_index=i,
```
sergiusens, line 214: personal preference comment, and I think I made it in the past in a different PR, but i and j should only be used for simple iterations, it's still used 10 lines below and hard to keep track of

charmcraft/commands/build.py
```diff
@@ -225,6 +242,22 @@ def run(self, bases_indices: Optional[List[int]] = None) -> List[str]:
 
         return charms
 
+    def build_charm_in_instance(self, instance: Executor, bases_index: int) -> str:
```
sergiusens, line 245: I would use a name different than build to not confuse in the future when parts lifecycle is introduced

charmcraft/commands/build.py
```diff
@@ -177,11 +183,7 @@ def run(self, bases_indices: Optional[List[int]] = None) -> List[str]:
         """
         charms: List[str] = []
 
-        if is_charmcraft_running_in_managed_mode():
-            if self.config.bases is None:
-                # XXX: 2021-06-16 Patterson temporary until we have the default base
-                raise CommandError("Bases are currently required in managed-mode.")
-
+        if self.config.bases:
             for i, bases_config in enumerate(self.config.bases):
```
sergiusens, line 187: given the size of this, can we separate selection/planning from execution?

charmcraft/commands/build.py
cjp256: Good point, will fix!

charmcraft/commands/build.py
cjp256: How about s/build/pack?

charmcraft/commands/build.py
cjp256: Sounds good, will do!

charmcraft/commands/build.py
cjp256: Done

charmcraft/commands/build.py
cjp256: Done, renamed to pack_charm_in_instance()

charmcraft/commands/build.py
```diff
@@ -177,41 +183,54 @@ def run(self, bases_indices: Optional[List[int]] = None) -> List[str]:
...
         charms: List[str] = []
 
-        if is_charmcraft_running_in_managed_mode():
-            if self.config.bases is None:
-                # XXX: 2021-06-16 Patterson temporary until we have the default base
-                raise CommandError("Bases are currently required in managed-mode.")
-
-            for i, bases_config in enumerate(self.config.bases):
-                if bases_indices and i not in bases_indices:
+        if self.config.bases:
```
facundobatista, line 186: Shouldn't we always have "bases"? Even if not mandatory yet, some default?

cjp256: Yes.  But that PR is upcoming.  At the moment this keeps behavior consistent.

facundobatista (approved):
Great, thanks!

sergiusens (approved):
Looks better, but I don't see my comment https://github.com/canonical/charmcraft/pull/411#discussion_r659027840 addressed. It's fine if it is a different PR.

PR #412, opened 2021-06-28
clean: introduce clean command (CRAFT-62)

cjp256:
Initial usage will clean any LXD containers created by Charmcraft
to build the project.  Future work will further extend functionality.

Add a caplog filter fixture to help with a common pattern of checking
logs for the module under test.

### DIFF
--- a/charmcraft/commands/clean.py
+++ b/charmcraft/commands/clean.py
@@ -1,0 +1,49 @@
+# Copyright 2021 Canonical Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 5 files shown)


sergiusens (approved):

charmcraft/commands/clean.py
```diff
@@ -0,0 +1,49 @@
...
+"""
+
+
+class CleanCommand(BaseCommand):
+    """Clean project artifacts."""
+
+    name = "clean"
+    help_msg = "Purge project artifacts"
+    overview = _overview
+    common = True
```
facundobatista, line 40: This will be something people use a lot?

facundobatista (approved):
Nice! Added a question.

charmcraft/commands/clean.py
cjp256: I'd imagine so...  it's used a fair bit on the Snapcraft side.

PR #414, opened 2021-06-28
config/build: set default bases configuration with DNs (CRAFT-265, CRAFT-266)

cjp256:
Set the default bases configuration to Ubuntu 20.04 for the host's
architecture.  This will effectively end legacy builds.

Add DN02 to deprecate projects without charmcraft.yaml.

Add DN03 to deprecate projects with bases configuration.

Add reset_deprecation_notices() autouse fixture to reset global value
found in deprecations module.

### DIFF
--- a/charmcraft/commands/build.py
+++ b/charmcraft/commands/build.py
@@ -146,7 +146,7 @@ class Builder:
         self.config = config
         self.metadata = parse_metadata_yaml(self.charmdir)
 
-    def build_charm(self, bases_config: Optional[BasesConfiguration]) -> str:
+    def build_charm(self, bases_config: BasesConfiguration) -> str:
         """Build the charm.
 
         :param bases_config: Bases configuration to use for build.
--- a/charmcraft/commands/build.py
+++ b/charmcraft/commands/build.py
@@ -185,7 +185,7 @@ class Builder:
         charms: List[str] = []
         is_managed_mode = is_charmcraft_running_in_managed_mode()
 
-        if not (self.charmdir / "charmcraft.yaml").exists() and not is_managed_mode:
+        if not (self.charmdir / "charmcraft.yaml").exists():
             notify_deprecation("dn02")
 
         for bases_index, bases_config in enumerate(self.config.bases):
--- a/charmcraft/commands/build.py
+++ b/charmcraft/commands/build.py
@@ -30,6 +30,7 @@ from craft_providers import Executor
 from charmcraft.bases import check_if_base_matches_host
 from charmcraft.cmdbase import BaseCommand, CommandError
 from charmcraft.config import Base, BasesConfiguration, Config
+from charmcraft.deprecations import notify_deprecation
 from charmcraft.env import (
     get_managed_environment_project_path,
     is_charmcraft_running_in_managed_mode,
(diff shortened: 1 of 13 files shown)


charmcraft/commands/build.py
```diff
@@ -182,65 +183,64 @@ def run(self, bases_indices: Optional[List[int]] = None) -> List[str]:
         :returns: List of charm files created.
         """
         charms: List[str] = []
+        is_managed_mode = is_charmcraft_running_in_managed_mode()
 
-        if self.config.bases:
-            for bases_index, bases_config in enumerate(self.config.bases):
-                if bases_indices and bases_index not in bases_indices:
+        if not (self.charmdir / "charmcraft.yaml").exists() and not is_managed_mode:
+            notify_deprecation("dn02")
```
facundobatista, line 189: This deprecation should come from `config.py`, which is the one handling `charmcraft.yaml`. I'm not sure why shouldn't be shown in managed mode...

charmcraft/config.py
```diff
@@ -328,7 +335,22 @@ def unmarshal(cls, obj: Dict[str, Any], project: Project):
...
+                # Set default bases to Ubuntu 20.04 to match strict snap's
+                # effective behavior.
+                bases = [
+                    {
+                        "name": "ubuntu",
+                        "channel": "20.04",
+                        "architectures": [get_host_architecture()],
+                    }
+                ]
+                if not is_charmcraft_running_in_managed_mode():
```
facundobatista, line 348: Same here. Why we don't want the message in managed mode?

charmcraft/commands/build.py
```diff
@@ -182,65 +183,64 @@ def run(self, bases_indices: Optional[List[int]] = None) -> List[str]:
...
-                raise CommandError(
-                    "No suitable 'build-on' environment found in any 'bases' configuration."
+            else:
+                logger.warning(
+                    "No suitable 'build-on' environment found in 'bases[%d]' configuration.",
+                    bases_index,
                 )
 
-        else:
-            charm_name = self.build_charm(None)
```
facundobatista, line 242: `build_charm` and `create_manifest` shouldn't accept None as bases anymore...

charmcraft/config.py
```diff
@@ -328,7 +335,22 @@ def unmarshal(cls, obj: Dict[str, Any], project: Project):
...
+                    notify_deprecation("dn03")
+
+            # Expand short-form bases if only the bases is a valid list. If it
+            # is not a valid list, parse_obj() will properly handle the error.
+            if isinstance(bases, list):
```
facundobatista, line 353: `bases` will always be a list, right?

facundobatista (changes requested):
Nice! Annotated some details inside.

charmcraft/deprecations.py
```diff
@@ -33,6 +33,8 @@
 # documentation)
 _DEPRECATION_MESSAGES = {
     "dn01": "Configuration keywords are now separated using dashes.",
+    "dn02": "A charmcraft.yaml configuration file is now required.",
+    "dn03": "Bases configuration is now required.",
```
sergiusens, line 37: Meta question: is it really required if not erroring?

Also,

- do we want charmcraft init to add a charmcraft.yaml?
- do we want to specify what it is being set to as a default in the message or documentation entry?

charmcraft/config.py
sergiusens: this is to not have the message printed twice, once from the outer shell and later in the inner run (maybe a code comment can help)

cjp256: Exactly

charmcraft/config.py
cjp256: Should be, but this is prior to the proper error checking which would happen just after this.

charmcraft/deprecations.py
cjp256: (1) yes, init should add a charmcraft.yaml, I can add that to the list
(2) it is in the documentation at the URL (https://discourse.charmhub.io/t/charmcraft-deprecation-notices/4652)

charmcraft/commands/build.py
cjp256: i put it here (for now at least) because of the statement in config that it can load for commands which do not require charmcraft.yaml.  I didn't think we would want DNs spewing for commands other than build/pack for now.

charmcraft/commands/build.py
cjp256: agreed

cjp256: Done!

charmcraft/deprecations.py
cjp256: Done (1) @ https://github.com/canonical/charmcraft/pull/415

charmcraft/config.py
facundobatista: Do we want *any* deprecation message in managed mode? Maybe the correct place for this "if" is inside the `notify_deprecation` function.

charmcraft/commands/build.py
facundobatista: Oh, you're right.

charmcraft/manifest.py
```diff
@@ -32,33 +32,22 @@
...
-        channel = os_platform.release
-        bases = [
-            {
-                "name": name,
-                "channel": channel,
-                "architectures": architectures,
-            }
-        ]
-    else:
+    if bases_config is not None:
```
facundobatista, line 50: Please add a comment here that this may be the case for bundles, thanks!

facundobatista (changes requested):
Two small details left.

charmcraft/config.py
cjp256: Oh, that's a better idea, I'll do that!

charmcraft/manifest.py
cjp256: Will do

cjp256: Done!

charmcraft/config.py
cjp256: Done!

sergiusens (approved):

facundobatista (approved):
Wonderful, thanks!

PR #415, opened 2021-06-28
init command: populate default charmcraft.yaml (CRAFT-330)

cjp256:
Initialize charmcraft.yaml with type "charm" and an Ubuntu 20.04
bases configuration.


cjp256:
Requires #414

charmcraft/templates/init/charmcraft.yaml.j2
```diff
@@ -0,0 +1,9 @@
+# Learn more about config at: https://juju.is/docs/sdk/config
```
jnsgruk, line 1: This link should probably be: https://juju.is/docs/sdk/charmcraft-config

cjp256: Yes indeed, thanks! Done!

sergiusens (approved):

facundobatista (approved):
Like it!

PR #416, opened 2021-06-29
Created the Language checker, first step in the analyze/lint structure (CRAFT-285)

facundobatista:


charmcraft/linters.py
```diff
@@ -0,0 +1,64 @@
...
+    Currently only Python is detected, if the following checks are true:
+
+    - the charm has a text dispatch with a python call
+    - the charm has a `.py` entry point
+    - the entry point file is executable
+    """
+
+    check_type = CheckType.trait
+    name = "language"
+    url = "https://juju.is/docs/sdk/charmcraft-analyze#heading--language"  # FIXME: confirm url
```
sergiusens, line 40: I would make individual pages from this, like
```suggestion
    url = "https://juju.is/docs/sdk/charmcraft-analyze/language"
```

charmcraft/linters.py
```diff
@@ -0,0 +1,64 @@
...
+    text = "The charm is written with Python."
+
+    # different result constatns
+    PYTHON = "python"
```
sergiusens, line 44: enums in python are pretty good these days, maybe group them with that like on https://github.com/canonical/craft-parts/blob/main/craft_parts/actions.py ?

charmcraft/linters.py
```diff
@@ -0,0 +1,64 @@
...
+
+    @classmethod
+    def run(cls, basedir):
+        """Run the proper verifications."""
+        # get the entrypoint from the last useful dispatch line
+        dispatch = basedir / "dispatch"
+        entrypoint_str = ""
+        if dispatch.exists():
+            last_line = None
+            for line in dispatch.open("rt", encoding="utf8"):
```
sergiusens, line 55: should there be a try/expect for encoding here?

charmcraft/linters.py
```diff
@@ -0,0 +1,64 @@
...
+    - the charm has a `.py` entry point
+    - the entry point file is executable
+    """
+
+    check_type = CheckType.trait
+    name = "language"
+    url = "https://juju.is/docs/sdk/charmcraft-analyze#heading--language"  # FIXME: confirm url
+    text = "The charm is written with Python."
+
+    # different result constatns
```
sergiusens, line 43:
```suggestion
    # different result constants
```

facundobatista: Yes, already spotted this one, thanks!

charmcraft/linters.py
facundobatista: Working on this, let's see what we can achieve with the documentation system..

charmcraft/linters.py
facundobatista: You say to build something like the following?
```
class Language:
    ....
    class Results(enum.Enum):
        PYTHON = "python"
        UNKNOWN = "unknown"
    ...
```
Then, to use it we would need to do:
```
>>> Language.Results.PYTHON.value
'python'
```

It's more verbose, but what I really dislike is having to access the `.value` attribute.

If you really want to group them, I'd use:
```
class Language:
    ....
    Result = namedtuple("Result", "python unknown")(python="python", unknown="unkwnon")
    ...
```
Then, it's just:
```
>>> Language.Result.python
'python'
```

What do you think?

sergiusens: namedtuple gets the job done too and is used across this code base so yeah, +1

facundobatista: Improved.

charmcraft/linters.py
facundobatista: Good catch! Also added support for permission errors.

charmcraft/linters.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+class Language:
+    """Check the language used to write the charm.
+
+    Currently only Python is detected, if the following checks are true:
+
+    - the charm has a text dispatch with a python call
+    - the charm has a `.py` entry point
+    - the entry point file is executable
```
cjp256, line 36: If directly executable, it should have an interpreter that should perhaps be validated?

charmcraft/linters.py
```diff
@@ -0,0 +1,67 @@
...
+                for line in fh:
+                    if line.strip():
+                        last_line = line
+                if last_line:
+                    entrypoint_str = shlex.split(last_line)[-1]
+        except (IOError, UnicodeDecodeError):
+            return cls.Result.unknown
+
+        entrypoint = basedir / entrypoint_str
+        if entrypoint.suffix == ".py" and os.access(entrypoint, os.X_OK):
```
cjp256, line 65: Should each check have its own error id?

charmcraft/linters.py
facundobatista: We assume that if it's a .py you put the proper shebang line inside. What would be a "valid interpreter"? `/usr/local/bin/python`? `/bin/python3.7`? `/usr/bin/pypy`? 

This fits more in the "ensure it will be run under Juju environment" than "detect if it's Python", probably we could add it in the future if needed, as another linter.

charmcraft/linters.py
facundobatista: The output to the user will be `language: python` or `language: unknown`, so no (we only provide "reasons" for warnings and errors)

charmcraft/linters.py
cjp256: Fair enough :+1:

charmcraft/linters.py
cjp256: Fair enough.  Though the executable bit seems to be out of scope for "python" vs "unknown".  Most python files aren't +x.

cjp256: (note that charmcraft/main.py is not+x ;))

facundobatista: This is not about "any possibility", but "recommended way of doing things". I can change the rules here, but we should change [the spec](https://discourse.charmhub.io/t/proposal-charmcraft-analyze/4792) first, so you should carry this discussion to there, if you want.

sergiusens (approved):

sergiusens (approved):
This new approach looks a lot easier to follow as well

cjp256 (approved):

PR #418, opened 2021-06-29
build/pack: pass --verbose or --quiet as needed (CRAFT-322)

cjp256:
When building with provided (managed) instance, append --verbose
or --quiet as required to match the parameter given to the host
charmcraft instance.

### DIFF
--- a/tests/commands/test_build.py
+++ b/tests/commands/test_build.py
@@ -78,16 +78,6 @@ def basic_project(tmp_path):
     yield tmp_path
 
 
-@pytest.fixture(autouse=True)
-def set_message_handler_mode():
-    message_handler.mode = message_handler.NORMAL
-
-    def set_mode(mode):
-        message_handler.mode = mode
-
-    return set_mode
-
-
 # --- Validator tests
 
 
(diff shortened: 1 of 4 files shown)


tests/commands/test_build.py
```diff
@@ -77,6 +78,16 @@ def basic_project(tmp_path):
...
+@pytest.fixture(autouse=True)
+def set_message_handler_mode():
+    message_handler.mode = message_handler.NORMAL
+
+    def set_mode(mode):
+        message_handler.mode = mode
+
+    return set_mode
+
+
```
facundobatista, line 90: Is this structure needed? Why not just doing below the following, instead of calling this?

```
monkeypatch.setattr(message_handler, 'mode', mode)
```

facundobatista (commented):
I like the code, let's talk about fixture/monkeypatch...

sergiusens (approved):

tests/commands/test_build.py
cjp256: That works, the fixture was more about also setting the default - i'll remove it though.

cjp256: Fixed, though had to update logsetup a bit and its tests to conform with using monkeypatch (being able to init to a level).

facundobatista (approved):
Great!

PR #419, opened 2021-06-29
providers: gracefully handle when LXD not installed (CRAFT-329)

cjp256:
When building, ensure LXD is installed if going to use a managed
environment.

When cleaning, check if LXD is installed before attempting to
clean project containers.

Add two interfaces in providers to support this:

- ensure_provider_is_available() to raise error if not available
(for cases like build where the provider may be required).

- is_provider_available() to check and return bool (for cases like
cleaning where the provider does not need to be available).

Future work can extend ensure_provider_is_available() to assist
the user in installation of the provider.


charmcraft/providers.py
```diff
@@ -40,6 +42,56 @@
...
+    for name in lxc.list_names(project=lxd_project, remote=lxd_remote):
+        match_regex = f"^charmcraft-{charm_name}-.+-.+-.+$"
+        if re.match(match_regex, name):
+            logger.debug("Deleting container: %s", name)
+            lxc.delete(
+                instance_name=name, force=True, project=lxd_project, remote=lxd_remote
+            )
+            deleted.append(name)
+        else:
+            logger.debug("Not deleting container: %s", name)
```
facundobatista, line 76: Please log with `%r` here and above; getting in one block or the other depends of the regex matching, if we need to debug this for sure we'd want the exact string.

facundobatista (approved):
Nice! Added a small requirement inside.

charmcraft/providers.py
cjp256: Done for both!

sergiusens (approved):
Looks good, but tok a while to review due to the move of clean_project_environments

PR #420, opened 2021-06-29
pack: fetch logs from managed instance on error (CRAFT-333)

cjp256:
Catch error when failing to pack a Charm, retrieving the logs from
the managed instance for easier inspection by the user.

- Update logger to write to /tmp/charmcraft.log so that it can be
readily retrieved by the host Charmcraft instance.  Provide this
path with a new env.get_managed_environment_log_path() interface.
Additionally, ensure the log file is truncated with mode='w' so
the instance logs are only for the last run.

- prefix logs in managed instance with "mm "

- add providers.capture_logs_from_instance() to providers to retrieve
  logs and send them to logger.debug()

- update build to use this mechanism

### DIFF
--- a/charmcraft/commands/build.py
+++ b/charmcraft/commands/build.py
@@ -40,6 +40,7 @@ from charmcraft.logsetup import message_handler
 from charmcraft.manifest import create_manifest
 from charmcraft.metadata import parse_metadata_yaml
 from charmcraft.providers import (
+    capture_logs_from_instance,
     ensure_provider_is_available,
     is_base_providable,
     launched_environment,
@@ -254,6 +255,10 @@ class Builder:
 
     def pack_charm_in_instance(self, instance: Executor, bases_index: int) -> str:
         """Pack instance in Charm."""
+        charm_name = format_charm_file_name(
+            self.metadata.name, self.config.bases[bases_index]
+        )
+
         try:
             cmd = ["charmcraft", "pack", "--bases-index", str(bases_index)]
 
@@ -264,16 +269,17 @@ class Builder:
 
             instance.execute_run(
                 cmd,
+                check=True,
                 cwd=get_managed_environment_project_path().as_posix(),
             )
         except subprocess.CalledProcessError as error:
+            capture_logs_from_instance(instance)
+
             raise CommandError(
                 f"Failed to build charm for bases index '{bases_index}'."
             ) from error
 
-        return format_charm_file_name(
-            self.metadata.name, self.config.bases[bases_index]
-        )
+        return charm_name
 
     def _load_juju_ignore(self):
         ignore = JujuIgnore(default_juju_ignore)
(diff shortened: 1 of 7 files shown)


charmcraft/commands/build.py
```diff
@@ -246,19 +247,28 @@ def run(self, bases_indices: Optional[List[int]] = None) -> List[str]:
...
         except subprocess.CalledProcessError as error:
+            # Capture log from instance and raise error.
+            instance_log_path = get_managed_environment_log_path()
+            local_log_path = pathlib.Path(self.charmdir, charm_name + ".log")
+            instance.pull_file(source=instance_log_path, destination=local_log_path)
+
             raise CommandError(
-                f"Failed to build charm for bases index '{bases_index}'."
+                f"Failed to build charm for bases index '{bases_index}'. The logs "
+                f"for this charm may be found in {str(local_log_path)!r}."
```
facundobatista, line 268: I don't like producing TWO log files for the user. One of them polluting the work directory doesn't sound nice either.

What if we get this log from the managed environment, and append its content to the "main log"?

facundobatista (changes requested):

charmcraft/commands/build.py
cjp256: I'm open to suggestion as I'm not sure this is ideal either, but I believe the current log implementation would need a bit of work to make that happen cleanly.

On the flip side, this only happens when there is an error and the output log will most likely be analyzed by the user. Having it conveniently located and concise is a benefit.

charmcraft/logsetup.py
```diff
@@ -76,10 +80,19 @@ def set_mode(self, mode):
...
+        if managed_mode:
+            self._log_filepath = str(get_managed_environment_log_path())
+        else:
+            _, self._log_filepath = tempfile.mkstemp(prefix="charmcraft-log-")
+
+        file_handler = logging.FileHandler(self._log_filepath, mode="w")
+
+        log_format = FORMATTER_DETAILED
+        if managed_mode:
+            log_format = "mm " + log_format
```
sergiusens, line 93: I would probably make this refer to a constant

```
FORMATTER_DETAILED_MANAGED = "%(asctime)s mm %(name)-40s %(levelname)-8s %(message)s"
```

facundobatista (approved):
Nice!

charmcraft/logsetup.py
cmatsuoka: Agree with Sergio that using this constant would be better.

cjp256: Done!

sergiusens (approved):
I personally would prefer time then mm but i seems like I am the only one :sweat_smile:

cjp256:
> I personally would prefer time then mm but i seems like I am the only one

the `mm` kinda gets lost with long timestamp.  it stands out more in front, but we can revisit this anytime.

PR #422, opened 2021-06-30
utils: introduce confirm_with_user() helper for prompting (CRAFT-336)

cjp256:

### DIFF
--- a/charmcraft/utils.py
+++ b/charmcraft/utils.py
@@ -20,8 +20,9 @@ import logging
 import os
 import pathlib
 import platform
+import sys
 from collections import namedtuple
-from stat import S_IXUSR, S_IXGRP, S_IXOTH, S_IRUSR, S_IRGRP, S_IROTH
+from stat import S_IRGRP, S_IROTH, S_IRUSR, S_IXGRP, S_IXOTH, S_IXUSR
 
 import attr
 import yaml
@@ -185,3 +186,26 @@ def get_host_architecture():
     """Get host architecture in deb format suitable for base definition."""
     os_platform = get_os_platform()
     return ARCH_TRANSLATIONS.get(os_platform.machine, os_platform.machine)
+
+
+def confirm_with_user(prompt, default=False) -> bool:
+    """Query user for yes/no answer.
+
+    If TTY is not interaction, returns default value.
+    If answer is empty (""), returns default value.
+
+    :returns: True if yes (Y,y,YES,yes), otherwise false.
+    """
+    if not sys.stdin.isatty():
+        return default
+
+    choices = " [Y/n]: " if default else " [y/N]: "
+    default_answer = "y" if default else "n"
+
+    reply = str(input(prompt + choices)).lower().strip() or default_answer
+    if reply[0] == "y":
+        return True
+    if reply[0] == "n":
+        return False
+    else:
+        return default
--- a/charmcraft/utils.py
+++ b/charmcraft/utils.py
@@ -191,21 +191,23 @@ def get_host_architecture():
 def confirm_with_user(prompt, default=False) -> bool:
     """Query user for yes/no answer.
 
-    If TTY is not interaction, returns default value.
-    If answer is empty (""), returns default value.
+    If stdin is not a tty, the default value is returned.
 
-    :returns: True if yes (Y,y,YES,yes), otherwise false.
+    If user returns an empty answer, the default value is returned.
+    returns default value.
+
+    :returns: True if answer starts with [yY], False if answer starts with [nN],
+        otherwise the default.
     """
     if not sys.stdin.isatty():
         return default
 
     choices = " [Y/n]: " if default else " [y/N]: "
-    default_answer = "y" if default else "n"
 
-    reply = str(input(prompt + choices)).lower().strip() or default_answer
-    if reply[0] == "y":
+    reply = str(input(prompt + choices)).lower().strip()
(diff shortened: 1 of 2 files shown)


charmcraft/utils.py
```diff
@@ -185,3 +186,26 @@ def get_host_architecture():
     """Get host architecture in deb format suitable for base definition."""
     os_platform = get_os_platform()
     return ARCH_TRANSLATIONS.get(os_platform.machine, os_platform.machine)
+
+
+def confirm_with_user(prompt, default=False) -> bool:
+    """Query user for yes/no answer.
+
+    If TTY is not interaction, returns default value.
```
cmatsuoka, line 194: Maybe "if stdin is not a terminal"?

charmcraft/utils.py
```diff
@@ -185,3 +186,26 @@ def get_host_architecture():
...
+    If answer is empty (""), returns default value.
+
+    :returns: True if yes (Y,y,YES,yes), otherwise false.
+    """
+    if not sys.stdin.isatty():
+        return default
+
+    choices = " [Y/n]: " if default else " [y/N]: "
+    default_answer = "y" if default else "n"
```
cmatsuoka, line 203: This looks a bit confusing, maybe refactor to test once for `default`?

charmcraft/utils.py
```diff
@@ -20,8 +20,9 @@
 import os
 import pathlib
 import platform
+import sys
 from collections import namedtuple
-from stat import S_IXUSR, S_IXGRP, S_IXOTH, S_IRUSR, S_IRGRP, S_IROTH
+from stat import S_IRGRP, S_IROTH, S_IRUSR, S_IXGRP, S_IXOTH, S_IXUSR
```
sergiusens, line 25: I would appreciate these more close to release delicate PRs not having these type of changes.

charmcraft/utils.py
```diff
@@ -185,3 +186,26 @@ def get_host_architecture():
     """Get host architecture in deb format suitable for base definition."""
     os_platform = get_os_platform()
     return ARCH_TRANSLATIONS.get(os_platform.machine, os_platform.machine)
+
+
+def confirm_with_user(prompt, default=False) -> bool:
```
sergiusens, line 191: argparse + click might be good thing to look into

charmcraft/utils.py
```diff
@@ -185,3 +186,26 @@ def get_host_architecture():
...
+    """Query user for yes/no answer.
+
+    If TTY is not interaction, returns default value.
+    If answer is empty (""), returns default value.
+
+    :returns: True if yes (Y,y,YES,yes), otherwise false.
+    """
+    if not sys.stdin.isatty():
```
sergiusens, line 199: I would also raise RuntimeError if used in managed-mode for now, maybe worth considering or are most runs except for the shell related ones running with stdin closed? In that case, it could be the conditional following this one :-)

charmcraft/utils.py
```diff
@@ -185,3 +186,26 @@ def get_host_architecture():
...
     return ARCH_TRANSLATIONS.get(os_platform.machine, os_platform.machine)
+
+
+def confirm_with_user(prompt, default=False) -> bool:
+    """Query user for yes/no answer.
+
+    If TTY is not interaction, returns default value.
+    If answer is empty (""), returns default value.
+
+    :returns: True if yes (Y,y,YES,yes), otherwise false.
```
cmatsuoka, line 197: Technically it returns True if answer starts with Y or y, False if it starts with N or n, otherwise default. Maybe state that as "True if yes, False if no, otherwise the specified default value"?

charmcraft/utils.py
cjp256: Fair point, though this shouldn't be called in managed-mode.  But a guard would be nice.

charmcraft/utils.py
cjp256: I must have bashed my head on keyboard :rofl:

sergiusens: _interactive_ might be the word that is generic enough

charmcraft/utils.py
cmatsuoka: This is probably isort in action.

charmcraft/utils.py
cjp256: I know click has `confirm()` but am unsure what you are asking here. Look into using click and replacing the argparse bits?

charmcraft/utils.py
cjp256: Updated docs!

charmcraft/utils.py
cjp256: Agreed, removed and simplified code a bit.

charmcraft/utils.py
cjp256: Improved docs!

cmatsuoka (approved):
Thanks Chris. Lgtm.

charmcraft/utils.py
cjp256: Done!

charmcraft/utils.py
cjp256: Yes it is.

charmcraft/utils.py
sergiusens: not asking, just a random thought in a PR. Sorry :-)

sergiusens (approved):

PR #423, opened 2021-06-30
providers: account for project path in instance name (CRAFT-334)

cjp256:
If Charmcraft were to build the same project [name] simultaneously in
two different directories, collisions on the containers in use may
occur.

To account for this, this PR uses the project directory's inode in the
container name.  This will ensure uniqueness for projects in multiple
paths, but also seamlessly handle project directory moves that don't
cross the filesystem boundary, the common case.

Note that 'clean' command will not affect other project directories,
so the clean must occur in the original directory.  To handle any
potentially unassociated containers, future work will include more
sweeping purge capabilities for the clean command.  This is not unique
to the introduction of the use of the directory inode, but is
exacerbated by it.

### DIFF
--- a/charmcraft/commands/clean.py
+++ b/charmcraft/commands/clean.py
@@ -41,9 +41,9 @@ class CleanCommand(BaseCommand):
 
     def run(self, parsed_args):
         """Run the command."""
-        metadata = parse_metadata_yaml(self.config.project.dirpath)
+        project_path = self.config.project.dirpath
+        metadata = parse_metadata_yaml(project_path)
         logger.debug("Cleaning project %r.", metadata.name)
 
-        clean_project_environments(metadata.name)
-
+        clean_project_environments(metadata.name, project_path)
         logger.info("Cleaned project %r.", metadata.name)
--- a/charmcraft/providers.py
+++ b/charmcraft/providers.py
@@ -66,6 +66,7 @@ def capture_logs_from_instance(instance: Executor) -> None:
 
 def clean_project_environments(
     charm_name: str,
+    project_path: pathlib.Path,
     *,
     lxd_project: str = "charmcraft",
     lxd_remote: str = "local",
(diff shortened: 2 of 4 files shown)


sergiusens (approved):
+1. I would prefer to have this as a built in to craft-providers. We should also look into adding a lock to avoid multiple runs in the same directory in a future PR.

cjp256:
> +1. I would prefer to have this as a built in to craft-providers. We should also look into adding a lock to avoid multiple runs in the same directory in a future PR.

Yeah, it'd be useful to move some of this functionality over into craft-providers at this point.

facundobatista (commented):
I wonder how multiplatform is to use the inode in the path. 

Maybe it's much easier to just add a random number to the path, to disambiguate it? e.g., the last 5 chars of `uuid.uuid4()`?

sergiusens:
Path uuid should be a nice global solution indeed

cjp256:
(1) I verified that Pathlib.stat().st_ino works on MacOS and Windows.  I suppose there may be filesystems it doesn't work on, but they'd likely be problematic for other reasons anyways.  I'll add that this implementation is scoped for LXD and Linux.

(2) I don't know what a "path uuid" is - just a UUID incorporating the path component in generation?

(3) Generating a completely random value or uuid would provide uniqueness, but would then require a way of mapping said project to random value. This would have to be done by dropping a config somewhere to provide the hints `project @ <path> = <this id>`.  Perfectly reasonable, but more work.  Instead of a file on the host, we could instead store this info in LXD's `user.user-data` configuration to track this.  Or add it to the `craft-image.conf` config, but that would require starting up containers/VMs to check it, so not great.

(4) We could encode the [optionally hashed] path into the name like we do for snapcraft's remote-build.  Con: Doesn't handle the common case of moving the project directory.

facundobatista (approved):
Thanks!

PR #425, opened 2021-07-01
providers: handle unexpected case where instance never logs (CRAFT-345)

cjp256:
If the managed instance never gets to creating the log, for whatever
reason, catch the FileNotFoundError and log it.


sergiusens (approved):

facundobatista (approved):
Like it!

PR #426, opened 2021-07-01
providers: install git as required dependency (CRAFT-346)

cjp256:
Git may be required to install python dependencies for projects.
Instead of using git from the snap, install the native git on
the host as a required dependency.  Future work on craft-parts
may aborb this dependency.


facundobatista (approved):
I like this; maybe put a `XXX` to remove it when we have `craft_parts` integrated

cjp256:
> I like this; maybe put a `XXX` to remove it when we have `craft_parts` integrated?

Done

sergiusens (approved):

PR #427, opened 2021-07-02
build: log reason to ignore build-on entry to info (CRAFT-350)

cjp256:
This information was being logged to debug, but should be made
available to user regardless if erroring out.  Log it as informational.

Also tweak the language to be more generic rather than intended
for managed/destructive mode (host isn't matched for provided
environments).

### DIFF
--- a/charmcraft/commands/build.py
+++ b/charmcraft/commands/build.py
@@ -234,8 +234,8 @@ class Builder:
                     charms.append(charm_name)
                     break
                 else:
-                    logger.debug(
-                        "Host does not match 'bases[%d].build-on[%d]' (%s)",
+                    logger.info(
+                        "Ignoring 'bases[%d].build-on[%d]': %s.",
                         bases_index,
                         build_on_index,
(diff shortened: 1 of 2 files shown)


charmcraft/commands/build.py
```diff
@@ -234,8 +234,8 @@ def run(self, bases_indices: Optional[List[int]] = None) -> List[str]:
                     charms.append(charm_name)
                     break
                 else:
-                    logger.debug(
-                        "Host does not match 'bases[%d].build-on[%d]' (%s)",
+                    logger.info(
+                        "Ignoring 'bases[%d].build-on[%d]': %s.",
```
sergiusens, line 238:
```suggestion
                        "Skipping 'bases[%d].build-on[%d]': %s.",
```

cjp256: Done

sergiusens (approved):

facundobatista (approved):
Like it!

PR #432, opened 2021-07-05
Removed stage filters in snapcraft.yaml.

facundobatista:


cmatsuoka (approved):

cjp256 (approved):
FWIW this change increases the amd64 snap size from 29M -> 49M.  While I think the existing filter patterns are a bit excessive, there is certainly room for some selective filtering to cut back the size.  Worth revisiting at some point IMO.

PR #435, opened 2021-07-05
Implemented the Framework checker (CRAFT-316)

facundobatista:
Note this is a checkout done from `facundobatista:analint-language-checker`, so the diff to master will include it.


charmcraft/linters.py
```diff
@@ -0,0 +1,180 @@
...
+            return cls.Result.unknown
+
+        entrypoint = basedir / entrypoint_str
+        if entrypoint.suffix == ".py" and os.access(entrypoint, os.X_OK):
+            return cls.Result.python
+        return cls.Result.unknown
+
+
+class Framework:
+    """Check on which framework the charm is based on.
```
sergiusens, line 85:
```suggestion
    """Check the framework the charm is based on.
```

charmcraft/linters.py
```diff
@@ -0,0 +1,180 @@
...
+    Currently it detects if the Operator Framework is used, if...
+
+    - the language trait is set to python
+    - the charm contains venv/ops
+    - the charm imports ops in the entry point.
+
+    ...or the Reactive Framework is used, if the charm...
+
+    - has a metadata.yaml with "name" in it
+    - has a reactive/<name>,py file that imports "charms.reactive"
```
sergiusens, line 96:
```suggestion
    - has a reactive/<name>.py file that imports "charms.reactive"
```

charmcraft/linters.py
```diff
@@ -0,0 +1,180 @@
...
+    - has a file inside "wheelhouse" dir whose name starts with "charms.reactive-"
```
sergiusens, line 97:
```suggestion
    - has a file name that starts with "charms-reactive-" inside the "wheelhouse" directory
```

charmcraft/linters.py
```diff
@@ -0,0 +1,180 @@
...
+    """
+
+    check_type = CheckType.trait
```
sergiusens, line 100: should we rename trait to attributes?

charmcraft/linters.py
```diff
@@ -0,0 +1,180 @@
...
+    name = "framework"
+    url = "https://juju.is/docs/sdk/charmcraft-analyze#heading--framework"
+    text = "The charm is based on the Operator Framework."
```
sergiusens, line 103: text should be multiple choice here, a reactive is not operator

charmcraft/linters.py
```diff
@@ -0,0 +1,180 @@
...
+    # different result constants
+    Result = namedtuple("Result", "operator reactive unknown")(
+        operator="operator", reactive="reactive", unknown=UNKNOWN
+    )
+
+    @staticmethod
+    def _get_imports(filepath: pathlib.Path) -> Generator[List[str], None, None]:
+        """Parse a Python filepath and yield its imports.
+
+        If the file does not exist or can't be parsed, return empty. Otherwise
```
sergiusens, line 114:
```suggestion
        If the file does not exist or cannot be parsed, return empty. Otherwise
```

charmcraft/linters.py
```diff
@@ -0,0 +1,180 @@
...
+    url = "https://juju.is/docs/sdk/charmcraft-analyze#heading--framework"
+    text = "The charm is based on the Operator Framework."
+
+    # different result constants
+    Result = namedtuple("Result", "operator reactive unknown")(
+        operator="operator", reactive="reactive", unknown=UNKNOWN
+    )
+
+    @staticmethod
+    def _get_imports(filepath: pathlib.Path) -> Generator[List[str], None, None]:
```
sergiusens, line 111: a better name that rings imports_iterator would be nice here

charmcraft/linters.py
```diff
@@ -0,0 +1,180 @@
...
+            return False
+
+        entrypoint = language_info["entrypoint"]
+        for import_parts in cls._get_imports(entrypoint):
+            if import_parts[0] == "ops":
+                return True
+        return False
+
+    @classmethod
+    def _check_reactive(cls, basedir: pathlib.Path) -> bool:
```
sergiusens, line 149: what i interesting here is that this should also be Language.Result.python and we might need to redo our python check

sergiusens (changes requested):
Looks pretty good. Text is the only real reason I marked this as I marked it; aside from some cosmetic typos and text change suggestions.

There's a thought about framwork reactive also being language python in our case that we might want to revisit

charmcraft/linters.py
facundobatista: Improved, thanks.

charmcraft/linters.py
facundobatista: Thanks.

charmcraft/linters.py
facundobatista: yes.

charmcraft/linters.py
facundobatista: Oh, right, this got old.

After some consideration, I think I will stop using these classes as classes, but instances. Then it's easier to provide different mechanisms, like a dynamic text here.

charmcraft/linters.py
facundobatista: Fixed, thanks.

charmcraft/linters.py
facundobatista: Let's go back to the drawing board?

facundobatista: Ok, we're not changing the Language result for this case.

charmcraft/linters.py
facundobatista: Decided to keep this.

charmcraft/linters.py
facundobatista: Fixed.

charmcraft/linters.py
facundobatista: Fixed.

charmcraft/linters.py
facundobatista: Improved.

cmatsuoka (approved):
Looks good, left a couple of comments but no showstoppers found.

tests/test_linters.py
```diff
@@ -0,0 +1,469 @@
...
+
+def test_language_no_entrypoint(tmp_path):
+    """Cannot find the entrypoint used in dispatch."""
+    dispatch = tmp_path / "dispatch"
+    dispatch.write_text(EXAMPLE_DISPATCH)
+    result = Language().run(tmp_path)
+    assert result == Language.Result.unknown
+
+
+def test_language_entrypoint_is_no_python(tmp_path):
```
sergiusens, line 83:
```suggestion
def test_language_entrypoint_is_not_python(tmp_path):
```

tests/test_linters.py
```diff
@@ -0,0 +1,469 @@
...
+    """The charm entrypoint is not executable."""
+    dispatch = tmp_path / "dispatch"
+    dispatch.write_text(EXAMPLE_DISPATCH)
+    entrypoint = tmp_path / "charm.py"
+    entrypoint.touch()
+    result = Language().run(tmp_path)
+    assert result == Language.Result.unknown
+
+
+def test_framework_no_useful_text():
```
sergiusens, line 109:
```suggestion
def test_framework_text_before_check_raises():
```

sergiusens (approved):

charmcraft/linters.py
```diff
@@ -0,0 +1,193 @@
...
+    - the entry point file is executable
+    """
+
+    check_type = CheckType.attribute
+    name = "language"
+    url = "https://juju.is/docs/sdk/charmcraft-analyze#heading--language"
+    text = "The charm is written with Python."
+
+    # different result constants
+    Result = namedtuple("Result", "python unknown")(python="python", unknown=UNKNOWN)
```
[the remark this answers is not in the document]
facundobatista: we discussed enums in the other PR, they are more verbose and cumbersome to use...

charmcraft/linters.py
```diff
@@ -0,0 +1,193 @@
...
+        except (IOError, UnicodeDecodeError):
+            return self.Result.unknown
+
+        entrypoint = basedir / entrypoint_str
+        if entrypoint.suffix == ".py" and os.access(entrypoint, os.X_OK):
+            return self.Result.python
+        return self.Result.unknown
+
+
+class Framework:
```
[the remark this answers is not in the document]
facundobatista: Considered this possibility... let's how everything looks when we add more and more classes here, if it makes sense to improve this naming or not...

tests/test_linters.py
facundobatista: improved.

tests/test_linters.py
facundobatista: Improved, slightly changed.

PR #437, opened 2021-07-07
Added a function to run all checkers (CRAFT-317).

facundobatista:
This is the "entry point" to the module, the only function to be used from outside.

Note this PR is based on the other two `analint-*` branches.


charmcraft/linters.py
```diff
@@ -191,3 +192,29 @@ def run(self, basedir: pathlib.Path) -> str:
             result = self.Result.unknown
         self.result = result
         return result
+
+
+# all checkers to run; the order here is important, as some checkers depend on the
```
sergiusens, line 197: maybe put this closer to shared_sate as they are somewhat intertwined.

sergiusens (approved):

cjp256 (approved):

PR #442, opened 2021-07-08
Fetch any defined schemas during charmcraft build

knkski:
The [serialized-data-interface library](https://github.com/canonical/serialized-data-interface) currently fetches any defined schemas at runtime, which [fails in proxy environments](https://github.com/canonical/serialized-data-interface/issues/9).

This change allows schemas to be fetched at build time instead.

Also includes commit that adds support for running tests via `tox`.

### DIFF
--- a/charmcraft/commands/build.py
+++ b/charmcraft/commands/build.py
@@ -17,6 +17,7 @@
 """Infrastructure for the 'build' command."""
 
 import errno
+import hashlib
 import logging
 import os
 import pathlib
@@ -26,6 +27,9 @@ import zipfile
 from typing import List, Optional
 
 from charmcraft import linters
+import requests
+import yaml
+
 from charmcraft.bases import check_if_base_matches_host
 from charmcraft.cmdbase import BaseCommand, CommandError
 from charmcraft.config import Base, BasesConfiguration, Config
@@ -38,7 +42,7 @@ from charmcraft.env import (
 from charmcraft.jujuignore import JujuIgnore, default_juju_ignore
 from charmcraft.logsetup import message_handler
 from charmcraft.manifest import create_manifest
-from charmcraft.metadata import parse_metadata_yaml
+from charmcraft.metadata import parse_metadata_yaml, CHARM_METADATA
 from charmcraft.providers import (
     capture_logs_from_instance,
     ensure_provider_is_available,
@@ -168,6 +172,7 @@ class Builder:
         linked_entrypoint = self.handle_generic_paths()
         self.handle_dispatcher(linked_entrypoint)
         self.handle_dependencies()
+        self.handle_schemas()
 
         linting_results = []
         # run linters, present them to the user according to their type, and fail if necessary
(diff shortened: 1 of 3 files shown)


charmcraft/commands/build.py
```diff
@@ -440,6 +444,27 @@ def handle_dispatcher(self, linked_entrypoint):
...
+
+        for rel_type in ("requires", "provides", "peers"):
+            for relation in metadata.get(rel_type, {}).values():
+                if "schema" in relation and isinstance(relation["schema"], str):
+                    schema = relation["schema"]
+                    try:
+                        response = requests.get(schema)
+                        response.raise_for_status()
+                    except requests.RequestException:
+                        raise CommandError(f"Couldn't fetch schema `{schema}`")
```
cjp256, line 462:
```suggestion
                        raise CommandError(f"Failed to fetch schema from {schema!r}")
```

charmcraft/commands/build.py
```diff
@@ -440,6 +444,27 @@ def handle_dispatcher(self, linked_entrypoint):
...
+    def handle_schemas(self):
+        """Handle fetching any defined schemas."""
+        logger.debug("Fetching any schemas from metadata.yaml")
+        metadata_yaml = self.charmdir / CHARM_METADATA
+
+        metadata = yaml.safe_load(metadata_yaml.read_text())
+
+        for rel_type in ("requires", "provides", "peers"):
+            for relation in metadata.get(rel_type, {}).values():
+                if "schema" in relation and isinstance(relation["schema"], str):
```
cjp256, line 456: I assume this is supposed to be a URL? Is there a reference for this somewhere?

If it isn't a string, should it be an error?

charmcraft/commands/build.py
```diff
@@ -440,6 +444,27 @@ def handle_dispatcher(self, linked_entrypoint):
...
+                    schema = relation["schema"]
+                    try:
+                        response = requests.get(schema)
+                        response.raise_for_status()
+                    except requests.RequestException:
+                        raise CommandError(f"Couldn't fetch schema `{schema}`")
+
+                    relation["schema"] = yaml.safe_load(response.text)
+
+        metadata_yaml.write_text(yaml.safe_dump(metadata), encoding="utf8")
```
cjp256, line 466: This approach to updating the YAML will result in the loss of comments.  It will also re-write the file in-place.  Can the schemas be stored separately?

tests/commands/test_build.py
```diff
@@ -2053,6 +2054,128 @@ def test_builder_with_jujuignore(tmp_path, config):
...
+    )
+
+    monkeypatch.setenv("CHARMCRAFT_MANAGED_MODE", "1")
+    with pytest.raises(CommandError, match="Couldn't fetch schema `bad-url`"), patch(
+        "charmcraft.commands.build.check_if_base_matches_host",
+        return_value=(True, None),
+    ):
+        builder.run()[0]
+
+
```
cjp256, line 2178: Consider adding a test where schema value is not a string to be interpreted as a URL.

charmcraft/commands/build.py
knkski: Yeah, it's supposed to be a URL, and here's a reference for how the SDI library handles it:

https://github.com/canonical/serialized-data-interface/blob/main/serialized_data_interface/utils.py#L40

Here's a reference for how it's used by a charm:

https://github.com/canonical/kfp-operators/blob/master/charms/kfp-api/metadata.yaml#L23

charmcraft/commands/build.py
knkski: Good point. The SDI library is doing this for reasons that aren't relevant with `charmcraft` supporting this itself. Updated to leave metadata alone and store the schemas in the built charm file under a `schemas` directory, let me know if that works for you.

sergiusens:
As discussed, this is part of a larger picture and the agreement is to iron out that big picture story first before moving forward.

PR #443, opened 2021-07-09
Don't run the ignored attributes and linters (CRAFT-326)

facundobatista:


cjp256 (approved):

charmcraft/config.py
```diff
@@ -105,6 +115,43 @@ def validate_relative_path(cls, value: str) -> str:
         return value
 
 
+class AttributeName(CustomStrictStr):
+    """Constrainted string which must be the name of an attribute from linters.CHECKERS."""
```
sergiusens, line 119:
```suggestion
    """Constrained string that must match the name of an attribute from linters.CHECKERS."""
```

charmcraft/config.py
```diff
@@ -105,6 +115,43 @@ def validate_relative_path(cls, value: str) -> str:
...
+
+    @classmethod
+    def validate_relative_path(cls, value: str) -> str:
+        """Validate attribute name."""
```
sergiusens, line 123:
```suggestion
    def validate_name_exists(cls, value: str) -> str:
        """Validate that the attribute checker exists."""
```

charmcraft/config.py
```diff
@@ -105,6 +115,43 @@ def validate_relative_path(cls, value: str) -> str:
...
+        return value
+
+
+class LinterName(CustomStrictStr):
+    """Constrainted string which must be the name of a linter from linters.CHECKERS."""
+
+    @classmethod
+    def validate_relative_path(cls, value: str) -> str:
+        """Validate attribute name."""
+        from charmcraft import linters  # import here to avoid cyclic imports
```
sergiusens, line 142:
```suggestion
    def validate_name_exists(cls, value: str) -> str:
        """Validate that the lint checker exists."""
```

charmcraft/linters.py
```diff
@@ -202,12 +202,20 @@ def run(self, basedir: pathlib.Path) -> str:
 ]
 
 
-def analyze(config: config.Config) -> List[CheckResult]:
+def analyze(config: config.Config, basedir: pathlib.Path) -> List[CheckResult]:
     """Run all checkers and linters."""
     all_results = []
     for checker_class in CHECKERS:
+        # do not run the ignored ones
```
sergiusens, line 209: I personally prefer reductions here, also, maybe we should have CHECKERS split in two groups in the future, could be done at module load time

sergiusens (approved):
Looks good, 2 comments about names that were confusing on first sight (together with the docstring) and a potential suggestion for a future change

charmcraft/config.py
facundobatista: Fixed/improved.

charmcraft/config.py
facundobatista: Ja, silly c&p, good catch! I changed it to `custom_validate`, as it's the generic custom validation called from `CustomStrictStr`.

charmcraft/linters.py
facundobatista: On one hand, `CHECKERS` should stay as a single list, as in the future attributes and linters may share state (analysis/explorations done by one of those that are useful to others, so will be stored in the shared_state to not repeat it).

We could split it in two groups at module load time, but you will be running that *everytime* the process starts (as the config imports linters), and it will be waste if the Builder is not used...

PR #445, opened 2021-07-12
Use linters' analyze from build command (CRAFT-318)

facundobatista:
Call linters' analyze from build command and deal with the results: so far only log the attribute ones (in the future it will deal also with warnings and errors), and use them for the manifest creation. Also changed the manifest creation function to properly save this new info.

Note: I had to fix checkers/analyze to save information in the shared state (a checkers integration test was missing).

Note 2: This PR depends on #443.

### DIFF
--- a/charmcraft/commands/build.py
+++ b/charmcraft/commands/build.py
@@ -25,6 +25,7 @@ import subprocess
 import zipfile
 from typing import List, Optional
 
+from charmcraft import linters
 from charmcraft.bases import check_if_base_matches_host
 from charmcraft.cmdbase import BaseCommand, CommandError
 from charmcraft.config import Base, BasesConfiguration, Config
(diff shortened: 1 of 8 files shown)


facundobatista:
Also added some names fixing (a leftover from previous branch).

charmcraft/commands/build.py
```diff
@@ -164,13 +165,33 @@ def build_charm(self, bases_config: BasesConfiguration) -> str:
...
 
-        create_manifest(self.buildpath, self.config.project.started_at, bases_config)
-
         linked_entrypoint = self.handle_generic_paths()
         self.handle_dispatcher(linked_entrypoint)
         self.handle_dependencies()
-        zipname = self.handle_package(bases_config)
 
+        linting_results = []
+        # run linters, present them to the user according to their type, and decide if go on
```
sergiusens, line 173: "decide if go on" is missing words, I would inverse the statement to not use so many words
```suggestion
        # run linters, present them to the user according to their type, and fail if necessary
```

or maybe just remove as there is not decision being made yet.

sergiusens (approved):

charmcraft/commands/build.py
facundobatista: thanks, improved

facundobatista: btw, left it, as I put the XXX below...

cmatsuoka (approved):
Looks good (and it doesn't seem to conflict with parts support integration).

cjp256 (approved):
LGTM, but would suggest moving the linter logic into its own method(s).

Issue #27, opened 2020-06-05
The build process should error out if metadata.yaml is not present


facundobatista:

mthaddon:
The item that says "Both `actions.yaml` and `config.yaml`, if present, need to be valid YAMLs" should be expanded a bit to say "and be considered valid by juju (e.g. correct config types)"

sergiusens:
What we do in Snapcraft, is one source of truth, and this is snapd, which grew a "snap pack" command with an additional "--check-skeleton" option to validate the integrity of the snap. Furthermore, we use review-tools to validate that this valid snap is acceptable in the Store

knkski:
Would it be possible to have an independently-versioned spec of any yaml files? By independently-versioned, I mean that it would be great to have a document like this to reference when writing metadata.yaml and other yaml files, instead of having to look at code to figure out valid input:

https://docs.docker.com/compose/compose-file/compose-versioning/

It might even be worth specifying the format using something like https://json-schema.org/, as then we could use the [warlock](https://pypi.org/project/warlock/) library to automatically deserialize the yaml files into Python objects.

PR #449, opened 2021-07-14
Added a linter to verify that the metadata is ok (CRAFT-322).

facundobatista:
Note that linters are still NOT enforced when packing (will do that in the next branch when I add the --force option to pack anyway).

Also note that this PR depends on the "analint 06" one.

### DIFF
--- a/charmcraft/linters.py
+++ b/charmcraft/linters.py
@@ -165,10 +165,8 @@ class Framework:
 
     def _check_reactive(self, basedir: pathlib.Path) -> bool:
         """Detect if the Reactive Framework is used."""
-        try:
-            entrypoint_name = parse_metadata_yaml(basedir).name
-        except Exception:
-            # file not found, corrupted, no name in it, etc.
+        entrypoint_name = shared_state[JujuMetadata.name]["name"]
+        if entrypoint_name is None:
             return False
 
         wheelhouse_dir = basedir / "wheelhouse"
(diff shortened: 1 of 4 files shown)


facundobatista:
I converted it to draft as I thought a better way of doing something...

charmcraft/metadata.py
```diff
@@ -52,10 +54,10 @@ def unmarshal(cls, obj: Dict[str, Any]):
             )
 
 
-def parse_metadata_yaml(charm_dir: pathlib.Path) -> Dict[str, Any]:
+def parse_metadata_yaml(charm_dir: pathlib.Path) -> CharmMetadata:
     """Parse project's metadata.yaml.
 
-    :returns: Metadata dictionary object, if it exists.
+    :returns: a CharmMetadata object; raises CommandError if metadata does not exist.
```
cjp256, line 60:
```suggestion
    :returns: a CharmMetadata object
    
    :raises CommandError: if metadata does not exist.
```

cjp256 (approved):
LGTM

charmcraft/linters.py
```diff
@@ -197,10 +195,48 @@ def run(self, basedir: pathlib.Path) -> str:
...
+
+    - the metadata.yaml is present
+    - it is a valid YAML file
+    - it has at least the following fields: name, summary, and description
+    """
+
+    check_type = CheckType.lint
+    name = "metadata"
+    url = "https://juju.is/docs/sdk/charmcraft-analyze#heading--metadata"
+    text = "Problems found with metadata.yaml file."
```
sergiusens, line 211: it would be nice to compose this, maybe text needs to be more succinct as well, like

- metadata.yaml: file missing
- metadata.yaml: not valid YAML
- metadata.yaml: missing 'name'

charmcraft/linters.py
```diff
@@ -197,10 +195,48 @@ def run(self, basedir: pathlib.Path) -> str:
...
+
+    # different result constants
+    Result = namedtuple("Result", "ok errors")(ok="ok", errors="errors")
+
+    def run(self, basedir: pathlib.Path) -> str:
+        """Run the proper verifications."""
+        try:
+            metadata = parse_metadata_yaml(basedir)
+        except Exception:
+            # file not found, corrupted, or mandatory "name" not present
```
sergiusens, line 221: is checking the name here only a side effect of charmcraft needing it?

sergiusens (approved):
Looks good, just a suggestion about finer grained "text"

sergiusens:
This is partially the first item in #27

charmcraft/linters.py
facundobatista: Yes. We cannot build a charm without knowing its name.

charmcraft/linters.py
facundobatista: I'll prepare a branch for this improvement in the future.

PR #450, opened 2021-07-14
parts: add charm plugin (CRAFT-89)

cmatsuoka:
Start integration with craft-parts by adding the charm plugin. This
plugin handles "charm" parts in `charmcraft.yaml`.
    

charmcraft/parts.py
```diff
@@ -0,0 +1,209 @@
...
+                work_dir=work_dir,
+                cache_dir=cache_dir,
+                venv_dir=venv_dir,
+            )
+            self._lcm.refresh_packages_list()
+        except PartsError as err:
+            raise CommandError(err)
+
+    @property
+    def prime_dir(self):
```
cjp256, line 195:
```suggestion
    def prime_dir(self) -> pathlib.Path:
```

charmcraft/parts.py
```diff
@@ -0,0 +1,209 @@
...
+
+        try:
+            self._lcm = LifecycleManager(
+                {"parts": all_parts},
+                application_name="charmcraft",
+                work_dir=work_dir,
+                cache_dir=cache_dir,
+                venv_dir=venv_dir,
+            )
+            self._lcm.refresh_packages_list()
```
cjp256, line 190: Might be an optimization we can do in the future, if using a provided environment we will have a refresh apt cache.

tests/test_parts.py
```diff
@@ -0,0 +1,91 @@
...
+
+from charmcraft import parts
+
+
+class TestPartValidation:
+    """Part data validation scenarios."""
+
+    def test_part_validation_happy(self):
+        data = {
+            "plugin": "make",
```
cjp256, line 28: `make` feels a bit odd as it's not owned by charmcraft. perhaps its going to be updated to charm plugin? :)

charmcraft/parts.py
```diff
@@ -0,0 +1,209 @@
...
+    if not plugin_name:
+        raise ValueError("'plugin' not defined")
+
+    plugin_class = plugins.get_plugin_class(plugin_name)
+
+    # validate plugin properties
+    plugin_class.properties_class.unmarshal(spec)
+
+    # validate common part properties
+    plugins.strip_plugin_properties(spec, plugin_name=plugin_name)
```
cjp256, line 165: I'd maybe make the copy just prior to this to be obvious what's going to nuke it, or comment why the copy is necessary above.  Or perhaps change strip_plugin_properties() to return a new dict instead. No action required, just a thought.

tests/test_parts.py
cmatsuoka: Changed to test using the charm plugin.

cjp256 (approved):

cmatsuoka:
Will propose a new version that executes a charm building tool from the plugin.

PR #451, opened 2021-07-15
Handle linters when building (CRAFT-325).

facundobatista:
Show errors and warnings. Fail if there are some errors, unless `--force` was indicated to `pack` (not to `build`, which we want to not touch until we remove it).

Note this depends on #449.

### DIFF
--- a/charmcraft/commands/build.py
+++ b/charmcraft/commands/build.py
@@ -153,25 +153,8 @@ class Builder:
         self.config = config
         self.metadata = parse_metadata_yaml(self.charmdir)
 
-    def build_charm(self, bases_config: BasesConfiguration) -> str:
-        """Build the charm.
-
-        :param bases_config: Bases configuration to use for build.
-
-        :returns: File name of charm.
-        """
-        logger.debug("Building charm in %r", str(self.buildpath))
-
-        if self.buildpath.exists():
-            shutil.rmtree(str(self.buildpath))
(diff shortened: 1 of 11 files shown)


charmcraft/commands/build.py
```diff
@@ -146,12 +146,53 @@ def __init__(self, args, config):
...
+                logger.info(template, result.name, result.text, result.url)
+        if linters.ERRORS in lint_results_by_outcome:
+            logger.info("Lint Errors:")
+            for result in lint_results_by_outcome[linters.ERRORS]:
+                logger.info(template, result.name, result.text, result.url)
+            if self.force_packing:
+                logger.info("Packing anyway as requested.")
+            else:
+                raise CommandError(
+                    "Exiting after lint errors (use --force to pack anyway).", retcode=2
```
sergiusens, line 193:
```suggestion
                    "Exiting after lint errors (use --force to override).", retcode=2
```

sergiusens (approved):

charmcraft/commands/build.py
cjp256: Maybe `Aborting due to lint errors ...` ?  Exiting is a bit ambiguous.

cjp256 (approved):
LGTM !

charmcraft/commands/build.py
facundobatista: Like it!

PR #452, opened 2021-07-18
build: create charm builder tool (CRAFT-373)

cmatsuoka:
Extract the build routine from the Builder class and create a stand-alone
charm builder tool. Invoke the tool from the Builder class passing all the
appropriate arguments to build charms. This tool will be invoked from the
charm plugin after parts integration.

The payload venv is now created from the user site of a staging venv to
cache the previous installation and avoid reinstalling packages that are
already installed. This is done to avoid spending a long time building
packages when binary wheels are disabled.

Charm building tests moved from `tests/commands/test_build.py` to
`tests/test_charm_builder.py`, replacing the `Builder` class with the
`CharmBuilder` class. Additional tests added to verify handling of
package dependencies and pip binaries arguments, and invocation
of the builder tool.


charmcraft/charm_builder.py
```diff
@@ -0,0 +1,357 @@
...
+        logger.debug("Installing dependencies")
+
+        if not self.python_packages and not self.requirements:
+            return
+
+        # create venv so packages can be cached. note: additional packages
+        # pip and setuptools will also be installed!
+        staging_venv_dir = self.charmdir / STAGING_VENV_DIRNAME
+        logger.debug("create staging venv in %r", str(staging_venv_dir))
+        retcode = _process_run([sys.executable, "-m", "venv", str(staging_venv_dir)])
```
facundobatista, line 253: Why not using `venv` as a module?

charmcraft/charm_builder.py
```diff
@@ -0,0 +1,357 @@
...
+        if retcode:
+            raise CommandError("problems creating venv")
+
+        retcode = _process_run(["pip3", "--version"])
```
facundobatista, line 257: Shouldn't you use the `pip` from the virtualenv, instead the one from the system?

charmcraft/charm_builder.py
```diff
@@ -0,0 +1,357 @@
...
+        """Handle from-directory and virtualenv dependencies."""
+        logger.debug("Installing dependencies")
+
+        if not self.python_packages and not self.requirements:
+            return
+
+        # create venv so packages can be cached. note: additional packages
+        # pip and setuptools will also be installed!
+        staging_venv_dir = self.charmdir / STAGING_VENV_DIRNAME
+        logger.debug("create staging venv in %r", str(staging_venv_dir))
```
facundobatista, line 252: Here and in other `logger.debug` calls: please start the sentence with uppercase.

charmcraft/charm_builder.py
```diff
@@ -0,0 +1,357 @@
...
+
+        retcode = _process_run(cmd)
+        if retcode:
+            raise CommandError("problems installing dependencies")
+
+
+def _process_run(cmd, **kwargs) -> int:
+    logger.debug("Running external command %s", cmd)
+    try:
+        ret = subprocess.run(cmd, **kwargs)
```
facundobatista, line 284: Shouldn't the output be captured and reported "in debug"?

charmcraft/charm_builder.py
```diff
@@ -0,0 +1,357 @@
...
+
+# The file name and template for the dispatch script
+DISPATCH_FILENAME = "dispatch"
+# If Juju doesn't support the dispatch mechanism, it will execute the
+# hook, and we'd need sys.argv[0] to be the name of the hook but it's
+# geting lost by calling this dispatch, so we fake JUJU_DISPATCH_PATH
+# to be the value it would've otherwise been.
+DISPATCH_CONTENT = """#!/bin/sh
+
+JUJU_DISPATCH_PATH="${{JUJU_DISPATCH_PATH:-$0}}" PYTHONPATH=lib:venv ./{entrypoint_relative_path}
```
facundobatista, line 48: This needs to be fixed if not using a "target" type of installation, but a virtualenv:

```
$ python3 -m venv foo
$ foo/bin/pip install requests
Collecting requests
  Using cached requests-2.26.0-py2.py3-none-any.whl (62 kB)
Collecting idna<4,>=2.5; python_version >= "3"
  Using cached idna-3.2-py3-none-any.whl (59 kB)
Collecting charset-normalizer~=2.0.0; python_version >= "3"
  Downloading charset_normalizer-2.0.3-py3-none-any.whl (35 kB)
Collecting certifi>=2017.4.17
  Using cached certifi-2021.5.30-py2.py3-none-any.whl (145 kB)
Collecting urllib3<1.27,>=1.21.1
  Using cached urllib3-1.26.6-py2.py3-none-any.whl (138 kB)
Installing collected packages: idna, charset-normalizer, certifi, urllib3, requests
Successfully installed certifi-2021.5.30 charset-normalizer-2.0.3 idna-3.2 requests-2.26.0 urllib3-1.26.6
$ PYTHONPATH=foo python3 -c "import requests;print(requests.__file__)"
/usr/lib/python3/dist-packages/requests/__init__.py
$ PYTHONPATH=foo/lib/python3.8/site-packages/ python3 -c "import requests;print(requests.__file__)"
/home/facundo/temp/tvenv/foo/lib/python3.8/site-packages/requests/__init__.py
```

charmcraft/charm_builder.py
```diff
@@ -0,0 +1,357 @@
...
+    def handle_dependencies(self):
+        """Handle from-directory and virtualenv dependencies."""
+        logger.debug("Installing dependencies")
+
+        if not self.python_packages and not self.requirements:
+            return
+
+        # create venv so packages can be cached. note: additional packages
+        # pip and setuptools will also be installed!
+        staging_venv_dir = self.charmdir / STAGING_VENV_DIRNAME
```
facundobatista, line 251: Note that here you're adding stuff to the project's directory, not inside the building one. This is something we never did so far.

facundobatista (changes requested):
There some problems with the changes beyond moving the code from one place to the other.

charmcraft/charm_builder.py
cmatsuoka: This is a temporary condition that exists only while the builder tool is invoked directly by the build command without the parts lifecycle. In this case, origin is charmdir and destination is buildpath. When invoked from the parts lifecycle, origin will be `<work_dir>/parts/charm/build` and destination will be `<work_dir>/parts/charm/install`, and the directory will not be created in charmdir.

charmcraft/charm_builder.py
cmatsuoka: The tool is executed using `polite_exec()` in the builder command, so its output is handled by the caller.

charmcraft/charm_builder.py
cmatsuoka: After the builder tool runs, only the virtual env `site.USER_SITE` directory is copied to the target `venv` directory, in such a way that the final charm payload remains with the same structure.

PR #457, opened 2021-07-21
Allow including resources when releasing charms directly in the upload command (CRAFT-381).

facundobatista:
Fixes 456.


cjp256 (approved):

jnsgruk:
Does this need a docs update on juju.is/docs/sdk?

facundobatista:
> Does this need a docs update on juju.is/docs/sdk?

I don't know, will check/decide.

sergiusens (commented):
I would assume that an error is required when using --resource without release

facundobatista:
> I would assume that an error is required when using --resource without release

Yes! But we need to talk about that. So far, in current situation, `--resource` will be ignored if no `--release`, so good enough for fixing the problem in `stable`.

sergiusens:
> > I would assume that an error is required when using --resource without release
> 
> Yes! But we need to talk about that. So far, in current situation, `--resource` will be ignored if no `--release`, so good enough for fixing the problem in `stable`.

ok, personally, I am not a fan of overloading the upload command with semantics for release.

sergiusens (approved):

facundobatista:
> > > I would assume that an error is required when using --resource without release
> > 
> > 
> > Yes! But we need to talk about that. So far, in current situation, `--resource` will be ignored if no `--release`, so good enough for fixing the problem in `stable`.
> 
> ok, personally, I am not a fan of overloading the upload command with semantics for release.

+1 to try to bring no more than this, but I think this is a must have. Having `--release` in the upload is super helpful to "upload and release", specially in CIs, unless everybody starts wrapping calls to both commands, parsing the revision from one to use it in the other. And if the charm has resources, releasing without them is breaking the charm.

PR #458, opened 2021-07-21
build: integrate craft-parts lifecycle (CRAFT-89, CRAFT-367, CRAFT-357)

cmatsuoka:
Create and run the craft-parts lifecycle, executing the charm builder
from a craft-parts charm build plugin.
    
### DIFF
--- a/charmcraft/commands/build.py
+++ b/charmcraft/commands/build.py
@@ -398,9 +398,7 @@ class Builder:
 
         return charm_name
 
-    def handle_package(
-        self, prime_dir, bases_config: Optional[BasesConfiguration] = None
-    ):
+    def handle_package(self, prime_dir, bases_config: Optional[BasesConfiguration] = None):
         """Handle the final package creation."""
         logger.debug("Creating the package itself")
         zipname = format_charm_file_name(self.metadata.name, bases_config)
--- a/charmcraft/commands/build.py
+++ b/charmcraft/commands/build.py
@@ -250,8 +250,10 @@ class Builder:
         """
         # add entrypoint
         if str(entrypoint) == entrypoint.name:
+            # the entry point is in the root of the project, just include it
             self._prime.append(str(entrypoint))
         else:
+            # the entry point is in a subdir, include the whole subtree
             self._prime.append(str(entrypoint.parts[0]))
 
         # add venv if there are requirements
--- a/charmcraft/commands/build.py
+++ b/charmcraft/commands/build.py
@@ -20,11 +20,10 @@ import logging
 import os
 import pathlib
 import subprocess
-import sys
 import zipfile
 from typing import List, Optional
 
-from charmcraft import charm_builder, linters
+from charmcraft import linters
 from charmcraft.bases import check_if_base_matches_host
 from charmcraft.cmdbase import BaseCommand, CommandError
 from charmcraft.config import Base, BasesConfiguration, Config
@@ -37,6 +36,7 @@ from charmcraft.env import (
 from charmcraft.logsetup import message_handler
 from charmcraft.manifest import create_manifest
 from charmcraft.metadata import parse_metadata_yaml
+from charmcraft.parts import PartsLifecycle, Step
 from charmcraft.providers import (
     capture_logs_from_instance,
     ensure_provider_is_available,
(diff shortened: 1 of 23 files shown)


charmcraft/parts.py
```diff
@@ -0,0 +1,243 @@
...
+
+    @classmethod
+    def get_build_snaps(cls) -> Set[str]:
+        """Return a set of required snaps to install in the build environment."""
+        return set()
+
+    def get_build_packages(self) -> Set[str]:
+        """Return a set of required packages to install in the build environment."""
+        return {
+            "python3-pip",
```
sergiusens, line 79: you can now remove the manual install of these done for the craft-providers integration

sergiusens: my comment was the other way around :-) Keep them here where they belong, and remove them from https://github.com/canonical/charmcraft/blob/7cdbeeeb569470b919e68b6c4212206b1e016649/charmcraft/providers.py#L324

cmatsuoka: Ah, I misread "craft-providers" there and parsed it as "craft-parts" for some reason. I guess I'm tired, fixing it.

charmcraft/commands/build.py
```diff
@@ -184,61 +205,66 @@ def build_charm(self, bases_config: BasesConfiguration) -> str:
...
+        """Add mandatory and optional charm files to the prime filter.
+
+        The prime filter should contain:
+        - The charm entry point, or the directory containing it if it's
+          not directly in the project dir.
+        - A set of mandatory charm files, including metadata.yaml, the
+          dispatcher and the hooks directory.
+        - A set of optional charm files.
+        """
+        # add entrypoint
```
facundobatista, line 251: It's somewhat explained in the docstring, but I would make it explicit in both blocks in the `if`...

```
        if str(entrypoint) == entrypoint.name:
            # the entry point is in the root of the project, just include it
            self._prime.append(str(entrypoint))
        else:
            # the entry point is in a subdir, include the whole subtree starting with it
            self._prime.append(str(entrypoint.parts[0]))
```

charmcraft/main.py
```diff
@@ -358,6 +359,7 @@ def main(argv=None):
     # process
     try:
         env.ensure_charmcraft_environment_is_supported()
+        setup_parts()
```
facundobatista, line 362: Is this setup expensive? Maybe we should setup parts only in the commands that use it? (for example, it doesn't make sense to setup parts if you do `charmcraft help` or `charmcraft version`).

charmcraft/parts.py
```diff
@@ -0,0 +1,243 @@
...
+
+    :param data: The part data to validate.
+    """
+    if not isinstance(data, dict):
+        raise TypeError("value must be a dictionary")
+
+    # copy the original data, we'll modify it
+    spec = data.copy()
+
+    plugin_name = spec.get("plugin", "")
```
facundobatista, line 148: No need for the default `""`, right? if you get None a ValueError will be raised below anyway...

facundobatista (approved):
Looks awesome! I annotated some details for you to consider, but +1!!

charmcraft/main.py
cmatsuoka: The setup is very lightweight and was designed to be called at the application start with practically no overhead.

charmcraft/commands/build.py
```diff
@@ -65,6 +65,23 @@
 MANDATORY_HOOK_NAMES = {"install", "start", "upgrade-charm"}
 HOOKS_DIR = "hooks"
 
+CHARM_FILES = [
```
sergiusens, line 68: consider moving this to the plugin

charmcraft/commands/build.py
```diff
@@ -65,6 +65,23 @@
...
+    "metadata.yaml",
+    DISPATCH_FILENAME,
+    HOOKS_DIR,
+]
+
+CHARM_OPTIONAL = [
```
sergiusens, line 74: consider moving this to the plugin

charmcraft/commands/build.py
```diff
@@ -184,61 +205,68 @@ def build_charm(self, bases_config: BasesConfiguration) -> str:
...
-        retcode = polite_exec(build_cmd)
-        if retcode:
-            raise CommandError("problems running charm builder")
+        entrypoint = self.entrypoint.relative_to(self.charmdir)
+
+        # add charm files to the prime filter
+        self._set_prime_filter(entrypoint)
+
+        # set source for buiding
+        self._charm_part["source"] = str(self.charmdir)
```
sergiusens, line 214: TODO review why this is not set in the default Config

charmcraft/commands/build.py
```diff
@@ -184,61 +205,68 @@ def build_charm(self, bases_config: BasesConfiguration) -> str:
...
             bases_config,
             linting_results,
         )
 
-        zipname = self.handle_package(bases_config)
+        zipname = self.handle_package(lifecycle.prime_dir, bases_config)
         logger.info("Created '%s'.", zipname)
         return zipname
 
+    def _set_prime_filter(self, entrypoint: pathlib.Path):
```
sergiusens, line 241: this most likely could all go into the plugin (entrypoint perhaps only once we have the plugin property for)

charmcraft/parts.py
```diff
@@ -0,0 +1,241 @@
...
+    plugin and source keywords can be used.
+
+    Extra files to be included in the charm payload must be listed under
+    the ``prime`` file filter.
+    """
+
+    properties_class = CharmPluginProperties
+
+    @classmethod
+    def get_build_snaps(cls) -> Set[str]:
```
sergiusens, line 72: TODO remove get_build_snaps

sergiusens (approved):

cmatsuoka:
Will check the impact in craft-parts of doing the transformations at plugin level and refactor that in a follow-up.

PR #455, opened 2021-07-20
Added an "analyze" command that exposes the linters/attributes results (CRAFT-319).

facundobatista:

PR #460, opened 2021-07-22
Isolated the linters/checkers, no more shared state between them (CRAFT-383).

facundobatista:
This is to avoid having problems when a "dependency checker" is ignored (or crashes), and then the "dependent checker" is missing the needed information.

The branch is big, but there's no really a change in functionality, I just:

- separated code into a function that is now used from two places
- parsed the metadata twice in two different places
- adapted all the tests

Note that this depends on [the previous branch](https://github.com/canonical/charmcraft/pull/455).


sergiusens (approved):
I suppose that order is no longer relevant, but nothing wrong with keeping it still (https://github.com/canonical/charmcraft/pull/460/files#diff-8c2f0f90a2be65dbfed7b57c53fec486604f281e95122e795046e260dd07de25R242)

cmatsuoka (approved):
Much nicer!

cjp256 (approved):

PR #462, opened 2021-07-23
Added a --format=json option to the analyze command (CRAFT-320)

facundobatista:


charmcraft/commands/analyze.py
```diff
@@ -75,6 +83,20 @@ def run(self, parsed_args):
...
+            info = [
+                {
+                    "name": r.name,
+                    "result": r.result,
+                    "url": r.url,
+                    "type": r.check_type,
+                }
+                for r in linting_results
+            ]
+            logger.info(json.dumps(info))
```
sergiusens, line 97: wouldn't we preffer to pretty print the output? `indent=4` to `json.dumps`

sergiusens (approved):
Looks good, I would advise to pretty print the output even if it is machine oriented.

charmcraft/commands/analyze.py
```diff
@@ -54,6 +57,11 @@ def fill_parser(self, parser):
             action="store_true",
             help="Force to run all checks, even those set to ignore in the configuration",
         )
+        parser.add_argument(
+            "--format",
+            choices=[JSON_FORMAT],
```
cmatsuoka, line 62: Will we offer other formats in the future?

cmatsuoka (approved):
Looks good.

charmcraft/commands/analyze.py
facundobatista: Maybe! No plans by now, though.

charmcraft/commands/analyze.py
facundobatista: Ooooooh, very nice!! Used it!

PR #466, opened 2021-07-26
pack: implement bundle plugin (CRAFT-360)

cmatsuoka:
Add the craft-parts bundle plugin and invoke the parts lifecycle from
the bundle packing command.

### DIFF
--- a/charmcraft/commands/build.py
+++ b/charmcraft/commands/build.py
@@ -218,6 +218,7 @@ class Builder:
         lifecycle = PartsLifecycle(
             self._parts,
             work_dir=self.buildpath,
+            ignore_local_sources=["*.charm"],
             entrypoint=entrypoint,
             requirements=[str(p) for p in self.requirement_paths],
         )
--- a/charmcraft/commands/pack.py
+++ b/charmcraft/commands/pack.py
@@ -17,49 +17,33 @@
 """Infrastructure for the 'pack' command."""
 
 import logging
+import os
+import pathlib
 import zipfile
 from argparse import Namespace
 
 from charmcraft.cmdbase import BaseCommand, CommandError
 from charmcraft.commands import build
 from charmcraft.manifest import create_manifest
+from charmcraft.parts import PartsLifecycle, Step
 from charmcraft.utils import SingleOptionEnsurer, load_yaml, useful_filepath
 
 logger = logging.getLogger(__name__)
 
 # the minimum set of files in a bundle
-MANDATORY_FILES = {"bundle.yaml", "manifest.yaml", "README.md"}
+MANDATORY_FILES = {"bundle.yaml", "README.md"}
 
 
-def build_zip(zippath, basedir, fpaths):
+def build_zip(zippath, prime_dir):
     """Build the final file."""
     zipfh = zipfile.ZipFile(zippath, "w", zipfile.ZIP_DEFLATED)
-    for fpath in fpaths:
-        zipfh.write(fpath, fpath.relative_to(basedir))
+    for dirpath, dirnames, filenames in os.walk(prime_dir, followlinks=True):
+        dirpath = pathlib.Path(dirpath)
+        for filename in filenames:
+            filepath = dirpath / filename
+            zipfh.write(str(filepath), str(filepath.relative_to(prime_dir)))
+
     zipfh.close()
-
-
-def get_paths_to_include(config):
-    """Get all file/dir paths to include."""
-    dirpath = config.project.dirpath
-    allpaths = set()
-
-    # all mandatory files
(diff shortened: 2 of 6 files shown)


sergiusens (approved):

charmcraft/commands/build.py
```diff
@@ -218,6 +218,7 @@ def build_charm(self, bases_config: BasesConfiguration) -> str:
         lifecycle = PartsLifecycle(
             self._parts,
             work_dir=self.buildpath,
+            ignore_local_sources=["*.charm"],
```
jnsgruk, line 221: Perhaps I'm misunderstanding this line, but should there be consideration for anything in the `.jujuignore` file too?

charmcraft/commands/pack.py
```diff
@@ -169,8 +155,13 @@ def _pack_charm(self, parsed_args):
 
     def _pack_bundle(self):
         """Pack a bundle."""
+        project = self.config.project
+        config_parts = self.config.parts.copy()
+        charm_part = config_parts.setdefault("bundle", {})
```
jnsgruk, line 160: This seems like an odd choice of variable name here? Should `charm_part` be `bundle_part`?

charmcraft/commands/pack.py
```diff
@@ -183,19 +174,28 @@ def _pack_bundle(self):
                 "file {!r}.".format(str(bundle_filepath))
```
jnsgruk, line 174: Could we link to a bundle spec in the error message to make this easier to solve?

charmcraft/commands/pack.py
```diff
@@ -183,19 +174,28 @@ def _pack_bundle(self):
...
+
+        # set source for buiding
+        charm_part["source"] = str(project.dirpath)
+
+        # run the parts lifecycle
+        logger.debug("Parts definition: %s", config_parts)
+        lifecycle = PartsLifecycle(
+            config_parts,
+            work_dir=project.dirpath / build.BUILD_DIRNAME,
+            ignore_local_sources=[bundle_name + ".zip"],
```
jnsgruk, line 192: As above, do we need to consider `.jujuignore`?

jnsgruk (approved):
Looks good, nice work. Ran the code on a sample using the snap, and all good.

charmcraft/commands/pack.py
sergiusens: good catch!

charmcraft/commands/pack.py
sergiusens: sounds like a sane idea, but the bundle reference https://juju.is/docs/sdk/bundle-reference looks more like a mix of mini-how-to's today. This might be catered to a previous bundle world as I don't see the name mentioned anywhere. Should we fix that first?

jnsgruk: Yeh, I think that bundle reference is a little out of date. I don't know what the defacto source of info on bundle spec is now? @jameinel?

charmcraft/commands/build.py
sergiusens: this should be handled in charm_builder.py (https://github.com/canonical/charmcraft/blob/9dc20872f1715585dfa23e2332bfece47126072f/charmcraft/charm_builder.py#L156) which is prior to this.

sergiusens: and that comes from

https://github.com/canonical/charmcraft/blob/9dc20872f1715585dfa23e2332bfece47126072f/charmcraft/charm_builder.py#L78

jnsgruk: Ah right, could very well be sorted already

charmcraft/commands/pack.py
cmatsuoka: Oh, oops, copy-and-paste error. This should be `bundle_part`.

charmcraft/commands/build.py
cmatsuoka: `.jujuignore` is handled by the charm build logic called from the parts lifecycle. The main purpose of this `*.charm` ignore filter is to prevent the lifecycle to consider the newly created charm file as part of the source, otherwise every time the charm file is created the lifecycle manager will consider the source outdated (because there's a new file there).

charmcraft/commands/pack.py
cmatsuoka: The current implementation of bundle creation doesn't parse `.jujuignore`. The ignore filter here prevents bundle files to include previous copies of themselves if they're created inside the bundle source directory.

charmcraft/commands/pack.py
cmatsuoka: We can add the link in a follow-up PR.

PR #473, opened 2021-08-02
snap: add configure hook with support for provider (CRAFT-363)

cjp256:
Enable (unused) configuration for 'provider', erroring if set to
anything other than multipass or lxd.  Update requirements
to include snap-helpers.

Example error:
```
$ snap set charmcraft provider=invalid
error: cannot perform the following tasks:
- Run configure hook of "charmcraft" snap (run hook "configure":
  Invalid snap configuration: provider 'invalid' is not supported.)
```

### DIFF
--- a/charmcraft/snap.py
+++ b/charmcraft/snap.py
@@ -16,8 +16,8 @@
 
 """Logic for snap configuration."""
 
-from dataclasses import dataclass
 import logging
+from dataclasses import dataclass
 from typing import Optional
 
 import snaphelpers
--- a/charmcraft/snap.py
+++ b/charmcraft/snap.py
@@ -16,16 +16,16 @@
 
 """Logic for snap configuration."""
 
+from dataclasses import dataclass
 import logging
 from typing import Optional
 
-import attr
 import snaphelpers
 
 logger = logging.getLogger(__name__)
 
 
-@attr.s(auto_attribs=True)
+@dataclass
 class CharmcraftSnapConfiguration:
     """Charmcraft's snap configuration options."""
 
--- a/charmcraft/snap.py
+++ b/charmcraft/snap.py
@@ -1,0 +1,63 @@
+# Copyright 2021 Canonical Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# For further info, check https://github.com/canonical/charmcraft
+
+"""Logic for snap configuration."""
+
+import logging
+from typing import Optional
+
+import attr
+import snaphelpers
+
+logger = logging.getLogger(__name__)
+
+
+@attr.s(auto_attribs=True)
+class CharmcraftSnapConfiguration:
+    """Charmcraft's snap configuration options."""
+
+    provider: Optional[str] = None
+
+
+def _get_config_key(*, snap_config: snaphelpers.SnapConfig, key: str, default=None):
+    """Get snap configuration for specified key.
+
+    :returns: Returns key's value or default if undefined.
+    """
+    try:
+        return snap_config.get(key)
+    except snaphelpers._conf.UnknownConfigKey:
+        return default
+
+
+def get_snap_configuration() -> CharmcraftSnapConfiguration:
+    """Get unvalidated snap configuration.
+
+    :returns: Current snap configuration.
+    """
+    snap_config = snaphelpers.SnapConfig()
+    provider = _get_config_key(snap_config=snap_config, key="provider")
+
+    return CharmcraftSnapConfiguration(provider=provider)
+
+
+def validate_snap_configuration(cfg: CharmcraftSnapConfiguration):
+    """Validate given snap configuration.
+
+    :raises ValueError: on error.
+    """
+    if cfg.provider is not None and cfg.provider not in ["lxd", "multipass"]:
+        raise ValueError(f"provider {cfg.provider!r} is not supported")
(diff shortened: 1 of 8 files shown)


cjp256:
1) If this is the approach we want to take, I can add unit tests for the configure script.
2) "passthrough" is required until stable Snapcraft supports environment for hooks (WIP).

snap/hooks/configure
```diff
@@ -0,0 +1,34 @@
...
+    try:
+        return snap_config.get(key)
+    except snaphelpers._conf.UnknownConfigKey:
+        return default
+
+
+def validate_provider(snap_config: snaphelpers.SnapConfig):
+    provider = read_config_key(snap_config, "provider")
+
+    if provider and provider not in ["lxd", "multipass"]:
```
sergiusens, line 18: multipass should only be valid once we add support for it

snap/snapcraft.yaml
```diff
@@ -120,9 +120,17 @@ parts:
...
       usr/lib/git-core: git/git-core
       usr/share/git-core/templates: git/templates
       usr/bin/git: bin/git
       lib/python3.8/site-packages: lib/
+
+hooks:
+  configure:
+    passthrough:
+      environment:
+        PATH: "$SNAP/bin:$PATH"
```
sergiusens, line 136: given the opportunity (this is it), I would restrict the path to just the things we need as this is a classic confined snap

setup.py
```diff
@@ -93,7 +94,10 @@
             "Programming Language :: Python :: 3",
         ],
         entry_points={
-            "console_scripts": ["charmcraft = charmcraft.main:main"],
+            "console_scripts": [
+                "charmcraft = charmcraft.main:main",
+                "charmcraft-configure-hook = charmcraft.snap_hooks:configure_main",
```
sergiusens, line 99: does this exist?

sergiusens (commented):
Looks good

snap/hooks/configure
cjp256: I put it in initially so we can toggle it once I start adding multipass support.  I expect this option would not be documented as supported until it is complete.

snap/snapcraft.yaml
cjp256: Good point, will fix!

setup.py
cjp256: Not anymore, will remove.

charmcraft/snap.py
```diff
@@ -0,0 +1,63 @@
...
+import logging
+from typing import Optional
+
+import attr
+import snaphelpers
+
+logger = logging.getLogger(__name__)
+
+
+@attr.s(auto_attribs=True)
```
sergiusens, line 28: why aren't you using dataclasses now?

snap/hooks/configure
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env python3
+
+import sys
+
+from charmcraft import snap
```
sergiusens, line 5: I would personally move all the logic in charmcraft.snap here. You can import snap.hooks.configure if you wanted from pytest (it may be inconvenient if the root is moved though). In any case, I would rename snap to snap_hooks or something similar.

sergiusens (approved):

charmcraft/snap.py
cjp256: In this case, either works just fine. I had just grepped the codebase and found only one instance of attr usage, but no dataclasses.  Just for consistency I stuck with attr.

If we want to standardize on dataclass here, I'll be happy to change this (and the other instance).

snap/hooks/configure
cjp256: The logic in charmcraft.snap is meant to be used by the charmcraft application at runtime to get/validate the snap configuration (to actually use it).

sergiusens: @cjp256 good point!

charmcraft/snap.py
sergiusens: attrs was used at the time only because of the runtime dependency. I would change here, the other one in a follow up

facundobatista: Yes, we used `attrs` because of dataclasses were not supported in 3.5. Let's go full dataclasses!

setup.py
```diff
@@ -93,7 +94,9 @@
             "Programming Language :: Python :: 3",
         ],
         entry_points={
-            "console_scripts": ["charmcraft = charmcraft.main:main"],
+            "console_scripts": [
+                "charmcraft = charmcraft.main:main",
+            ],
```
facundobatista, line 99: This change is not really a change, right?

facundobatista (changes requested):
Annotated a couple of details. Also, a question: Is this config already used from the "production" code to decide the provider?

snap/snapcraft.yaml
cjp256: Done!

setup.py
cjp256: Fixed!

charmcraft/snap.py
cjp256: :+1:

setup.py
cjp256: Nope, just leftover formatting delta from prior change. Will fix

charmcraft/snap.py
cjp256: other change done @ https://github.com/canonical/charmcraft/pull/483

setup.py
cjp256: Fixed

cjp256:
> Annotated a couple of details. Also, a question: Is this config already used from the "production" code to decide the provider?

No.  Future work will actually enable usage of it.

charmcraft/snap.py
```diff
@@ -0,0 +1,63 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# For further info, check https://github.com/canonical/charmcraft
+
+"""Logic for snap configuration."""
+
+from dataclasses import dataclass
```
facundobatista, line 19: Very small detail: we group the `from ... import` after the `import`s (without a blank line between).

facundobatista (approved):
Wonderful! Annotated a small detail. Thanks!

charmcraft/snap.py
cjp256: Fixed (and why i like to use isort :rofl:)

PR #474, opened 2021-08-02
charmcraft: perform LXD checks before use (CRAFT-420)

cjp256:
Invoke ensure_lxd_is_ready() interface before using LXD
provider. This consolidates a couple of existing checks
done elsewhere as well as a new one for ensuring the user
has the necessary permissions to use LXD.

Minor simplification of imports used in providers module
to improve and update mocks accordingly.

Example error:
```
$ /snap/bin/charmcraft pack
LXD requires additional permissions.
Please ensure that the user is in the 'lxd' group.
(full execution logs in '/home/test1/snap/charmcraft/common/charmcraft-log-0xgcist7')
```

### DIFF
--- a/charmcraft/providers.py
+++ b/charmcraft/providers.py
@@ -116,12 +116,12 @@ def ensure_provider_is_available() -> None:
                 lxd.install()
             except lxd.LXDInstallationError as error:
                 raise CommandError(
-                    "Failed to install LXD. Please visit https://snapcraft.io/lxd for "
+                    "Failed to install LXD. Visit https://snapcraft.io/lxd for "
                     "instructions on how to install the LXD snap for your distribution"
                 ) from error
         else:
             raise CommandError(
-                "LXD is required, but not installed. Please visit https://snapcraft.io/lxd for "
+                "LXD is required, but not installed. Visit https://snapcraft.io/lxd for "
                 "instructions on how to install the LXD snap for your distribution"
             )
 
--- a/charmcraft/providers.py
+++ b/charmcraft/providers.py
@@ -27,16 +27,10 @@ from typing import Dict, List, Optional, Tuple, Union
 
 from craft_providers import Executor, bases, lxd
 from craft_providers.actions import snap_installer
-from craft_providers.lxd import LXDInstallationError
-from craft_providers.lxd import installer as lxd_installer
-from craft_providers.lxd.remotes import configure_buildd_image_remote
 
 from charmcraft.cmdbase import CommandError
 from charmcraft.config import Base
-from charmcraft.env import (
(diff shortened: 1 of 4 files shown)


charmcraft/providers.py
```diff
@@ -112,26 +106,29 @@ def ensure_provider_is_available() -> None:
...
+        if confirm_with_user(
+            "LXD is required, but not installed. Do you wish to install LXD "
+            "and configure it with the defaults?",
+            default=False,
+        ):
+            try:
+                lxd.install()
+            except lxd.LXDInstallationError as error:
+                raise CommandError(
+                    "Failed to install LXD. Please visit https://snapcraft.io/lxd for "
```
sergiusens, line 119: Remove Please, start with Visit -> https://discourse.ubuntu.com/t/tone-of-voice/18883

charmcraft/providers.py
```diff
@@ -112,26 +106,29 @@ def ensure_provider_is_available() -> None:
...
+                    "instructions on how to install the LXD snap for your distribution"
+                ) from error
+        else:
             raise CommandError(
-                "Failed to install LXD. Please visit https://snapcraft.io/lxd for "
+                "LXD is required, but not installed. Please visit https://snapcraft.io/lxd for "
```
sergiusens, line 124: same

sergiusens (changes requested):
Looks generally good, tone of voice needs adjusting.

charmcraft/providers.py
cjp256: Ah yes, I questioned it when I wrote it way back when and never did reconsider. :thinking:  I think I just feel compelled to be polite :rofl:  will fix (but note that this was just an indent change, not new text).

cjp256: done!

charmcraft/providers.py
cjp256: done!

sergiusens (approved):

charmcraft/providers.py
```diff
@@ -112,26 +106,29 @@ def ensure_provider_is_available() -> None:
...
-        "and configure it with the defaults?",
-        default=False,
-    ):
-        try:
-            lxd_installer.install()
-        except LXDInstallationError as error:
+    if not is_provider_available():
+        if confirm_with_user(
+            "LXD is required, but not installed. Do you wish to install LXD "
+            "and configure it with the defaults?",
```
facundobatista, line 112: Not worth for this branch, but for the future: for readability in these cases please write the long message outside the `if`. Thanks!

facundobatista (approved):
Looks great!

PR #475, opened 2021-08-02
Fixed a test to work ok also when running tests in developer mode.

facundobatista:


tests/test_main.py
```diff
@@ -418,8 +418,10 @@ def test_main_environment_is_supported_error(
     mock_is_charmcraft_running_in_supported_environment.return_value = False
```
cjp256, line 418: I messed this up I think.  I'd suggest:
- replacing `mock_is_charmcraft_running_in_supported_environment` with `mock_ensure_charmcraft_environment_is_supported` instead of trying to twiddle the external module checks to cause it to fail.
- use that mock to raise `CommandError("foo")` as the `side_effect`.

cjp256: If you want I can do it.  But approving as-is regardless.

cjp256 (approved):

tests/test_main.py
facundobatista: Nice idea! Improved as suggested.

sergiusens (approved):

cjp256 (approved):
LGTM, thanks!

PR #476, opened 2021-08-02
Fixed permissions bits after extracting the charm's content to analyze (CRAFT-429)

facundobatista:


tests/commands/test_analyze.py
```diff
@@ -76,6 +76,29 @@ def fake_analyze(passed_config, passed_basedir, *, override_ignore_config):
     assert fake_analyze_called
 
 
+@pytest.mark.parametrize("modebits", [0o777, 0o750, 0o444])
+def test_expanded_charm_permissions(config, tmp_path, monkeypatch, modebits):
+    """Check that the expanded charm keep original permissions."""
```
sergiusens, line 81:
```suggestion
    """Check that the expanded charm keeps original permissions."""
```

charmcraft/commands/analyze.py
```diff
@@ -61,23 +61,37 @@ def fill_parser(self, parser):
...
+        tmpdir = pathlib.Path(tempfile.mkdtemp())
         try:
-            zf = zipfile.ZipFile(filepath)
-            zf.extractall(path=tmpdir)
+            zf = zipfile.ZipFile(str(filepath))
+            zf.extractall(path=str(tmpdir))
         except Exception as exc:
             raise CommandError(
-                "Cannot open the indicated charm file {!r}: {!r}.".format(filepath, exc)
+                "Cannot open the indicated charm file {!r}: {!r}.".format(str(filepath), exc)
```
sergiusens, line 72:
```suggestion
                "Cannot open charm file {!r}: {!r}.".format(str(filepath), exc)
```

sergiusens: Mayb f-string :-)

sergiusens (approved):

charmcraft/commands/analyze.py
```diff
@@ -61,23 +61,37 @@ def fill_parser(self, parser):
...
             )
 
+        # fix permissions as extractall does not keep them (see bpo15795)
```
cjp256, line 75: googling bpo15795 found nothing :-/

```suggestion
        # fix permissions as extractall does not keep them (see https://bugs.python.org/issue15795)
```

Very interesting.

cjp256 (approved):
With suggested changes, :+1: Thanks!

tests/commands/test_analyze.py
facundobatista: thanks

charmcraft/commands/analyze.py
facundobatista: Improved.

charmcraft/commands/analyze.py
facundobatista: Ack.

PR #484, opened 2021-08-04
providers: refactor to better support multiple providers (CRAFT-362)

cjp256:
With the upcoming introduction of Multipass, the providers module
would grow significantly in complexity.

Convert providers module into a package, splitting out functionality
into well scoped modules.

Introduce a common Provider class to be the general interface for
any Charmcraft-supported providers, and refactor the LXD-specifics
into a new subclass LXDProvider.  Add interface get_provider() for
users of providers to get the correct provider (only returns an
instance of LXDProvider at this time).

There should be no change in behaviors with this commit, the large diff
is to accompany the shuffling of the whole module into a new package,
folding this functionality into the new provider classes, and updating
the test mocks.


sergiusens (approved):
I would just rename `providers._helpers` to something for specific

facundobatista (approved):
Great, thanks!

cjp256:
> I would just rename `providers._helpers` to something for specific

Done (_logs), thanks!

PR #488, opened 2021-08-09
packaging, providers: update craft-parts to 1.0 (CRAFT-452)

cmatsuoka:
Update craft-parts version to 1.0 in package setup and requirements,
and remove unnecessary installation of sudo in provider setup.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -133,6 +133,31 @@ jobs:
           test -f "destructive-mode-tests_ubuntu-$VERSION_ID-amd64.charm"
           popd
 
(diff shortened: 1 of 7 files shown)


cjp256:
Can you add a set of integration / smoke tests which install build packages?

cmatsuoka:
Added s smoke test using parts and build requires, both installing the package and not installing it if it's already present on the system (allowing charmcraft to run without superuser privileges in the latter case).

facundobatista (approved):
Nice!

cjp256 (approved):
Thanks!!

sergiusens (approved):

PR #491, opened 2021-08-11
providers: introduce multipass support (CRAFT-409)

cjp256:
Add support for Multipass similar to LXD.

To determine provider on linux:

- check CHARMCRAFT_PROVIDER environment variable if in
  developer mode.

- check snap configuration if running as snap.

- fall back to platform default.

Update execute_run() to use cwd=pathlib.Path(x) interface.

Add unit tests and a smoke test for Linux & Multipass.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -117,8 +117,8 @@ jobs:
           sg lxd -c "charmcraft -v clean"
           popd
 
-          mkdir -p another-directory
-          pushd another-directory
+          mkdir -p another-test
+          pushd another-test
           sg lxd -c "charmcraft -v pack --project-dir ../charm-smoke-test"
           test -f *.charm
           sg lxd -c "charmcraft -v clean --project-dir ../charm-smoke-test"
--- a/charmcraft/providers/_get_provider.py
+++ b/charmcraft/providers/_get_provider.py
@@ -17,12 +17,55 @@
 """Build environment provider support for charmcraft."""
 
 import logging
+import os
+import sys
+
+from charmcraft.snap import get_snap_configuration
+from charmcraft.cmdbase import CommandError
+from charmcraft.env import is_charmcraft_running_from_snap, is_charmcraft_running_in_developer_mode
 
 from ._lxd import LXDProvider
+from ._multipass import MultipassProvider
 
 logger = logging.getLogger(__name__)
 
 
+def _get_platform_default_provider() -> str:
+    if sys.platform == "linux":
+        return "lxd"
+
+    return "multipass"
+
+
 def get_provider():
-    """Get the configured or appropriate provider for the host OS."""
-    return LXDProvider()
+    """Get the configured or appropriate provider for the host OS.
+
+    If platform is not Linux, use Multipass.
+
+    If platform is Linux:
+    (1) use provider specified with CHARMCRAFT_PROVIDER if running
+        in developer mode,
+    (2) use provider specified with snap configuration if running
+        as snap,
+    (3) default to platform default (LXD on Linux).
+
+    :return: Provider instance.
+    """
+    provider = None
+
+    if is_charmcraft_running_in_developer_mode():
+        provider = os.getenv("CHARMCRAFT_PROVIDER")
+
+    if provider is None and is_charmcraft_running_from_snap():
+        snap_config = get_snap_configuration()
+        provider = snap_config.provider if snap_config else None
+
+    if provider is None:
+        provider = _get_platform_default_provider()
+
+    if provider == "lxd":
+        return LXDProvider()
+    elif provider == "multipass":
+        return MultipassProvider()
+
+    raise CommandError(f"Unsupported provider specified {provider!r}.")
--- a/charmcraft/providers/_get_provider.py
+++ b/charmcraft/providers/_get_provider.py
@@ -20,9 +20,9 @@ import logging
 import os
 import sys
 
-from charmcraft.snap import get_snap_configuration
 from charmcraft.cmdbase import CommandError
 from charmcraft.env import is_charmcraft_running_from_snap, is_charmcraft_running_in_developer_mode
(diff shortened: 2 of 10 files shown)


.github/workflows/tests.yaml
```diff
@@ -164,3 +161,12 @@ jobs:
...
           fi
+      - name: Linux Multipass Provider Tests
+        run: |
+          sudo snap set charmcraft provider=multipass
+
+          pushd "$(mktemp -d)"
+          charmcraft -v pack --project-dir ../charm-smoke-test
+          test -f *.charm
+          charmcraft -v clean --project-dir ../charm-smoke-test
+          popd
```
sergiusens, line 172: I know it is probably an irrelevant question, but is popd needed here? If it really is then it probably needs to be trapped.

cjp256: Trapped?  In this case I don't think the popd is required because its just one (last) test, but the pattern is appropriate I think.  Either way I removed this test because it has to run on macos to get nested virt.

sergiusens: trap is like go's defer, you would do something like:

```
mkdir foo
trap "rm -rf foo" EXIT
pushd foo
exit 1
```

cjp256: AFAIK any command that fails will result in an abort of the remainder of the test

cjp256: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#using-a-specific-shell

charmcraft/providers/_get_provider.py
```diff
@@ -17,12 +17,55 @@
...
+        in developer mode,
+    (2) use provider specified with snap configuration if running
+        as snap,
+    (3) default to platform default (LXD on Linux).
+
+    :return: Provider instance.
+    """
+    provider = None
+
+    if is_charmcraft_running_in_developer_mode():
```
sergiusens, line 56: personally I would restrict this to snapctl for now, is there a need for this envirnonment variable now? What are the user stories solved with it?

cjp256: It's the developer story, which is why I restrict the env var to developer mode (CHARMCRAFT_DEVELOPER=y).  When developing on it, it prevents me from having to rebuild every change as a snap...

sergiusens (approved):

charmcraft/providers/_multipass.py
```diff
@@ -0,0 +1,175 @@
...
+        if not self.is_provider_available():
+            return deleted
+
+        inode = str(project_path.stat().st_ino)
+
+        for name in self.multipass.list():
+            match_regex = f"^charmcraft-{charm_name}-{inode}-.+-.+-.+$"
+            if re.match(match_regex, name):
+                logger.debug("Deleting Multipass VM %r.", name)
+                self.multipass.delete(
```
facundobatista, line 76: This is always successful? May we hit the situation of deleting one env, failing to delete the second, and need to delete the third?

facundobatista (approved):
Approved with a couple of inline questions...

charmcraft/providers/_multipass.py
```diff
@@ -0,0 +1,175 @@
...
+
+        :returns: List of containers deleted.
+        """
+        deleted: List[str] = []
+
+        # Nothing to do if provider is not installed.
+        if not self.is_provider_available():
+            return deleted
+
+        inode = str(project_path.stat().st_ino)
```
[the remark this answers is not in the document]
sergiusens: I suspect this is not needed due to the format string below

charmcraft/providers/_multipass.py
sergiusens: this should be try, and catched with a message

charmcraft/providers/_multipass.py
```diff
@@ -0,0 +1,175 @@
...
+
+        environment = self.get_command_environment()
+        base_configuration = CharmcraftBuilddBaseConfiguration(
+            alias=alias, environment=environment, hostname=instance_name
+        )
+        instance = multipass.launch(
+            name=instance_name,
+            base_configuration=base_configuration,
+            image_name=base.channel,
+            cpus=2,
```
sergiusens, line 161: can you come up with a proposal to customize this?

charmcraft/providers/_multipass.py
```diff
@@ -0,0 +1,175 @@
...
+            disk_gb=64,
+            mem_gb=2,
```
sergiusens, line 163: can you come up with a proposal to customize this?

sergiusens (approved):

charmcraft/providers/_multipass.py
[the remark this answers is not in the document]
cjp256: Ah yes, leftover from previous usage. Will remove

charmcraft/providers/_multipass.py
cjp256: will fix

charmcraft/providers/_multipass.py
cjp256: Sure thing.

We can support the options in snap configuration without issue, but as the primary use case for Multipass is not Linux, we should probably support either environment variables (such as CHARMCRAFT_MULTIPASS_MEMORY=4) or command-line parameters (such as --multipass-memory=4).

I assume environment variables would be best?

charmcraft/providers/_multipass.py
cjp256: Fixed for both this and list. (and added tests)

charmcraft/providers/_multipass.py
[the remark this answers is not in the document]
cjp256: Fixed

charmcraft/providers/_multipass.py
cjp256: Fixed!

PR #495, opened 2021-08-12
pack: add support for --debug, --shell, and --shell-after (CRAFT-408)

cjp256:
Add flags to support debugging Charm building in target environment.
Execute this shell within the context of managed or destructive mode
to make behavior consistent regardless of provider usage.

--debug will pop a shell on failure in the lifecycle.

--shell will pop a shell instead of building a charm.

--shell will pop a shell after building the charm.


cmatsuoka (approved):
This should work for shell/shell-after until we solve environment variable issues in craft-parts.

charmcraft/commands/build.py
```diff
@@ -145,6 +146,9 @@ def __init__(self, args, config):
         self.entrypoint = args["entrypoint"]
         self.requirement_paths = args["requirement"]
         self.force_packing = args["force"]
+        self.debug = args["debug"]
+        self.shell = args["shell"]
+        self.shell_after = args["shell_after"]
```
facundobatista, line 151: Let's talk about these three command line options for this functionality.

On one hand, is this the same `--debug` that we'll use to control "extreme verbosity" in the tool output, or we have names collision?

On the other hand, what about combinations between these options? I don't see any control of the user not passing `--shell` and `--debug` at the same time, for example, and one makes the other to not work, in the code. 

Also I don't like that `--shell-after` is explicitly after, while `--shell` does not indicate when it happens. Also, `debug` will open a shell, while the name does not indicate that at all...

What if we only add a `--shell=WHEN` option, with three possible values: `before`, `after`, and `exception`?

charmcraft/shell.py
```diff
@@ -0,0 +1,32 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def launch_shell(*, cwd: Optional[pathlib.Path] = None) -> None:
+    """Launch a user shell for debugging environment.
+
+    :param cwd: Working directory to start user in.
+    """
+    subprocess.run(["bash"], check=False, cwd=cwd)
```
facundobatista, line 32: Does it worth to have a whole new module, only to have inside ONE function, that even has ONE line?

Do we plan for this module to grow in the future?

facundobatista (dismissed):

charmcraft/commands/build.py
cmatsuoka: The `--debug` clash is something we need to address. It will happen in all tools using craft-cli and providers+parts.

charmcraft/shell.py
cjp256: I placed it in its own module to prevent unnecessary coupling between different modules I expected to use it.  I had originally expected that `parts` module would also use it, but that's no longer the case.  OTOH, I expect it to get a little more complex in a cross-OS future.

I can consolidate it into `commands/build` and import from `commands/pack` ?

facundobatista (approved):
Great, thanks!

PR #496, opened 2021-08-13
build: fix parts workdir in managed mode (CRAFT-459)

cmatsuoka:
When building in managed mode use workdir outside of project dir to
prevent leaks to the external environment.

### DIFF
--- a/charmcraft/commands/build.py
+++ b/charmcraft/commands/build.py
@@ -28,7 +28,6 @@ from charmcraft.bases import check_if_base_matches_host
 from charmcraft.cmdbase import BaseCommand, CommandError
 from charmcraft.config import Base, BasesConfiguration, Config
 from charmcraft.deprecations import notify_deprecation
-from charmcraft.env import is_charmcraft_running_in_managed_mode
 from charmcraft.logsetup import message_handler
 from charmcraft.manifest import create_manifest
 from charmcraft.metadata import parse_metadata_yaml
@@ -202,7 +201,7 @@ class Builder:
 
         self._handle_deprecated_cli_arguments()
 
-        if is_charmcraft_running_in_managed_mode():
+        if env.is_charmcraft_running_in_managed_mode():
             work_dir = env.get_managed_environment_home_path()
         else:
             work_dir = self.buildpath
@@ -315,7 +314,7 @@ class Builder:
         """
         charms: List[str] = []
 
-        managed_mode = is_charmcraft_running_in_managed_mode()
+        managed_mode = env.is_charmcraft_running_in_managed_mode()
         if not managed_mode and not destructive_mode:
             self.provider.ensure_provider_is_available()
 
--- a/charmcraft/commands/build.py
+++ b/charmcraft/commands/build.py
@@ -23,16 +23,12 @@ import subprocess
 import zipfile
 from typing import List, Optional
 
-from charmcraft import linters, parts
+from charmcraft import env, linters, parts
 from charmcraft.bases import check_if_base_matches_host
 from charmcraft.cmdbase import BaseCommand, CommandError
 from charmcraft.config import Base, BasesConfiguration, Config
 from charmcraft.deprecations import notify_deprecation
-from charmcraft.env import (
-    get_managed_environment_home_path,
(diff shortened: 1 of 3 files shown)


cmatsuoka:
@balbirthomas Could you verify if this version addresses the build vs git clean issue you reported earlier?

balbirthomas:
> @balbirthomas Could you verify if this version addresses the build vs git clean issue you reported earlier?

Hi, Is there an edge release for this ?

charmcraft/commands/build.py
```diff
@@ -23,16 +23,12 @@
...
 from charmcraft.bases import check_if_base_matches_host
 from charmcraft.cmdbase import BaseCommand, CommandError
 from charmcraft.config import Base, BasesConfiguration, Config
 from charmcraft.deprecations import notify_deprecation
-from charmcraft.env import (
-    get_managed_environment_home_path,
-    get_managed_environment_project_path,
-    is_charmcraft_running_in_managed_mode,
-)
+from charmcraft.env import is_charmcraft_running_in_managed_mode
```
facundobatista, line 31: Why not call this one also through `env.`?

charmcraft/commands/build.py
```diff
@@ -206,6 +202,11 @@ def build_charm(self, bases_config: BasesConfiguration) -> str:
 
         self._handle_deprecated_cli_arguments()
 
+        if is_charmcraft_running_in_managed_mode():
+            work_dir = env.get_managed_environment_home_path()
```
facundobatista, line 206: Maybe the `logger.debug` call five lines above should be fixed?

facundobatista (approved):
Looks fine, added a couple of inline questions. Thanks!

charmcraft/commands/build.py
cmatsuoka: No specific reason, will update to call using `env.`.

charmcraft/commands/build.py
cmatsuoka: Indeed. Will fix to display `work_dir` instead.

cmatsuoka:
> Hi, Is there an edge release for this ?

@facundo Do we create snap artifacts for charmcraft pull requests? Otherwise the solution for @balbirthomas will be to check out the repository and build a snap locally.

facundobatista:
> > Hi, Is there an edge release for this ?
> 
> @facundo Do we create snap artifacts for charmcraft pull requests? Otherwise the solution for @balbirthomas will be to check out the repository and build a snap locally.

We do not. Every commit in `master` goes to `edge`, but nothing before it (automatically). You *could* prepare a snap for this PR and release it to an edge's branch, if needed...

sergiusens:
@facundobatista @cmatsuoka the snap is a store artifact, https://github.com/canonical/charmcraft/pull/496/checks?check_run_id=3340795540

sergiusens:
we can look into snapcore/actions-publish if we want to publish on edge iif the PR comes from this repository

sergiusens (approved):

balbirthomas:
Hi @cmatsuoka I tried building charmcraft from git repo but build failed with
```                                                                     
Pulling python3 
+ snapcraftctl pull
'charmcraft' has dependencies that need to be staged: python3
Skipping pull python3 (already ran)
Building python3 
+ snapcraftctl build
+ install -D -m 0755 /root/project/snap/local/sitecustomize.py /root/parts/python3/install/usr/lib/python3.8/sitecustomize.py
install: cannot stat '/root/project/snap/local/sitecustomize.py': No such file or directory
Failed to run 'override-build': Exit code was 1.
Run the same command again with --debug to shell into the environment if you wish to introspect this failure.
```
This could be a problem with my setup since I typically do not build snaps. I will be happy to test this any time you have an edge release. Unfortunately at the present time I can not devote more time to building from git.

cmatsuoka:
> This could be a problem with my setup since I typically do not build snaps. I will be happy to test this any time you have an edge release. Unfortunately at the present time I can not devote more time to building from git.

Thanks. We will land this PR and the resulting build should appear in edge. If you notice something wrong you can notify us before the next release so we can fix it.

PR #499, opened 2021-08-17
commands/build: separate planning from execution (CRAFT-331)

cjp256:
Determine bases that can be built first (in new `plan()` method)
then run through the process of building.

### DIFF
--- a/charmcraft/commands/build.py
+++ b/charmcraft/commands/build.py
@@ -21,7 +21,7 @@ import os
 import pathlib
 import subprocess
 import zipfile
-from typing import List, Optional
+from typing import List, Optional, Tuple
 
 from charmcraft import env, linters, parts
 from charmcraft.bases import check_if_base_matches_host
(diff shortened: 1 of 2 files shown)


charmcraft/commands/build.py
```diff
@@ -300,10 +300,10 @@ def _set_prime_filter(self):
...
                 self._prime.append(fn)
 
-    def run(
-        self, bases_indices: Optional[List[int]] = None, destructive_mode: bool = False
-    ) -> List[str]:
-        """Run build process.
+    def plan(
+        self, *, bases_indices: Optional[List[int]], destructive_mode: bool, managed_mode: bool
+    ) -> List[Tuple[BasesConfiguration, Base, int, int]]:
+        """Plan build porocess.
```
cmatsuoka, line 306: Typo here in process.

charmcraft/commands/build.py
```diff
@@ -371,11 +349,62 @@ def run(
...
             raise CommandError(
                 "No suitable 'build-on' environment found in any 'bases' configuration."
             )
 
+        charms = []
+        for bases_config, build_on, bases_index, build_on_index in build_plan:
+            logger.debug(
+                "Building for 'bases[%d][%d]'.",
+                bases_index,
+                build_on_index,
```
cmatsuoka, line 396: Maybe the arguments could be in a single line?

cmatsuoka (approved):
Looks good! (but there's a typo do fix)

charmcraft/commands/build.py
```diff
@@ -312,20 +312,7 @@ def run(
 
         :returns: List of charm files created.
```
facundobatista, line 313: This is not longer true, right?

facundobatista (approved):
Like it. Annotated a small detail.

charmcraft/commands/build.py
cjp256: done!

charmcraft/commands/build.py
cjp256: Fixed and improved the entire doc block

charmcraft/commands/build.py
cjp256: Indeed, I rewrote the entire doc string, thanks!

PR #505, opened 2021-08-23
Use specific ArgumentParsingError for argument parsing errors.

facundobatista:
This is in preparation for the integration of craft-cli handling messages. This way, the original CommandError is not overloaded just to react differently when the only intended result is to show a "command parsing help" message to the user.

Also the HelpCommand itself is now simplified to show a message when the intended result is to show the message (the command itself functioning well) and only raise an exception when there is actually an error.

### DIFF
--- a/charmcraft/main.py
+++ b/charmcraft/main.py
@@ -30,6 +30,10 @@ from charmcraft.parts import setup_parts
 logger = logging.getLogger(__name__)
 
 
+class ArgumentParsingError(Exception):
+    """Flag an argument parsing error: just show the message."""
+
+
 class HelpCommand(BaseCommand):
     """Special internal command to produce help and usage messages.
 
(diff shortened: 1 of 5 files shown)


sergiusens (approved):

charmcraft/main.py
```diff
@@ -345,6 +353,10 @@ def main(argv=None):
         setup_parts()
         dispatcher = Dispatcher(argv[1:], COMMAND_GROUPS)
         retcode = dispatcher.run()
+    except ArgumentParsingError as err:
+        print(err)
```
cjp256, line 357: I'd suggest printing to stderr instead.

charmcraft/main.py
```diff
@@ -30,6 +30,10 @@
 logger = logging.getLogger(__name__)
 
 
+class ArgumentParsingError(Exception):
+    """Flag an argument parsing error: just show the message."""
```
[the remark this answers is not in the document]
facundobatista: improved.

charmcraft/main.py
facundobatista: yes! good catch

facundobatista: improved

tests/test_help.py
```diff
@@ -353,25 +353,48 @@ def test_command_help_text_loneranger(config):
...
+def test_tool_exec_no_arguments_help():
+    """Execute charmcraft without any option at all."""
+    with patch("charmcraft.helptexts.get_full_help") as mock:
+        mock.return_value = "test help"
+        with pytest.raises(ArgumentParsingError) as cm:
+            dispatcher = Dispatcher([], COMMAND_GROUPS)
+            dispatcher.run()
+    error = cm.value
+
+    # check the given information to the builder
```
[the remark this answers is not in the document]
facundobatista: The one that builds the "help text" that will be presented later to the user; improved that comment and others in the code...

charmcraft/main.py
sergiusens: Good catch

tests/test_help.py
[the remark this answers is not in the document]
sergiusens: sorry for being vague, I was implying that the code comment could be improved :-)
Very nitpicky, feel free to ignore.

charmcraft/main.py
facundobatista: Also added a test for this, as I found it wasn't covered.

Issue #198, opened 2020-11-15
`charmcraft init` would be better off with explicit directory name


markshuttle:
The combination of --project-dir and --name in `charmcraft init` is rather awkward.

Could we make it that the directory is the only unqualified argument to init, and that --name is then an optional qualified argument along with --author?

So:

```
charmcraft init
charmcraft init directory
charmcraft init --name foo
charmcraft init directory --name foo
```

sed-i:
I also was wondering if there's a reason for `charmcraft init --project-dir` not to auto create the project dir (as in `mkdir -p` / `os.makedirs(..., exist_ok=True)`).

Currently if the dir doesn't exist,
```console
charmcraft internal error! FileNotFoundError: [Errno 2] No such file or directory: ...
```

jnsgruk:
I also find this a little awkward - to me a more sane default would be:

`charmcraft init test-charm`

Where that command would create the `test-charm` directory and populate it with the template

facundobatista:
So, let's keep `--project-dir` for the standard way of pointing to project directories, and now that we have the fix to create the directory if not there, it shouldn't be surprising anymore. Let's see if in the future a similar need/situation arises.

PR #508, opened 2021-08-27
Support --project-dir pointing to a missing directory (CRAFT-474).

facundobatista:
I just fixed this when loading the config, so at least `charmcraft` does not crash.

I considered doing this verification in `main.py`, when loading the argument. The verification there would make more sense, as it's a "command line argument" verification, and we could raise a proper ArgumentParsingError indicating that the directory does not exist.

However, we *may want* for `--project-dir` to point to a directory that is not there, for the `init` command. See issue #198, where it claims that if the directory indicated to `init` is not there, we should create it.

So, in the end, I decided that the best behaviour is to just "accept" the parameter pointing to nowhere, and load a default config with that info. Note that the flag `config_provided` is `False` in this case, so if the executed command does need a real config, it will finish in error anyway.

One way to test this behaviour change is by issuing:

    charmcraft version --project-dir /whatever

### DIFF
--- a/charmcraft/config.py
+++ b/charmcraft/config.py
@@ -276,7 +276,10 @@ class BasesConfiguration(
 class Project(ModelConfigDefaults):
     """Internal-only project configuration."""
 
-    dirpath: pydantic.DirectoryPath
+    # do not verify that `dirpath` is a valid existing directory; it's used externally as a dir
+    # to load the config itself (so we're really do the validation there), and we want to support
+    # the case of a missing directory (and still load a default config structure)
+    dirpath: pathlib.Path
     config_provided: bool = False
 
     # this timestamp will be used in several places, even sent to Charmhub: needs to be UTC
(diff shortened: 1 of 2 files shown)


cmatsuoka (approved):

sed-i:
With the latest charmcraft this is still/again broken

```sh
> $ charmcraft version
1.2+165.ge36c73a

> $ charmcraft init --project-dir doesntexistyet
charmcraft internal error! FileNotFoundError: [Errno 2] No such file or directory: '/home/tux/charms/doesntexistyet' (full execution logs in /home/tux/snap/charmcraft/common/charmcraft-log-0bhcjdhl)
```
[charmcraft.log](https://github.com/canonical/charmcraft/files/7187632/charmcraft.log)

facundobatista:
Opened [this](https://github.com/canonical/charmcraft/pull/543) to fix it. Thanks!

PR #515, opened 2021-09-01
setup: unlink target file before rename (CRAFT-495)

cjp256:
On Windows we get an error when using rename to overwrite an existing
file:

FileExistsError: [WinError 183] Cannot create a file when that file
 already exists: 'charmcraft\\version.py~' -> 'charmcraft\\version.py'

Unlink the target first if it exists.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -62,6 +62,8 @@ extras_require = {
 
 version_path = Path("charmcraft/version.py")
 version_backup = Path("charmcraft/version.py~")
+if version_backup.exists():
+    version_backup.unlink()
 version_path.rename(version_backup)
 try:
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -62,6 +62,8 @@
 
 version_path = Path("charmcraft/version.py")
 version_backup = Path("charmcraft/version.py~")
+if version_backup.exists():
+    version_backup.unlink()
```
cmatsuoka, line 66: It could be interesting to add a comment explaining why we are doing this, otherwise someone could try to remove these lines in a future refactoring.

sergiusens (approved):
looks good, add Claudio's code comment though

PR #516, opened 2021-09-01
commands/init: fix handling for platforms without pwd (CRAFT-496)

cjp256:
Future work can add similar support for other platforms. Example
code for Windows:

https://sjohannes.wordpress.com/2010/06/19/win32-python-getting-users-display-name-using-ctypes/

### DIFF
--- a/charmcraft/commands/init.py
+++ b/charmcraft/commands/init.py
@@ -79,6 +79,8 @@ def _get_author_from_user() -> str:
     """
     if pwd is not None:
         author = _get_users_full_name_gecos()
+    else:
+        author = None
 
     if not author:
         raise CommandError(
--- a/charmcraft/commands/init.py
+++ b/charmcraft/commands/init.py
@@ -93,7 +93,7 @@ def _get_author_from_user() -> str:
     """
     if sys.platform == "win32":
         author = _get_users_full_name_windows()
-    else:
+    elif pwd is not None:
         author = _get_users_full_name_gecos()
 
     if not author:
--- a/charmcraft/commands/init.py
+++ b/charmcraft/commands/init.py
@@ -16,11 +16,9 @@
 
 """Infrastructure for the 'init' command."""
 
-import ctypes
 import logging
 import os
 import re
-import sys
 from datetime import date
 from typing import Optional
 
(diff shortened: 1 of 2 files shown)


charmcraft/commands/init.py
```diff
@@ -90,8 +89,12 @@ def run(self, args):
 
         if args.author is None:
             try:
+                # Not all platforms support pwd, so we do a late import here and handle
+                # the potential import error.
+                import pwd
+
                 gecos = pwd.getpwuid(os.getuid()).pw_gecos.split(",", 1)[0]
-            except KeyError:
+            except (KeyError, ImportError):
```
facundobatista, line 97: I don't like doing the import here in the middle. 

Maybe isolating a `get_gecos` call? It will help to test this...

sergiusens (commented):
I would do something like this at the top

```python
def get_gecos_windows():
    raise CommandError("Cannot automatically determine Author on Windows.")


def get_gecos_posix():
    ....


try:
    import pwd
    get_gecos = get_gecos_posix
except ImportError:
    get_gecos = get_gecos_windows
```

cjp256:
Added support to query information from Windows

charmcraft/commands/init.py
cjp256: Done!

sergiusens (approved):

charmcraft/commands/init.py
```diff
@@ -18,13 +18,24 @@
...
 from datetime import date
+from typing import Optional
 
 from charmcraft.cmdbase import BaseCommand, CommandError
 from charmcraft.utils import make_executable, get_templates_environment
 
+try:
+    import ctypes
+except ImportError:
+    ctypes = None
```
facundobatista, line 32: What this is for? Is there any platform where `ctypes` does not exist?

facundobatista (changes requested):

charmcraft/commands/init.py
cjp256: Ah yes, I forgot. I'm only used to seeing it on Windows :D

charmcraft/commands/init.py
```diff
@@ -59,6 +60,48 @@
...
+
+    try:
+        return pwd.getpwuid(os.getuid()).pw_gecos.split(",", 1)[0]
+    except KeyError:
+        return None
+
+
+def _get_users_full_name_windows() -> str:
+    """Get user's full name on Windows."""
+    import ctypes
```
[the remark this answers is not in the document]
cjp256: I considered os.getlogin() which is the same thing, but that doesn't get the user's full name.  I would suggest keeping it as-is.  It's quite widely available code on stack overflow, etc.

If you want something using another library:
```
pip install pywin32
>>> import win32api
>>> import win32net
>>> user_info = win32net.NetUserGetInfo(None,   win32api.GetUserName(), 2)
>>> print(user_info["full_name"])
Chris Patterson
```

charmcraft/commands/init.py
```diff
@@ -18,13 +18,24 @@
...
 
+try:
+    import ctypes
+except ImportError:
+    ctypes = None
+
+try:
+    import pwd
+except ImportError:
+    pwd = None
```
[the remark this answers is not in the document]
cjp256: Added check

charmcraft/commands/init.py
cjp256: Done

facundobatista (changes requested):

charmcraft/commands/init.py
```diff
@@ -59,6 +64,30 @@
...
+        return None
+
+
+def _get_author_from_user() -> str:
+    """Get the author's name by querying system's user information.
+
+    :raises CommandError: If unable to determine author.
+    """
+    if pwd is not None:
+        author = _get_users_full_name_gecos()
```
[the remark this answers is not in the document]
cjp256: ?

The next line checks if author is empty/none and raises an error?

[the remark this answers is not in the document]
cjp256: Fixed!

facundobatista (approved):
Awesome, thanks!

PR #519, opened 2021-09-01
utils: add arch translation for Windows' AMD64 (CRAFT-499)

cjp256:


facundobatista (approved):
I like the change, but question: we will only support Windows under amd64?

sergiusens (approved):

sergiusens:
@facundobatista eventually yes, but the machine architecture translation would be normalized to amd64 in any case (similar to snaps on Fedora using amd64 even though they use a different label). I think that was the spirit of your question.

cjp256:
If the question was about the other arch translations, I couldn't find them documented anywhere so I stuck with the only one I can test :D

PR #520, opened 2021-09-01
providers/logs: close temporary file before overwriting (CRAFT-500)

cjp256:
Windows fails to use the temp file because it is in use.  Close the
open descriptor before writing to the path.

Switch to NamedTemporaryFile for simpler mocking (mkstemp requires
closing the handle via os.close()).


facundobatista:
Not sure where the problem is, couldn't make it fail:

In one Python process:

```
>>> import tempfile
>>> fd, name = tempfile.mkstemp()
>>> fd
3
>>> name
'C:\\Users\\User\\AppData\\Local\\Temp\\tmp4d63umfe'
>>>
```

Without closing that one, in other Python process:

```
>>> import os
>>> os.path.exists('C:\\Users\\User\\AppData\\Local\\Temp\\tmp4d63umfe')
True
>>> open('C:\\Users\\User\\AppData\\Local\\Temp\\tmp4d63umfe', "wt")
<_io.TextIOWrapper name='C:\\Users\\User\\AppData\\Local\\Temp\\tmp4d63umfe' mode='wt' encoding='cp1252'>
>>> _.write("foobar")
6
>>>
```

facundobatista (approved):

PR #521, opened 2021-09-01
make required files executable (CRAFT-501)

cjp256:
Windows does not support executable bits or fchmod().  Check if
system is posix before setting exe bit.

### DIFF
--- a/charmcraft/charm_builder.py
+++ b/charmcraft/charm_builder.py
@@ -188,8 +188,7 @@ class CharmBuilder:
             )
             with dispatch_path.open("wt", encoding="utf8") as fh:
                 fh.write(dispatch_content)
-                if os.name == "posix":
-                    make_executable(fh)
+                make_executable(fh)
 
         # bunch of symlinks, to support old juju: verify that any of the already included hooks
         # in the directory is not linking directly to the entrypoint, and also check all the
--- a/charmcraft/charm_builder.py
+++ b/charmcraft/charm_builder.py
@@ -188,7 +188,8 @@ class CharmBuilder:
             )
             with dispatch_path.open("wt", encoding="utf8") as fh:
                 fh.write(dispatch_content)
-                make_executable(fh)
+                if os.name == "posix":
+                    make_executable(fh)
 
         # bunch of symlinks, to support old juju: verify that any of the already included hooks
         # in the directory is not linking directly to the entrypoint, and also check all the
--- a/charmcraft/commands/init.py
+++ b/charmcraft/commands/init.py
@@ -133,7 +133,7 @@ class InitCommand(BaseCommand):
                 fh.write(out)
                 for todo in _todo_rx.findall(out):
                     todos.append((template_name, todo))
-                if template_name in executables:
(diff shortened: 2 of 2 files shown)


facundobatista (changes requested):
I don't like this change. Avoiding just to call the function leaves the caller broken.

What we need here is to ensure we're calling `make_executable` *only in a posix system*, and fail if it's not, *unless a flag is passed*, so it's explicit in the caller that is ok for it to not be really executed. 

Then, we would need to change the call in `init` command to pass that flag, but also to re-call it on the same file at building time (so when the charm is packed, that file will be in its correct state).

cjp256:
> I don't like this change. Avoiding just to call the function leaves the caller broken.
> 
> What we need here is to ensure we're calling `make_executable` _only in a posix system_, and fail if it's not, _unless a flag is passed_, so it's explicit in the caller that is ok for it to not be really executed.
> 
> Then, we would need to change the call in `init` command to pass that flag, but also to re-call it on the same file at building time (so when the charm is packed, that file will be in its correct state).

Agreed, done!

charmcraft/commands/init.py
```diff
@@ -133,7 +133,7 @@ def run(self, args):
                 fh.write(out)
                 for todo in _todo_rx.findall(out):
                     todos.append((template_name, todo))
-                if template_name in executables:
+                if template_name in executables and os.name == "posix":
```
facundobatista, line 136: It's simpler and faster to check if `posix` above when building the `executables` list, where it's also a good place to indicate what's going on with a comment, like "we will set these to be executables if posix only, as other platforms doesn't support it".

Also missing a test.

charmcraft/charm_builder.py
```diff
@@ -188,7 +188,8 @@ def handle_dispatcher(self, linked_entrypoint):
             )
             with dispatch_path.open("wt", encoding="utf8") as fh:
                 fh.write(dispatch_content)
-                make_executable(fh)
+                if os.name == "posix":
```
facundobatista, line 191: Isn't this always the case?

In any case, we can NOT do this: if we build a charm without an executable `dispatch`, it's a broken charm.

facundobatista (changes requested):

charmcraft/charm_builder.py
cjp256: In a future world of Windows support, the dispatch file would likely be an `.exe` file, we just wouldn't need to set an executable bit.

cjp256: (and in that case, os.name would be "nt")

charmcraft/commands/init.py
cjp256: The relevant tests update is actually in another PR:
https://github.com/canonical/charmcraft/pull/516/files#diff-134db82b983286583fedb40b7d946fd37d08efba163ea4b54e36ee1628f004f3R133

facundobatista (approved):
Great!

sergiusens (approved):

PR #522, opened 2021-09-01
providers: add support to install Charmcraft snap from store (CRAFT-502)

cjp256:
On Linux, the default behavior is to inject the host snap into the target
environment.

On other platforms, the Charmcraft snap is installed from the Snap Store.

When installing the snap from the Store, we check if the user specifies a
channel, using CHARMCRAFT_INSTALL_SNAP_CHANNEL=<channel>.  If unspecified,
we use the default determined by get_managed_environment_snap_channel_default().

On Linux, the user may specify this environment variable to force Charmcraft
to install the snap from the Store rather than inject the host snap.

### DIFF
--- a/charmcraft/env.py
+++ b/charmcraft/env.py
@@ -21,6 +21,7 @@ import distutils.util
 import os
 import pathlib
 import sys
+from typing import Optional
 
 from charmcraft.cmdbase import CommandError
 
@@ -38,6 +39,25 @@ def get_managed_environment_log_path():
 def get_managed_environment_project_path():
     """Path for project when running in managed environment."""
     return get_managed_environment_home_path() / "project"
+
+
+def get_managed_environment_snap_channel_default() -> str:
+    """Default channel to use when installing Charmcraft snap from Snap Store.
+
+    May someday be updated to use a track known to be compatible with the running
+    version.  For now, we default to default track's stable.
+
+    :returns: Channel string.
+    """
+    return "stable"
+
+
+def get_managed_environment_snap_channel() -> Optional[str]:
+    """User-specified channel to use when installing Charmcraft snap from Snap Store.
+
+    :returns: Channel string if specified, else None.
+    """
+    return os.getenv("CHARMCRAFT_INSTALL_SNAP_CHANNEL")
 
 
 def is_charmcraft_running_from_snap():
--- a/charmcraft/env.py
+++ b/charmcraft/env.py
@@ -41,17 +41,6 @@ def get_managed_environment_project_path():
     return get_managed_environment_home_path() / "project"
 
 
-def get_managed_environment_snap_channel_default() -> str:
-    """Channel to use when installing Charmcraft snap from Snap Store.
-
-    May someday be updated to use a track known to be compatible with the running
-    version.  For now, we default to default track's stable.
-
-    :returns: Channel string.
-    """
-    return "stable"
-
-
 def get_managed_environment_snap_channel() -> Optional[str]:
     """User-specified channel to use when installing Charmcraft snap from Snap Store.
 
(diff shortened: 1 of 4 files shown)


charmcraft/env.py
```diff
@@ -21,10 +21,30 @@
...
 
+def get_managed_environment_snap_channel_default() -> str:
+    """Default channel to use when installing Charmcraft snap from Snap Store.
+
+    May someday be updated to use a track known to be compatible with the running
+    version.  For now, we default to latest/stable.
+
+    :returns: Channel string.
+    """
+    return "latest/stable"
```
sergiusens, line 37: Ensure this is default track ready

```suggestion
    return "stable"
```

sergiusens (approved):
Assuming the test that can now pass is fixed and the default channel change takes place

charmcraft/env.py
cjp256: Good point!

cjp256: Done

charmcraft/env.py
```diff
@@ -40,6 +41,25 @@ def get_managed_environment_project_path():
...
 
+def get_managed_environment_snap_channel_default() -> str:
+    """Channel to use when installing Charmcraft snap from Snap Store.
+
+    May someday be updated to use a track known to be compatible with the running
+    version.  For now, we default to default track's stable.
+
+    :returns: Channel string.
+    """
+    return "stable"
```
facundobatista, line 52: Can't this be just a module's level "constant"?

charmcraft/env.py
```diff
@@ -40,6 +41,25 @@ def get_managed_environment_project_path():
...
+
+
+def get_managed_environment_snap_channel() -> Optional[str]:
+    """User-specified channel to use when installing Charmcraft snap from Snap Store.
+
+    :returns: Channel string if specified, else None.
+    """
+    return os.getenv("CHARMCRAFT_INSTALL_SNAP_CHANNEL")
```
facundobatista, line 60: is this new env var properly documented?

charmcraft/providers/_buildd.py
```diff
@@ -40,6 +47,45 @@ class CharmcraftBuilddBaseConfiguration(bases.BuilddBase):
...
+            snap_channel = get_managed_environment_snap_channel_default()
+
+        if snap_channel:
+            try:
+                snap_installer.install_from_store(
+                    executor=executor, snap_name="charmcraft", channel=snap_channel, classic=True
+                )
+            except snap_installer.SnapInstallationError as error:
+                raise bases.BaseConfigurationError(
+                    brief="Failed to install Charmcraft snap from store into target environment.",
```
facundobatista, line 77: Please include the channel in this error, it will be helpful to debug when the channel is mispelled in the env var.

facundobatista (changes requested):

charmcraft/env.py
cjp256: I left it as a function because it may have more complex logic in the future. For example, specifying a track that matches the host Charmcraft version to ensure compatibility, etc.

charmcraft/env.py
cjp256: I will once it merges.

charmcraft/providers/_buildd.py
cjp256: Done!

charmcraft/env.py
facundobatista: But you're *expecting* for this to be more complex? or just in case? 

If there's no more functionality *planned* to put it there, make it simpler.

charmcraft/providers/_buildd.py
facundobatista: thanks

charmcraft/env.py
cjp256: Done! (Removed get_managed_environment_snap_channel_default())

facundobatista (approved):
Thanks!

PR #523, opened 2021-09-01
logsetup: close log descriptor before unlinking (CRAFT-503)

cjp256:
Attempting to unlink the open file on Windows results in:

PermissionError: [WinError 32] The process cannot access the file because it is
being used by another process.

Save off the file handle and close it before unlinking.

### DIFF
--- a/charmcraft/logsetup.py
+++ b/charmcraft/logsetup.py
@@ -86,7 +86,9 @@ class _MessageHandler:
         if managed_mode:
             self._log_filepath = str(get_managed_environment_log_path())
         else:
-            _, self._log_filepath = tempfile.mkstemp(prefix="charmcraft-log-")
+            fd, self._log_filepath = tempfile.mkstemp(prefix="charmcraft-log-")
+            # Logger will re-open as needed.
+            os.close(fd)
 
         file_handler = logging.FileHandler(self._log_filepath, mode="w")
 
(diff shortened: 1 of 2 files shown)


charmcraft/logsetup.py
```diff
@@ -103,17 +104,25 @@ def _set_filehandler(self):
         self._file_logger.addHandler(file_handler)
         self._file_logger.debug("Starting charmcraft version %s", __version__)
 
+    def _cleanup_log(self):
+        if self._log_fd is not None:
+            os.close(self._log_fd)
+        self._log_fd = None
```
facundobatista, line 110: Why not telling the whole logging system to release the file handlers? It can be done through [shutdown](https://docs.python.org/3/library/logging.html#logging.shutdown).

facundobatista (changes requested):

charmcraft/logsetup.py
cjp256: The handle is opened with mkstemp() and we need to close it.  Anyways, isn't this all going to be replaced in the near future with craft-cli? :D

sergiusens (approved):

charmcraft/logsetup.py
facundobatista: Probably we should close the file descriptor opened by mkstemp anyway, the moment after creating the temp file, not storing to the future. In any case, even if we close the filedescriptor (after creating the temp file, or at the end), we should tell `logging` to shutdown, right? otherwise we won't be able to unlink the file anyway....

facundobatista (approved):
Awesome!

PR #524, opened 2021-09-02
tests: skip broken tests when running on Windows (CRAFT-504)

cjp256:
As a starting point for Windows testing, we disable all tests which fail
on Windows.  Over time we should update the tests to either work on
Windows, or mark them as specifically unsupported on Windows.

Tests are disabled entirely for the test_parts module, as craft-parts
does not yet support Windows.  In practice in may not be much effort
to enable it for these tests.

The rest of the tests are disabled on a per-test basis.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -13,7 +13,7 @@ jobs:
   run-tests:
     strategy:
       matrix:
-        os: [macos-10.15, ubuntu-18.04, ubuntu-20.04]
+        os: [macos-10.15, ubuntu-18.04, ubuntu-20.04, windows-2019]
         python-version: [3.8, 3.9]
 
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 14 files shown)


facundobatista (approved):
Looks great!

sergiusens (commented):
Are there places where something like `@pytest.mark.xfail(sys.platform == "win32", reason="requires implementation")` could be used instead? Skipping craft-parts dependents entirely I am fine with, just the other smaller ones

.github/workflows/tests.yaml
```diff
@@ -13,7 +13,7 @@ jobs:
   run-tests:
     strategy:
       matrix:
-        os: [macos-10.15, ubuntu-18.04, ubuntu-20.04]
```
sergiusens, line 16: I suspect Windows would need its own entry or bash installed or make the steps conditional

cjp256:
> Are there places where something like `@pytest.mark.xfail(sys.platform == "win32", reason="requires implementation")` could be used instead? Skipping craft-parts dependents entirely I am fine with, just the other smaller ones

Absolutely. I think as each module gets touched on in the future that determination should be made and the text & marker should be updated accordingly.  I believe most tests _should_ pass on Windows, even if not needed - there is just a minor amount of work that is required.

.github/workflows/tests.yaml
cjp256: Already done :D

cjp256:
@facundo can you update the repo to require the Windows tests to pass after merging?

sergiusens (approved):

PR #525, opened 2021-09-02
parts: execute lifecycle from the project dir (CRAFT-505)

cmatsuoka:
Craft-parts plugins may specify sources as relative paths, and these
are assumed to be relative to the current directory. To allow Charmcraft
to be executed from an arbitrary location on the filesystem, change to
the project directory before executing the parts lifecycle.


cjp256 (approved):
LGTM

charmcraft/parts.py
```diff
@@ -237,9 +237,11 @@ def __init__(
         all_parts: Dict[str, Any],
         *,
         work_dir: pathlib.Path,
+        project_dir: pathlib.Path,
```
sergiusens, line 240: can you document this, since it is right next to work_dir (add docstring), which I suspect answers my question in the general comment

sergiusens (approved):
Can you confirm if the expectation of running with --destructive-mode is that the parts created artifacts should be in the directory where charmcraft is run from or where the charmcraft project lives and can you confirm that this change at least does not change that?

charmcraft/parts.py
cmatsuoka: The parts processing work directory root is defined as an absolute path from the charm project directory (`charmdir/build`) if running in destructive mode, or as the user home directory if running on a managed environment.
```
        if env.is_charmcraft_running_in_managed_mode():
            work_dir = env.get_managed_environment_home_path()
        else:
            work_dir = self.buildpath
```

cmatsuoka: Added parameter information to the `PartsLifecycle` docstring.

charmcraft/parts.py
```diff
@@ -279,6 +290,7 @@ def run(self, target_step: Step) -> None:
                     self._lcm.clean(Step.BUILD, part_names=["charm"])
                     self._lcm.reload_state()
 
+            logger.debug("Executing parts lifecycle in %s", self._project_dir)
```
facundobatista, line 293: We log all paths doing repr of the str, like:
```
logger.debug("Executing parts lifecycle in %r",str( self._project_dir))
```

facundobatista (approved):
+1, but please check comment. Thanks!

charmcraft/parts.py
cmatsuoka: Fixing to keep consistency.

PR #528, opened 2021-09-03
windows: build installers for Windows (CRAFT-21)

cjp256:


facundobatista (approved):
All makes sense. Don't know if it 100% works, but we can polish details on first release.

sergiusens (approved):

PR #530, opened 2021-09-07
Implement the 'close' command to close channels in Charmhub (CRAFT-493).

### LINKED ISSUE #232: Implement `charmcraft close`
There needs to be a way to close channels, bringing over syntax from snapcraft, the following syntax should be possible:

    charmcraft close <name> <channel>...

facundobatista:
Fixes #232.


charmcraft/commands/store/__init__.py
```diff
@@ -560,6 +560,43 @@ def run(self, parsed_args):
...
+
+        The channel is made up of `track/risk/branch` with both the track and
+        the branch as optional items, so formally:
+
+          [track/]risk[/branch]
+
+        Channel risk must be one of stable, candidate, beta or edge. The
+        track defaults to `latest` and branch has no default.
+
+        Closing a channel will take you through login if needed.
```
sergiusens, line 580: we will need to update this if we go down the agreed path of saying "Not logged in" if the user has never logged in before.

charmcraft/commands/store/__init__.py
```diff
@@ -560,6 +560,43 @@ def run(self, parsed_args):
...
+
+    def fill_parser(self, parser):
+        """Add own parameters to the general parser."""
+        parser.add_argument("name", help="The name of charm or bundle")
+        parser.add_argument("channel", help="The channel to close")
+
+    def run(self, parsed_args):
+        """Run the command."""
+        store = Store(self.config.charmhub)
+        revision = None  # revision in None will actually close the channel
```
sergiusens, line 593:
```suggestion
        revision = None  # revision None will actually close the channel
```

charmcraft/commands/store/__init__.py
```diff
@@ -560,6 +560,43 @@ def run(self, parsed_args):
...
+        channels = [parsed_args.channel]  # the API accepts multiple channels, we have only one
+        resources = []  # no really used when closing channels
```
sergiusens, line 595:
```suggestion
        resources = []  # not really used when closing channels
```

sergiusens (approved):
Looks good, just a few literal nits

charmcraft/commands/store/__init__.py
facundobatista: IIUC we should ask automatically for login in all operations except the auth-related ones (like `whoami`).

charmcraft/commands/store/__init__.py
facundobatista: yes, thanks!

cmatsuoka (approved):
LGTM.

PR #531, opened 2021-09-07
Support the case of tracks without releases in 'status' command (CRAFT-507).

facundobatista:


cmatsuoka (approved):

sergiusens (approved):
Looks good. I am a bit itchty on displaying the base for the first line only, but that is unrelated

PR #533, opened 2021-09-07
Improved the list-lib command's help (CRAFT-508).

### LINKED ISSUE #377: `list-lib` should show you what command to run to fetch a library
Currently if you're looking for a library for a specific charm using charmcraft (e.g. `charmcraft list-lib nginx-ingress-integrator`) you get output like this:
```
Library name    API    Patch
ingress         0      9
```
To actually fetch that library you have to run `charmcraft fetch-lib charms.nginx_ingress_integrator.v0.ingress`. This isn't obvious from the `list-lib` command.

facundobatista:
Added an example of how the output would look like, and how to build a fetch-lib command from that. Fixes #377.

### DIFF
--- a/charmcraft/commands/store/__init__.py
+++ b/charmcraft/commands/store/__init__.py
@@ -1106,10 +1106,10 @@ class FetchLibCommand(BaseCommand):
         The first time a library is downloaded the command will create the needed
         directories to place it, subsequent fetches will just update the local copy.
 
-        You can specify the library to update/download building its fully qualified
-        name with the charm and library names, and the desired API version. For
-        example, to fetch the API version 3 of library 'somelib' from charm
-        `specialcharm`, do:
+        You can specify the library to update or download by building its fully
+        qualified name with the charm and library names, and the desired API
+        version. For example, to fetch the API version 3 of library 'somelib'
+        from charm `specialcharm`, do:
 
         $ charmcraft fetch-lib charms.specialcharm.v3.somelib
         Library charms.specialcharm.v3.somelib version 3.7 downloaded.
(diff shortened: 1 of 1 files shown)


charmcraft/commands/store/__init__.py
```diff
@@ -1211,6 +1211,20 @@ class ListLibCommand(BaseCommand):
...
+        Library name    API    Patch
+        my_great_lib    0      3
+        my_great_lib    1      0
+        other_lib       0      5
+
+        To fetch one of the shown libraries you can use the fetch-lib command, indicating
+        the library and which API version to get:
+
+        $ charmcraft fetch-lib charms.my-charm.v0.my_great_lib
+        Library charms.my-charm.v0.my_great_lib version 1.0 downloaded.
```
sergiusens, line 1227: I would constrain the help on to use fetch-lib to its own help with that same text

```suggestion
        To fetch one of the shown libraries you can use the fetch-lib command.
```

facundobatista: Trimmed this help and improved the fetch-lib one.

charmcraft/commands/store/__init__.py
```diff
@@ -1105,6 +1105,17 @@ class FetchLibCommand(BaseCommand):
 
         The first time a library is downloaded the command will create the needed
         directories to place it, subsequent fetches will just update the local copy.
+
+        You can specify the library to update/download building its fully qualified
+        name with the charm and library names, and the desired API version. For
```
cmatsuoka, line 1110: This was a bit hard to parse the first time I read it. Maybe change to "...library to update or download by building its fully qualified name..."?

cmatsuoka (approved):

sergiusens (approved):

charmcraft/commands/store/__init__.py
facundobatista: thanks! improved

PR #537, opened 2021-09-14
First step in isolating the help texts module from the app itself (CRAFT-530).

facundobatista:
This includes:

- not having the general summary or app name anymore hardcoded in the module

- moved the functions at module level to be methods inside a class; this is for the system to be able to be "initiated" with the app specific information. 

- those moved in functions are exactly the same, with some minor structural changes:

    - `get_full_help` and `get_detailed_help` do not receive the "command groups" 

    - `get_usage_message` now formats `USAGE` passing also the `appname`

    - `get_full_help` and `get_detailed_help` indents the "general summary" (as the app text doesn't need to be passed indented)

**Note** on initiating `Helper`: currently there's no guard on using it before calling its `init`, as I'm not 100% sure the form this will have when both this help texts module and the dispatcher will be in the craft-cli library, but for sure the final form will be more robust.

**Note 2**: the PR is too big, but most of it is just moving code around, and zillion changes in the texts because of two functions not receiving command groups anymore, and the need to init the helper before usage.

### DIFF
--- a/charmcraft/helptexts.py
+++ b/charmcraft/helptexts.py
@@ -22,41 +22,18 @@ from operator import attrgetter
 # max columns used in the terminal
 TERMINAL_WIDTH = 72
 
-# the summary of the whole program, already indented so it represents the real
-# "columns spread", for easier human editing.
-GENERAL_SUMMARY = """
-    Charmcraft helps build, package and publish operators on Charmhub.
-
-    Together with the Python Operator Framework, charmcraft simplifies
-    operator development and collaboration.
-
-    See https://charmhub.io/publishing for more information.
-"""
-# XXX Facundo 2020-09-10: we should add an extra (separated) line to the summary with:
-#   See <url> for additional documentation.
-# Related issue: https://github.com/canonical/charmcraft/issues/161
-
 # generic intro and outro texts
 HEADER = """
 Usage:
-    charmcraft [help] <command>
+    {appname} [help] <command>
 """
 
 USAGE = """\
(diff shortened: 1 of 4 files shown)


charmcraft/helptexts.py
```diff
@@ -82,207 +59,223 @@ def _build_item(title, text, title_space):
...
-    """Produce the text for the default help.
-
-    - command_groups: list of grouped commands, as it's defined in the main
-      module
-
-    - global_options: options defined at charmcraft level (not in the commands),
-      with the (options, description) structure
-
-    The help text has the following structure:
+class Helper:
```
sergiusens, line 62: I would prefer just
```suggestion
class Help:
```

or HelpBuilder, HelpConstructor

charmcraft/helptexts.py
```diff
@@ -82,207 +59,223 @@ def _build_item(title, text, title_space):
...
+    """Produce the different helping texts."""
```
sergiusens, line 63:
```suggestion
    """Produce the different help texts."""
```

cmatsuoka (approved):
Diff is a bit confusing due to changes in indents, but the overall idea seems sane to me.

charmcraft/helptexts.py
facundobatista: HelpBuilder it is

charmcraft/helptexts.py
facundobatista: fixed, thanks

sergiusens (approved):

PR #539, opened 2021-09-15
env: enable darwin (CRAFT-535)

sergiusens:
It has the same mechanics as Windows but with much more similarities
at the platform level to Linux.


cmatsuoka (approved):

facundobatista (approved):
Nice!

PR #540, opened 2021-09-16
More isolation of the helptext module and simplification of its usage (CRAFT-528)

facundobatista:
In detail:

- Stop passing (again) the application name when calling `get_usage_message`

- Moved two tests from "help suite" to "cmdbase suite" as they were more about the command's content than the help infrastructure

- Fixed/improved tests so they don't depend anymore on `COMMAND_GROUPS` nor `VersionCommand`, which are app-specific

### DIFF
--- a/charmcraft/helptexts.py
+++ b/charmcraft/helptexts.py
@@ -73,12 +73,21 @@ class HelpBuilder:
         self.general_summary = general_summary
         self.command_groups = command_groups
 
-    def get_usage_message(self, fullcommand, error_message):
+    def get_usage_message(self, extra_command, error_message):
         """Build a usage and error message.
 
-        The fullcommand is the command used by the user (`charmcraft`, `charmcraft build`, etc),
-        and the error message is the specific problem in the given parameters.
+        The extra_command is the extra string used after the application name to build the
+        full command that will be shown in the usage message; for example, having an
+        application name of "someapp":
+        - if extra_command is "" it will be shown "Try 'appname -h' for help".
+        - if extra_command is "version" it will be shown "Try 'appname version -h' for help"
+
+        The error message is the specific problem in the given parameters.
         """
+        if extra_command:
+            fullcommand = f"{self.appname} {extra_command}"
+        else:
+            fullcommand = self.appname
         return USAGE.format(
             appname=self.appname, fullcommand=fullcommand, error_message=error_message
         )
--- a/charmcraft/helptexts.py
+++ b/charmcraft/helptexts.py
@@ -30,7 +30,7 @@ Usage:
 
 USAGE = """\
 Usage: {appname} [options] command [args]...
-Try '{fullcommand} -h' for help.
+Try '{full_command} -h' for help.
 
 Error: {error_message}
 """
@@ -73,23 +73,23 @@ class HelpBuilder:
         self.general_summary = general_summary
         self.command_groups = command_groups
 
-    def get_usage_message(self, extra_command, error_message):
+    def get_usage_message(self, error_message, command=""):
         """Build a usage and error message.
 
-        The extra_command is the extra string used after the application name to build the
+        The command is the extra string used after the application name to build the
         full command that will be shown in the usage message; for example, having an
         application name of "someapp":
-        - if extra_command is "" it will be shown "Try 'appname -h' for help".
-        - if extra_command is "version" it will be shown "Try 'appname version -h' for help"
+        - if command is "" it will be shown "Try 'appname -h' for help".
+        - if command is "version" it will be shown "Try 'appname version -h' for help"
 
         The error message is the specific problem in the given parameters.
         """
-        if extra_command:
-            fullcommand = f"{self.appname} {extra_command}"
+        if command:
+            full_command = f"{self.appname} {command}"
         else:
-            fullcommand = self.appname
+            full_command = self.appname
         return USAGE.format(
-            appname=self.appname, fullcommand=fullcommand, error_message=error_message
+            appname=self.appname, full_command=full_command, error_message=error_message
         )
 
     def get_full_help(self, global_options):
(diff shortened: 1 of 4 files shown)


charmcraft/helptexts.py
```diff
@@ -73,12 +73,21 @@ def init(self, appname, general_summary, command_groups):
         self.general_summary = general_summary
         self.command_groups = command_groups
 
-    def get_usage_message(self, fullcommand, error_message):
+    def get_usage_message(self, extra_command, error_message):
```
cmatsuoka, line 76: Would `subcommand` be a better name in this case?

charmcraft/helptexts.py
```diff
@@ -73,12 +73,21 @@ def init(self, appname, general_summary, command_groups):
...
+        The extra_command is the extra string used after the application name to build the
+        full command that will be shown in the usage message; for example, having an
+        application name of "someapp":
+        - if extra_command is "" it will be shown "Try 'appname -h' for help".
+        - if extra_command is "version" it will be shown "Try 'appname version -h' for help"
+
+        The error message is the specific problem in the given parameters.
         """
+        if extra_command:
+            fullcommand = f"{self.appname} {extra_command}"
```
cmatsuoka, line 88: Maybe `full_command` to be consistent with `extra_command`, `error_message`, etc?

charmcraft/main.py
```diff
@@ -86,7 +86,7 @@ def run(self, parsed_args, all_commands):
         if parsed_args.command_to_help not in all_commands:
             # asked help on a command that doesn't exist
             msg = "no such command {!r}".format(parsed_args.command_to_help)
-            help_text = help_builder.get_usage_message("charmcraft", msg)
+            help_text = help_builder.get_usage_message(extra_command="", error_message=msg)
```
cmatsuoka, line 89: `extra_command` could be an optional argument defaulting to "", unless you want to be very explicit about it.

cmatsuoka (approved):
Overall implementation seems sane, added suggestions for variable renaming.

tests/test_cmdbase.py
```diff
@@ -87,3 +87,33 @@ class TestClass(BaseCommand):
...
     with pytest.raises(NotImplementedError):
         tc.run([])
+
+
+# -- tests for strings in commands
+
+
+@pytest.mark.parametrize("command", all_commands)
+def test_aesthetic_help_msg(command):
+    """All the real commands help msg start with uppercase and doesn't end with a dot."""
```
sergiusens, line 97: This looks more like an integration test.

```suggestion
    """All real command help msgs start with uppercase and do not end with a dot."""
```

tests/test_cmdbase.py
```diff
@@ -87,3 +87,33 @@ class TestClass(BaseCommand):
...
+    msg = command.help_msg
+    assert msg[0].isupper() and msg[-1] != "."
+
+
+@pytest.mark.parametrize("command", all_commands)
+def test_aesthetic_args_options_msg(command, config):
+    """All the real commands args help messages start with uppercase and dont' end with a dot."""
```
sergiusens, line 104: This looks more like an integration test.

```suggestion
    """All real command args help messages start with uppercase and do not end with a dot."""
```

tests/test_help.py
```diff
@@ -16,65 +16,50 @@
...
     return help_builder
 
 
-@pytest.mark.parametrize("command", all_commands)
-def test_aesthetic_help_msg(command):
-    """All the real commands help msg start with uppercase and doesn't end with a dot."""
-    msg = command.help_msg
-    assert msg[0].isupper() and msg[-1] != "."
-
+# -- bulding of the "usage" help
```
sergiusens, line 35:
```suggestion
# -- building "usage" help
```

tests/test_help.py
```diff
@@ -16,65 +16,50 @@
...
-
-        def add_argument(self, *args, **kwargs):
-            """Verify that all commands have a correctly formatted help."""
-            help_msg = kwargs.get("help")
-            assert help_msg, "The help message must be present in each option"
-            assert help_msg[0].isupper() and help_msg[-1] != "."
+def test_get_usage_message_with_command(help_builder):
+    """Check the general "usage" text passing a command."""
+    help_builder.init("testapp", "general summary", [])
+    text = help_builder.get_usage_message("build", "bad parameter for the build")
```
sergiusens, line 41: I know this is just a test, but just in case :-)

```suggestion
    text = help_builder.get_usage_message("build", "bad parameter for build")
```

tests/test_help.py
```diff
@@ -16,65 +16,50 @@
...
+    expected = textwrap.dedent(
+        """\
+        Usage: testapp [options] command [args]...
+        Try 'testapp build -h' for help.
 
-    command("group", config).fill_parser(FakeParser())
+        Error: bad parameter for the build
```
sergiusens, line 47:
```suggestion
        Error: bad parameter for build
```

tests/test_help.py
```diff
@@ -16,65 +16,50 @@
...
+    """
+    )
+    assert text == expected
 
 
-def test_get_usage_message(help_builder):
-    """Check the general "usage" text."""
+def test_get_usage_message_no_command(help_builder):
+    """Check the general "usage" text not passing a command."""
```
sergiusens, line 54:
```suggestion
    """Check the general "usage" text when not passing a command."""
```

sergiusens (approved):
Looks good, some string comments

charmcraft/helptexts.py
facundobatista: It could be just `command`, as in reality currently charmcraft just passes the command (e.g. `pack`) or nothing, when the usage help requested is "general".

I could probably make it `""` by default so it's optional.

What do you think?

charmcraft/main.py
facundobatista: Ah, yes, I just talked about this in the other comment. I'll do it.

charmcraft/helptexts.py
facundobatista: I went forward with this.

tests/test_help.py
facundobatista: yes

PR #541, opened 2021-09-16
charm_builder: build dependencies locally (CRAFT-537)

cmatsuoka:
Comply with the requirement of not installing binary dependencies,
building them locally instead. The process may take a long time to
complete, so use the venv module to install requirements to a persistent
staging virtualenv so that dependencies already installed won't be
reinstalled (and rebuilt) when the charm is repacked.

Note: This solution installs additional dependencies to venv
(setup-tools, pip, and pkg_resources).
    
### DIFF
--- a/charmcraft/charm_builder.py
+++ b/charmcraft/charm_builder.py
@@ -22,7 +22,9 @@ import logging
 import os
 import pathlib
 import shutil
+import site
 import subprocess
+import venv
 from typing import List
 
 from charmcraft.cmdbase import CommandError
@@ -33,6 +35,7 @@ from charmcraft.utils import make_executable
 # Some constants that are used through the code.
 WORK_DIRNAME = "work_dir"
 VENV_DIRNAME = "venv"
+STAGING_VENV_DIRNAME = "staging-venv"
 
 # The file name and template for the dispatch script
 DISPATCH_FILENAME = "dispatch"
@@ -134,7 +137,10 @@ class CharmBuilder:
                 rel_path = rel_basedir / name
                 abs_path = abs_basedir / name
 
-                if self.ignore_rules.match(str(rel_path), is_dir=True):
+                if name == STAGING_VENV_DIRNAME:
+                    logger.debug("Ignoring staging venv directory: %r", STAGING_VENV_DIRNAME)
+                    ignored.append(pos)
+                elif self.ignore_rules.match(str(rel_path), is_dir=True):
                     logger.debug("Ignoring directory because of rules: %r", str(rel_path))
                     ignored.append(pos)
                 elif abs_path.is_symlink():
@@ -224,20 +230,23 @@ class CharmBuilder:
 
         # virtualenv with other dependencies (if any)
         if self.requirement_paths:
-            _process_run(["pip3", "--version"])
+            staging_venv_dir = self.charmdir / STAGING_VENV_DIRNAME
+            envbuilder = venv.EnvBuilder(symlinks=True, with_pip=True)
+            envbuilder.create(staging_venv_dir)
+            pip_cmd = str(staging_venv_dir / "bin" / "pip3")
 
-            venvpath = self.buildpath / VENV_DIRNAME
-            cmd = [
-                "pip3",
-                "install",  # base command
-                "--target={}".format(venvpath),  # put all the resulting files in that specific dir
-            ]
+            _process_run([pip_cmd, "--version"])
+
+            cmd = [pip_cmd, "install", "--no-binary", ":all:"]  # base command
             if _pip_needs_system():
                 logger.debug("adding --system to work around pip3 defaulting to --user")
                 cmd.append("--system")
             for reqspath in self.requirement_paths:
                 cmd.append("--requirement={}".format(reqspath))  # the dependencies file(s)
             _process_run(cmd)
+
+            # The charm builder is executed with PYTHONUSERBASE set to the staging venv dir
+            shutil.copytree(site.USER_SITE, self.buildpath / VENV_DIRNAME)
 
 
 def _pip_needs_system():
--- a/charmcraft/charm_builder.py
+++ b/charmcraft/charm_builder.py
@@ -22,7 +22,6 @@ import logging
 import os
 import pathlib
 import shutil
-import site
 import sys
 import subprocess
 from typing import List
@@ -229,7 +228,7 @@ class CharmBuilder:
         # virtualenv with other dependencies (if any)
         if self.requirement_paths:
             staging_venv_dir = self.charmdir / STAGING_VENV_DIRNAME
-            _process_run(["python3", "-m", "venv", str(staging_venv_dir)])
+            _process_run([sys.executable, "-m", "venv", str(staging_venv_dir)])
             pip_cmd = str(_find_venv_bin(staging_venv_dir, "pip3"))
 
             _process_run([pip_cmd, "--version"])
@@ -239,11 +238,10 @@ class CharmBuilder:
                 cmd.append("--requirement={}".format(reqspath))  # the dependencies file(s)
             _process_run(cmd)
 
-            # The charm builder is executed with PYTHONUSERBASE set to the staging venv dir
-            # so that site.USER_SITE points to the inner directory inside STAGING_VENV_DIRNAME
-            # where the packages are actually installed. This is the directory that dispatch
-            # expects as /venv.
-            shutil.copytree(site.USER_SITE, self.buildpath / VENV_DIRNAME)
+            # copy the virtualvenv site-packages directory to /venv in charm
+            basedir = pathlib.Path(STAGING_VENV_DIRNAME)
+            site_packages_dir = _find_venv_site_packages(basedir)
+            shutil.copytree(site_packages_dir, self.buildpath / VENV_DIRNAME)
 
 
 def _find_venv_bin(basedir, exec_base):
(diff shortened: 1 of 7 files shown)


charmcraft/charm_builder.py
```diff
@@ -224,21 +230,24 @@ def handle_dependencies(self):
...
+            envbuilder = venv.EnvBuilder(symlinks=True, with_pip=True)
+            envbuilder.create(staging_venv_dir)
+            pip_cmd = str(staging_venv_dir / "bin" / "pip3")
+
+            _process_run([pip_cmd, "--version"])
+
+            cmd = [pip_cmd, "install", "--no-binary", ":all:"]  # base command
             if _pip_needs_system():
                 logger.debug("adding --system to work around pip3 defaulting to --user")
                 cmd.append("--system")
```
cmatsuoka, line 243: Can this be removed now that we're installing to a virtualenv?

charmcraft/charm_builder.py
```diff
@@ -224,21 +229,23 @@ def handle_dependencies(self):
...
-                "install",  # base command
-                "--target={}".format(venvpath),  # put all the resulting files in that specific dir
-            ]
+            staging_venv_dir = self.charmdir / STAGING_VENV_DIRNAME
+            _process_run(["python3", "-m", "venv", str(staging_venv_dir)])
+            pip_cmd = str(staging_venv_dir / "bin" / "pip3")
+
+            _process_run([pip_cmd, "--version"])
+
+            cmd = [pip_cmd, "install", "--no-binary", ":all:"]  # base command
```
facundobatista, line 238: If requirements changed, how are we reflecting those changes in the staging venv dir?

facundobatista (changes requested):
Nice!! I added some comments inside, thanks!

charmcraft/charm_builder.py
```diff
@@ -224,21 +229,23 @@ def handle_dependencies(self):
...
-            _process_run(["pip3", "--version"])
-
-            venvpath = self.buildpath / VENV_DIRNAME
-            cmd = [
-                "pip3",
-                "install",  # base command
-                "--target={}".format(venvpath),  # put all the resulting files in that specific dir
-            ]
+            staging_venv_dir = self.charmdir / STAGING_VENV_DIRNAME
+            _process_run(["python3", "-m", "venv", str(staging_venv_dir)])
```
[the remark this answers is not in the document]
cmatsuoka: It was originally implemented using `venv.create(symlinks=True, with_pip=True)`, however for some reason ensurepip fails silently when invoked from the current interpreter and the resulting virtualenv has no `bin/pip` or `bin/pip3`. Calling as an external process fixes this issue.

Filenames with spaces should be handled by Python's `Popen` because the argument is passed as a single `argv` member instead of being split by the shell into multiple arguments (unlike `popen(3)` that takes a command string and invokes a shell). On Windows `Popen` calls `CreateProcess()` which takes a single command string, but [according to the documentation](https://docs.python.org/3/library/subprocess.html#subprocess.Popen) it should be converted taking whitespaces and other special characters into consideration (but please verify the documentation to confirm that my interpretation is correct):

> On Windows, if args is a sequence, it will be converted to a string in a manner described in [Converting an argument sequence to a string on Windows](https://docs.python.org/3/library/subprocess.html#converting-argument-sequence). This is because the underlying CreateProcess() operates on strings.

This simple test works as expected on POSIX:
```
>>> import subprocess
>>> x=subprocess.Popen(["ls", "/home/claudio/test with spaces"])
>>> foo
```

charmcraft/charm_builder.py
```diff
@@ -224,21 +229,23 @@ def handle_dependencies(self):
...
+            pip_cmd = str(staging_venv_dir / "bin" / "pip3")
+
+            _process_run([pip_cmd, "--version"])
+
+            cmd = [pip_cmd, "install", "--no-binary", ":all:"]  # base command
             if _pip_needs_system():
                 logger.debug("adding --system to work around pip3 defaulting to --user")
```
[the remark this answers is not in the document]
cmatsuoka: I was tempted to remove this, but I waited to confirm it first. My understanding is that this verification was added to work around the stock Debian pip that defaults to `--user`, but since we're using the venv pip it seems that this is not needed anymore. I'll drop it.

charmcraft/charm_builder.py
```diff
@@ -224,21 +229,23 @@ def handle_dependencies(self):
...
                 cmd.append("--system")
             for reqspath in self.requirement_paths:
                 cmd.append("--requirement={}".format(reqspath))  # the dependencies file(s)
             _process_run(cmd)
 
+            # The charm builder is executed with PYTHONUSERBASE set to the staging venv dir
+            shutil.copytree(site.USER_SITE, self.buildpath / VENV_DIRNAME)
```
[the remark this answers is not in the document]
cmatsuoka: `site.USER_SITE` resolves to `staging-venv/lib/python3.8/site-packages`, which is where the installed packages are located. It corresponds to what dispatch expects to find in `venv`. I'll add a comment to the code to clarify this.

charmcraft/charm_builder.py
cmatsuoka: If requirements are added, only the new requirements are installed and old requirements are not reinstalled. If requirements are removed, they'll remain installed on venv and a rebuild is needed to clean them (this is similar to what happens when source files are removed and sources need to be re-pulled).

charmcraft/charm_builder.py
```diff
@@ -134,7 +136,10 @@ def handle_generic_paths(self):
                 rel_path = rel_basedir / name
                 abs_path = abs_basedir / name
 
-                if self.ignore_rules.match(str(rel_path), is_dir=True):
+                if name == STAGING_VENV_DIRNAME:
```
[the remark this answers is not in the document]
cmatsuoka: Extended `ignore_rules` and removed special handling for this directory:
```
DEBUG    :: Translated .jujuignore 1 "/staging-venv" => "/staging\-venv\Z"
...
DEBUG    :: Building charm in '/home/claudio/charmcraft/charmcraft/testdir/build/parts/charm/install'
DEBUG    :: Linking in generic paths
DEBUG    :: Ignoring directory because of rules: 'staging-venv'
```

charmcraft/charm_builder.py
```diff
@@ -224,21 +229,23 @@ def handle_dependencies(self):
...
-
-            venvpath = self.buildpath / VENV_DIRNAME
-            cmd = [
-                "pip3",
-                "install",  # base command
-                "--target={}".format(venvpath),  # put all the resulting files in that specific dir
-            ]
+            staging_venv_dir = self.charmdir / STAGING_VENV_DIRNAME
+            _process_run(["python3", "-m", "venv", str(staging_venv_dir)])
+            pip_cmd = str(staging_venv_dir / "bin" / "pip3")
```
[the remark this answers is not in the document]
cmatsuoka: Using a modified version of `find_venv_dir` that does the appropriate processing based on platform instead of verifying if the directory exists will probably suit us better and also be more test-friendly since the venv creation process is mocked up. I'll prepare a patch.

charmcraft/charm_builder.py
sergiusens: This is the main benefit of using a venv instead of --target, only changes get reflected, that said, we need an integration test for this and we probably want to add -U

charmcraft/charm_builder.py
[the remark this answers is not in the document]
sergiusens: one of the points of confusing I keep raising is that the "venv" dir for a charm is not so much a venv that you would expect from traditional python. I would consider a future rename to python-site-packages or python-packages.

charmcraft/parts.py
```diff
@@ -99,7 +101,11 @@ def get_build_commands(self) -> List[str]:
         """Return a list of commands to run during the build step."""
         options = cast(CharmPluginProperties, self._options)
 
-        build_env = dict(LANG="C.UTF-8", LC_ALL="C.UTF-8")
+        build_env = dict(
+            LANG="C.UTF-8",
+            LC_ALL="C.UTF-8",
+            PYTHONUSERBASE=self._part_info.part_build_dir / charm_builder.STAGING_VENV_DIRNAME,
```
sergiusens, line 107: I don't see why this is needed considering that this is determined by running the interpreter in the venv and everything else is determined from that symlink

sergiusens (commented):
Looks good, but I have questions

charmcraft/parts.py
cmatsuoka: It's only being used is to recover the `lib/pythonX.Y/site-packages` directory name, so I think we can drop it and use `site.USER_SITE` relative to `site.USER_BASE`?

charmcraft/charm_builder.py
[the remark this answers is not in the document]
cmatsuoka: This was changed to find the venv site-packages based on version instead of using `site.USER_SITE`.

charmcraft/parts.py
cmatsuoka: Dropped `PYTHONUSERBASE` and `site.USER_SITE`, now obtaining the path to `site-packages` based on version information.

charmcraft/charm_builder.py
cmatsuoka: Adding `--upgrade` to the list of pip command line options.

cmatsuoka: Regarding tests: we have a unit test that ensures pip is called with the expected arguments, and tests that verify if ops is present in the resulting charm package's venv (the latter runs during the snap smoke tests in both managed and destructive modes). Installing only the changes is part of pip's behavior, we could design a test to verify that but then we would be testing pip instead of our code.

charmcraft/charm_builder.py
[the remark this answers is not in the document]
facundobatista: Ok!

charmcraft/charm_builder.py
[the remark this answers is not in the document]
facundobatista: ack

charmcraft/charm_builder.py
facundobatista: Great!

facundobatista (approved):
Great, thanks!

sergiusens (approved):

PR #543, opened 2021-09-20
Support a missing directory in the init command (CRAFT-541).

facundobatista:
Also removed a bunch of "series" being passed that had no effect.

### DIFF
--- a/tests/commands/test_init.py
+++ b/tests/commands/test_init.py
@@ -14,6 +14,7 @@
 #
 # For further info, check https://github.com/canonical/charmcraft
 
+import datetime
 import os
 import subprocess
 import sys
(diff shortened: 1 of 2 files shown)


tests/commands/test_init.py
```diff
@@ -165,4 +167,22 @@ def test_gecos_missing_user_information(config):
         mock_pwd.side_effect = KeyError("no user")
         msg = "Unable to automatically determine author's name, specify it with --author"
         with pytest.raises(CommandError, match=msg):
-            cmd.run(Namespace(name="my-charm", author=None, series="k8s", force=False))
+            cmd.run(Namespace(name="my-charm", author=None, force=False))
+
+
+def test_missing_directory(tmp_path, config):
+    """If the indicated directory does not exist, create it."""
+    init_dir = tmp_path / "foo" / "bar"
```
sed-i, line 175: Just in case "foo/bar" exist from another test,

```suggestion
    init_dir = tempfile.mkdtemp(dir=tmp_path)
```

facundobatista: No, `tmp_path` is guaranteed to be an empty new directory on each test.

sergiusens (approved):
Looks good, the pr/issue's intent might have clearer if the message was "create missing parent directories on init" or something like that :-)

cmatsuoka (approved):

PR #544, opened 2021-09-20
tests: warn if LXD cannot be refresh in 20.04's GHA

sergiusens:
LXD is preinstalled on the 20.04 images, it can still work without
refreshing to the latest.


cmatsuoka (approved):

facundobatista (approved):
Thanks!

PR #545, opened 2021-09-20
For bundles, do not show a deprecation message if 'bases' not included, and forbid its usage.

### LINKED ISSUE #479: The control of `bases` in `charmcraft.yaml` should be different for bundles
Currently it complains that there's no `bases`:

```
$ charmcraft revisions facutest-bundle-1
DEPRECATED: Bases configuration is now required.
See https://discourse.charmhub.io/t/4652#heading--dn03 for more information.
Revision    Version    Created at    Status
1           1          2021-02-03    released
```

But bundles do NOT have bases. So this check should be removed. And ANOTHER check should be implemented, to actually error out if the bundle's config has `bases` included (to avoid confusions).

Thanks!

facundobatista:
Fixes #479.

### DIFF
--- a/charmcraft/config.py
+++ b/charmcraft/config.py
@@ -343,6 +343,17 @@ class Config(ModelConfigDefaults, validate_all=False):
         validate_part(item)
         return item
 
+    @pydantic.validator("bases", pre=True)
+    def validate_bases_presence(cls, bases, values):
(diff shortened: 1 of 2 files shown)


sergiusens (approved):

cmatsuoka (approved):
Looks good.

PR #550, opened 2021-09-22
Transformed the help management to be implemented not using a command (CRAFT-542).

facundobatista:
This has several benefits:

- The help handling is now more straightforward

- The COMMAND_GROUPS constant does not include anything related to help anymore (more app isolation ot a future lib)

- We're able to provide better messages when help was "requested wrong" (e.g. with too many parameters or an incorrect command); IMO this is very important, as we're able to guide the user better in a situation when they're a little lost (there is a reason they are asking for help...)

- All help handling happens on the "initial args pre-parsing", and not one part there and one part when the command is run (this is transparent for the user, but useful for when Dispatcher will be in other lib)

We have only one minor drawbacks: 'help' itself, as a command, is not listed in the "basic commands group" anymore.

Note I added a ProvideHelpException exception to interrupt the normal flow of execution and finish the process providing a guiding text when **help was requested correctly**. This was a simple way to achieve this branch targets without moving a lot of code around, it may change in the future to be more simple when we isolate the Dispatcher further.

### DIFF
--- a/charmcraft/main.py
+++ b/charmcraft/main.py
@@ -22,7 +22,7 @@ import sys
 from collections import namedtuple
 
 from charmcraft import config, env
-from charmcraft.cmdbase import BaseCommand, CommandError
+from charmcraft.cmdbase import CommandError
 from charmcraft.commands import build, clean, init, pack, store, version, analyze
 from charmcraft.helptexts import help_builder
 from charmcraft.logsetup import message_handler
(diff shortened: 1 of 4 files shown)


charmcraft/main.py
```diff
@@ -315,19 +299,26 @@ def _pre_parse_args(self, sysargs):
...
+                help_text = self._get_requested_help(cmd_args)
+                raise ProvideHelpException(help_text)
+
             if command not in self.commands:
                 msg = "no such command {!r}".format(command)
                 help_text = help_builder.get_usage_message(msg)
                 raise ArgumentParsingError(help_text)
         else:
-            # no command!
+            # no command indicated!
```
sergiusens, line 321:
```suggestion
            # no command passed!
```

tests/test_help.py
```diff
@@ -390,36 +386,79 @@ def test_tool_exec_full_help(sysargv, caplog):
...
+    help_builder.init("testapp", "general summary", [])
+    with pytest.raises(ArgumentParsingError) as cm:
+        Dispatcher(sysargv, [])
+
+    expected = textwrap.dedent(
+        """\
+        Usage: testapp [options] command [args]...
+        Try 'testapp -h' for help.
+
+        Error: Too many parameters when requesting help; pass a command, '--all', or leave it empty
```
sergiusens, line 461: I am not fan of "leave it empty" but I cannot find other words for it

sergiusens (approved):
Looks good, I do find it a bit confusing as the variable names don't follow the convention for commands, options and arguments seen in other CLI implementations and their conventions. A varname refactor could perhaps help.

tests/test_help.py
facundobatista: I had the same struggle. In 2nd place came "or nothing."

cmatsuoka (approved):

Issue #551, opened 2021-09-22
[charmcraft 1.3]: Cannot pack a charm with Jinja2 as dependency


heitorPB:
Charmcraft 1.3 fails to pack our [slurmd charm](https://github.com/omnivector-solutions/slurm-charms). Reverting to 1.2 works again. 

Here's part of the log:
```
[...]
2021-09-22 17:22:58,259  craft_parts.executor.step_handler        DEBUG    :: :: Collecting jinja2
2021-09-22 17:22:58,261  craft_parts.executor.step_handler        DEBUG    :: ::   Using cached Jinja2-3.0.1.tar.gz (267 kB)
2021-09-22 17:22:58,388  craft_parts.executor.step_handler        DEBUG    :: ::     ERROR: Command errored out with exit status 1:
2021-09-22 17:22:58,388  craft_parts.executor.step_handler        DEBUG    :: ::      command: /root/parts/charm/build/staging-venv/bin/python3 -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-install-nbidbuh_/jinja2/setup.py'"'"'; __file__='"'"'/tmp/pip-install-nbidbuh_/jinja2/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /tmp/pip-install-nbidbuh_/jinja2/pip-egg-info
2021-09-22 17:22:58,388  craft_parts.executor.step_handler        DEBUG    :: ::          cwd: /tmp/pip-install-nbidbuh_/jinja2/
2021-09-22 17:22:58,388  craft_parts.executor.step_handler        DEBUG    :: ::     Complete output (35 lines):
2021-09-22 17:22:58,388  craft_parts.executor.step_handler        DEBUG    :: ::     Traceback (most recent call last):
2021-09-22 17:22:58,388  craft_parts.executor.step_handler        DEBUG    :: ::       File "<string>", line 1, in <module>
2021-09-22 17:22:58,388  craft_parts.executor.step_handler        DEBUG    :: ::       File "/tmp/pip-install-nbidbuh_/jinja2/setup.py", line 4, in <module>
2021-09-22 17:22:58,388  craft_parts.executor.step_handler        DEBUG    :: ::         setup(
[...]

PR #553, opened 2021-09-23
parts,charm_builder: add charm-python-packages plugin property (CRAFT-552)

cmatsuoka:
Packages listed in the `charm-python-packages` property are installed
before dependencies listed in the requirements file. This mirrors
`python-packages` that exists in the python plugin.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -162,6 +162,19 @@ jobs:
           unzip -l build-packages-test_*.charm | grep "venv/ops/charm.py"
           popd
 
+          mkdir -p python-packages-test
+          pushd python-packages-test
+          charmcraft -v init --author testuser
+          sed -i "s|20.04|$VERSION_ID|g" charmcraft.yaml
+          cat <<- EOF >> charmcraft.yaml
+          parts:
+            charm:
+              charm-python-packages: [bump2version]
+          EOF
+          sg lxd -c "charmcraft -v pack"
+          unzip -l python-packages-test_*.charm | grep "venv/bumpversion/__init__.py"
+          popd
+
           sudo snap set charmcraft provider=lxd
           sudo snap set charmcraft provider=multipass
           if sudo snap set charmcraft provider=invalid; then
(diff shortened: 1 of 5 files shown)


cmatsuoka:
@heitorPB, could you check if the jinja2 installation works correctly using `charm-python-packages: [markupsafe]`?

facundobatista (approved):
100% awesome!

heitorPB:
I wonder if the `charm-python-packages` entry should be in the `bases` instead of `parts`? We do support multiple OSes and we will eventually encounter a situation that we need to add a dependency to the build system for one OS but not for the other.

Same comment applies for `charm.parts.build-packages`.

heitorPB:
> @heitorPB, could you check if the jinja2 installation works correctly using `charm-python-packages: [markupsafe]`?

Is this availabe in the `latest/edge` snap? How do I run it locally?

cmatsuoka:
> I wonder if the `charm-python-packages` entry should be in the `bases` instead of `parts`? We do support multiple OSes and we will eventually encounter a situation that we need to add a dependency to the build system for one OS but not for the other.

For Python packages you can add a platform conditional such as:
```
$ pip install "markupsafe;sys_platform=='darwin'"
Ignoring markupsafe: markers 'sys_platform == "darwin"' don't match your environment
```

> Same comment applies for `charm.parts.build-packages`.

In this case we need to add a more advanced grammar processing similar to the one used in Snapcraft.

cmatsuoka:
> Is this availabe in the `latest/edge` snap? How do I run it locally?

Hmm, edge snaps are built from the main branch so you can either create a snap locally (by checking out the branch and running `snapcraft`) or run from the checked out tree in destructive mode. Creating a local snap should be easier and safer.

heitorPB:
Could we use this plugin to update a specific library? It appears that the problem with #551 is `setuptools` is not the latest version.

cmatsuoka:
> Could we use this plugin to update a specific library? It appears that the problem with #551 is `setuptools` is not the latest version.

Ah, that's an interesting find. Yes, setuptools should be upgraded if listed in `charm-python-packages`.

cmatsuoka:
Landing it. It should be available in edge tomorrow.

Issue #310, opened 2021-04-23
Charmcraft should validate all yaml files when building a charm


heitorPB:
Yesterday i added a buggy action to our charms. This was the problematic part, in our `actions.yaml`:
```
foo:
  title: 'Runs foo."
```

Charmcraft successfully built the charm. No warnings/errors given.

But when deploying a bundle that uses this charm, I got this message:

```bash
$ juju deploy ./bundle.yaml
ERROR cannot deploy bundle: yaml: line 60: found unexpected end of stream
```

Our CI was failing with the same message. I had no idea how to debug this. Our bundle has 30 lines, the error was on line 60 (?)...

Charmcraft should validate all charm files before building the charm. A message like `ERROR invalid actions.yaml` would be great.

jameinel:
In this particular case, I think you have 'almost valid yaml', where you
open with a single quote, and everything beyond until the end of the file
is treated as just string content. Which is why the error is at 'line 60'
because you didn't close your string.
I agree that having charmcraft build just ensure that all the .yaml files
are valid yaml is reasonable. I think there is a request from charmcraft
such that Juju could actually validate according to a schema, rather than
trying to map Juju's schema in charmcraft. But it could certainly just try
to safe_load the yaml and make sure it is properly formatted.

facundobatista:
This is implemented already. It runs automatically on building time, and the `analyze` command can be used to explicitly test it.

facundobatista:
We'll create two linters, one for each file, that will only check that they are valid YAML files. No content will be verified. 

Related, there is #547 where it's being discussed what to check in `config,yaml` content.

PR #556, opened 2021-09-28
Added linters to verify actions.yaml and config.yaml (CRAFT-554).

### LINKED ISSUE #547: `charmcraft analyze` - check for invalid `config.yaml`
Follow on from [canonical/operator#618](https://github.com/canonical/operator/pull/618). 

At present, it is possible to successfully `charmcraft pack` on a charm with an empty `config.yaml`, i.e. one that does not contain an `options:` map. Given that Juju will refuse to deploy a charm with such a `config.yaml`, we should probably try to catch this using the `charmcraft analyze` machinery that is invoked when a charm is packed.

As I understand, you can completely omit `config.yaml`, but if it exists it must have a top level key named `options`.

facundobatista:
Fixes #547 and #310.

### DIFF
--- a/charmcraft/linters.py
+++ b/charmcraft/linters.py
@@ -240,7 +240,7 @@ class JujuActions:
 
     check_type = CheckType.lint
     name = "juju-actions"
-    url = "https://juju.is/docs/sdk/charmcraft-analyze#heading--juju-actions"  # FIXME
+    url = "https://juju.is/docs/sdk/charmcraft-analyze#heading--juju-actions"
     text = "The actions.yaml file is not a valid YAML file."
 
     # different result constants
@@ -274,7 +274,7 @@ class JujuConfig:
 
     check_type = CheckType.lint
     name = "juju-config"
-    url = "https://juju.is/docs/sdk/charmcraft-analyze#heading--juju-config"  # FIXME
+    url = "https://juju.is/docs/sdk/charmcraft-analyze#heading--juju-config"
 
     # different result constants
     Result = namedtuple("Result", "ok errors")(ok=OK, errors=ERRORS)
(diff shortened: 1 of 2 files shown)


facundobatista:
FTR, these two new linters were added to the [documentation](https://discourse.charmhub.io/t/analyzers-and-linters/4796).

charmcraft/linters.py
```diff
@@ -233,10 +235,86 @@ def run(self, basedir: pathlib.Path) -> str:
         return result
 
 
+class JujuActions:
+    """Check that the actions.yaml file is valid YAML if it exists."""
+
+    check_type = CheckType.lint
+    name = "juju-actions"
+    url = "https://juju.is/docs/sdk/charmcraft-analyze#heading--juju-actions"
```
cmatsuoka, line 243: Fixed a typo and slightly rephrased the linked text.

charmcraft/linters.py
```diff
@@ -233,10 +235,86 @@ def run(self, basedir: pathlib.Path) -> str:
...
+    The file is considered valid if the following checks are true:
+
+    - has an 'options' key
+    - it is a dictionary
+    - each item inside has the mandatory 'type' key
+    """
+
+    check_type = CheckType.lint
+    name = "juju-config"
+    url = "https://juju.is/docs/sdk/charmcraft-analyze#heading--juju-config"
```
cmatsuoka, line 277: Slightly rephrased the linked text to fix a typo.

cmatsuoka (approved):
The verifications match the valid file description.

charmcraft/linters.py
facundobatista: thanks!

PR #557, opened 2021-09-28
Removed the command group from the command itself to simplify the whole structure (CRAFT-555).

facundobatista:
This is a **big** branch, but 99% are silly and repetitive test changes.

I'll add comments around real code changes, to flag them.

### DIFF
--- a/charmcraft/helptexts.py
+++ b/charmcraft/helptexts.py
@@ -265,7 +265,7 @@ class HelpBuilder:
 
         # recommend other commands of the same group
         for group_name, _, command_classes in self.command_groups:
-            if group_name == command.group:
+            if any(isinstance(command, command_class) for command_class in command_classes):
                 break
         else:
             raise RuntimeError("Internal inconsistency in commands groups")
--- a/charmcraft/main.py
+++ b/charmcraft/main.py
@@ -183,19 +183,19 @@ class Dispatcher:
         for _cmd_group, _, _cmd_classes in commands_groups:
             for _cmd_class in _cmd_classes:
                 if _cmd_class.name in commands:
-                    _stored_class, _ = commands[_cmd_class.name]
+                    _stored_class = commands[_cmd_class.name]
                     raise RuntimeError(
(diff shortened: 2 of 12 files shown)


charmcraft/helptexts.py
```diff
@@ -265,7 +265,7 @@ def get_command_help(self, command, arguments):
 
         # recommend other commands of the same group
         for group_name, _, command_classes in self.command_groups:
-            if group_name == command.group:
+            if any(isinstance(command, command_class) for command_class in command_classes):
```
facundobatista, line 268: This was the real only place where the `group` attribute was used: to find the command group for a given command, when presenting help. 

The new code iterates on all the commands, but this is not a problem as the list will always be small.

charmcraft/main.py
```diff
@@ -183,19 +183,19 @@ def _get_commands_info(self, commands_groups):
...
                 if _cmd_class.name in commands:
-                    _stored_class, _ = commands[_cmd_class.name]
+                    _stored_class = commands[_cmd_class.name]
                     raise RuntimeError(
                         "Multiple commands with same name: {} and {}".format(
                             _cmd_class.__name__, _stored_class.__name__
                         )
                     )
-                commands[_cmd_class.name] = (_cmd_class, _cmd_group)
+                commands[_cmd_class.name] = _cmd_class
```
facundobatista, line 192: Here and a couple of lines above: stop having the group in the internal structure, as it's no longer used.

charmcraft/main.py
```diff
@@ -183,19 +183,19 @@ def _get_commands_info(self, commands_groups):
...
         return commands
 
     def _load_command(self, command_name, cmd_args, charmcraft_config):
         """Load a command."""
-        cmd_class, group = self.commands[command_name]
-        cmd = cmd_class(group, charmcraft_config)
+        cmd_class = self.commands[command_name]
+        cmd = cmd_class(charmcraft_config)
```
facundobatista, line 198: Here and below: stop instantiating the command with the group.

PR #562, opened 2021-10-04
Formalize the command groups to be proper objects, not a tuple of values (CRAFT-560).

facundobatista:
This is a requirement for when the command groups will be defined and used by different codes (lib and app), after the split of part of `charmcraft` code to `craft-cli`.

While doing this, I also removed the group's "codename" and just left the "description", which is what should be used for showing to the user.

### DIFF
--- a/charmcraft/helptexts.py
+++ b/charmcraft/helptexts.py
@@ -119,9 +119,9 @@ class HelpBuilder:
 
         # collect common commands
         common_commands = []
-        for group, _, commands in self.command_groups:
-            max_title_len = max(len(group), max_title_len)
-            for cmd in commands:
+        for command_group in self.command_groups:
+            max_title_len = max(len(command_group.name), max_title_len)
+            for cmd in command_group.commands:
                 if cmd.common:
                     common_commands.append(cmd)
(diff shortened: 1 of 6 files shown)


charmcraft/helptexts.py
```diff
@@ -140,9 +140,9 @@ def get_full_help(self, global_options):
         textblocks.append("\n".join(common_lines))
 
         grouped_lines = ["Commands can be classified as follows:"]
-        for group, _, commands in sorted(self.command_groups):
-            command_names = ", ".join(sorted(cmd.name for cmd in commands))
-            grouped_lines.extend(_build_item(group, command_names, max_title_len))
+        for command_group in sorted(self.command_groups, key=attrgetter("name")):
+            command_names = ", ".join(sorted(cmd.name for cmd in command_group.commands))
+            grouped_lines.extend(_build_item(command_group.name, command_names, max_title_len))
```
facundobatista, line 145: This is the only "functionality change"; it's a bug fix, as previously the "internal groupe code name" was exposed to the user.

sergiusens (approved):
Looks good, in the future I would probably come up with a different taxonomy (e.g. resources, lifecycle, general charmhub, ...)

PR #563, opened 2021-10-06
Improved the Dispatcher to work with default global args and support extra ones in the future (CRAFT-565).

facundobatista:
This is a small improvement (the dispatcher supporting the "extra global args", for which tests were added) and two groups of code moved around (no tests changes needed, as there is no functionality changes really). All part of the plan of splitting this functionality into `craft-cli` lib.

The code moved around or changed is:

- better names! `_Global` transformed into `GlobalArgument` as this will be part of the lib's public API, and `GLOBAL_ARGS` transformed into `_DEFAULT_GLOBAL_ARGS` which will be part of the internal lib's functionality

- moved the functionality of three help-related functions to be part of the Dispatcher (as they need to access/use the "final global args")

Note that the "project dir" global arg is still "default", but will be moved to be "extra" when the Dispatcher is isolated from the configuration loading.

### DIFF
--- a/charmcraft/main.py
+++ b/charmcraft/main.py
@@ -95,18 +95,30 @@ COMMAND_GROUPS = [
 
 
 # global options: the name used internally, its type, short and long parameters, and help text
-_Global = namedtuple("Global", "name type short_option long_option help_message")
-GLOBAL_ARGS = [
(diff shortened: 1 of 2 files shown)


cmatsuoka (approved):
Looks good!

PR #564, opened 2021-10-07
Moved the config loaded (with proper dir specification) outside of the Dispatcher (CRAFT-561).

facundobatista:
The PR is big, but most of it are mechanical changes in the tests. For example in `test_help.py`, because the moment where the exception is raised to provide help changed location (previously everything happened on Dispatch instantiaton, and now it may happen when `pre_parse_args` or `load_command`. Also, `--project-dir` is not included by default anymore in the tests.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -26,9 +26,14 @@ jobs:
         uses: actions/setup-python@v2
         with:
           python-version: ${{ matrix.python-version }}
-      - name: Install/update pip, wheel, and setuptools
+      - name: Install/update pip, wheel, and setuptools on non Windows
+        if: ${{ matrix.os != 'windows-2019' }}
         run: |
           pip install -U pip wheel setuptools
+      - name: Install/update pip, wheel, and setuptools on Windows
+        if: ${{ matrix.os == 'windows-2019' }}
+        run: |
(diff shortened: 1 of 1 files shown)


charmcraft/main.py
```diff
@@ -353,12 +338,34 @@ def main(argv=None):
...
+        # load the dispatcher and put everything in motion
+        dispatcher = Dispatcher(COMMAND_GROUPS, extra_global_options)
+        global_args = dispatcher.pre_parse_args(argv[1:])
+        loaded_config = config.load(global_args["project_dir"])
+        command = dispatcher.load_command(loaded_config)
+        if command.needs_config and not loaded_config.project.config_provided:
+            raise ArgumentParsingError(
+                "The specified command needs a valid 'charmcraft.yaml' configuration file (in "
+                "the current directory or where specified with --project-dir option); see "
+                "the reference: https://discourse.charmhub.io/t/charmcraft-configuration/4138"
```
sergiusens, line 365:
```suggestion
                "The specified command needs a valid 'charmcraft.yaml' configuration file (in "
                "the current directory or in the specified --project-dir option); see "
                "the reference: https://discourse.charmhub.io/t/charmcraft-configuration/4138"
```

PR #581, opened 2021-10-29
snap: remove apt-transport-https

cjwatson:
HTTPS support was moved into apt itself in version 1.5, so staging
apt-transport-https separately hasn't been necessary since core18.


cmatsuoka (approved):
Thanks Colin. I think Snapcraft is also including it, will check.

sergiusens (approved):

sergiusens:
ii  apt-transport-https 2.0.6        all          transitional package for https su>

PR #583, opened 2021-11-01
Rewrite of all tests that assert outputs to improve messages testing  using new machinery.

facundobatista:


cmatsuoka (approved):
Looks good. I'll update the emitter fixture in rockcraft tests to be consistent with these changes.

PR #584, opened 2021-11-01
build: ignore venv in project source (CRAFT-591)

cmatsuoka:
If a project source contains an entry called `venv`, it's mandatory
to ignore it to be able to pack the charm. Instead of requiring
`.jujuignore` to list it, add it to the list of ignored local source
entries when creating the parts lifecycle.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -145,6 +145,7 @@ jobs:
           pushd build-packages-test
           charmcraft -v init --author testuser
           sed -i "s|20.04|$VERSION_ID|g" charmcraft.yaml
+          echo "ignore me" > venv
           touch extra_file.txt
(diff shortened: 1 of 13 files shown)


facundobatista:
Why not add it to `default_juju_ignore` in `charmcraft/jujuignore.py`? Furthermore, what's the whole point of having `ignore_local_sources` if we already have that other mechanism?

cmatsuoka:
> Why not add it to `default_juju_ignore` in `charmcraft/jujuignore.py`? Furthermore, what's the whole point of having `ignore_local_sources` if we already have that other mechanism?

I'm totally fine with declaring it in `charmcraft/jujuignore.py`, will prepare a patch to do so.

facundobatista (approved):
+1, thanks!

PR #585, opened 2021-11-02
Added a --trace global option that sets the Emitter level in TRACE.

facundobatista:


charmcraft/main.py
```diff
@@ -123,6 +123,13 @@ class ProvideHelpException(Exception):
         "--quiet",
         "Only show warnings and errors, not progress",
     ),
+    GlobalArgument(
+        "trace",
+        "flag",
+        "-t",
+        "--trace",
+        "Show all information needed to trace any internal behaviour",
```
cmatsuoka, line 131: Maybe "any" isn't strictly true, only those that have been instrumented will display information.

cmatsuoka (approved):

charmcraft/main.py
facundobatista: Improved, thanks

PR #588, opened 2021-11-02
charm_builder: install binary python packages (CRAFT-601)

cmatsuoka:
Charm plugin property `charm-binary-python-packages` can be used to
list binary packages to install from PyPI. Packages may still be
installed from sources if they're only available in source format.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -61,4 +61,4 @@ tomli==1.2.2
 typing-extensions==3.10.0.2
 urllib3==1.26.7
 zipp==3.6.0
-git+https://github.com/canonical/craft-cli.git@4af19f9c0da733321dc754be1180aea28f3feeb1
+git+git://github.com/canonical/craft-cli.git@4af19f9c0da733321dc754be1180aea28f3feeb1
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -45,7 +45,7 @@ python-dateutil==2.8.2
 pytz==2021.3
 pyxdg==0.27
 PyYAML==6.0
-regex==2021.10.23
+regex==2021.11.2
 requests==2.26.0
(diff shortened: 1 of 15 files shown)


requirements-dev.txt
```diff
@@ -61,4 +61,4 @@ tomli==1.2.2
 typing-extensions==3.10.0.2
 urllib3==1.26.7
 zipp==3.6.0
-git+git://github.com/canonical/craft-cli.git@4af19f9c0da733321dc754be1180aea28f3feeb1
+git+https://github.com/canonical/craft-cli.git@4af19f9c0da733321dc754be1180aea28f3feeb1
```
facundobatista, line 64: Why this change?

snap/snapcraft.yaml
```diff
@@ -72,6 +72,7 @@ parts:
     plugin: nil
     build-packages:
       - python3-dev
+      - git
```
facundobatista, line 75: What this is needed for?

facundobatista (changes requested):
Added a couple of questions, thanks!

requirements-dev.txt
cmatsuoka: Github is deprecating access using the unencrypted git protocol (see https://github.blog/2021-09-01-improving-git-protocol-security-github/). This is not effective yet but there was a brownout on Nov 02 when this PR was submitted. All entries using non-published packages will be removed before the next release.

snap/snapcraft.yaml
cmatsuoka: This was needed to be able to install packages using git+https and can be removed when those lines are removed. I will revert these changes since this brownout period ended.

facundobatista (approved):
Great, thanks!

PR #591, opened 2021-11-08
Revisited all error usage to adapt to new craft-cli context.

facundobatista:
In detail:

- ensured that when the error comes from a sub-process, use `from err` (no need for another error type, it's just handled by `craft-cli`)

- started to use `details` attribute in some errors, to provide extra info

- returned 1 from commands that actually end ok (raising CommandError) but show error messages that come from an unsuccesful call to the Store

- support `craft-store` raising StoreServerError and transform that correctly


cmatsuoka (approved):
Looks good.

PR #602, opened 2021-11-17
Added an export option to login to save the obtained credentials (CRAFT-641).

facundobatista:
**Note**: this PR depends on the #601 one.

### DIFF
--- a/tests/commands/test_store_api.py
+++ b/tests/commands/test_store_api.py
@@ -150,6 +150,10 @@ def test_not_logged_in_warns_alternate_auth(monkeypatch):
 
 def test_login(client_mock, config):
     """Simple login case."""
+    # set up a response from client's login
+    acquired_credentials = "super secret stuff"
+    client_mock.login = MagicMock(return_value=acquired_credentials)
+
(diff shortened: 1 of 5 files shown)


tests/commands/test_store_api.py
```diff
@@ -150,6 +150,10 @@ def test_not_logged_in_warns_alternate_auth(monkeypatch):
 
 def test_login(client_mock, config):
     """Simple login case."""
+    # set up a response from client's login
+    acquired_credentials = "super secret stuff"
```
sergiusens, line 154: maybe for _interface_ clarity make this a base64 encoded string.

sergiusens (approved):
Looks good, added a very minor comment and just to be clear, I only reviewed https://github.com/canonical/charmcraft/pull/602/commits/4ca7aff9a31b74d1d8a045bc21899f54e5ee0142

tests/commands/test_store_api.py
facundobatista: Actually charmcraft does not care about it. It may be a zip binary in the future. It's really obscure to Charmcraft...

PR #603, opened 2021-11-18
Added restrictive options to login command (CRAFT-640).

facundobatista:
**Note**: this PR depends on #601 and #602 .

### DIFF
--- a/charmcraft/commands/store/__init__.py
+++ b/charmcraft/commands/store/__init__.py
@@ -28,6 +28,7 @@ from operator import attrgetter
 
 import yaml
 from craft_cli import emit
+from craft_store import attenuations
 from craft_store.errors import NotLoggedIn
 from humanize import naturalsize
 from tabulate import tabulate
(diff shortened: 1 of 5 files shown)


charmcraft/commands/store/__init__.py
```diff
@@ -104,11 +114,82 @@ def fill_parser(self, parser):
...
+                    "can only be used when credentials are exported."
+                )
+        if parsed_args.permission is not None:
+            invalid = set(parsed_args.permission) - VALID_ATTENUATIONS
+            if invalid:
+                invalid_text = ", ".join(map(repr, sorted(invalid)))
+                raise CommandError(f"Invalid permission: {invalid_text}.")
+
+        # restrictive options, mapping the names between what is used in Namespace (singular,
+        # even if it ends up being a list) and the more natural ones used in the Store layer
```
cmatsuoka, line 178: I had to read this comment a few times to understand it, then it became very obvious (to the point I would consider to removing it altogether).

cmatsuoka (approved):
Looks good, assuming that the store protocol is implemented correctly.

charmcraft/commands/store/store.py
```diff
@@ -54,6 +54,15 @@
 }
 POLL_DELAY = 1
 
+# default restrictions to get auth credentials
+AUTH_DEFAULT_TTL = 3600 * 30
```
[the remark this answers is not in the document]
facundobatista: I guess it's fine, as it's what was used before (Sergio set this value initially). We can change it in the future anyway....

charmcraft/commands/store/__init__.py
facundobatista: I like it, at some point explains why a "translation" is needed, we can improve it later.

PR #609, opened 2021-12-07
store: handle store errors in whoami command (CRAFT-701)

cmatsuoka:
Define a `allow_login` parameter to the store client decorator so
that the `whoami` command can use it to handle errors without
automatically logging the user in.

Example error messages:

- If no external credentials are used, and the user is not logged in:
```
$ charmcraft whoami
You are not logged in to Charmhub.
```

- If credentials are used, and the response code is 401:
```
$ CHARMCRAFT_AUTH="$(cat ~/creds-expired)" charmcraft whoami
Provided credentials are no longer valid for Charmhub. Regenerate them and try again.
```

### DIFF
--- a/charmcraft/commands/store/store.py
+++ b/charmcraft/commands/store/store.py
@@ -126,41 +126,52 @@ def _get_hostname() -> str:
     return hostname
 
 
-def _store_client_wrapper(method):
+def _store_client_wrapper(auto_login=True):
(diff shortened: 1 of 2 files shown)


facundobatista (approved):
Wonderful, thanks!

sergiusens (approved):
I am not fond of the word "Regenerate" but the rest looks good

PR #612, opened 2021-12-09
Show more information about current credentials in the whoami command (CRAFT-651).

facundobatista:
For this I had to update the `craft-store` version, which implied a couple of unexpected API changes (very small, so I just left them here).

### DIFF
--- a/charmcraft/commands/store/store.py
+++ b/charmcraft/commands/store/store.py
@@ -31,7 +31,9 @@ from charmcraft.cmdbase import CommandError
 from charmcraft.commands.store.client import Client, ALTERNATE_AUTH_ENV_VAR
 
 # helpers to build responses from this layer
-User = namedtuple("User", "name username userid")
+Account = namedtuple("Account", "name username id")
+Package = namedtuple("Package", "id name type")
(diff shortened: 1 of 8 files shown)


charmcraft/commands/store/store.py
```diff
@@ -210,10 +212,20 @@ def whoami(self):
...
+        else:
+            packages = [
+                Package(type=pkg["type"], name=pkg.get("name"), id=pkg.get("id"))
+                for pkg in response["packages"]
+            ]
+        result = MacaroonInfo(
+            account=account,
+            packages=packages,
+            channels=response["channels"],
+            permissions=response["permissions"],
```
sergiusens, line 228: is it possible for charmhub to not return these keys at all

sergiusens (approved):
Looks good, just one observation on potential KeyError

charmcraft/commands/store/store.py
facundobatista: Nop, check the [API Spec for the used endpoint](https://api.charmhub.io/docs/default.html#macaroon_info):

```
    "required": [
        "account",
        "channels",
        "packages",
        "permissions"
    ],
```

They may be `null`, though, which is supported in the code...

PR #613, opened 2021-12-14
plugins: reactive using charm tool (CRAFT-699)

sergiusens:


tests/test_parts.py
```diff
@@ -177,6 +177,117 @@ def test_invalid_properties(self):
...
+    def test_get_build_commands(self, tmp_path):
+        assert self._plugin.get_build_commands() == [
+            "rm -f charmcraft.yaml",
+            # exit on errors (code 200), warnings (code 100) allowed
+            'charm proof || test "$?" -ge 200 && exit 1',
+            f'mkdir -p "{tmp_path}/parts/foo/install"',
+            f'ln -sf . "{tmp_path}/parts/foo/install/fake-project"',
+            f'charm build -o "{tmp_path}/parts/foo/install" || test "$?" -ge 200 && exit 1',
+            f'rm -f "{tmp_path}/parts/foo/install/fake-project"',
+        ]
```
facundobatista, line 249: We need to test this better.

This is the core of the plugin, and the test is just an adapted "copy and paste" of above's strings. I'm not worried about current lines (they look fine... but again, they are untested), but this is not useful for future changes, in the sense that if I need to change something in these commands, I would just modify the real code and then "blindly" modify these lines accordingly.

Let's talk about this. I chatted several times with Claudio too about this "limitation" of having "untested bash script lines" as the core of the plugins.

One way of overcoming this limitation is:
- have a `reactive_plugin.py` file.
- that file is imported from `parts.py`, which uses `ReactivePluginEnvironmentValidator`, `ReactivePlugin` etc.
- `get_build_commands` all is does is something like `python3 somepath/reactive_plugin.py`
- `if __name__ == "__main__"` in `reactive_plugin.py` will just run another function
- that function is properly tested

facundobatista (changes requested):
All looks nice except the plugin's core commands, I tried to explain myself in the comment, let's talk if not clear enough.

Thanks!

tests/test_parts.py
sergiusens: https://github.com/canonical/charmcraft/pull/613/commits/b95defda988521dda1d147e60ee940a65892af70 has the same code in a new module, new code upcoming.

charmcraft/reactive_plugin.py
```diff
@@ -0,0 +1,180 @@
...
+    """Build a charm using charm tool.
+
+    The charm tool is used to build reactive charms, the build process
+    is as follows:
+
+    - Remove any charmcraft.yaml from the build directory (occurs for
+      local in-tree builds)
+
+    - Run charm proof to ensure the charm
+      would build with no errors (warnings are allowed) - - Create a
```
facundobatista, line 140: it looks an item got together with the previous one here

charmcraft/reactive_plugin.py
```diff
@@ -0,0 +1,180 @@
...
+      symlink to where charm
```
facundobatista, line 141: and maybe the sentence is not finished?

charmcraft/reactive_plugin.py
```diff
@@ -0,0 +1,180 @@
...
+        if call_error.returncode >= 200:
+            raise call_error
+    finally:
+        charm_build_dir.unlink()
+
+
+if __name__ == "__main__":
+    try:
+        build(charm_name=sys.argv[1], build_dir=Path(sys.argv[2]), install_dir=Path(sys.argv[3]))
+    except subprocess.CalledProcessError as call_error:
```
facundobatista, line 178: Don't leak `CalledProcessError` and the magic number `200` out of `build`.

It's get simpler, even: make `build` to return the return code (see comments above)

charmcraft/reactive_plugin.py
```diff
@@ -0,0 +1,180 @@
...
+    charmcraft_yaml = build_dir / "charmcraft.yaml"
+    if charmcraft_yaml.exists():
+        charmcraft_yaml.unlink()
+
+    # Verify the charm is ok from a charm tool point of view.
+    try:
+        subprocess.run(["charm", "proof"], check=True)
+    except subprocess.CalledProcessError as call_error:
+        if call_error.returncode >= 200:
+            raise call_error
```
facundobatista, line 158: as per comment below, here you would do `return call_error.returncode`.

charmcraft/reactive_plugin.py
```diff
@@ -0,0 +1,180 @@
...
+        charm_build_dir.symlink_to(install_dir, target_is_directory=True)
+
+    try:
+        subprocess.run(["charm", "build", "-o", build_dir], check=True)
+    except subprocess.CalledProcessError as call_error:
+        if call_error.returncode >= 200:
+            raise call_error
+    finally:
+        charm_build_dir.unlink()
+
```
facundobatista, line 173: as per comment below, here you would do `return 0`.

facundobatista (changes requested):
Added a couple of simple details.

Thanks!

facundobatista (approved):
Looks awesome, thanks!

PR #614, opened 2021-12-15
Exploration of splitting argparsing out from charmcraft into another lib (CRAFT-704).

facundobatista:
This is to define API/interaction to move the functionality into `craft-cli`.

Detail of what is included in this big refactor:

The big diff is mostly that I just moved stuff from one place to the other:

- the whole `helptext` module to the fake directory

- the dispatch and global commands to a new `dispatcher` module (leaving charmcraft-specific code in `main.py`)

- the `BaseCommand` class into the `dispatcher` module too

- the two errors from `main.py` into a `errors` module (that later will be merged with `craft_cli` one)

Note that tests are adjusted/moved/copied around to reflect the changes just described.

There are functionality changes, though; these are marked with comments in the PR, but here's a summary:

- moved a couple of "generic commands tests" (aesthetic stuff over their strings) from `test_cmdbase` into `test_main`

- the `test_basecommand_mandatory_attributes` was challenging, as it verified that some mandatory attributes (which is responsibility of the library) are defined in all the commands (which are application specific); to untangle this I...
    - declared some abstract attributes in BaseCommand
    - added tests for it so we ensure that no command could subclass it without defining those attributes
    - removed the `test_basecommand_mandatory_attributes` itself

- The `help_builder` init is not done "by the app" anymore, but by the Dispatcher (had to adjust its `__init__` signature properly)


tests/test_main.py
```diff
@@ -568,5 +223,32 @@ def test_commands(cmd_name):
...
+            """Return self, as it is used to add arguments too."""
+            return self
+
+        def add_argument(self, *args, **kwargs):
+            """Verify that all commands have a correctly formatted help."""
+            help_msg = kwargs.get("help")
+            assert help_msg, "The help message must be present in each option"
+            assert help_msg[0].isupper() and help_msg[-1] != "."
+
+    command(config).fill_parser(FakeParser())
```
facundobatista, line 254: These last two tests are what was moved from `test_cmdbase`.

fake_craft_cli/dispatcher.py
```diff
@@ -0,0 +1,338 @@
...
+
+    @property
+    @abstractmethod
+    def help_msg(self):
+        """A one line help for user documentation."""
+
+    @property
+    @abstractmethod
+    def overview(self):
+        """Ah longer multi-line text with the whole command description."""
```
facundobatista, line 320: This is how I made these attributes "mandatory".

tests/test_fakecraftcli_dispatcher.py
```diff
@@ -0,0 +1,468 @@
...
+
+def test_basecommand_mandatory_attribute_name():
+    """BaseCommand subclasses must override the name attribute."""
+
+    class TestClass(BaseCommand):
+        help_msg = "help message"
+        overview = "fake overview"
+
+    with pytest.raises(TypeError):
+        TestClass("config")
```
facundobatista, line 446: Here and below are the new tests to check that attributes are mandatory in subclasses.

charmcraft/main.py
```diff
@@ -369,7 +116,12 @@ def main(argv=None):
         setup_parts()
 
         # load the dispatcher and put everything in motion
-        dispatcher = Dispatcher(COMMAND_GROUPS, extra_global_options)
+        dispatcher = Dispatcher(
+            "charmcraft",
+            COMMAND_GROUPS,
+            summary=GENERAL_SUMMARY,
+            extra_global_args=extra_global_options,
+        )
```
facundobatista, line 124: Changed signature to "hide" `help_builder` init inside Dispatcher.

fake_craft_cli/dispatcher.py
```diff
@@ -0,0 +1,338 @@
...
+    The subclass must be declared in the corresponding section of main.COMMAND_GROUPS,
+    and will receive and store this group on instantiation (if overriding `__init__`, the
+    subclass must pass it through upwards).
+    """
+
+    common = False
+    needs_config = False
+
+    def __init__(self, config):
+        self.config = config
```
facundobatista, line 305: **OPEN QUESTION**: All commands in all apps will have the config as an attribute? 

This is **very** handy, and Charmcraft usage of config was coded around this, but in other apps it may be "against the grain" of how config is handed there. Maybe the BaseCommand in the lib should NOT have this, and Charmcraft should subclass it to add it.

Thoughts? Thanks!

sergiusens: if config could be none, this would be similar to click's concept of "context"

sergiusens:
Looks good, it would really help to get a feel of this if a short tutorial (or code snippet to later use in the tutorial) is presented

facundobatista:
Formally proposed in #635 .

PR #619, opened 2022-01-07
setup, packaging: pin cryptography package and update dependencies

cmatsuoka:
cryptography is not needed as a direct dependency, but it will be brought in
newer versions from other dependencies, and the new ones need a Rust compiler
which is not present in armhf, ppc64el and s390x architectures.


facundobatista (approved):
Thanks!!

PR #624, opened 2022-01-12
Massive renaming of CommandError to CraftError (CRAFT-709).

facundobatista:
The only "real changes" are the import changes from cmdbase to craft_cli, and of course the removal of the no longer used CommandError definition.

### DIFF
--- a/charmcraft/charm_builder.py
+++ b/charmcraft/charm_builder.py
@@ -25,9 +25,8 @@ import sys
 import subprocess
 from typing import List
 
-from craft_cli import emit, EmitterMode
+from craft_cli import emit, EmitterMode, CraftError
 
-from charmcraft.cmdbase import CommandError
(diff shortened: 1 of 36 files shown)


cmatsuoka:
Would it be worth to have an intermediate `CharmcraftError` below `CraftError` so that any application-wide custom adjustments could be done in this derived class?

facundobatista:
We didn't identify such a need (that's why we're removing CommandError instead of renaming it). In reality CraftError already have a lot of more functionality than what was used in Charmcraft before (design more inherited from extended capabilities in Snapcraft), so the direction that I foresee here is to start using that functionality from Charmcraft itself (e.g. "docs_url").

cmatsuoka (approved):
This is just a direct replacement and shouldn't create any problems If there's no need for an intermediate error class.

PR #626, opened 2022-01-14
add prometheus charm pack test

simskij:
To avoid regressions like the one described in https://github.com/canonical/charmcraft/issues/625, this adds a test where we build the prometheus-k8s-operator, which at the moment is the most complex charm of the observability team.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -232,6 +232,7 @@ jobs:
             echo "configure script failure"
             exit 1
           fi
+          sudo snap set charmcraft provider=lxd
       - name: Test building the prometheus-k8s-charm
         run: |
           git clone https://github.com/canonical/prometheus-k8s-operator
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yaml
```diff
@@ -232,7 +232,13 @@ jobs:
             echo "configure script failure"
             exit 1
           fi
-
+      - name: Test building the prometheus-k8s-charm
```
facundobatista, line 235: Hello @simskij ! Thanks for this.

This test is failing because multipass is not available, I think because the previous step left it "dirty".

To try it, may you add the following line as the last thing in the previous step? (between the `fi` and your new code):
```
sudo snap set charmcraft provider=lxd
```
Thanks!

facundobatista (changes requested):
Suggested a change to see if we fix the environment. Thanks!

.github/workflows/tests.yaml
simskij: Sure, let's try it 👍🏽

simskij:
Ready for review 👍🏽

facundobatista (approved):
Wonderful! We'll add this test here, and in a future move it to the proper integration tests infrastructure (that needs to be setup/created).

PR #627, opened 2022-01-17
Added git into the build packages list, as it's needed by pip.

facundobatista:


cmatsuoka (approved):
This was also needed during github's unsecure access brownouts.

PR #628, opened 2022-01-19
Proof of concept for packing declarative charms.

facundobatista:
To use this PoC run `charmcraft pack` in a directory with the following two files:

**charmcraft.yaml**
```
type: "charm"
bases:
  - build-on:
      - name: "ubuntu"
        channel: "20.04"
    run-on:
      - name: "ubuntu"
        channel: "20.04"
parts:
  charm:
    source: .  # FIXME we should not need this
    plugin: declarative
```

**manifest.yaml**
```
name: my_charm
summary: Declarative manifest.
description: An example declarative manifest to try the Charmcraft's PoC to build it.

requires:
  relations:
    interface: mongodb

environment:
  - name: SPRING_DATA_MONGODB_URI
    template: '{{relations.consumes.datasource.app.replica_set_uri}}'
```

### DIFF
--- a/charmcraft/commands/build.py
+++ b/charmcraft/commands/build.py
@@ -129,7 +129,8 @@ class Builder:
 
         self.buildpath = self.charmdir / BUILD_DIRNAME
         self.config = config
-        self.metadata = parse_metadata_yaml(self.charmdir)
+        #self.metadata = parse_metadata_yaml(self.charmdir)
+        self.metadata_name = "declarativepoc"
 
         if self.config.parts:
             self._parts = self.config.parts.copy()
@@ -217,7 +218,7 @@ class Builder:
             self._parts,
             work_dir=work_dir,
             project_dir=self.charmdir,
-            project_name=self.metadata.name,
+            project_name=self.metadata_name,
             ignore_local_sources=["*.charm"],
         )
         lifecycle.run(Step.PRIME)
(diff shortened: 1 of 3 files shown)


charmcraft/commands/build.py
```diff
@@ -129,7 +129,8 @@ def __init__(self, args, config):
 
         self.buildpath = self.charmdir / BUILD_DIRNAME
         self.config = config
-        self.metadata = parse_metadata_yaml(self.charmdir)
+        #self.metadata = parse_metadata_yaml(self.charmdir)
+        self.metadata_name = "declarativepoc"
```
facundobatista, line 133: A Charmcraft refactor will be needed here.

The problem is that in current code, the `metadata.yaml` is needed in several places to start/use the craft_parts machinery, and for declarative charms (at least if we stick to the "plugin" model), the `metadata.yaml` will be available only after running the declarative plugin (by craft_parts).

charmcraft/declarative_plugin.py
```diff
@@ -0,0 +1,133 @@
...
+    """
+    emit.trace(f"Declarative building started from dir {project_dir}.")
+
+    manifest_filepath = project_dir / "manifest.yaml"
+    if not manifest_filepath.exists():
+        raise CraftError("Cannot find mandatory 'manifest.yaml' file.")
+    manifest_content = yaml.safe_load(manifest_filepath.read_text())
+    emit.trace("Manifest loaded ok.")
+
+    # TODO: we need to define a proper versioned location for this
```
facundobatista, line 80: Big important decision here, about where the "cnb operator" will be located. Versions are important.

PR #635, opened 2022-01-24
Use all functionality from craft-cli (CRAFT-710).

facundobatista:


cmatsuoka (approved):
This rebase of Charmcraft code on the argument parsing provided by craft-cli looks good (and the fact that the craft-cli implementation is based on code extracted from Charmcraft itself suggests that there are no missing features or behavior changes.)

Issue #582, opened 2021-11-01
Allow emitter to be paused


cmatsuoka:
There are situations where we may allow a process to execute with complex terminal output. A typical example would be a craft tool re-execution inside a build provider, which would use craft-cli to report the execution progress. In this case, we don't want the spinner from the outer instance to appear during the execution of the inner instance, which can have its own spinner. A possible implementation could be a context manager that pauses the emitter.

PR #651, opened 2022-02-04
Several small improvements after testing and inspecting logs.

facundobatista:
In detail:

- removed legacy support for bases configuration being None

- improved the message when launching container/VM (suggested by Mark)

- better logging of jujuignored rules

### DIFF
--- a/charmcraft/commands/build.py
+++ b/charmcraft/commands/build.py
@@ -70,21 +70,14 @@ def _format_bases_config(bases_config: BasesConfiguration) -> str:
     return "_".join([_format_run_on_base(r) for r in bases_config.run_on])
 
 
-def format_charm_file_name(
-    charm_name: str, bases_config: Optional[BasesConfiguration] = None
-) -> str:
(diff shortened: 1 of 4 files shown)


charmcraft/commands/build.py
```diff
@@ -434,7 +427,10 @@ def pack_charm_in_instance(
         if self.shell_after:
             cmd.append("--shell-after")
 
-        emit.progress(f"Launching environment to pack for base {build_on}")
+        emit.progress(
+            f"Launching environment to pack for base {build_on} "
+            "(may take a while the first time but it's reusable)"
```
cmatsuoka, line 432: I think the environment launch time in charmcraft may appear longer than what it actually is because the actual charm build process is silent. It's possible that if more progress is printed during the charm build, it may seem faster.

Note that this will require the outer terminal control to be paused so that the inner instance can print messages (see issue #582)

cmatsuoka (approved):

charmcraft/commands/build.py
facundobatista: Ack, thanks!

PR #670, opened 2022-02-11
errors: catch CredentialsNotParseable and CraftStoreErrors

mr-cal:
Catch all CraftStoreErrors while initializing the craft_store library.


charmcraft/commands/store/store.py
```diff
@@ -181,7 +181,7 @@ class Store:
     def __init__(self, charmhub_config):
         try:
             self._client = Client(charmhub_config.api_url, charmhub_config.storage_url)
-        except craft_store.errors.NoKeyringError as error:
+        except craft_store.errors.CraftStoreError as error:
```
sergiusens, line 184: I believe Facundo's comment is related to catching it here https://github.com/canonical/charmcraft/blob/feacef7aadc4f21f3f8f259f898f2ec62fe9fd70/charmcraft/main.py#L154

sergiusens: as a new except

sergiusens (approved):
I am assuming most things work here

PR #675, opened 2022-02-16
linters: module name from metadata name

sergiusens:

### DIFF
--- a/tests/test_linters.py
+++ b/tests/test_linters.py
@@ -321,7 +321,10 @@ def test_framework_operator_no_ops_imported(tmp_path, monkeypatch, import_line):
         "from charms.reactive.stuff import Stuff",
     ],
 )
-def test_framework_reactive_used_ok(tmp_path, monkeypatch, import_line):
+@pytest.mark.parametrize(
+    "charm_name, charm_module", [("foobar", "foobar.py"), ("foo-bar", "foo_bar.py")]
+)
+def test_framework_reactive_used_ok(tmp_path, import_line, charm_name, charm_module):
     """The reactive framework was used.
 
     Parametrized args:
@@ -329,10 +332,10 @@ def test_framework_reactive_used_ok(tmp_path, monkeypatch, import_line):
     """
     # metadata file with needed name field
     metadata_file = tmp_path / "metadata.yaml"
-    metadata_file.write_text("name: foobar")
+    metadata_file.write_text(f"name: {charm_name}")
 
     # a Python file that imports charms.reactive
-    entrypoint = tmp_path / "reactive" / "foobar.py"
+    entrypoint = tmp_path / "reactive" / charm_module
     entrypoint.parent.mkdir()
     entrypoint.write_text(f"{import_line}")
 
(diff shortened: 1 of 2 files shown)


tests/test_linters.py
```diff
@@ -16,27 +16,15 @@
...
-    JujuActions,
-    JujuConfig,
-    JujuMetadata,
-    Language,
-    UNKNOWN,
-    analyze,
-    check_dispatch_with_python_entrypoint,
-)
-
+from charmcraft.linters import (CHECKERS, FATAL, IGNORED, UNKNOWN, CheckType,
```
facundobatista, line 24: This format change is because a new black? Can we stick to the old format, please?

Furthermore, why this part of the diff anyway? I don't see any changes in the import list (and comparing them is something this new format makes very difficult).

tests/test_linters.py
```diff
@@ -321,18 +309,19 @@ def test_framework_operator_no_ops_imported(tmp_path, monkeypatch, import_line):
...
     - import_line: different ways to express the import
     """
     # metadata file with needed name field
     metadata_file = tmp_path / "metadata.yaml"
-    metadata_file.write_text("name: foobar")
+    metadata_file.write_text(f"name: {charm_name}")
 
     # a Python file that imports charms.reactive
-    entrypoint = tmp_path / "reactive" / "foobar.py"
+    entrypoint = tmp_path / "reactive" / "{}.py".format(charm_name.replace("-", "_"))
```
facundobatista, line 324: Don't just duplicate the tested code here, is error prone (if the original code has a difficult to see bug, we're just reproducing it here). 

Be explicit! Parametrize passing both charm and module names. Thanks!

facundobatista (changes requested):
+1 to the code, two small details annotated inside. Thanks!

tests/test_linters.py
sergiusens: New black, yes. Reverted for now

tests/test_linters.py
sergiusens: Fixed

tests/test_linters.py
```diff
@@ -321,18 +321,19 @@ def test_framework_operator_no_ops_imported(tmp_path, monkeypatch, import_line):
         "from charms.reactive.stuff import Stuff",
     ],
 )
-def test_framework_reactive_used_ok(tmp_path, monkeypatch, import_line):
+@pytest.mark.parametrize("charm_name", [("foobar", "foobar"), ("foo-bar", "foo_bar")])
+def test_framework_reactive_used_ok(tmp_path, import_line, charm_name):
```
facundobatista, line 325:
```
@pytest.mark.parametrize("charm_name,charm_module", [("foobar", "foobar"), ("foo-bar", "foo_bar")])
def test_framework_reactive_used_ok(tmp_path, import_line, charm_name, charm_module):
```

facundobatista (changes requested):
There's a better way than accessing items in the tuple.

tests/test_linters.py
sergiusens: done

facundobatista (approved):
Awesome, thanks!

Issue #679, opened 2022-02-16
Charmcraft should allow resource revisions == 0 when specifying uploads


facundobatista:
Example of what should be a valid command:

```
charmcraft release etcd --revision=670 --channel=latest/edge --resource=core:0
```

PR #682, opened 2022-02-18
Allow for resources with a rev of 0 during `charmcraft release`

addyess:
addreses issue #679


facundobatista (approved):
Awesome, thank you very much!

PR #684, opened 2022-02-23
Add name argument for upload command

### LINKED ISSUE #681: Add a --name parameter to upload command (to use a different name than the specified in metadata.yaml)
With the recent transition of the old charm store to charmhub all charm uploads now need to be handled. In our case charms are available to our customers through the old charm store URLs (e.g. `cs:~anbox-charmers/ams`) and the `metadata.yaml` file includes `name: ams`. Migration to charmhub flat names is in progress but we will have customers using the charm store URLs for some more time and need to make sure we continue to support them.

The problem now is that charmcraft forces us to rename the charm in `metadata.yaml` when we want to upload it to charmhub. WIth the import `cs:~anbox-charmers/ams` became `anbox-charmers-ams` on charmhub but the charm still contains `name: anbox` in `metadata.yaml` and has to continue to do so for backward compatibility and as we plan to move it in the near future to the flat name `ams` on charmhub.

The problem is specifically in the implementation of the upload command where it derives name to upload to on charmhub from the name set in `metadata.yaml` (see https://github.com/canonical/charmcraft/blob/main/charmcraft/commands/store/__init__.py#L64) which will fails as in our case it will derive `ams` rather than `anbox-charmers-ams` and with that will upload to the wrong charm or fail if permissions to upload to the `ams` are not given.

To workaround we added a `--name` argument to the upload command in a custom charmcraft we're using for the time being which allows to upload to a specific charm on charmhub which doesn't match the name specified in `metadata.yaml`.

It would be great if we can get this solved in charmcraft proper.

morphis:
In some cases the name of the charm a user wants to upload is not the
same as the one specified in metadata.yaml. Specifically for charms
imported into charmhub from the old charmstore this isn't true if they
formerly namespaces (e.g. cs:~foo/bar become foo-bar but charm name
remains foo in metadata.yaml).

To allow charms to be still uploaded in this case a --name option is
added to the upload command which allows a user to specify the exact
name of the charm he wants to upload to rather than letting charmcraft
automatically extract the name from the metadata.yaml file

Fixes #681

### DIFF
--- a/charmcraft/commands/store/__init__.py
+++ b/charmcraft/commands/store/__init__.py
@@ -473,6 +473,11 @@ class UploadCommand(BaseCommand):
             help="The channel(s) to release to (this option can be indicated multiple times)",
         )
         parser.add_argument(
+            "--name",
+            type=str,
+            help="Name of the charm or bundle on Charmhub to upload to",
+        )
+        parser.add_argument(
             "--resource",
             action="append",
             type=ResourceOption(),
@@ -507,7 +512,10 @@ class UploadCommand(BaseCommand):
 
     def run(self, parsed_args):
         """Run the command."""
-        name = get_name_from_zip(parsed_args.filepath)
+        if parsed_args.name:
+            name = parsed_args.name
+        else:
+            name = get_name_from_zip(parsed_args.filepath)
         self._validate_template_is_handled(parsed_args.filepath)
         store = Store(self.config.charmhub)
         result = store.upload(name, parsed_args.filepath)
--- a/tests/commands/test_store_commands.py
+++ b/tests/commands/test_store_commands.py
@@ -750,7 +750,7 @@ def test_upload_call_ok(emitter, store_mock, config, tmp_path):
 
     test_charm = tmp_path / "mystuff.charm"
     _build_zip_with_yaml(test_charm, "metadata.yaml", content={"name": "mycharm"})
-    args = Namespace(filepath=test_charm, release=[])
+    args = Namespace(filepath=test_charm, release=[], name=None)
     retcode = UploadCommand(config).run(args)
     assert retcode == 0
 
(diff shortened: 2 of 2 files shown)


hloeung:
@facundobatista or anyone else from the team able to take a look?

charmcraft/commands/store/__init__.py
```diff
@@ -472,6 +472,11 @@ def fill_parser(self, parser):
             action="append",
             help="The channel(s) to release to (this option can be indicated multiple times)",
         )
+        parser.add_argument(
+            "--name",
+            type=str,
+            help="Name of the charm on Charmhub to upload to",
```
facundobatista, line 478: it can be a bundle too, so maybe say "Name of the charm/bundle on ..."?

charmcraft/commands/store/__init__.py
```diff
@@ -508,6 +513,8 @@ def _validate_template_is_handled(self, filepath):
     def run(self, parsed_args):
         """Run the command."""
         name = get_name_from_zip(parsed_args.filepath)
+        if hasattr(parsed_args, "name"):
+            name = parsed_args.name
```
facundobatista, line 517: `parsed_args` will always have `name`, in `None` if the user didn't pass it. So you could just do `if parsed_args.name:`...

charmcraft/commands/store/__init__.py
```diff
@@ -508,6 +513,8 @@ def _validate_template_is_handled(self, filepath):
     def run(self, parsed_args):
         """Run the command."""
         name = get_name_from_zip(parsed_args.filepath)
```
facundobatista, line 515: This is not necessary if the user indicated the name through the option.

tests/commands/test_store_commands.py
```diff
@@ -893,6 +893,22 @@ def test_upload_charm_with_init_template_todo_token(tmp_path, config):
         UploadCommand(config).run(args)
 
 
+def test_upload_with_different_name_than_in_metadata(emitter, store_mock, config, tmp_path):
+    """Simple upload, success result."""
```
facundobatista, line 897: This docstring is misleading, adapt it to this test (explain the case that is being tested).

facundobatista (changes requested):
Great, thanks for this PR!!! I annotated a couple of details inside. Thanks again!

facundobatista (approved):
Awesome, thank you!

PR #698, opened 2022-03-04
_get_lib_info should generate importable paths.

### LINKED ISSUE #284: Running `charmcraft fetch-lib` on a charm with dashes should fetch to a path with underscores
If you have a library for a charm (such as nginx-ingress-integrator) that has dashes in the name, and run `charmcraft fetch-lib charms.nginx-ingress-integrator.v0.charms` it should still fetch to a path on disk that has underscores rather than dashes, so it's importable.

jguedez:
Fixes #284.

### DIFF
--- a/charmcraft/commands/store/__init__.py
+++ b/charmcraft/commands/store/__init__.py
@@ -922,9 +922,13 @@ def _get_lib_info(*, full_name=None, lib_path=None):
         # dir and Python extension.
         #    e.g.: charms.mycharm.v4.foo -> lib/charms/mycharm/v4/foo.py
         try:
-            charmsdir, importable_charm_name, v_api, libfile = full_name.split(".")
+            charmsdir, charm_name, v_api, libfile = full_name.split(".")
         except ValueError:
             raise _BadLibraryNameError(full_name)
+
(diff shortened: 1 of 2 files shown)


charmcraft/commands/store/__init__.py
```diff
@@ -922,7 +922,10 @@ def _get_lib_info(*, full_name=None, lib_path=None):
         # dir and Python extension.
         #    e.g.: charms.mycharm.v4.foo -> lib/charms/mycharm/v4/foo.py
         try:
-            charmsdir, importable_charm_name, v_api, libfile = full_name.split(".")
+            charmsdir, charm_name, v_api, libfile = full_name.split(".")
+
+            # the lib full_name includes the charm_name which might not be importable (dashes)
+            importable_charm_name = create_importable_name(charm_name)
```
facundobatista, line 928: This comment and code are fine, but they should be placed outside the try/except structure.

facundobatista (changes requested):
Awesome, thanks for this PR!! I annotated a small detail to change. Thanks!

charmcraft/commands/store/__init__.py
jguedez: Thanks for the review. I updated the commit with the change.

PR #714, opened 2022-03-16
Dependencies handling related bug fixes and improvements

facundobatista:
Specifically...

In `charm_builder.py`:

- Fixed the type for requirement files: all paths should be pathlib.Path, not strings

- Stopped CharmBuilder receiving `python_packages` and `binary_python_packages` that could be lists or None; now they are always lists (that may be empty, it's fine)

- Improved argument parsing:

    - removed the `default=None` in arguments definitions, as it is... well.. the default

    - removed the `metavar` in arguments definitions, it's only used for help text (this is not a user exposed script) and they were being used wrong anyway (the correct use is to better identify the option, and the metavar strings were repeated)

    - for file/dir paths, the argument parsing now returns pathlib.Paths directly

In `commands/build.py`:

- The "venv" directory were added to the prime filter if requirements existed; that was missing the possibility that maybe requirement files didn't exist, but python packages (binary or not) did

### DIFF
--- a/charmcraft/charm_builder.py
+++ b/charmcraft/charm_builder.py
@@ -308,44 +308,38 @@ def _parse_arguments() -> argparse.Namespace:
     parser = argparse.ArgumentParser()
     parser.add_argument(
         "--entrypoint",
-        metavar="filename",
         default="src/charm.py",
+        type=pathlib.Path,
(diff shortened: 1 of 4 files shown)


cmatsuoka (approved):
+1

PR #726, opened 2022-03-22
Show lifecycle step progress (and related fix for always getting the instance logs).

facundobatista:
The main change in this branch is showing progress on each step, so the user may be aware of what goes on. 

As that happens in the instance, I needed to use the new Emitter's `.pause()` functionality. That led to a related fix that had to go into this branch, which is always getting the instance log, on error *and* on execution success (this is because as the emitter is being paused we lose getting the instance output from the terminal... IOW, would never capture the instance logs, not even in verbose, unless we have an error of course).

Also, two output related improvements:

- now the output from the inner charmcraft is directly shown in the screen (the outer one emitter is paused) so for each base we pack it puts the "Charms packed: (...)" message; I added code for this to not happen, so the only "final message" that is shown is from the outer charmcraft, with all the packed bases

- as I was testing different runs/output combinations, I found that the message of the created zip file is better to be shown as intermediate (so it appears with the timestamp in verbose mode)


sergiusens (approved):
Looks good to me, but Claudio has more context so let's wait for him

cmatsuoka (commented):
Craft-cli usage looks correct according to previous discussion, ~~my only point is that lifecycle step progress is not being reported. If you see~~ https://github.com/canonical/charmcraft/blob/eb77143a79ac4966a764fc06517fbf3722e67abb/charmcraft/parts.py#L320 ~~all actions are executed in a singe run so you get a single message "Executing parts lifecycle" instead of breaking it down in steps. It could just be a misleading PR title tho. If this behavior is intended, go ahead and merge, otherwise if you want to show lifecycle step progress you need to execute each action separately and emit a progress message before executing each action. Snapcraft does that in https://github.com/snapcore/snapcraft/blob/7c2ac002781701ab7d789565365b313653097162/snapcraft/parts/parts.py#L120~~

Sorry, this was already fixed. Land it!

PR #728, opened 2022-03-25
parts: update craft-parts and enable craftctl

cmatsuoka:
Use craft-parts 1.4.0 with support to craftctl. Users can now invoke
`craftctl default` from part scriptlets to run the overriden step's
built-in handler.

---
CRAFT-944

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -62,6 +62,7 @@ apps:
       # help git find its stuff
       GIT_TEMPLATE_DIR: $SNAP/git/templates
       GIT_EXEC_PATH: $SNAP/git/git-core
+      PATH: $SNAP/libexec/charmcraft:$PATH
 
 grade: stable
 confinement: classic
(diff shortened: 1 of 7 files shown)


snap/snapcraft.yaml
```diff
@@ -129,6 +130,12 @@ parts:
       usr/bin/git: bin/git
       lib/python3.8/site-packages: lib/
 
+  craftctl:
+    plugin: dump
+    source: libexec
+    organize:
+      craftctl: libexec/charmcraft/craftctl
```
sergiusens, line 137: this doesn't hurt, but I believe this was only in snapcraft to avoid snapcraft not being able to bootstrap itself.

sergiusens: ah, ic the PATH trick

sergiusens (approved):

snap/snapcraft.yaml
cmatsuoka: Otherwise it executes with the wrong interpreter and doesn't find the necessary packages.

Issue #732, opened 2022-03-29
Metadata v1 charms should be classified as invalid


SimonRichardson:
Now the charmstore cut-over has happened, all charms using `charmcraft pack` or `charmcraft build` should not allow metadata v1. Instead, all new deployments should use metadata v2. This will help deprecate the old style by using the new charmcraft tool and we can in the future remove metadata v1 charms that haven't been updated.

lengau:
Since Charmcraft is now writing the metadata field for most charms, I think this is no longer necessary.

PR #735, opened 2022-03-30
Manifest Linter

SimonRichardson:
The following implements a manifest linter that ensures that you don't
have both a metadata series and a manifest base. By doing this, we
can make the Juju side of the code simpler in the future. If all within
the charmhub store and locally deployed charms with Juju only have one
source of truth for the series information it makes determination easy.

Additionally, Juju wants to move away from series terminology and this
will help with that, as in the future we would want to deprecate
metadata.series field.

Fixes https://github.com/canonical/charmcraft/issues/732

### DIFF
--- a/charmcraft/linters.py
+++ b/charmcraft/linters.py
@@ -236,6 +236,46 @@ class JujuMetadata:
         return result
 
 
+class JujuManifest:
+    """Check that the manifest.yaml file is valid YAML if it exists and that
+    if it contains a base, that the metadata YAML does not contain a series
+    field.
+    """
+
+    check_type = CheckType.lint
+    name = "manifest"
+    url = "https://juju.is/docs/sdk/charmcraft-analyze#heading--manifest"
+    text = "Problems found with manifest.yaml file."
+
+     # different result constants
+    Result = namedtuple("Result", "ok errors")(ok=OK, errors=ERRORS)
+
+    def run(self, basedir: pathlib.Path) -> str:
+        """Run the proper verifications."""
+        from charmcraft.manifest import parse_manifest_yaml
+
+        try:
+            manifest = parse_manifest_yaml(basedir)
+        except Exception:
+            # file not found, corrupted, or mandatory "name" not present
+            return self.Result.errors
+
+        if manifest is None or manifest.bases is None or len(manifest.bases) == 0:
+            return self.Result.ok
+
+        try:
+            metadata = parse_metadata_yaml(basedir)
+        except Exception:
+            # metadata.yaml file not found, corrupted, or mandatory "name" not present
+            return self.Result.errors
+
+        if metadata.series is not None and len(metadata.series) > 0:
+            # manifest.yaml contains bases and metadata.yaml file contains a series, only one is required 
+            return self.Result.errors
+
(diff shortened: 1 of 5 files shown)


SimonRichardson:
I've signed the CLA.

sergiusens:
This looks good, but this is also one of those "you broke my build" PRs, is there any communication around this?

facundobatista (changes requested):
The manifest is written by Charmcraft itself, so it doesn't make sense to have a linter that verifies what we just wrote.

I agree with the intention (not having both series and bases), aiming to deprecate "metadata v1", but IMO we should take a different course of action...

First, we could add the necessary tests to ensure that generated information on packing is correct. But note that if the user uses `series` in metadata, the only thing we could do is to stop putting `bases` in `manifest.yaml`, and I'm -0 on producing "some manifests with `bases` and some without", it's inconsistent and may bring issues in other ecosystem components now or in the future. I'm +1 to keep the manifest as is today, and do what I'm describing next...

Second, we should add restrictions on having `series` in `metadata.yaml`. We could enhance the Metadata linter, and produce a warning (not an error). That said, I think a more sensible path should be adding a proper "deprecation notice" saying we will forbid the usage of `series` in metadata in the future, and enforce it in a couple of months (changing Charmcraft's major version number).

SimonRichardson:
> The manifest is written by Charmcraft itself.

You can't guarantee this. I could write a manifest myself and not charmcraft. The solution here is to always validate any manifest yaml file.

sergiusens:
My response got lost, so here I go again.

The part about metadata.yaml coming out of charmcraft shouldn't be a concern, or rather, even if it is originally generated by charmcraft, someone could change it following old documentation. Another scenario for this, is someone importing their project from a prior charming world.

Now, this is a breaking change, as folks with currently working charms will suddenly have their builds broken (for something that works today). We should work out a communication period where we warn folks that this would change.

jnsgruk:
We have the Charmcraft deprecation notices already, could this be something that lands and produces a warning on `pack` for the next couple of releases?

facundobatista:
@SimonRichardson also, thanks for this work! I should have said that in the first comment, sorry :)

facundobatista:
> We have the Charmcraft deprecation notices already, could this be something that lands and produces a warning on `pack` for the next couple of releases?

We totally can issue a deprecation notice, and then enforce the change some months later.

What I don't understood from your question is if you mean to do that when...
- finding a `series` key in metadata.yaml at build time
- finding a combination of "series" and "bases" in manifest.yaml

I stand corrected on my opinion after Sergio and Simon comments, so considering everything I think the plan should be:

- introduce a deprecation notice if `series` found in `metadata.yaml`; some months later, just error out.
- introduce a new Manifest linter that produces a `warning` if finds both `series` and `bases`, with a message specifying that this will be enforced in the future; some months later, when the deprecation above is enforced, this `warning` should become an `error`

We want the double check because, even if the `manifest.yaml` can be manually created, 99% of people will just be using `pack`, and we should fail as earlier as possible.

What do you think?

sergiusens:
We will need to revisit this one, but we will revisit having some form of checks for metadata

PR #740, opened 2022-04-01
Pause the Emitter when asking the user for input (CRAFT-952).

facundobatista:


sergiusens (approved):
Looks good, do you have confirmation that it works on Windows with this? If not, I would like to test that before merging

facundobatista:
Yes, I tested it in windows.

PR #745, opened 2022-04-08
Updated Craft Providers to 1.2.0, which includes the option to do warmup (CRAFT-959).

facundobatista:


sergiusens (approved):

charmcraft/providers/_buildd.py
```diff
@@ -109,5 +107,28 @@ def setup(
         :raises BaseConfigurationError: on other unexpected error.
         """
         super().setup(executor=executor, retry_wait=retry_wait, timeout=timeout)
+        self._setup_charmcraft(executor=executor)
+
+    def warmup(
```
cmatsuoka, line 112: Is this method actually used at runtime or is it provided only for testing convenience? I don't see it being called by the application.

cmatsuoka (approved):

PR #748, opened 2022-04-13
Show better timestamps in commands outputs (CRAFT-873).

### LINKED ISSUE #643: Add exact time to the output of the `revisions` command
With publishing soon being integrated with CI it would be useful to have a more detailed timestamp than just the date, because occasionally there is more than one revision per day, and it's also unclear which timezone is reported, so "the same day" might mean different things in different places.
```
> $ charmcraft revisions loki-k8s
Revision    Version    Created at    Status
2           2          2021-08-30    approved
1           1          2021-08-30    approved
```

facundobatista:
Also improved how errors are shown in the Emitter fixture.

Fixes #643 .

### DIFF
--- a/charmcraft/utils.py
+++ b/charmcraft/utils.py
@@ -240,6 +240,6 @@ def format_timestamp(dt: datetime.datetime) -> str:
 
     Always in UTC.
     """
-    # convert to UTC no matter the timezone `dt` has
+    # convert to UTC from whatever timezone `dt` has
     dtz = datetime.datetime.fromtimestamp(time.mktime(dt.utctimetuple()))
     return dtz.strftime("%Y-%m-%dT%H:%M:%SZ")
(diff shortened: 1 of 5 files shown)


cmatsuoka (approved):

charmcraft/utils.py
```diff
@@ -227,3 +229,17 @@ def confirm_with_user(prompt, default=False) -> bool:
...
+def format_timestamp(dt: datetime.datetime) -> str:
+    """Convert a datetime object (with or without timezone) to a string.
+
+    The format is
+
+        <DATE>T<TIME>Z
+
+    Always in UTC.
+    """
+    # convert to UTC no matter the timezone `dt` has
```
cmatsuoka, line 243: Maybe explain somewhere where our datetimes come from and why we can ignore the timezone?

facundobatista: Mm... we're not ignoring the timezone.

facundobatista: There I improved the comment

PR #753, opened 2022-04-20
Pause the emitter while launching the user-requested shell (CRAFT-871).

facundobatista:
Took the opportunity to fix how flake8 issues were reported.


sergiusens (approved):
nice

PR #755, opened 2022-04-25
The JujuMetadata linter now exposes the problem found, if any (CRAFT-377).

### LINKED ISSUE #641: All linters should produce an error text in function to what they found
If the linter finds a problem in a file, the error text should not be static (JujuMetadata currently only produces `Problems found with metadata.yaml file.`) but dinamically according to what the problem is.

facundobatista:
For that I had to separate the `parse_metadata_yaml` in two (reading and yaml-decoding it, and actually verifying the format), to have control on the issues found.

Fixed #641 .


charmcraft/linters.py
```diff
@@ -215,25 +215,30 @@ class JujuMetadata:
...
-        # no need to verify "name" as it's mandatory in the metadata parsing
-        if metadata.summary and metadata.description:
-            result = self.Result.ok
-        else:
-            result = self.Result.errors
-        return result
+        # check required attributes
+        for field in ["name", "summary", "description"]:
+            if field not in metadata:
+                self.text = f"Error in metadata.yaml: must have a {field!r} attribute."
```
sergiusens, line 239:
```suggestion
        missing_fields = [f for f in ["name", "summary", "description"] if f not in metadata]
        if missing_fields:
            self.text = f"Error in metadata.yaml: fields for {missing_fields} are missing."
```

Maybe humanize missing_fields in self.text.

sergiusens: Line below would be overindented when applying this....

sergiusens (commented):
looks good, a suggestion about showing all potentially missing fields together

charmcraft/linters.py
facundobatista: Improved, thanks!

sergiusens (approved):

Issue #730, opened 2022-03-25
[1.5] No keyring error with --export


heitorPB:
Hello,

I am trying to export a limited auth token to use with Github actions, but I had no success so far:

```bash
$ CHARMCRAFT_AUTH="" charmcraft login --export=gh.auth --charm slurmd --charm slurmdbd --charm slurmctld --charm slurmrestd --permission package-manage
No keyring found to store or retrieve credentials from.
Full execution log: '/home/h/snap/charmcraft/common/cache/charmcraft/log/charmcraft-20220325-145605.991757.log'

$ cat /home/h/snap/charmcraft/common/cache/charmcraft/log/charmcraft-20220325-145605.991757.log
2022-03-25 14:56:05.992 Starting charmcraft version 1.5.0
2022-03-25 14:56:05.992 Raw pre-parsed sysargs: args={'help': False, 'verbose': False, 'quiet': False, 'trace': False, 'project_dir': None} filtered=['login', '--export=gh.auth', '--charm', 'slurmd', '--charm', 'slurmdbd', '--charm', 'slurmctld', '--charm', 'slurmrestd', '--permission', 'package-manage']
2022-03-25 14:56:05.992 General parsed sysargs: command='login' args=['--export=gh.auth', '--charm', 'slurmd', '--charm', 'slurmdbd', '--charm', 'slurmctld', '--charm', 'slurmrestd', '--permission', 'package-manage']
2022-03-25 14:56:05.992 Couldn't find config file '/home/h/projects/slurm-charms/.github/workflows/charmcraft.yaml'
2022-03-25 14:56:05.993 Command parsed sysargs: Namespace(bundle=None, channel=None, charm=['slurmd', 'slurmdbd', 'slurmctld', 'slurmrestd'], export=PosixPath('gh.auth'), permission=['package-manage'], ttl=None)
2022-03-25 14:56:05.993 System details: OSPlatform(system='ubuntu', release='20.04', machine='x86_64'); Environment: None
2022-03-25 14:56:06.069 No keyring found to store or retrieve credentials from.
2022-03-25 14:56:06.071 Traceback (most recent call last):
2022-03-25 14:56:06.071   File "/snap/charmcraft/834/lib/charmcraft/commands/store/store.py", line 179, in __init__
2022-03-25 14:56:06.071     self._client = Client(charmhub_config.api_url, charmhub_config.storage_url)
[...]

PR #760, opened 2022-04-28
login command: --export is ephemeral

sergiusens:
This means no keyring issues

#730


facundobatista (approved):
thanks!

PR #767, opened 2022-05-12
Always validate the entrypoint, no matter its source.

### LINKED ISSUE #761: The `pack` command does not validate the entrypoint unless it's specified
If `pack` is run without using the `--entrypoint` option, even if it defaults to `src/charm.py`, it's not running any validation on it (e.g. that is executable).

facundobatista:
The code is somewhat entangled, because we're currently supporting different sources for entrypoint: configuration, command line option (deprecated), and the default when the developer didn't specify anything.

Besides the several explicit cases that I started to validate, I added one validation for the case of writing the absolute path in the config, which was missing, surely because when the parts lifecycle and plugin were introduced the entrypoint path started to be always relative, when before it was always absolute.

There's a remaining duplicated validation that will disappear when deprecations are enforced. This whole code will be quite more simple after that.

Fixes #761 .


cmatsuoka (approved):
Looks good.

PR #768, opened 2022-05-12
Include an 'actions' dir if present.

### LINKED ISSUE #764: The `actions` directory is not being included by default
When dealing with charm actions, Juju sometimes expect to find scripts to fulfill those actions in a particular directory called `actions` in the charm's root directory. 

From the Charmcraft POV, that directory may or may not exist, but if it does, it must be included in the resulting `.charm` (exactly like `actions.yaml`.

Currently, it's not included by default.

facundobatista:
Fixes #764.


charmcraft/commands/build.py
```diff
@@ -57,6 +57,7 @@
     "LICENSE",
     "icon.svg",
     "README.md",
+    "actions",
```
sergiusens, line 60: from our conversations last week, this list and CHARM_FILES is the list of things I would not do at the plugin level, but instead as an action of packing.

facundobatista: I don't remember that conversation, maybe we were talking about different things? Let's talk again.

That said, maybe we can include this change to fix the problem, and aim for a refactor later?

facundobatista: As we discussed in the call, the approach coded here is the correct.

PR #771, opened 2022-05-19
Fix the tests in the init template

jnsgruk:
The introduction of more realistic container networking in the test suite broke the init template tests. I know we're looking at updating the template anyway, but this should be considered a bug fix in the mean time.

In order for the `charmcraft` tests to pass, the `ops` version in `requirements-dev.txt` also needed bumping.

Perhaps we could introduce a CI test that runs charmcraft init; pip install -r requirements-dev.txt; ./run_tests to ensure this doesn't get broken again?


jnsgruk:
@facundobatista this should be good to go now

facundobatista (approved):
Wonderful, thanks!!

PR #775, opened 2022-05-23
Pass --force to the actual build command

axinojolais:
Without this, passing --force is useless

### DIFF
--- a/charmcraft/commands/build.py
+++ b/charmcraft/commands/build.py
@@ -444,6 +444,9 @@ class Builder:
 
         if self.shell_after:
             cmd.append("--shell-after")
+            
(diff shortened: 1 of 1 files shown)


facundobatista (changes requested):
Hello! Sorry for taking so long to review this! We need a test case. I'll be happy to add it if you can't, thanks!

axinojolais:
> Hello! Sorry for taking so long to review this! We need a test case. I'll be happy to add it if you can't, thanks!

Hello ! Please do then, I'm won't have the time to get back to this for a while.

Thanks !

facundobatista:
Overriden by #789.

PR #779, opened 2022-05-31
Added programmatic outputs to all non store-related commands.

facundobatista:
This is the first of two branches, the second will complete the rest of commands that needs this new feature.

### DIFF
--- a/charmcraft/cmdbase.py
+++ b/charmcraft/cmdbase.py
@@ -39,14 +39,14 @@ class BaseCommand(craft_cli.BaseCommand):
 
     needs_config = False
 
-    def format_content(self, format_, content):
+    def format_content(self, fmt, content):
         """Format the content."""
-        if format_ == JSON_FORMAT:
+        if fmt == JSON_FORMAT:
             return json.dumps(content, indent=4)
         raise ValueError("Specified format not supported.")
 
     def include_format_option(self, parser):
-        """Add the 'format' option to the indicated parser."""
+        """Add the 'format' option to this parser."""
         parser.add_argument(
             "--format",
             choices=[JSON_FORMAT],
--- a/charmcraft/cmdbase.py
+++ b/charmcraft/cmdbase.py
@@ -16,7 +16,11 @@
 
 """Infrastructure for common base commands functionality."""
 
+import json
+
 import craft_cli
+
+JSON_FORMAT = "json"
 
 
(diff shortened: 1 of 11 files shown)


charmcraft/cmdbase.py
```diff
@@ -27,6 +31,24 @@ class BaseCommand(craft_cli.BaseCommand):
...
     The subclass must be declared in the corresponding section of main.COMMAND_GROUPS.
+
+    If the command may produce the result in a programmatic-friendly format, it
+    should call the 'include_format_option' method to properly affect the parser and
+    then emit only one message with the result of the 'format_content' method.
     """
 
     needs_config = False
+
+    def format_content(self, format_, content):
```
sergiusens, line 42: Personally I would prefer fmt to format_ :-)

charmcraft/cmdbase.py
```diff
@@ -27,6 +31,24 @@ class BaseCommand(craft_cli.BaseCommand):
...
+        """Format the content."""
+        if format_ == JSON_FORMAT:
+            return json.dumps(content, indent=4)
+        raise ValueError("Specified format not supported.")
+
+    def include_format_option(self, parser):
+        """Add the 'format' option to the indicated parser."""
```
sergiusens, line 49:
```suggestion
        """Add the 'format' option to this parser."""
```

tests/commands/test_pack.py
```diff
@@ -765,3 +774,67 @@ def test_charm_builder_infrastructure_called(config):
...
+    config.set(type="charm")
+
+    builder_instance_mock = MagicMock()
+    builder_instance_mock.run.return_value = ["mystuff1.charm", "mystuff2.charm"]
+    with patch("charmcraft.commands.build.Builder") as builder_class_mock:
+        builder_class_mock.return_value = builder_instance_mock
+        PackCommand(config).run(args)
+
+    if formatted:
+        emitter.assert_json_output({"charm": ["mystuff1.charm", "mystuff2.charm"]})
```
sergiusens, line 815: After seeing this, maybe "charm" should be renamed to "charms" as it is a list

sergiusens (approved):
looks good, I cannot put a pin on it, but it looks like with some thinking we could make this a craft-cli primitive.

Saome random comments inline

charmcraft/cmdbase.py
facundobatista: good idea

PR #780, opened 2022-06-03
Second batch of commands with added --format=json for programmatic output.

facundobatista:
Added comments in the PR for those formats that do NOT follow the spec (I'd later fix the document if we all agree that the new format is better).
---
CRAFT-1101

### DIFF
--- a/charmcraft/commands/store/__init__.py
+++ b/charmcraft/commands/store/__init__.py
@@ -246,23 +246,38 @@ class WhoamiCommand(BaseCommand):
     """
     )
 
+    def fill_parser(self, parser):
+        """Add own parameters to the general parser."""
+        self.include_format_option(parser)
+
     def run(self, parsed_args):
         """Run the command."""
         store = Store(self.config.charmhub)
         try:
             macaroon_info = store.whoami()
         except CredentialsUnavailable:
-            emit.message("You are not logged in to Charmhub.")
+            if parsed_args.format:
+                info = {"logged": False}
+                emit.message(self.format_content(parsed_args.format, info))
+            else:
+                emit.message("You are not logged in to Charmhub.")
             return
 
-        emit.message(f"name: {macaroon_info.account.name}")
-        emit.message(f"username: {macaroon_info.account.username}")
-        emit.message(f"id: {macaroon_info.account.id}")
+        human_msgs = []
+        prog_info = {"logged": True}
+
+        human_msgs.append(f"name: {macaroon_info.account.name}")
+        prog_info["name"] = macaroon_info.account.name
+        human_msgs.append(f"username: {macaroon_info.account.username}")
+        prog_info["username"] = macaroon_info.account.username
+        human_msgs.append(f"id: {macaroon_info.account.id}")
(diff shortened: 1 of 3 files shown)


charmcraft/commands/store/__init__.py
```diff
@@ -246,41 +246,67 @@ class WhoamiCommand(BaseCommand):
...
-        emit.message(f"id: {macaroon_info.account.id}")
+        human_msgs = []
+        prog_info = {"logged": True}
+
+        human_msgs.append(f"name: {macaroon_info.account.name}")
+        prog_info["name"] = macaroon_info.account.name
+        human_msgs.append(f"username: {macaroon_info.account.username}")
+        prog_info["username"] = macaroon_info.account.username
+        human_msgs.append(f"id: {macaroon_info.account.id}")
+        prog_info["id"] = macaroon_info.account.id
```
facundobatista, line 274: The spec consider the first three attributes as mandatory. That is not useful when not logged, so I made them optional and added a `logged` attribute that the developer can rely on.

charmcraft/commands/store/__init__.py
```diff
@@ -246,41 +246,67 @@ class WhoamiCommand(BaseCommand):
...
+                    pkg_info = []
                     for item in grouped[package_type]:
                         if item.name is not None:
-                            emit.message(f"- name: {item.name}")
+                            human_msgs.append(f"- name: {item.name}")
+                            pkg_info.append({"name": item.name})
                         elif item.id is not None:
-                            emit.message(f"- id: {item.id}")
+                            human_msgs.append(f"- id: {item.id}")
+                            pkg_info.append({"id": item.id})
```
facundobatista, line 296: The spec is sub-defined here. For each package I put a dict instead of just the package name, to indicate which kind of value it holds (the name or the id).

charmcraft/commands/store/__init__.py
```diff
@@ -838,17 +929,32 @@ def run(self, parsed_args):
                         version = revno = resources = (
                             "↑" if release_shown_for_this_track_base else "-"
                         )
+                        prog_version = prog_revno = prog_resources = None
+                        prog_status = "tracking" if release_shown_for_this_track_base else "closed"
```
facundobatista, line 933: Need to specify in the spec how "not released" and "tracked" is expressed.

charmcraft/commands/store/__init__.py
```diff
@@ -816,16 +888,35 @@ def run(self, parsed_args):
...
 
             # bases are shown alphabetically ordered
-            for base in sorted(releases_by_base):
+            sorted_bases = sorted(
+                releases_by_base, key=lambda b: b and (b.name, b.channel, b.architecture)
+            )
+            for base in sorted_bases:
                 releases_by_channel = releases_by_base[base]
-                shown_base = base
+                if base is None:
```
facundobatista, line 907: Make explicit in the doc that 'base' can be None

cmatsuoka (approved):
Thanks, the comments were very helpful to understand the changes.

PR #798, opened 2022-06-29
Removed the long time deprecated 'build' command (CRAFT-1109).

facundobatista:
The Validator, also used from `pack`, no longer makes sense:

- bases_index: moved this control to the Pack command.

- `from`: this was filled with the result of --project-dir, which already verified that the directory existed

- `destructive_mode`, `debug`, `shell`, `shell_after` and `force`: just converted an already bool to bool

Left for next PR: move the `Builder` class and helper function/constants to the `pack.py` file (which would be a huge diff, specially because of tests, but without semantic changes).

### DIFF
--- a/charmcraft/commands/build.py
+++ b/charmcraft/commands/build.py
@@ -14,7 +14,9 @@
 #
 # For further info, check https://github.com/canonical/charmcraft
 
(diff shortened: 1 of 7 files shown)


cmatsuoka (approved):
Looks good.

PR #800, opened 2022-06-30
many: reword docstrings and code comments

sergiusens:

---

CRAFT-697


mr-cal (approved):
+1

PR #801, opened 2022-07-04
Adapted project to the new Craft CLI verbosity levels (CRAFT-1143).

facundobatista:
These are the changes in detail:

- lot of changes from calling `.trace` to calling `.debug` (except that changed one call to use `verbose` and left `trace` in a couple of places where auth headers were shown)

- changed message/intermediate to progress/permanent (and added permanent=True to some calls that we converted to progress some branches ago)

- simplified how charmcraft is started in the managed instance (as we now have `--verbosity` to set any mode

- initiated `emit` in `BRIEF` now

- updated/refreshed all dependencies versions (not only Craft CLI)


charmcraft/commands/build.py
```diff
@@ -416,14 +416,8 @@ def pack_charm_in_instance(
...
-        cmd = ["charmcraft", "pack", "--bases-index", str(bases_index)]
-
-        if emit.get_mode() == EmitterMode.VERBOSE:
-            cmd.append("--verbose")
-        elif emit.get_mode() == EmitterMode.QUIET:
-            cmd.append("--quiet")
-        elif emit.get_mode() == EmitterMode.TRACE:
-            cmd.append("--trace")
+        mode = emit.get_mode().name.lower()
+        cmd = ["charmcraft", "pack", "--bases-index", str(bases_index), f"--verbosity={mode}"]
```
sergio-costas, line 420: This is just a personal opinion, but personally I prefer the explicit way.

At least I would add a comment explaining what is being converted to what...

sergio-costas (changes requested):
With the exception of that comment, it looks good to me.

charmcraft/commands/build.py
facundobatista: Hello! 

Sorry, not following your reasoning. New code is not implicit, but uses the new `--verbosity` option which has the benefit of setting the mode directly without needing to know about or pass different options.

Also, FTR, `--trace` does not exist anymore.

sergio-costas (approved):
LGTM.

cmatsuoka (approved):
Changes look sensible, but I didn't run the updated version to examine the actual output.

PR #813, opened 2022-07-25
Update snap install to use --classic

NucciTheBoss:
Same as #805, but reauthored to pass CLA check


facundobatista (approved):
Wonderful, thanks!

PR #817, opened 2022-08-04
docs(README): Replace obsoleted build command

merkata:
The README refers to the process of building a charm by invoking
`charmcraft build`, but that command is (no longer) available.
Looking at the code, a transition is being made from build.py to pack.py
and the command to build the charm is also `pack`.
I've renamed it to reflect this for the current version (2.0.0).


facundobatista (approved):
Wonderful, thanks!

PR #824, opened 2022-08-11
Improve developer experience

jdkandersson:
- Add vscode settings file to get it to include the lib folder in its Python search path for easier autocomplete and syntax highlighting for any libraries
- Add link to the hello world tutorial in the readme file for newly generated charms
- Add comment above the logger with the juju command that can be used to retrieve the logs once the model is deployed


facundobatista (changes requested):
Hello! Thanks for this PR! Some comments:

- we'll not be adding in this template specific code for each editor people may use, so please remove the vscode related changes. That said, in the future we will have "init profiles", so we may grow an init template for vscode... but surely not in this template and not now, sorry.

- I like the link to the "hello world" documentation, but not sure that the proper place for it is in the produced README. Maybe in the `init` command help message? Anyway, I'm +0 here.

- I like the comment regarding the logs.

Thanks again! Regards,

facundobatista (approved):
Nice comment, thanks!

PR #835, opened 2022-08-18
providers: delete instances via Executor object

mr-cal:
Previously, `charmcraft` bypassed `craft-providers` `Executor` object to delete instances.
This low-level delete call causes issues when `craft-providers` modifies the name of the actual instance  (which occurs [here](https://github.com/canonical/craft-providers/pull/134)).

Now, a build plan is created and instances are deleted one-by-one via the `Executor` object.

(CRAFT-1250)


facundobatista (changes requested):
Hey, great work! 

I have some "big topics" to discuss about this PR:

- don't like "clean" importing from "build", maybe "plan" should be in a third module, like "charmcraft/providers.py"?

- "plan" returns a tuple of stuff, as this is used from several places now and it's kind of "independant", what about providing a better structure (e.g. namedtuple)

- isn't `clean_project_environments` almost equal now in _lxd and _multipass?

Let's talk later, thanks!

mr-cal:
I believe I addressed all your concerns, so this is ready for review again!

facundobatista (approved):
Awesome! Thank you very much!

PR #837, opened 2022-08-22
create_build_plan: refactor

mr-cal:
### Overview
- `create_build_plan` is now in a common location so other commands can access it
- `create_build_plan` returns a list of named tuples
- Added a comprehensive set of unit tests.  Previously, this function was only being tested indirectly.

### Rationale for the new file `charmcraft/providers/providers.py`
As the new `craft-providers` interface is built out, the directory `charmcraft/providers/` will be split into 2 places:
1. `charmcraft/providers.py` for charmcraft-specific code
2. The `craft-providers` library

The code in the `charmcraft/providers/` directory will be split piece-by-piece, in a series of small PRs.

Once all the code is finally split, the file `charmcraft/providers/providers.py` will be moved to `charmcraft/providers.py`. Then, the directory will be removed. Finally, all references to the intermediate file`charmcraft/providers/providers.py` will be replaced with `charmcraft/providers.py`.

(CRAFT-1272)

### DIFF
--- a/charmcraft/commands/build.py
+++ b/charmcraft/commands/build.py
@@ -22,18 +22,18 @@ import os
 import pathlib
 import subprocess
 import zipfile
-from typing import List, Optional, Tuple
+from typing import List, Optional
 
 from craft_cli import emit, CraftError
 
 from charmcraft import env, linters, parts
-from charmcraft.bases import check_if_base_matches_host
 from charmcraft.charm_builder import DISPATCH_FILENAME, HOOKS_DIR
 from charmcraft.config import Base, BasesConfiguration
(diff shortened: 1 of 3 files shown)


facundobatista:
# Once all the code is finally split, the file charmcraft/providers/providers.py will be moved 
# to charmcraft/providers.py. Then, the directory will be removed. Finally, all references 
# to the intermediate filecharmcraft/providers/providers.py will be
# replaced with charmcraft/providers.py

AHHHH WONDERFUL

facundobatista (approved):
Wonderful, thanks!!

PR #842, opened 2022-08-23
build: use namedtuple for build plan

mr-cal:
Minor follow up to https://github.com/canonical/charmcraft/pull/837 to take advantage of the new `namedtuple`.


facundobatista (approved):
+1

PR #852, opened 2022-08-30
requirements: update craft-providers to 1.4.1

mr-cal:
Bump `craft-providers` to `1.4.1` to fix a bug in the truncation of long hostnames when creating an instance.


facundobatista (approved):
+1

PR #854, opened 2022-08-31
Created the profile option for the init command (CRAFT-1317).

facundobatista:
It's optional and currently only accepts its default, "simple", which uses a template very similar to what we had so far, evolved according to [this template from Jon](https://github.com/canonical/template-operator/tree/charmcraft-templates/default).


charmcraft/templates/init-simple/CONTRIBUTING.md.j2
```diff
@@ -0,0 +1,45 @@
...
+### Deploy
+
+```bash
+# Create a model
+juju add-model dev
+# Enable DEBUG logging
+juju model-config logging-config="<root>=INFO;unit=DEBUG"
+# Deploy the charm
+juju deploy ./{{ name }}.charm \
+    --resource httpbin-image=kennethreitz/httpbin \
```
jnsgruk, line 44: My fault, we should probably drop this bit

jnsgruk:
Looks like some of the changes are a commit or two behind the latest commit on [the branch](https://github.com/canonical/template-operator/tree/charmcraft-templates/default) - which explains why you had the flake8 problem I think. Either way this looks good! :D

facundobatista:
As I included `tox` in the requirements file, I updated them, and that lead into a known conflict between black and flake8. Flake8 is the responsible of this, the "internet solution" is to ignore that error, and while trying to do that I found that we had duplicate configuration for it, and fixed the situation.

tests/commands/test_init.py
```diff
@@ -100,24 +105,23 @@ def test_force(tmp_path, config):
 def test_bad_name(config):
     cmd = InitCommand(config)
     with pytest.raises(CraftError):
-        cmd.run(Namespace(name="1234", author="שראלה ישראל", force=False))
+        cmd.run(Namespace(name="1234", author="שראלה ישראל", force=False, profile=DEFAULT_PROFILE))
```
tigarmo, line 108: Seems a bit verbose to have to add the `profile` parameter in unrelated tests, doesn't it? Maybe it's just me
It feels a bit awkward to have to pass something that has a default

tigarmo (approved):
looks good to me, for whatever that's worth ;)

tests/commands/test_init.py
tigarmo: Sure. The only suggestion that comes to me now, if reducing the amount of `profile=DEFAULT_PROFILE` repetition is ever desired, would be to have a `create_parser()` function/fixture that can reuse the `fill_parser()` code that already exists. Then this line could be `cmd.run(parser.parse_args(namespace=Namespace(name="..."))` and the default would be applied
Just some good ol' _Friday bikeshedding_

charmcraft/templates/init-simple/CONTRIBUTING.md.j2
facundobatista: Done.

charmcraft/templates/init-simple/tox.ini.j2
```diff
@@ -0,0 +1,76 @@
...
+    isort {[vars]all_path}
+    black {[vars]all_path}
+
+[testenv:lint]
+description = Check code against coding style standards
+deps =
+    black
+    flake8<5
+    flake8-docstrings
+    flake8-copyright
```
jnsgruk, line 40: I think we should remove this (will comment elsewhere on same topic)

```suggestion
```

charmcraft/templates/init-simple/tox.ini.j2
```diff
@@ -0,0 +1,76 @@
...
+    isort
+commands =
+    isort {[vars]all_path}
+    black {[vars]all_path}
+
+[testenv:lint]
+description = Check code against coding style standards
+deps =
+    black
+    flake8<5
```
jnsgruk, line 38: If we remove this dependency, then the correct version of `flake8` will get pulled in by `pyproject-flake8` and will save us an update later :)

```suggestion
```

charmcraft/templates/init-simple/pyproject.toml.j2
```diff
@@ -0,0 +1,39 @@
...
+# Ignore W503, E501 because using black creates errors with this
+# Ignore D107 Missing docstring in __init__
+ignore = ["W503", "E501", "D107"]
+# D100, D101, D102, D103: Ignore missing docstrings in tests
+per-file-ignores = ["tests/*:D100,D101,D102,D103,D104"]
+docstring-convention = "google"
+# Check for properly formatted copyright header in each file
+copyright-check = "True"
+copyright-author = "{{ author }}"
+copyright-regexp = "Copyright\\s\\d{4}([-,]\\d{4})*\\s+%(author)s"
```
jnsgruk, line 39: Recommend removing this - may not be so applicable outside of Canonical?
```suggestion
```

charmcraft/templates/init-simple/metadata.yaml.j2
```diff
@@ -21,3 +21,5 @@ resources:
   httpbin-image:
     type: oci-image
     description: OCI image for httpbin (kennethreitz/httpbin)
+    # Included for simplicity in integration tests
```
jnsgruk, line 24: This file needs overhauling slightly - please check the template here: https://github.com/canonical/template-operator/blob/charmcraft-templates/default/metadata.yaml

charmcraft/templates/init-simple/charmcraft.yaml.j2
```diff
@@ -1,5 +1,9 @@
+# Copyright {{ year }} {{ author }}
+# See LICENSE file for licensing details.
+
 # Learn more about charmcraft.yaml configuration at:
 # https://juju.is/docs/sdk/charmcraft-config
+#
```
jnsgruk, line 6: Some updated comment wording from Daniele in my branch: https://github.com/canonical/template-operator/blob/charmcraft-templates/default/charmcraft.yaml

charmcraft/templates/init-simple/README.md.j2
```diff
@@ -19,6 +18,9 @@ TODO: Include a link to the default image your charm uses
 
 ## Contributing
 
-Please see the [Juju SDK docs](https://juju.is/docs/sdk) for guidelines
+<!-- TEMPLATE-TODO: Change this URL to be the correct Github path to CONTRIBUTING.md-->
```
jnsgruk, line 21: This file needs overhauling in line with the branch here: https://github.com/canonical/template-operator/blob/charmcraft-templates/default/README.md

jnsgruk (changes requested):
Some more changes to be in line with the proposed template - `config.yaml` also needs updating, and `actions.yaml` removing. 

Can you do another scrub on this branch please and check we're in line? https://github.com/canonical/template-operator/tree/charmcraft-templates/default

PR #855, opened 2022-09-02
Improved init "simple" templates using the correct source

facundobatista:
There are some slight differences between [the templates in the correct source](https://github.com/canonical/template-operator/tree/charmcraft-templates/default) and files after producing a new charm running `init` ([full real diff](https://pastebin.ubuntu.com/p/rjdwsn8TCV/)):

- `charmcraft.yaml`: better indentation
- `.gitignore`: added a newline at the end of last line
- `tox.ini`: used `pdb` in `PYTHONBREAKPOINT`
- several replacements of `{{ year }} {{ author }}` by actual values
- several replacements of `{{ name }}` by the charm name
- several replacements of `{{ class_name }}` by the charm class name
- removed the final `.` in a couple of URLs (the final dot would make the URL invalid)
- fixed some style changes (complained by pep257, black, etc):
    - removed spaces in a blank line
    - improved a couple of docstrings to use the imperative form
    - collapsed some code to be in the same line
- kept from previous code a helping message that informs how to see the charm generated logs using Juju

Also removed the TODO messages from the response of the command. This output may be improved next week.


jnsgruk (approved):
Thanks for the iteration on this @facundobatista :)

charmcraft/templates/init-simple/metadata.yaml.j2
```diff
@@ -1,25 +1,43 @@
...
+# The charm package name, no spaces (required)
+# See https://juju.is/docs/sdk/naming#heading--naming-charms for guidance.
 name: {{ name }}
-display-name: |
-  TEMPLATE-TODO: fill out a display name for the Charmcraft store
+ 
+# The following metadata are human-readable and will be published prominently on Charmhub.
+
+# (Recommended)
+display-name: Charm Template
```
sergiusens, line 11:
```suggestion
display-name: Display Name
```

I would even consider uppercase dash separation of name as useful alternative

charmcraft/templates/init-simple/metadata.yaml.j2
```diff
@@ -1,25 +1,43 @@
...
+  # You may remove this if your charm will run without a workload sidecar container.
   httpbin-image:
     type: oci-image
-    description: OCI image for httpbin (kennethreitz/httpbin)
-    # Included for simplicity in integration tests
+    description: OCI image for httpbin
+    # The upstream-source field is ignored by Juju. It is included here as a reference
+    # so the integration testing suite knows which image to deploy during testing. This field
+    # is also used by the 'canonical/charming-actions' Github action for automated releasing.
     upstream-source: kennethreitz/httpbin
```
sergiusens, line 43: I find it interesting that we are allowing anything to go into the store, these fields should be gated or we won't be able to sustainably grow this in the future

charmcraft/templates/init-simple/metadata.yaml.j2
facundobatista: @jnsgruk your call ^

charmcraft/templates/init-simple/tests/unit/test_charm.py.j2
```diff
@@ -4,42 +4,25 @@
...
-from ops.model import ActiveStatus
+import ops.testing
+from ops.model import ActiveStatus, BlockedStatus, WaitingStatus
 from ops.testing import Harness
 
 from charm import {{ class_name }}
 
+# Enable more accurate simulation of container networking.
+# For more information, see https://juju.is/docs/sdk/testing#heading--simulate-can-connect
+ops.testing.SIMULATE_CAN_CONNECT = True
```
sergiusens, line 16: this should probably be a pytest fixture autoused at the module scope, in the future someone might find themselves scratching their heads if they do something like `tests.integration.test_...` and this setting might get in the way

charmcraft/templates/init-simple/charmcraft.yaml.j2
```diff
@@ -1,14 +1,11 @@
...
-# Learn more about charmcraft.yaml configuration at:
-# https://juju.is/docs/sdk/charmcraft-config
-#
-type: "charm"
+type: charm
 bases:
   - build-on:
-    - name: "ubuntu"
+    - name: ubuntu
       channel: "20.04"
```
sergiusens, line 8: are we not ready for 22.04?

facundobatista: @jnsgruk any particular detail why we're not in Jammy here?

charmcraft/templates/init-simple/tests/unit/test_charm.py.j2
facundobatista: good idea!

facundobatista: actually I changed to make it similar to current test format (put it in the `setUp` of the class)

charmcraft/templates/init-simple/metadata.yaml.j2
facundobatista: Noted, will bring this conversation outside the PR to a wider audience.

charmcraft/templates/init-simple/metadata.yaml.j2
jnsgruk: Personally I think it was fine as is 🤷🏼‍♂️

charmcraft/templates/init-simple/charmcraft.yaml.j2
jnsgruk: Good question, I'd go with jammy.

I've not confirmed 💯 that it'll deploy, but pretty sure it will. Might just be worth checking.

charmcraft/templates/init-simple/metadata.yaml.j2
facundobatista: I'll keep the original. Thanks!

charmcraft/templates/init-simple/charmcraft.yaml.j2
facundobatista: Checking everything, including deployment.

facundobatista: Oh, the template was broken (bad `metadata.yaml`), but after the fix it deploys ok!

FTR, with this diff:
```diff
$ git diff --no-ext-diff
diff --git a/charmcraft/templates/init-simple/charmcraft.yaml.j2 b/charmcraft/templates/init-simple/charmcraft.yaml.j2
index 40eb799d..900d34a2 100644
--- a/charmcraft/templates/init-simple/charmcraft.yaml.j2
+++ b/charmcraft/templates/init-simple/charmcraft.yaml.j2
@@ -5,7 +5,7 @@ type: charm
 bases:
   - build-on:
     - name: ubuntu
-      channel: "20.04"
+      channel: "22.04"
     run-on:
     - name: ubuntu
-      channel: "20.04"
+      channel: "22.04"
diff --git a/charmcraft/templates/init-simple/metadata.yaml.j2 b/charmcraft/templates/init-simple/metadata.yaml.j2
index 6b7eaab5..43a751c5 100644
--- a/charmcraft/templates/init-simple/metadata.yaml.j2
+++ b/charmcraft/templates/init-simple/metadata.yaml.j2
@@ -28,7 +28,7 @@ description: |
 # Your workload’s containers.
 containers:
   httpbin:
-    resource: httpbin
+    resource: httpbin-image
 
 # This field populates the Resources tab on Charmhub.
 resources:
```

PR #857, opened 2022-09-05
Removed two simple XXXs from the code.

facundobatista:
The first is simply gone by putting the import where it corresponds (now that class is defined in the library).

The second is gone because "format in JSON" is well tested now.


mr-cal (approved):
+1!

PR #859, opened 2022-09-07
Added a kubernetes profile to init command (CRAFT-1318).

facundobatista:
Also fixed "simple" template (don't know why these changes didn't get into the previous PR).

### DIFF
--- a/charmcraft/templates/init-kubernetes/.gitignore.j2
+++ b/charmcraft/templates/init-kubernetes/.gitignore.j2
@@ -1,0 +1,1 @@
+../init-simple/.gitignore.j2
--- a/charmcraft/templates/init-kubernetes/metadata.yaml.j2
+++ b/charmcraft/templates/init-kubernetes/metadata.yaml.j2
@@ -1,10 +1,10 @@
 # This file populates the Overview on Charmhub.
-# See https://juju.is/docs/sdk/metadata-reference for a checklist and guidance.
+# See https://juju.is/docs/some-url-to-be-determined/ for a checklist and guidance.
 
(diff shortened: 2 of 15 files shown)


charmcraft/templates/init-kubernetes/metadata.yaml.j2
```diff
@@ -0,0 +1,40 @@
+# This file populates the Overview on Charmhub.
+# See https://juju.is/docs/some-url-to-be-determined/ for a checklist and guidance.
```
jnsgruk, line 2: We should remove this line
```suggestion
```

jnsgruk (approved):

charmcraft/templates/init-kubernetes/metadata.yaml.j2
facundobatista: done!

charmcraft/templates/init-kubernetes/.gitignore.j2
```diff
@@ -0,0 +1 @@
+../init-simple/.gitignore.j2
```
tigarmo, line 1: so does this syntax mean to use the contents of the other file? i.e., is it a shortcut to Jinja2's `include`?

tigarmo (approved):

charmcraft/templates/init-kubernetes/.gitignore.j2
facundobatista: simpler! Jinja2 just sees the file, and for the human checking it it's clearly the same file...

PR #860, opened 2022-09-09
Improved names used in charmbuilder script to reflect better the reality.

facundobatista:


charmcraft/charm_builder.py
```diff
@@ -365,13 +365,13 @@ def _parse_arguments() -> argparse.Namespace:
         help="The charm entry point. Default is 'src/charm.py'.",
     )
     parser.add_argument(
-        "--charmdir",
+        "--builddir",
```
mr-cal, line 368: How much will this affect the end user?

I wonder if it is worth having a deprecation notice or notifying users to use the new arguments.  The `--builddir` parameter keeping the same name but changing its meaning could be confusing.

mr-cal: Discussed offline - `charm_builder` isn't called by the user.

facundobatista: no final user involved! :) this is deep in the execution layers of `pack` command

mr-cal (approved):
+1

Issue #809, opened 2022-07-18
Reactive plugin doesn't allow for other `charm` cli arguments.


addyess:
The `charm-tools` builder `charm` allows for many build switches, some which are necessary to produce reproducible stable builds.

It doesn't seem that this plugin allows for any such options to be specified. 

For instance, one cannot override the `layer-index` used during the build with this cli switch, so only the default is available.
```bash
  -i LAYER_INDEX, --layer-index LAYER_INDEX
                        One or more index URLs used to look up layers,
                        separated by commas. Can include the token DEFAULT,
                        which will be replaced by the default index
                        (https://juju.github.io/layer-index/). E.g.:
                        https://my-site.com/index/,DEFAULT
```

https://github.com/canonical/charmcraft/blob/a2205cb1270220c03d244814468ea812e984c01c/charmcraft/reactive_plugin.py#L159-L165

it would be beneficial to reactive charms for `charmcraft` to build with these switches primarily as CLI switches or at a minimum `charmcraft.yaml` configuration options.

fnordahl:
Bump, especially now that juju/charm-tools#620 has landed we would like to be able to use the reactive plugin with binary wheels to fully make use of the power charmcraft + Launchpad builders + Charmhub gives us in terms of maintaining charms across multiple series and architectures.

PR #863, opened 2022-09-14
reactive_plugin: Support arguments to charm build command

fnordahl:
This patch adds the ability to pass arguments to the `charm build` command when using the reactive plugin.

Example charmcraft.yaml:
```yaml
type: charm

parts:
  charm:
    plugin: reactive
    reactive-charm-build-arguments:
      - --binary-wheels-from-source
    build-packages:
      - tox
      - git
      - python3-dev
      - libffi-dev
    source: src/
    build-snaps:
      - charm/latest/edge
    build-environment:
      - CHARM_INTERFACES_DIR: $CRAFT_PROJECT_DIR/interfaces/
      - CHARM_LAYERS_DIR: $CRAFT_PROJECT_DIR/layers/
```
Fixes: #809

### DIFF
--- a/charmcraft/reactive_plugin.py
+++ b/charmcraft/reactive_plugin.py
@@ -19,7 +19,7 @@ import shlex
 import subprocess
 import sys
 from pathlib import Path
-from typing import Any, Dict, List, Optional, Set
+from typing import Any, Dict, List, Optional, Set, cast
 
 from craft_cli import emit
 from craft_parts import plugins
@@ -30,6 +30,7 @@ class ReactivePluginProperties(plugins.PluginProperties, plugins.PluginModel):
     """Properties used to pack reactive charms using charm-tools."""
 
     source: str
+    reactive_charm_build_arguments: Optional[List[str]] = []
 
     @classmethod
     def unmarshal(cls, data: Dict[str, Any]):
@@ -119,6 +120,8 @@ class ReactivePlugin(plugins.Plugin):
 
     def get_build_commands(self) -> List[str]:
         """Return a list of commands to run during the build step."""
+        options = cast(ReactivePluginProperties, self._options)
+
         command = [
             sys.executable,
             "-I",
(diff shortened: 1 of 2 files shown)


charmcraft/reactive_plugin.py
```diff
@@ -113,19 +115,24 @@ def get_build_environment(self) -> Dict[str, str]:
...
+        options = cast(ReactivePluginProperties, self._options)
+
         command = [
             sys.executable,
             "-I",
             __file__,
             self._part_info.project_name,
             str(self._part_info.part_build_dir),
             str(self._part_info.part_install_dir),
+            json.dumps(options.reactive_charm_build_arguments),
```
facundobatista, line 127: We can't just use whatever the user specified in the YAML, otherwise we would be sending "together" options that should be different parts of the command. 

IOW, if the user specified in the YAML...
```
    reactive-charm-build-arguments:
      - --layer-index 3
```
, we should NOT execute charm with `["charm", "build", "--layer-index 3", "-o", builddir]`, but with `["charm", "build", "--layer-index", "3", "-o", builddir]`.

You can use `shlex.split` to separate the strings from the YAML to be suitable for the command line usage.

facundobatista (changes requested):
Hello! Thanks for this work!

I like the general idea of having an optional list of arguments that the user can freely setup so the `charm build` command is modified. 

And I like how this is handled in this PR, with the exception of two details, one is commented inline, and this one is more generic: it looks weird  to use JSON to send "an obscure set of parameters" to the script; we can simply use the fact that they are all "command line parameters traveling".

I have a proposal to improve that situation, let me know what do you think: use all extra arguments like normal arguments to the script, something like:
```
command = [
    sys.executable,
    "-I",
    __file__,
    self._part_info.project_name,
    str(self._part_info.part_build_dir),
    str(self._part_info.part_install_dir),
]
command.extend(options.reactive_charm_build_arguments))
```

...and then in the script just get all parameters to the build...

```
    returncode = build(
        charm_name=sys.argv[1],
        build_dir=Path(sys.argv[2]),
        install_dir=Path(sys.argv[3]),
        build_extra_arguments=sys.argv[4:],
    )
```

(the command building you did in `build` will remain exactly as you proposed).

Thanks again!!

facundobatista:
@swalladge right now I'm working on a reboot of the documentation for this plugin and the overall charmcraft configuration... it should be ready in the next days

PR #865, opened 2022-09-15
Finally isolated charm_builder.py as a generic plugin script (CRAFT-1032).

### LINKED ISSUE #750: Convert charm_builder.py to a regular external script
Currently it uses Craft CLI, but that is because legacy reasons (it was integrated to the rest of Charmcraft code before), and it relies in its functionality but as it's run as a script, it's not used (e.g., showing progress). Furthermore, the added timestamp dirties the logs.

So instead of using the Emitter, we should set up a logging infrastructure that will just show the message to the terminal (no timestamp, no log level, etc), and transform the 14 `emit.trace` calls in `logger.debug` and the 5 `emit.progress` calls in `logger.info`. By default the logger level should be INFO, but if the script is run with a `--trace` flag it should change to DEBUG.

When preparing the script to run, in the `charm` plugin, we should use that flag in some cases:
- if Charmcraft is run in quiet or normal mode, `charm_builder.py` should be called without the flag (nothing will be seen, its INFO lines will be saved into the logs)
- if Charmcraft is run in verbose mode, `charm_builder.py` should be called without the flag (its INFO lines will be seen and saved into the logs)
- if Charmcraft is run in trace mode, `charm_builder.py` should be called **with** the flag (its INFO and DEBUG lines will be seen and saved into the logs)

We also should stop raising CraftError as it just means a crash.

facundobatista:
This will stop its execution from showing some weirdness in the logs and improve the general structure of the code (stop mixing code run from the plugin with the rest of Charmcraft).

After a long talk with Claudio yesterday we decided to convert all outputs to plain prints to stdout, again, as a regular plugin script.

Also converted the CraftError exceptions to Runtime ones (fully isolation from Craft CLI) and improved the message.

Fixes #750 .

### DIFF
--- a/charmcraft/charm_builder.py
+++ b/charmcraft/charm_builder.py
@@ -14,7 +14,10 @@
 #
 # For further info, check https://github.com/canonical/charmcraft
 
-"""The charm package builder."""
+"""The charm package builder.
+
+This is a standalone script run by the Charm Plugin.
+"""
 
 import argparse
 import errno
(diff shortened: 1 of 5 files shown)


mr-cal (approved):
Looks good!

PR #866, opened 2022-09-15
reactive_plugin: Fix charm tools version validation

### LINKED ISSUE #864: reactive plugin charm tools version validation code is britle
The charm tools may soon remove the now redundant legacy charm store code, when that happens the output from the version command may no longer use the format expected here:
https://github.com/canonical/charmcraft/blob/dbf82a0fbaa6633248bb213d3943603c1e7c50a6/charmcraft/reactive_plugin.py#L64-L65

I would recommend either using the `--format json` argument or looking for the `charm-tools` string in the output rather than relying on there being a newline character there.

Ref: https://github.com/juju/charm-tools/blob/17812f0e00ad634f8440053d43d1a5370a71fa50/charmtools/version.py#L15-L25

fnordahl:
The current charm-tools version validation code is britle and may soon break as charm-tools evolves.

Fixes #864

### DIFF
--- a/charmcraft/reactive_plugin.py
+++ b/charmcraft/reactive_plugin.py
@@ -14,6 +14,7 @@
 
 """Charmcraft's reactive plugin for craft-parts."""
 
+import json
 import shlex
 import subprocess
 import sys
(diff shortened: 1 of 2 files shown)


facundobatista (approved):
Hello! Thanks for this work, it's awesome!

PR #867, opened 2022-09-15
ci: add GHA for spread

sergiusens:
---
CRAFT-1368

### DIFF
--- a/.github/workflows/spread.yaml
+++ b/.github/workflows/spread.yaml
@@ -1,0 +1,46 @@
+name: Spread
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+  schedule:
+    - cron: "0 0 */2 * *"
+
+jobs:
+  snap-build:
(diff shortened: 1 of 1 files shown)


.github/workflows/spread.yaml
```diff
@@ -0,0 +1,46 @@
...
+  pull_request:
+  push:
+    branches:
+      - main
+  schedule:
+    - cron: "0 0 */2 * *"
+
+jobs:
+  snap-build:
+    runs-on: ubuntu-20.04
```
facundobatista, line 12: Why not jammy?

facundobatista (approved):
+1, but please check comment.

.github/workflows/spread.yaml
sergiusens: I would rather not, this is a GHA provided instance and 20.04 is stable; this is mostly a platform for us to run snapcraft which creates a lxd container with the appropriate target for the base.

PR #869, opened 2022-09-16
reactive_plugin: Properly check returncode from charm tool

### LINKED ISSUE #804: The reactive plugin is not correctly checking if `charm` failed
The `charm` tool used by the reactive plugin has a complex return code specification.

While Charmcraft is correctly checking that the return code >= 200 is an error, it's failing to verify that <0 is also an error.

Furthermore, it should log in debug the return code from the execution (no matter which one it is).

All said above applies for both `charm` calls.

fnordahl:
Also always log the returncode.

Fixes #804

### DIFF
--- a/charmcraft/reactive_plugin.py
+++ b/charmcraft/reactive_plugin.py
@@ -21,7 +21,6 @@ import sys
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Set, cast
 
-from craft_cli import emit
 from craft_parts import plugins
 from craft_parts.errors import PluginEnvironmentValidationError
 
@@ -76,7 +75,7 @@ class ReactivePluginEnvironmentValidator(plugins.validator.PluginEnvironmentVali
                     reason=f"invalid charm tools version {version_data}",
                 )
             tools_version = f"{tool_name} {version_data[tool_name]['version']} ({version_data[tool_name]['git']})"
-            emit.debug(f"found {tools_version}")
+            print(f"found {tools_version}")
         except ValueError as err:
             raise PluginEnvironmentValidationError(
                 part_name=self._part_name,
--- a/charmcraft/reactive_plugin.py
+++ b/charmcraft/reactive_plugin.py
@@ -139,6 +139,26 @@ class ReactivePlugin(plugins.Plugin):
         return [" ".join(shlex.quote(i) for i in command)]
 
 
+def run_charm_tool(args: List[str]):
+    """Run the charm tool, log and check exit code."""
+    result_classification = "SUCCESS"
+    exc = None
+
+    print(f"charm tool execution command={args}")
+    try:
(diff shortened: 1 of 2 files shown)


fnordahl:
Putting this up as a draft as I need to fix the tests.

charmcraft/reactive_plugin.py
```diff
@@ -179,17 +200,24 @@ def build(
...
+        return call_error.returncode
     finally:
         charm_build_dir.unlink()
 
     return 0
 
 
 if __name__ == "__main__":
+    logpath = get_managed_environment_log_path()
+    emit.init(
```
facundobatista, line 214: Do not use Emitter from the plugin's subprocess. We actually went away from that model in `charm_builder.py` a few days ago, towards a more generic "plugin subprocess" model as always have been used by Snapcraft.

Just `print()` all the progress/debug info that you need, it will be handled properly.

charmcraft/reactive_plugin.py
```diff
@@ -140,6 +141,26 @@ def get_build_commands(self) -> List[str]:
...
+        completed_process = subprocess.run(args, check=True)
+    except subprocess.CalledProcessError as call_error:
+        exc = call_error
+        if call_error.returncode < 100 or call_error.returncode >= 200:
+            result_classification = "ERROR"
+            raise
+        else:
+            result_classification = "WARNING"
+    finally:
+        emit.debug(
```
facundobatista, line 159: This would need to be converted to a `print` call. 

Also, please do this small change: print the command to be executed at the beginning of this function, which gets us two benefits: *this* message will be shorter and simpler, and we will have the command that is being executed in case that it takes too long (or "forever").

charmcraft/reactive_plugin.py
```diff
@@ -140,6 +141,26 @@ def get_build_commands(self) -> List[str]:
...
+    exc = None
+
+    try:
+        completed_process = subprocess.run(args, check=True)
+    except subprocess.CalledProcessError as call_error:
+        exc = call_error
+        if call_error.returncode < 100 or call_error.returncode >= 200:
+            result_classification = "ERROR"
+            raise
+        else:
```
facundobatista, line 156: this "else" is not needed, as the "block of True" of the if will not allow these lines to continue.

facundobatista (changes requested):
Awesome, thanks for this work! I annotated a couple of suggestions. Thanks again!

charmcraft/reactive_plugin.py
fnordahl: Ah, that makes it much simpler, thank you for sharing that development. :+1:

charmcraft/reactive_plugin.py
fnordahl: :+1:

facundobatista (approved):
Wonderful, thanks!!

PR #871, opened 2022-09-20
Support both image id and digest when uploading oci-image resources (CRAFT-1324).

### LINKED ISSUE #853: The upload-resource command should accept images id or digests
(all this is for OCI images, file resources are fine)

Currently is mixed: accepts "ids" (and no point in checking remotely in that case) but the docs say "digest".

It should:

- accept "id" or digest, these two should be fine: `charmcraft upload-resource mychar myresource ...`
    `--image=3e42dd4e79c7`
    `--image=sha256:9bc34afe08ca30ef179404318cdebe6430ceda35a4ebe4b67d10789b17bdf7c4`
- the distinction between id or digest should happen checking if it starts with `sha`
- if it's an ID
    - just upload from local (no point in checking remotely); however inform it in "debug"
    - get the image info from local dockerd as it currently does
- if it's a Digest
    - do the "check remote, if not there check local" dance as today
    - get the image info from local dockerd searching *all* the images (`/images/json`) and checking which one (if any) has the indicated digest in one of the `RepoDigests` items
- command documentation should be improved to reflect this
- forum post about uploading resource should also be improved

facundobatista:
Also be very explicit on which one is used when. It results on being quicker if the id is used (no check on remote registry) and better messages to the user. Which were improved also: not all needed to be permanent, and actually ended in error if the image to upload is not found and nothing can be done.

Fixes #853.

### DIFF
--- a/charmcraft/commands/store/registry.py
+++ b/charmcraft/commands/store/registry.py
@@ -290,12 +290,12 @@ class LocalDockerdInterface:
     def __init__(self):
         self.session = requests_unixsocket.Session()
 
-    def get_image_info(self, digest: str) -> Union[dict, None]:
-        """Get the info for a specific image.
+    def get_image_info_from_id(self, image_id: str) -> Union[dict, None]:
+        """Get the info for a specific image using its id.
 
-        Returns None to flag that the requested digest was not found by any reason.
+        Returns None to flag that the requested id was not found by any reason.
         """
-        url = self.dockerd_socket_baseurl + "/images/{}/json".format(digest)
+        url = self.dockerd_socket_baseurl + "/images/{}/json".format(image_id)
         try:
             response = self.session.get(url)
         except requests.exceptions.ConnectionError:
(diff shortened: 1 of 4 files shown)


charmcraft/commands/store/registry.py
```diff
@@ -290,12 +290,12 @@ class LocalDockerdInterface:
     def __init__(self):
         self.session = requests_unixsocket.Session()
 
-    def get_image_info(self, digest: str) -> Union[dict, None]:
-        """Get the info for a specific image.
+    def get_image_info_from_id(self, image_id: str) -> Union[dict, None]:
+        """Get the info for a specific image using its id.
 
-        Returns None to flag that the requested digest was not found by any reason.
+        Returns None to flag that the requested id was not found by any reason.
```
mr-cal, line 296: pedantic: `found by any reason` -> `found for any reason` 

The same sentence is in the next method's docstring too.

charmcraft/commands/store/registry.py
```diff
@@ -313,9 +313,31 @@ def get_image_info(self, digest: str) -> Union[dict, None]:
...
+        try:
+            response = self.session.get(url)
+        except requests.exceptions.ConnectionError:
+            emit.debug(
+                "Cannot connect to /var/run/docker.sock , please ensure dockerd is running.",
+            )
+            return
+
+        if response.status_code != 200:
+            emit.debug(f"Bad response when validation local image: {response.status_code}")
```
mr-cal, line 331: pedantic: `when validation local image` -> `when validating local image`

mr-cal (approved):
Looks great!

charmcraft/commands/store/registry.py
facundobatista: thanks! that was a silly mistake

charmcraft/commands/store/registry.py
facundobatista: typo! thanks

PR #874, opened 2022-09-21
Use incremental delays (with very small first ones) during the polling of upload status (CRAFT-894).

### LINKED ISSUE #79: Implement a backoff when polling the Store for an upload status
Currently we're polling every a fixed time, for ever.

We need to:
- implement a slight backoff algorithm
- implement a timeout after N retries (big one, after snapcraft experience)

Actually, it doesn't need to be a real "algorithm", it's better to manually design the user experience.

Proposal: A first test real quick, in case the Store fast approved the charm, with two more tests very fast too. Then some more tests going multi-second. Then keep trying every 3s for a **total timeout** of one minute:

```
>>> poll_delays = [.2, .3, .5, 1, 1, 1, 2] + [3] * 18
>>> sum(poll_delays)
60.0
```

facundobatista:
Fixes #79.

Also sneaked in a change of `imp` to `importlib` library (as the former is deprecated).

### DIFF
--- a/charmcraft/commands/store/store.py
+++ b/charmcraft/commands/store/store.py
@@ -1,4 +1,4 @@
-# Copyright 2020-2021 Canonical Ltd.
+# Copyright 2020-2022 Canonical Ltd.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


mr-cal (approved):
Very clean - nice work!

PR #889, opened 2022-10-06
spread: add large external charm pack tests

sergiusens:
These are cronned to run on Wednesdays and Sundays at midnight.

Their only purpose is to run "pack".

### DIFF
--- a/tests/spread/charms/k8s-operator/task.yaml
+++ b/tests/spread/charms/k8s-operator/task.yaml
@@ -1,0 +1,29 @@
+summary: pack external k8s operator charms
+manual: true
+systems:
+  - ubuntu-22.04-64
+kill-timeout: 30m
+
+environment:
+  CHARM/alertmanager: https://github.com/canonical/alertmanager-k8s-operator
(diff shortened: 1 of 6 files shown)


tests/spread/charms/k8s-operator/task.yaml
```diff
@@ -0,0 +1,29 @@
...
+  CHARM/postgresql: https://github.com/canonical/postgresql-k8s-operator
+  CHARM/prometheus: https://github.com/canonical/prometheus-k8s-operator
+  CHARM/traefik: https://github.com/canonical/traefik-k8s-operator
+  CHARM/mattermost: https://git.launchpad.net/charm-k8s-mattermost
+  CHARM/discourse: https://git.launchpad.net/charm-k8s-discourse
+
+prepare: |
+  git clone --depth=1 "${CHARM}" charm
+
+restore: |
```
tigarmo, line 20: could all these `restore`s eventually become a single `restore-each`? They are all very similar

sergiusens: there's more to into this suite that might not fit, let's finish the work and find the common grounds

PR #892, opened 2022-10-06
(internal) Remove CharmcraftBuilddBaseConfiguration and use BuilddBase directly (CRAFT-1278, CRAFT-1356).

facundobatista:


mr-cal (approved):
+1

PR #894, opened 2022-10-06
(internal) Moved user confirmation on provider installation (CRAFT-1401)

facundobatista:


charmcraft/providers/_multipass.py
```diff
@@ -43,6 +42,12 @@ class MultipassProvider(Provider):
     :param multipass: Optional Multipass client to use.
     """
 
+    name = "LXD"
```
mr-cal, line 45:
```suggestion
    name = "Multipass"
```

mr-cal (approved):

PR #903, opened 2022-10-11
(charmhub) Support including collaborations when listing registered names (CRAFT-878).

### LINKED ISSUE #546: The `names` command should allow listing those "collaborated" charms
Currently it only lists those charms *owned* by the developer.

We could add an `--include-collaborations` option (or something shorter, but remember we have bash autocompletion, so...) to also list the charms where the developer is a collaborator.

When this option is used a new column will be shown, exposing the Publisher's display name.

Implementation-wise, the Charmhub endpoint needs to be modified this way: `/v1/charm?include-collaborations=true`

facundobatista:
Fixes #546.

### DIFF
--- a/charmcraft/commands/store/__init__.py
+++ b/charmcraft/commands/store/__init__.py
@@ -401,6 +401,10 @@ class ListNamesCommand(BaseCommand):
         seen by any user, while `private` items are only for you and the
         other accounts with permission to collaborate on that specific name.
 
+        The --include-collaborations option can be included to also list those
+        names you collaborate with; in that case the publisher will be included
+        in the output.
+
         Listing names will take you through login if needed.
     """
     )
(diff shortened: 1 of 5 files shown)


charmcraft/commands/store/__init__.py
```diff
@@ -409,26 +413,36 @@ class ListNamesCommand(BaseCommand):
     def fill_parser(self, parser):
         """Add own parameters to the general parser."""
         self.include_format_option(parser)
+        parser.add_argument(
+            "--include-collaborations",
+            action="store_true",
+            help="Include those names that you are a collaborator of",
```
mr-cal, line 419: `Include those names that you are a collaborator` -> `Include names that you are a collaborator of` 

(`Include the names you are a collaborator of` reads well too)

mr-cal (approved):
Nice work!

charmcraft/commands/store/__init__.py
facundobatista: Awesome, thanks! Improved.

PR #916, opened 2022-10-27
(charmhub) Provide a better error to the user on a login attempt when already having credentials (CRAFT-1269)

### LINKED ISSUE #833: Provide a better error description and suggest solution for the "double login" case
Craft Store does not allow to "login" if it finds credentials in the keyring. Currently the received error is confusing:

```
craft-store error: Credentials found for 'charmcraft' on 'api.charmhub.io'.    
```

Furthermore, when the user reads "credentials found" they move on and issue another store related command that may also fail because the credentials are expired! so they try to login again and ... (loop)

We should provide a better error and possible resolution:

```
Cannot login because credentials were found in your system (which may be no longer valid, though)
Recommended resolution: Please logout before login again 
```

(and the original error should appear in the logs only)

facundobatista:
Fixes #833.

### DIFF
--- a/charmcraft/commands/store/store.py
+++ b/charmcraft/commands/store/store.py
@@ -25,6 +25,7 @@ from functools import wraps
 import craft_store
 from craft_cli import emit, CraftError
 from craft_store import attenuations, endpoints
+from craft_store.errors import CredentialsAlreadyAvailable
 from dateutil import parser
 
 from charmcraft.commands.store.client import Client, ALTERNATE_AUTH_ENV_VAR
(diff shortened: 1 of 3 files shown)


charmcraft/commands/store/store.py
```diff
@@ -207,7 +208,15 @@ def login(self, permissions=None, ttl=None, charms=None, bundles=None, channels=
...
             kwargs["packages"] = packages
 
-        return self._client.login(**kwargs)
+        try:
+            credentials = self._client.login(**kwargs)
+        except CredentialsAlreadyAvailable as exc:
+            raise CraftError(
+                "Cannot login because credentials were found in your system "
+                "(which may be no longer valid, though).",
+                resolution="Please logout before login again.",
```
mr-cal, line 217: `Please logout before login again.` -> `Please logout before logging in again.` 

(or to keep "login" in simple tense: `Please logout first, then login again.`)

mr-cal (approved):
Looks great!
I added 1 minor grammar suggestion

PR #917, opened 2022-10-27
Show (with value hidden) the auth environment variable in system details.

### LINKED ISSUE #914: The presence of CHARMCRAFT_AUTH should be logged
Currently we're logging all CHARMCRAFT_* environment variables, but specifically excluding `CHARMCRAFT_AUTH` to not leak secrets.

However, it's very useful, for debugging purposes, know that `CHARMCRAFT_AUTH` is present or not, so we should log it with a different value, something like...

```
System details: (...); Environment: DESKTOP_SESSION='plasma', CHARMCRAFT_AUTH='<hidden>', ...
```

facundobatista:
Fixes #914.

### DIFF
--- a/charmcraft/main.py
+++ b/charmcraft/main.py
@@ -102,7 +102,7 @@ EXTRA_ENVIRONMENT = ("DESKTOP_SESSION", "XDG_CURRENT_DESKTOP")
 
 def _get_system_details():
     """Produce details about the system."""
-    # prepare the useful environment variables: all CHARMCRAFT* (hidding AUTH keys)
+    # prepare the useful environment variables: all CHARMCRAFT* (hiding AUTH keys)
     # and desktop/session
     useful_env = {
         name: value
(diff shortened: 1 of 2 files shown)


tests/test_main.py
```diff
@@ -334,8 +334,8 @@ def test_systemdetails_charmcraft_environment():
     )
 
 
-def test_systemdetails_no_auth():
-    """System details specifically excluding secrets."""
+def test_systemdetails_hidden_auth():
+    """System details specifically hidding secrets."""
```
mr-cal, line 338: same as above

charmcraft/main.py
```diff
@@ -102,16 +102,20 @@
 
 def _get_system_details():
     """Produce details about the system."""
-    # prepare the useful environment variables: all CHARMCRAFT* (except AUTH keys)
+    # prepare the useful environment variables: all CHARMCRAFT* (hidding AUTH keys)
```
mr-cal, line 105: `hidding` -> `hiding`

mr-cal (approved):
Looks great!

charmcraft/main.py
facundobatista: Oh, thanks!

PR #925, opened 2022-11-08
providers: set lxd project to 'charmcraft'

mr-cal:
Set the project to `charmcraft` instead of the default project `default`.  

This fixes a regression from migrating to the new craft-providers interface.


facundobatista (approved):
Thanks!

PR #926, opened 2022-11-08
Made pack & upload tests more robust.

facundobatista:
Avoided to get "last" revision because the uploaded one may not be the latest one (because other concurrently running test may have uploaded other one).


tests/spread/store/bundle-upload-and-release/task.yaml
```diff
@@ -35,18 +35,23 @@ execute: |
   test -f $BUNDLE_DEFAULT_NAME.zip
 
   # upload
```
mr-cal, line 37: pedantic: `# upload` -> `# upload and get uploaded revision` to match the other file

mr-cal (approved):

PR #927, opened 2022-11-08
Improved code to be more robust also when DST is active.

facundobatista:


mr-cal (approved):
Unit tests pass in my timezone!

PR #932, opened 2022-11-25
feat: replace flake8 + plugins and isort with ruff

jnsgruk:
We have previously standardised on using `flake8` with a number of plugins, and `pyproject-flake8` to enable us to configure all of our linting tools in one place (`pyproject.toml`)

Over the past months, there have been two major releases of `flake8`, and both have caused various issues with the various plugins and their dependencies, leading to instability in CI pipelines.

I've been following [ruff](https://github.com/charliermarsh/ruff) for a while. It's a Python linter, written in Rust, that combines a lot of the rules we were using from these other plugins. It's actively maintained, and all under one project, reducing the complexity of the dependency tree a little. It's got adoption from some pretty large projects like FastAPI and pydantic.

Proposing here for discussion.

### DIFF
--- a/charmcraft/templates/init-kubernetes/tests/unit/test_charm.py.j2
+++ b/charmcraft/templates/init-kubernetes/tests/unit/test_charm.py.j2
@@ -4,10 +4,10 @@
 import unittest
 
 import ops.testing
+from charm import {{ class_name }}
 from ops.model import ActiveStatus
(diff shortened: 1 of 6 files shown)


facundobatista (approved):
Wonderful, thanks! Removed a couple of extra blank lines so "current flake8 from outside" does not complain.

PR #933, opened 2022-11-28
Update the charmcraft status --format

### LINKED ISSUE #794: Clarify `charmcraft status --help` text for `--format`
#784 implements structured output of `charmcraft status`, but the help text on it is a bit misleading:
```
base ❯ time charmcraft status --help                               
Usage:
    charmcraft status [options] <name>

...

Options:
...
             --format:  Produce the result formatted as a JSON string
```

The above makes it seem like `charmcraft status charmname --format` will export in JSON, but really it wants `--format json`.  Easy to fix by changing the option string to say it needs an extra arg, one of `json` or whatever else is available.

rahulsinghcanonical:
This commit add the option to format the output data in yaml format and also updates the text for charmcraft status --help for --format option.

Fixes #794

### DIFF
--- a/charmcraft/cmdbase.py
+++ b/charmcraft/cmdbase.py
@@ -21,6 +21,7 @@ import json
 import craft_cli
 
 JSON_FORMAT = "json"
+FORMAT_HELP_STR = "Produce the result in the specified format (currently only 'json')"
 
 
 class BaseCommand(craft_cli.BaseCommand):
@@ -50,5 +51,5 @@ class BaseCommand(craft_cli.BaseCommand):
         parser.add_argument(
             "--format",
             choices=[JSON_FORMAT],
-            help="Produce the result formatted as a JSON string",
+            help=FORMAT_HELP_STR,
         )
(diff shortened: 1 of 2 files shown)


facundobatista:
Hello @rahulsinghcanonical , thanks for the PR!

+1 to fix the help message... I'd remove the uncertainty of using the "like", though, something more assertive like `Produce the formatted result in the specified format (currently only "json")`.

Regarding the YAML support, we should discuss first if we want that. What's the use case? Is JSON not enough to have it consumed programmatically? You should open an Issue if you want to push for this functionality, but I'd remove it from this PR.

Thanks!!!

rahulsinghcanonical:
@facundobatista Thank you for the review. Yes, I agree we should have a separate commits for bug fix and having a new feature (if there is use case for it). I have removed the `yaml` formatting code from this commit.

charmcraft/cmdbase.py
```diff
@@ -21,6 +21,7 @@
 import craft_cli
 
 JSON_FORMAT = "json"
+FORMAT_HELP_STR = "Produce the formatted result in the specified format (currently only 'json')"
```
mr-cal, line 24: I think this reads better as `Produce the result in the specified format (currently only 'json')`

rahulsinghcanonical: @mr-cal yes, I agree. Updated the string.

facundobatista (approved):
+1, thanks!!

PR #943, opened 2022-12-07
Move dependency updates to Renovate

lengau:
This sets up Renovate as a GitHub action (using [Renovatebot's Github Action](https://github.com/renovatebot/github-action)) running every 2 hours.

The bot is currently configured (.github/renovate.json) create PRs on the weekend (up to 5 concurrent PRs) and will only automerge between 01:00 and 07:00 UTC. Patch level updates will be auto-merged, as will minor updates to dev dependencies.

This won't update any GitHub workflows the way it would if we used the GitHub app (see [this job](https://github.com/lengau/charmcraft/actions/runs/3638761284/jobs/6141344859) for an example of the warning output it gives re: workflows).


lengau:
Cancelling this because we might already have the renovate app so we won't need to use the action. New PR incoming to test that.

PR #952, opened 2022-12-12
Include requirements-* files in Renovate config

lengau:


lengau:
Opened in wrong spot sorry

PR #953, opened 2022-12-12
Update Renovate config to match all requirements-*.txt files.

lengau:
Based on [this comment](https://github.com/canonical/charmcraft/pull/948#pullrequestreview-1213392441)

If we later need to ignore specific requirements files we can add an `ignorePaths` array to the base config.


facundobatista (approved):
Let's try this!

PR #955, opened 2022-12-13
Refine renovate configuration

lengau:
* ~~`requirements-dev.txt` is considered dev dependencies~~ Removed after further discussion
* Patches get automerged at top priority
* GitHub actions get a higher priority than most dependencies
* de-prioritise major updates below minor dev updates
* de-prioritise major dev dependencies as the lowest priority

Fixes CRAFT-1533


.github/renovate.json5
```diff
@@ -3,31 +3,51 @@
   extends: ["config:base"],
   labels: ["dependencies"],  // For convenient searching in GitHub
   pip_requirements: {
-    fileMatch: ["(^|/)requirements([\\w-]*)\\.txt$"]
+    // Match every requirements file except requirements-dev.txt, which is handled with a regex manager below.
+    fileMatch: ["(^|\\/)requirements(-(d|[^d].*|de|d[^e].*|de[^v].*|dev\\w+|))?\\.txt$"]
```
tigarmo, line 7: this looks curious, is there a particular reason for this regex structure? Would something like `^(?!requirements-dev\.txt)requirements(-\S+)?\.txt$` work?

.github/renovate.json5
```diff
@@ -36,6 +56,14 @@
       matchStrings: [
         "# renovate: datasource=(?<datasource>\\S+)\n\\s+(?<depName>.*?)==(?<currentValue>.*?)\\n"
       ]
+    },
+    {
+      // Make everything in requirements-dev.txt a dev dependency.
```
tigarmo, line 61: I'm not sure I follow: can you explain what it means to be a "dev dependency" in the context of this configuration?

.github/renovate.json5
tigarmo: Oh I see. Maybe it's possible to use `fileMatch` in conjunction with [ignorePaths](https://docs.renovatebot.com/configuration-options/#ignorepaths) to explicitly exclude requirements-dev.txt?

.github/renovate.json5
lengau: In this context it means this file itself will be updated with development dependencies, not with regular dependencies. This was the best compromise I could think of to still group updates to things like black/pylint/flake8/etc. without creating a list of packages that are development dependencies somewhere.

In practice, this file will be updated primarily in PRs made through the rules on lines 31-39 and 45-50. By de-prioritising dev dependency PRs, I *believe* I've made it so higher priority PRs can scoop up updates in this file (e.g. updating requests in `requirements.txt` should still update it here too), but I'm not entirely sure on that.

The intention, though, is that minor updates to dev tools get auto-merged, and likewise minor updates to regular dependencies get auto-merged into `requirements-dev` but not `requirements` (though once we're through with the initial barrage of updates these PRs should be created at the same time), so we'll be using the newer minor updates on our own machines before they get included in the standard package.

I hope that makes sense. I feel like some of this probably needs to get added to the comments here, but I'm not sure what's useful and what's just yammering, so I'd love some help turning this braindump into a usable explanation.

.github/renovate.json5
lengau: Good thought! I thought I'd tried that yesterday before creating this monster of a regex, but plugging it in now didn't make the validator yell at me so I'll get back to you on it.

.github/renovate.json5
```diff
@@ -41,10 +60,11 @@
   timezone: "Etc/UTC",
   automergeSchedule: "after 1 am and before 7 am",
   schedule: "every weekend",
-  prConcurrentLimit: 5, // No more than 5 open PRs at a time.
+  //prConcurrentLimit: 5, // No more than 5 open PRs at a time.
+  prConcurrentLimit: 0, // <-- TEMPORARY!
```
tigarmo, line 64: what does this do?

lengau: Unlimited concurrent PRs. It appears to be recommended for when you start if there's a lot of stuff out of date and you have the CI capacity.

tigarmo (approved):
thanks!

PR #957, opened 2022-12-15
Manually refreshed dependencies.

facundobatista:


lengau (approved):

requirements-dev.txt
```diff
@@ -1,72 +1,80 @@
...
 six==1.16.0
-snap-helpers==0.2.0
+snap-helpers==0.3.1
 snowballstemmer==2.2.0
-tabulate==0.8.10
+tabulate==0.9.0
 toml==0.10.2
 tomli==2.0.1
-tox==3.25.1
+tox==4.0.11
```
lengau, line 73: Tox 4.0 has a fairly different configuration setup - I'm willing to bet it's what's causing the test failures. If we pin it <4.0 for the moment it should fix the breakages for now.

lengau (changes requested):

facundobatista:
Yeah, found that tox is the problem. But it's super weird, it does not complain about the "old tox.ini", it just... makes `codespell` fail?  https://pastebin.canonical.com/p/RRZhSndfrm/

tigarmo:
That command line `codespell /tmp/newproy/.` is kind of awkward, maybe `--skip /tmp/newproy/.tox` is not skipping `/tmp/newproy/./.tox`?

facundobatista:
That's part of the reason, yes, see fix: https://github.com/canonical/charmcraft/pull/958/files

That said, `FOO` and `FOO/.` should be the same path. 

But it was not that it misunderstood that, it kind of broke the command parsing and made codespell to ignore the rest of the parameters.

Will try to use this fix and new tox.

charmcraft/templates/init-simple/tox.ini.j2
```diff
@@ -40,7 +40,7 @@ deps =
 commands =
     # uncomment the following line if this charm owns a lib
     # codespell {[vars]lib_path}
-    codespell {toxinidir}/. \
+    codespell {toxinidir} \
```
lengau, line 43: The `tox.ini` files in `init-machine` and `init-kubernetes` probably need this too.

lengau: Ope, never mind - Jon pointed out that those are symlinks, which my tooling has changed the way it shows so I didn't notice.

facundobatista: No, they are symlinks to the "simple" one

mr-cal (approved):

PR #958, opened 2022-12-15
Update tox.ini.j2

jnsgruk:
Fixes an issue with codespell and later versions of tox

### DIFF
--- a/charmcraft/templates/init-simple/tox.ini.j2
+++ b/charmcraft/templates/init-simple/tox.ini.j2
@@ -40,7 +40,7 @@ deps =
 commands =
     # uncomment the following line if this charm owns a lib
     # codespell {[vars]lib_path}
-    codespell {toxinidir}/. \
+    codespell {toxinidir} \
               --skip {toxinidir}/.git \
               --skip {toxinidir}/.tox \
               --skip {toxinidir}/build \


charmcraft/templates/init-simple/tox.ini.j2
```diff
@@ -40,7 +40,7 @@ deps =
 commands =
     # uncomment the following line if this charm owns a lib
     # codespell {[vars]lib_path}
-    codespell {toxinidir}/. \
+    codespell {toxinidir} \
```
lengau, line 43: This probably needs to be changed in `init-machine/tox.ini.j2` and `init-kubernetes/tox.ini.j2` as well.

jnsgruk: Pretty sure they're just symlinks to here, no?

lengau:
@jnsgruk Ope, you're right! Sorry about that - the new version of PyCharm isn't italicising symlinks anymore so I missed that.

charmcraft/templates/init-simple/tox.ini.j2
lengau: Ope, you're right! Sorry about that - the new version of PyCharm isn't italicising symlinks anymore so I missed that.

lengau (approved):

facundobatista:
This was superseded by https://github.com/canonical/charmcraft/pull/957

PR #990, opened 2023-01-18
Move charm library helper functions to their own modules.

lengau:
Fixes CRAFT-1557

### DIFF
--- a/tests/commands/test_store_charmlibs.py
+++ b/tests/commands/test_store_charmlibs.py
@@ -17,6 +17,7 @@
 """Tests for store helpers commands (code in store/charmlibs.py)."""
 import hashlib
 import pathlib
+import re
 import sys
 
 import pytest
@@ -265,7 +266,7 @@ class TestGetLibInfo:
     def test_invalid_metadata_field(self, charm_path, fields, match_template):
         """Some metadata field is not really valid."""
         test_path = self.create_lib(**fields)
-        match = match_template.format(str(test_path))
+        match = match_template.format(re.escape(str(test_path)))
         with pytest.raises(CraftError, match=match):
             get_lib_info(lib_path=test_path)
 
@@ -303,7 +304,7 @@ class TestGetLibInfo:
     def test_api_patch_invalid(self, lib_fields, match_template, field):
         """Invalid values for API or patch versions."""
         test_path = self.create_lib(**lib_fields)
-        match = match_template.format(str(test_path), field)
+        match = match_template.format(re.escape(str(test_path)), field)
         with pytest.raises(CraftError, match=match):
             get_lib_info(lib_path=test_path)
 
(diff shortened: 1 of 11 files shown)


tests/commands/test_store_charmlibs.py
```diff
@@ -0,0 +1,331 @@
...
+        [
+            pytest.param("LIBID = 99", id="Non-string"),
+            pytest.param("LIBID = 'moño'", id="Non-ASCII"),
+            pytest.param("LIBID = ''", id="empty"),
+        ],
+    )
+    def test_libid_invalid(self, metadata_id):
+        """The ID is not really a string."""
+        test_path = self.create_lib(metadata_id=metadata_id)
+        match = r"Library .+ metadata field LIBID must be a non-empty ASCII string\."
```
facundobatista, line 329: Here and in other tests, do not build generic strings when checking for errors or other messages exposed to the user.

For example in this case, if we introduce a bug in the code and the error message is `Library foobarBROKEN metadata field LIBID must be a non-empty ASCII string.` this still will pass, and it shouldn't.

Yes, the chances are low, but we have to be extra careful with messages shown to the final user.

tests/commands/test_store_charmlibs.py
```diff
@@ -0,0 +1,331 @@
...
+    )
+    def test_api_patch_invalid(self, lib_fields, match):
+        """Invalid values for API or patch versions."""
+        test_path = self.create_lib(**lib_fields)
+        with pytest.raises(CraftError, match=match):
+            get_lib_info(lib_path=test_path)
+
+    def test_metadata_api_different_path_api(
+        self,
+    ):
```
facundobatista, line 311: This should be one line.

tests/commands/test_store_charmlibs.py
```diff
@@ -0,0 +1,331 @@
...
+            pytest.param(b"", ValueError, "invalid literal", id="empty-bytes"),
+        ],
+    )
+    def test_exceptions(self, raw_value, error_class, error_regex):
+        """Test various exceptions that can be passed through."""
+        with pytest.raises(error_class, match=error_regex):
+            get_positive_int(raw_value)
+
+
+class TestGetLibInfo:
```
facundobatista, line 56: We're trying to move away from class-structured tests in pytest. Is there specific functionality that you need here?

facundobatista (changes requested):
Thanks for this work!

There's a problem with this PR, though. PRs should be small, so we can review them... "PRs that move stuff around" are intrinsically big, but it's fine because you don't have to review them

This PR is a) huge  b) has a lot of changes. While most of the changes I agree, there are some changes that need to be reverted (see inline comments).

Please, propose a PR with *just moving code around* (big, but no real review needed), and then after that, if you want to do changes, more PRs (small and reviewable).

Thanks!!

facundobatista:
There quite some changes that are welcomed there, and I would expect in another PR (e.g. using parametrization to merge tests that are too similar).

That said, of course *some* changes are ok in the "move stuff around", like different imports, or removing underscores because of transforming something private in public (so it can live in other module).

Thanks!

lengau:
@facundobatista Agreed. I'm going to close this PR and redo it as two.

PR #998, opened 2023-01-23
refactor: isolate the charm lib internals extractor

facundobatista:
This is a preparation PR before changing the charmlib parsing to use AST. It does not change any code, just move it around, including tests (some of which are collapsed using parametrize).


lengau (approved):
Looks good to me!

PR #1000, opened 2023-01-24
refactor: get_lib_internals now uses AST (CRAFT-1558)

facundobatista:


charmcraft/commands/store/charmlibs.py
```diff
@@ -75,62 +67,69 @@ def create_charm_name_from_importable(charm_name: str) -> str:
...
+
+    # walk the AST "first layer", find assignments to the key fields and validate them
+    metadata = {}
+    for node in ast.iter_child_nodes(tree):
+        if isinstance(node, ast.Assign):
+            for target in node.targets:
+                if target.id in FIELDS:
+                    validator, error_msg = FIELDS[target.id]
+                    if not isinstance(node.value, ast.Constant):
+                        raise CraftError(error_msg.format(str(lib_path)))
```
lengau, line 102: Is this something we might want to consider a different message for? E.g. if a user thinks they're being very smart and makes the file say:

```python
LIBAPI = int(open('VERSION').read().split('.')[0])
```

we might want to let them know that the message needs to be a constant rather than just evaluating to an `int >= 0`.

lengau (approved):
Looks good!

charmcraft/commands/store/charmlibs.py
```diff
@@ -75,62 +67,69 @@ def create_charm_name_from_importable(charm_name: str) -> str:
...
-    except (ValueError, UnicodeDecodeError):
-        raise CraftError(bad_libid_msg.format(str(lib_path)))
-    if not libid or not isinstance(libid, str):
-        raise CraftError(bad_libid_msg.format(str(lib_path)))
-
-    content = lib_path.read_text()
+    # hash the file excluding those lines where appear keys used for version control
+    metadata_vcs_fields = (b"LIBAPI", b"LIBPATCH", b"LIBID")
+    hasher = hashlib.sha256()
+    with lib_path.open("rb") as fh:
```
lengau, line 123: Pretty sure the Windows test failure is occurring because of this. I *think* forcing `newline="\n"` should do the trick.

facundobatista: It was the newline. But `newline` argument does not work in binary mode, so I explicitly replaced/fixed it and added a comment about having the same hash for the same file in all platforms.

facundobatista: Thanks!!

charmcraft/commands/store/charmlibs.py
facundobatista: Nice idea.

facundobatista: Actually, it was more complicated that sounded :p 

The problem is when the user does `LIBPATCH = -1` (which may be more probable than a twisted expression as you used), and that ends up in an error like "you need to put a constant". That's because `-1` is actually an unary operation...

So I just improved the message to be explicit about needing a constant, and it should work.

Thanks!!!

lengau (approved):
Looks great!

Issue #1002, opened 2023-01-26
Ensure the charm libraries related commands work correctly when command not issued in project's root dir


facundobatista:
Most of the current charm libraries infrastructure depend on being standing in the root directory of the project, as the "lib" directory has a specific structure that is validated or used to detect libraries.

We need to ensure that all commands for charm libraries (fetch, publish, etc.) work correctly when standing somewhere else and `--project-dir` being used.

PR #1003, opened 2023-01-26
pack: include the charm libraries dependencies (CRAFT-1559)

facundobatista:
There are two situations where these dependencies need to be used:

- in the Builder when it decides if the `venv` directory needs to be primed

- in charm_builder when dependencies are actually handled (hashed, installed, etc.)

I created a helper to collect all the dependencies, used from those two places. Located it in the charmlibs module for convenience (initially I thought putting it in `charmcraft/utils` but hit a circular import).

Alternatively, these dependencies could have been collected where other dependencies are defined or dynamically included: in the CharmPluginProperties class. This will make "pydeps" management much more similar to standard python packages or requirements files, but there are several drawbacks of this path:

- a new `charm_libs_deps` or similar key would have need to be included, even if that key should not be exposed to the user as a real config key

- an extra validator would need to be added to collect all PYDEPS values, which would be called *twice* (when config is validated, and when the plugin is actually run), so there is no drawback in calling the collection twice in the current model

- furthermore, that validator would have been executed (and all PYDEPS collected) when the config is validated *for any command* (e.g. `charmcraft names`)

Also, I include here these needed changes in the charm libraries:

- included PYDEPS in the parsing of a library internals

- added a `root` parameter to the `get_libs_from_tree` function; notably, this function relied on being on the project's root directory, which is suspicious, so I opened #1002 to ensure everything is fine (or fix it)

- moved the `emit` usage out of `get_libs_from_tree` as it is now used also from the charm_builder.py, called from the plugin, as a standard subprocess that does not use the Craft CLI infrastructure.

### DIFF
--- a/charmcraft/charm_builder.py
+++ b/charmcraft/charm_builder.py
@@ -30,6 +30,7 @@ import subprocess
 from typing import List
 
 from charmcraft import instrum
+from charmcraft.commands.store.charmlibs import collect_charmlib_pydeps
 from charmcraft.env import get_charm_builder_metrics_path
 from charmcraft.jujuignore import JujuIgnore, default_juju_ignore
 from charmcraft.utils import make_executable
@@ -85,6 +86,9 @@ class CharmBuilder:
         self.requirement_paths = requirements
         self.ignore_rules = self._load_juju_ignore()
         self.ignore_rules.extend_patterns([f"/{STAGING_VENV_DIRNAME}"])
+
+        self.charmlib_deps = collect_charmlib_pydeps(builddir)
+        print("Collected charmlib dependencies:", self.charmlib_deps)
 
     def build_charm(self) -> None:
         """Build the charm."""
@@ -230,6 +234,7 @@ class CharmBuilder:
             all_deps.append(req_file.read_text())
         all_deps.extend(self.binary_python_packages)
         all_deps.extend(self.python_packages)
+        all_deps.extend(self.charmlib_deps)
         deps_mashup = "".join(map(repr, all_deps))
         deps_hash = hashlib.sha1(deps_mashup.encode("utf8")).hexdigest()
         return deps_hash
@@ -265,10 +270,21 @@ class CharmBuilder:
                     cmd.append("--requirement={}".format(reqspath))  # the dependencies file(s)
                 _process_run(cmd)
 
+            if self.charmlib_deps:
+                # install charmlibs python dependencies
+                cmd = [pip_cmd, "install", "--upgrade", "--no-binary", ":all:"]  # base command
+                cmd.extend(self.charmlib_deps)  # the python packages to install
+                _process_run(cmd)
+
     def handle_dependencies(self):
         """Handle from-directory and virtualenv dependencies."""
         print("Handling dependencies")
-        if not (self.requirement_paths or self.binary_python_packages or self.python_packages):
+        if not (
+            self.requirement_paths
+            or self.binary_python_packages
+            or self.python_packages
+            or self.charmlib_deps
+        ):
             print("No dependencies to handle")
             return
 
--- a/charmcraft/commands/store/charmlibs.py
+++ b/charmcraft/commands/store/charmlibs.py
@@ -75,7 +75,7 @@ def get_lib_internals(lib_path: pathlib.Path) -> LibInternals:
     There are two kind of metadata fields, simple constant fields and PYDEPS. The simple
     constant fields are also the mandatory ones: LIBAPI, LIBPATCH and LIBID.
     """
-    content = lib_path.read_text()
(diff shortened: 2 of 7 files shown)


charmcraft/commands/store/charmlibs.py
```diff
@@ -244,6 +270,15 @@ def get_libs_from_tree(charm_name=None):
                 for libfile in sorted(v_dir.glob("*.py")):
                     local_libs_data.append(get_lib_info(lib_path=libfile))
 
-    found_libs = [lib_data.full_name for lib_data in local_libs_data]
-    emit.debug(f"Libraries found under {str(base_dir)!r}: {found_libs}")
+    os.chdir(current_directory)
```
tigarmo, line 273: might want to put this in a `finally:` block (not sure if the functions called before this can raise)

charmcraft/commands/store/charmlibs.py
```diff
@@ -244,6 +270,15 @@ def get_libs_from_tree(charm_name=None):
...
     return local_libs_data
+
+
+def collect_charmlib_pydeps(basedir: pathlib.Path):
```
tigarmo, line 277:
```suggestion
def collect_charmlib_pydeps(basedir: pathlib.Path) -> Set[str]:
```

charmcraft/commands/store/charmlibs.py
```diff
@@ -220,7 +242,7 @@ def get_lib_info(*, full_name=None, lib_path=None):
     )
 
 
-def get_libs_from_tree(charm_name=None):
+def get_libs_from_tree(charm_name: Optional[str] = None, root: Optional[pathlib.Path] = None):
```
tigarmo, line 245: add the return type here too

tests/commands/test_store_charmlibs.py
```diff
@@ -350,4 +376,114 @@ def test_getlibinternals_libid_bad_value(tmp_path, value, monkeypatch):
...
 
 
+def test_getlibinternals_pydeps_complex(tmp_path, monkeypatch):
+    """The PYDEPS field can be multiline, unicode, different quotes."""
+    monkeypatch.chdir(tmp_path)
+    test_path = _create_lib(
+        pydeps="""PYDEPS = [
+        'foo',
+        "bar",
+        "moño",
```
tigarmo, line 386: good thinking testing non-ascii text

tests/test_charm_builder.py
```diff
@@ -1155,9 +1165,11 @@ def mock_build_charm(self):
     fake_argv += ["-r" "reqs1.txt", "--requirement", "reqs2.txt"]
     with patch.object(sys, "argv", fake_argv):
         with patch("charmcraft.charm_builder.CharmBuilder.build_charm", new=mock_build_charm):
-            with pytest.raises(SystemExit) as raised:
-                charm_builder.main()
+            with patch("charmcraft.charm_builder.collect_charmlib_pydeps") as mock_collect_pydeps:
```
tigarmo, line 1168: out of curiosity: have you considered using the `mocker` fixture instead of "accumulating" context managers? Might make it easier to read.

charmcraft/commands/store/charmlibs.py
facundobatista: No, it should be a `print`, as `charm_builder.py` is run as any other script and its stdout collected (see the `prints` in the rest of the file too)

charmcraft/charm_builder.py
```diff
@@ -86,6 +87,9 @@ def __init__(
         self.ignore_rules = self._load_juju_ignore()
         self.ignore_rules.extend_patterns([f"/{STAGING_VENV_DIRNAME}"])
 
+        self.charmlib_deps = collect_charmlib_pydeps(builddir)
+        print("Collected charmlib dependencies:", self.charmlib_deps)
```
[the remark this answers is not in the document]
facundobatista: No, it should be a `print`, as `charm_builder.py` is run as any other script and its stdout collected (see the `prints` in the rest of the file too)

charmcraft/commands/store/charmlibs.py
facundobatista: good idea

charmcraft/commands/store/charmlibs.py
facundobatista: +1

charmcraft/commands/store/charmlibs.py
facundobatista: +1

tests/commands/test_store_charmlibs.py
facundobatista: Yes! I don't remember any package with non-ascii chars but AFAIK they are not forbidden.

tigarmo (approved):
nice, ty

PR #1010, opened 2023-01-30
Update tox.ini in `charmcraft init` template

carlcsaposs-canonical:


charmcraft/templates/init-simple/tox.ini.j2
```diff
@@ -2,27 +2,27 @@
...
+env_list = format, lint, unit
 
 [vars]
-src_path = {toxinidir}/src/
-tst_path = {toxinidir}/tests/
-;lib_path = {toxinidir}/lib/charms/operator_name_with_underscores
-all_path = {[vars]src_path} {[vars]tst_path} 
+src_path = {tox_root}/src
+tests_path = {tox_root}/tests
+;lib_path = {tox_root}/lib/charms/operator_name_with_underscores
```
lengau, line 12: Is this intentionally commented out? If so it should probably be noted to uncomment it down in the lint testenv

charmcraft/templates/init-simple/tox.ini.j2
```diff
@@ -2,27 +2,27 @@
 # See LICENSE file for licensing details.
 
 [tox]
-skipsdist=True
+no_package = True
 skip_missing_interpreters = True
-envlist = fmt, lint, unit
+env_list = format, lint, unit
```
lengau, line 7: These updates will cause tox to fail if the user has tox <= 4.0. I the user installed tox through apt on any current version of Ubuntu that'll be the case, but you can have tox >=3.8.0 (available in apt on focal and later) handle that with a bootstrap venv:

```suggestion
env_list = format, lint, unit
minversion=4.0.0
requires = 
    tox>=4.0.0
```

lengau (approved):

jnsgruk (approved):
Nice, thanks @carlcsaposs-canonical!

PR #1011, opened 2023-01-31
Add `__init__.py` files to test subdirectories in template

carlcsaposs-canonical:
Enables Python files in tests/integration/ and tests/unit/ to import other Python files in that directory

After removing the top-level directory from PYTHONPATH in 0333f4f529a0aace62a88865876e613eda01df11, Python modules in tests/ cannot import from each other without `__init__.py`

An `__init__.py` was not added to tests/ since that would cause pytest to add the top-level repository directory to PYTHONPATH. (More info: https://doc.pytest.org/en/latest/explanation/goodpractices.html#choosing-an-import-mode)

### DIFF
--- a/charmcraft/templates/init-kubernetes/tests/integration/__init__.py.j2
+++ b/charmcraft/templates/init-kubernetes/tests/integration/__init__.py.j2
@@ -1,0 +1,1 @@
+../../../init-simple/tests/integration/__init__.py.j2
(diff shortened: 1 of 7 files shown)


facundobatista:
> However, I'm not sure if the default `charmcraft init` project should add `__init__.py` files to make test modules importable from other test modules—it seems like that behavior might be confusing & discourage the use of pytest fixtures. In other words, using a separate Python module for helper functions is one possible approach but I don't know if it should be the default approach.

Hello! Trying to see if I understood correctly: you say that you're not sure about the changes included in this PR? That's why it's in Draft? Thanks!

sergiusens:
We probably DO want __init__.py's for pyright (and I believe mypy) to go in and scan the package

PR #1025, opened 2023-02-15
deps: update python-apt

lengau:
Uses the latest source packages for python3-apt from each Ubuntu base's updates repository.


cmatsuoka (approved):
Thanks! We also need to change those in other tools that use craft-parts to have consistent versions across all the tools.

PR #1032, opened 2023-02-23
Don't reference non-existent --all argument in publish-lib help

benhoyt:
I noticed the `--all` argument doesn't actually exist on the `publish-lib` command, contradicting the help output. Fix that.


benhoyt:
FWIW I just signed the Canonical CLA ... after 2.5 years at Canonical. :-)

lengau (approved):
Thanks for this!

I've updated this from main, which should catch the signed CLA now as the tests re-run.

PR #1033, opened 2023-02-23
Transform the few remaining tests inside classes to regular functions.

facundobatista:


tests/test_reactive_plugin.py
```diff
@@ -53,101 +53,114 @@ def broken_charm_exe(tmp_path):
...
+def plugin(tmp_path, plugin_properties):
+    project_dirs = craft_parts.ProjectDirs(work_dir=tmp_path)
+    spec = {
+        "plugin": "reactive",
+        "source": str(tmp_path),
+        "reactive-charm-build-arguments": [
+            "--charm-argument",
+            "--charm-argument-with argument",
+        ],
+    }
```
facundobatista, line 80: I don't like duplicating the `spec` here, but couldn't find other way of doing this.

tigarmo: is just because of `tmp_path`? You could have a `spec` fixture that receives `tmp_path` and returns this dict

tigarmo (approved):

tests/test_reactive_plugin.py
facundobatista: good idea, did that!

PR #1035, opened 2023-02-27
Allow use charmcraft non-build commands without set bases

syu-w:
This allowed user use command like login/logout, register, names, version, etc. without fully configured charmcraft.yaml.

For command need to use "bases", set "needs_config_bases = True" in the command class.


charmcraft/cmdbase.py
```diff
@@ -39,6 +41,7 @@ class BaseCommand(craft_cli.BaseCommand):
     """
 
     needs_config = False
+    needs_config_bases = False
```
tigarmo, line 44: I understand that this implementation is what was agreed on, but to me it's awkward that we are adding "escape hatches" to bypass issues on the project file. Wouldn't it make more sense to not load the project at all, if the command doesn't need it?
My fear is this: what if tomorrow another restriction is added to charmcraft.yaml? We are going to add another `needs_config_...` to handle it.
@sergiusens can we discuss alternatives here? Or should we just do this quick-and-dirty?

syu-w: The things is current implementations all check configuration files before getting commands. If we change it to check the basic command first and then decide whether to load the corresponding configuration file, this problem can be eliminated.

facundobatista: @tigarmo What do you mean "not load the project at all"?

The config needs to be parsed, the command needs to be chosen, and at that point decisions can be made. It's not super super cheap, but if you'll end up in an error, timings are not *that* important.

Regarding the how and when to take the decision, it can be done two ways I think:
- take the decision in the config structure using commands attributes (this is the one proposed here); it looks very organized, but may not scale well if different commands have different needs.
- take the decision in the command itself; for example, the first thing the pack command could do is `if self.config.bases is None: raise CraftError("Bases is needed for packing")`; it's not centralized in one place, but it's very flexible.

charmcraft/config.py
```diff
@@ -381,21 +382,34 @@ def expand_short_form_bases(cls, bases: List[Dict[str, Any]]) -> None:
...
+                    raise CraftError(format_pydantic_errors(pydantic_errors))
+
+                # Print invalid base configuration error and continues.
+                # The invalid base will be cleared.
+                # Command should check it if want to use "bases" in "charmcraft.yaml".
+                print(format_pydantic_errors(pydantic_errors), file=sys.stderr)
+                base.clear()
+                base["build-on"] = []
+                base["run-on"] = []
+                continue
```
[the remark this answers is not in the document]
syu-w: OK

charmcraft/config.py
```diff
@@ -404,29 +418,23 @@ def unmarshal(cls, obj: Dict[str, Any], project: Project):
             # base configurations.  Doing it here rather than a Union
             # type will simplify user facing errors.
             bases = obj.get("bases")
```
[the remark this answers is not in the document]
syu-w: We don't need to expand the bases?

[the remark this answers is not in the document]
syu-w: Oh, I understand what you mean.

charmcraft/cmdbase.py
```diff
@@ -53,3 +56,22 @@ def include_format_option(self, parser):
             choices=[JSON_FORMAT],
             help=FORMAT_HELP_STR,
         )
+
+    def _check_config(self, config_file: bool = False, bases: bool = False) -> None:
+        """Check if requested config content exists.
```
mr-cal, line 61: can you add the exceptions to the docstring? e.g.:
```
:raises ArgumentParsingError: <description>
:raises CraftError: <description>
```

charmcraft/cmdbase.py
```diff
@@ -53,3 +56,22 @@ def include_format_option(self, parser):
...
+
+        - config_file: request a valid 'charmcraft.yaml' file.
```
mr-cal, line 63: The meaning of "request" isn't clear for either of these variables, it hints that something is returned. 

I would reword this to something like "If true, check if a valid charmcraft.yaml file exists."

charmcraft/cmdbase.py
```diff
@@ -38,8 +43,6 @@ class BaseCommand(craft_cli.BaseCommand):
     then emit only one message with the result of the 'format_content' method.
     """
 
-    needs_config = False
```
mr-cal, line 41: Can you remove this from the docstring?

mr-cal (changes requested):
The implementation looks good.  It looks like what we discussed offline.

I added a few comments, primarily about docstrings.  

Also, can you look at `charmcraft/commands/analyze.py`?  I see that a class that still sets the `needs_config` variable.

mr-cal (approved):
Looks great, thanks!

PR #1037, opened 2023-02-28
charm libraries: allow anonymous fetching

syu-w:
If the keyring is unavailable, use a fake key.

### DIFF
--- a/charmcraft/commands/store/client.py
+++ b/charmcraft/commands/store/client.py
@@ -47,10 +47,34 @@ def build_user_agent():
     )
 
 
+class AnonymousClient:
+    """Lightweight layer that access public store data."""
+
+    def __init__(self, api_base_url: str, storage_base_url: str):
+        self.api_base_url = api_base_url.rstrip("/")
+        self.storage_base_url = storage_base_url.rstrip("/")
+        self._http_client = craft_store.http_client.HTTPClient(user_agent=build_user_agent())
+
+    def request_urlpath_text(self, method: str, urlpath: str, *args, **kwargs) -> str:
+        """Return a request.Response to a urlpath."""
+        return self._http_client.request(method, self.api_base_url + urlpath, *args, **kwargs).text
+
+    def request_urlpath_json(self, method: str, urlpath: str, *args, **kwargs) -> Dict[str, Any]:
+        """Return .json() from a request.Response to a urlpath."""
+        response = self._http_client.request(method, self.api_base_url + urlpath, *args, **kwargs)
+
+        try:
+            return response.json()
+        except JSONDecodeError as json_error:
+            raise CraftError(
+                f"Could not retrieve json response ({response.status_code} from request"
+            ) from json_error
+
+
 class Client(craft_store.StoreClient):
     """Lightweight layer above StoreClient."""
 
-    def __init__(self, api_base_url, storage_base_url, ephemeral=False):
+    def __init__(self, api_base_url: str, storage_base_url: str, ephemeral: bool = False):
         self.api_base_url = api_base_url.rstrip("/")
         self.storage_base_url = storage_base_url.rstrip("/")
 
--- a/charmcraft/commands/store/store.py
+++ b/charmcraft/commands/store/store.py
@@ -28,7 +28,7 @@ from craft_store import attenuations, endpoints
 from craft_store.errors import CredentialsAlreadyAvailable
 from dateutil import parser
 
-from charmcraft.commands.store.client import Client, ALTERNATE_AUTH_ENV_VAR
+from charmcraft.commands.store.client import AnonymousClient, Client, ALTERNATE_AUTH_ENV_VAR
 
 # helpers to build responses from this layer
 Account = namedtuple("Account", "name username id")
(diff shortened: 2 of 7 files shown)


charmcraft/commands/store/store.py
```diff
@@ -175,7 +175,9 @@ def error_decorator(self, *args, **kwargs):
 class Store:
     """The main interface to the Store's API."""
 
-    def __init__(self, charmhub_config, ephemeral=False):
+    def __init__(self, charmhub_config, ephemeral=False, needs_auth=True):
+        if not needs_auth and not os.getenv(ALTERNATE_AUTH_ENV_VAR):
+            os.environ[ALTERNATE_AUTH_ENV_VAR] = "Tk9LRVk="
```
tigarmo, line 180: What is this `"Tk9LRVk="` string? Is the environment variable needed? I don't think we should be polluting the environment like this

syu-w: `Tk9LRVk=` is `NOKEY` in base64. There is no way to bypass the key check since it is in `craft_store.StoreClient` `__init__`. It is not known what command will be invoked until it is initialized and return a client for others to call.

Rewrite `craft_store.StoreClient` and `BaseClient` are possible. Move the Auth part out of the `__init__` and check that for each calls. However this requires modifying all functions and cannot be checked at initialization time.

tigarmo: Right, I took a look at the code and I see what you're saying. I still think we should find some middle ground so I'm setting up some time where we can discuss this
Thanks!

tests/spread/store/libraries/task.yaml
```diff
@@ -55,3 +55,9 @@ execute: |
   rm -rf lib/
   charmcraft fetch-lib $lib_fqdn
   test -f $lib_filepath
```
sergiusens, line 57: this test is no longer needed

charmcraft/commands/store/client.py
```diff
@@ -47,6 +47,34 @@ def build_user_agent():
     )
 
 
+class AnonymizedClient(craft_store.http_client.HTTPClient):
```
tigarmo, line 50: Possibly `AnonymousClient` is a better name; "Anonymized" makes me think that the client strips identifying info from the requests somehow, which isn't the case

charmcraft/commands/store/client.py
```diff
@@ -47,6 +47,34 @@ def build_user_agent():
...
+    """Lightweight layer that access public store data."""
+
+    def __init__(self, api_base_url, storage_base_url, ephemeral=False):
+        self.api_base_url = api_base_url.rstrip("/")
+        self.storage_base_url = storage_base_url.rstrip("/")
+        del ephemeral  # Placeholder
```
tigarmo, line 56: ? what is this for? It looks to me like the `ephemeral` parameter is auth-related, so this class doesn't need the parameter at all right?

tigarmo: also this init is missing type annotations

charmcraft/commands/store/client.py
```diff
@@ -47,6 +47,34 @@ def build_user_agent():
...
+
+        super().__init__(
```
tigarmo, line 58: I'm thinking now that maybe this class should contain an `HTTPClient` instance, as opposed to being one. I say this because now this `AnonymizedClient` has a bunch of public methods like 'get' and 'post' that shouldn't be used directly; the public api should only be those two `request_urlpath_text|json` right?
@facundobatista what do you think?

tests/commands/test_store_api.py
```diff
@@ -78,6 +90,27 @@ def test_client_init_ephemeral(config):
     ]
 
 
+# -- tests for anonymized client usage
+
+
+def test_anonymized_client_init(config):
+    """Check that the client is initiated ok even without config."""
+    with patch("charmcraft.commands.store.store.AnonymizedClient") as anonymized_client_mock:
```
tigarmo, line 98: is this mock necessary? What happens if you don't mock it and instead just check the `_client` attribute of the created store?

tests/commands/test_store_client.py
```diff
@@ -135,6 +136,26 @@ class _Client(Client, _StoreClientMock):
     auth_patch.stop()
 
 
+@pytest.fixture
+def anonymized_client_class():
+    """Return a anonymized client instance."""
+    with patch("charmcraft.commands.store.store.AnonymizedClient"):
```
tigarmo, line 142: this feels like a lot of mocking, it's hard to tell what's going on. Can we mock just `HTTPClient.request()`? It's the only method that's used right?
also you can use the `mocker` fixture to simplify these mocks quite a bit

tigarmo (changes requested):

tests/commands/test_store_client.py
```diff
@@ -354,3 +375,86 @@ def test_alternate_auth_logout_forbidden(client_class, monkeypatch):
         "Cannot logout when using alternative auth through CHARMCRAFT_AUTH environment variable."
     )
     assert str(cm.value) == expected_error
+
+
+def test_anonymized_client_init():
```
tigarmo, line 380: I like these test cases, looks very complete :+1:

charmcraft/commands/store/client.py
syu-w: OK

tests/commands/test_store_api.py
syu-w: That will also works, but I reused exists test case from the normal client.

tests/commands/test_store_client.py
syu-w: Yes, that can be done

tigarmo (approved):
Nice, thanks for the work!

sergiusens (approved):
This works for now, but we might eventually need to use the anon client and the auth one at the same time

PR #1044, opened 2023-03-07
command: add the initial promote-bundle command

lengau:
This is a draft still because I'm still working on tests at various levels (not yet included)


charmcraft/commands/store/__init__.py
```diff
@@ -722,6 +737,222 @@ def run(self, parsed_args):
         emit.message(msg.format(*args))
 
 
+class PromoteBundleCommand(BaseCommand):
+    """Promote a bundle in the Store."""
+
+    name = "promote-bundle"
+    needs_config = True
```
syu-w, line 744: No longer use `needs_config`.

charmcraft/commands/store/__init__.py
```diff
@@ -722,6 +737,222 @@ def run(self, parsed_args):
...
+            "-e",
+            "--exclude",
+            action="append",
+            default=[],
+            help="Any charms to exclude from the promotion process",
+        )
+
+    def run(self, parsed_args: "Namespace") -> None:
+        """Run the command."""
+        # Validation...
```
syu-w, line 774: Needs `self._check_config(config_file=True)`
We don't need `bases` here?

charmcraft/commands/store/__init__.py
```diff
@@ -722,6 +737,222 @@ def run(self, parsed_args):
...
+        Promote a bundle to another channel in the Store.
+
+
+        """  # TODO (amlowe): complete this description.
+    )
+
+    def fill_parser(self, parser: "ArgumentParser") -> None:
+        """Add promote-bundle parameters to the general parser."""
+        parser.add_argument("source_channel", help="The channel from which to promote the bundle")
+        parser.add_argument("target_channel", help="The target channel for the promoted bundle")
```
syu-w, line 757: Maybe `from_channel` `to_channel` better? Or just `from`, `to`.

charmcraft/commands/store/__init__.py
```diff
@@ -722,6 +737,222 @@ def run(self, parsed_args):
...
+                non_charms.append(charm_name)
+        if invalid_charms:
+            charm_list = utils.humanize_list(invalid_charms, "and")
+            raise CraftError(
+                "The following entities do not exist or you are not a collaborator on them: "
+                f"{charm_list}"
+            )
+        if non_charms:
+            non_charm_list = utils.humanize_list(non_charms, "and")
+            raise CraftError(f"The following store entities are not charms: {non_charm_list}")
```
syu-w, line 877: Does the API able to tell the charm exists or not?
I think something like:

The following charms you own/collaboration will be promoted:
xxxx: beta -> xxxx: stable
xxxx: beta -> xxxx: stable

The following charms you do not own/collaboration will be ignored:
xxxx: beta

So we don't need to ask user to exclude everything.

charmcraft/commands/store/__init__.py
lengau: `needs_config` isn't used directly, but it overrides the `needs_config` in `BaseCommand`. Commands with `needs_config = False` don't have to be run with a project directory specified (or in a project directory).

charmcraft/commands/store/__init__.py
lengau: That's what I originally had. I changed it because the spec used the "source" and "target" language and it was easier for me when implementing the spec. @sergiusens is there any reason to leave this with source/target language or can I call them `from_channel` and `to_channel`?

charmcraft/commands/store/__init__.py
lengau: The API we're using won't reveal any charms you can't write to, which is why we can't distinguish charms the user doesn't own or isn't a collaborator on from charms that don't exist. We could retrieve those charms the same way the website does, but I don't think that's a stable public API. If we want to do that we'll probably need to get the store team to add an endpoint to the developer API. The `exclude` option still needs to exist for excluding promotions on charms the user already owns, but we could detect invalid charms and simply exclude them instead. 

For the MVP, I opted for maximum explicitness after discussing it with Facundo. This particular behaviour is undefined in the spec, so we'll probably want to discuss the options with stakeholders. The way I'd imagine we'd implement it would be with a pair of arguments (`--auto-exclude` and `--no-auto-exclude`) to prefer an answer, defaulting to either asking when interactive or failing (equivalent to `--no-auto-exclude`) when not interactive.

charmcraft/commands/store/__init__.py
```diff
@@ -722,6 +737,222 @@ def run(self, parsed_args):
...
+            raise CraftError(
+                "Invalid bundle config; missing a 'name' field indicating the bundle's name in "
+                f"file {str(bundle_path)!r}."
+            )
+        emit.progress("Determining charms to promote")
+        charms = [c["charm"] for c in bundle_config["applications"].values()]
+        errant_excludes = []
+        for excluded in parsed_args.exclude:
+            try:
+                charms.remove(excluded)
```
cmatsuoka, line 840: Instead of creating a full list of charms and then removing the excluded ones, wouldn't it be better to create the initial list without the excluded items?

lengau: The best I can come up with without sacrificing the error checking is:

```python
        charms = []
        excluded_charms = set()
        for application in bundle_config["applications"].values():
            if application["charm"] in parsed_args.exclude:
                excluded_charms.add(application["charm"])
            else:
                charms.append(application["charm"])
        errant_excludes = set(parsed_args.exclude) - excluded_charms
```

I'm not sure it gains us anything, but I'm happy to put it (or a modification) in instead since I don't see it losing us anything either.

charmcraft/commands/store/__init__.py
lengau: (Marking this as resolved for now, but it might be something to discuss in a further design meeting)

charmcraft/commands/store/__init__.py
lengau: [resolved after direct conversation]

cmatsuoka (approved):
Looks good, thanks Alex.

sergiusens (approved):

PR #1045, opened 2023-03-08
plugin: make charm plugin base aware

syu-w:
This allows use charm plugins in CentOS.


syu-w:
This one needs `is_yum_based()` from `craft_parts`, not released yet.

charmcraft/parts.py
```diff
@@ -149,13 +151,32 @@ def get_build_snaps(cls) -> Set[str]:
...
+        if platform.is_deb_based():
+            return {
+                "python3-dev",
+                "python3-pip",
+                "python3-setuptools",
+                "python3-venv",
+                "python3-wheel",
+            }
+        elif platform.is_yum_based():
+            if os_utils.OsRelease().version_id() == "7":
```
cmatsuoka, line 163: This works because only CentOS is listed as yum-based. A more robust way to do this verification would be to check both the distribution name and version here (just to be sure).

cmatsuoka: Also `is_yum_based` is in an unreleased version of craft-parts, maybe we should land this PR in a separate branch that uses craft parts from main? (some fixes are still necessary before relasing stable 1.19.0)

cmatsuoka: Or we can wait until the features are available in a point release.

syu-w:
Updated. And I am able to build charms successful in centos.

cmatsuoka (approved):
Thanks!

PR #1049, opened 2023-03-20
requirements: update craft-parts to 1.19.0

cmatsuoka:
Craft-parts 1.19.0 contains initial support for the yum package manager and adds a source handler for rpm packages.


lengau (approved):
Sorry this took me so long to review! The complex and bulky changes here required a lot of thought.

PR #1050, opened 2023-03-20
cli: add how to use exported credentials in help

syu-w:


charmcraft/commands/store/__init__.py
```diff
@@ -103,7 +103,17 @@ class LoginCommand(BaseCommand):
     def fill_parser(self, parser):
         """Add own parameters to the general parser."""
         parser.add_argument(
-            "--export", type=pathlib.Path, help="The file to save the credentials to"
+            "--export",
+            type=pathlib.Path,
+            help=(
+                "Export the Charmhub unencrypted secret credentials to a file. "
```
mr-cal, line 109: I believe we try to keep the help for flags to a single short sentence (this a just a conjecture as I can't find any specifications supporting this).  I think this type of information is more appropriate in the overview.

charmcraft/commands/store/__init__.py
```diff
@@ -87,6 +87,17 @@ class LoginCommand(BaseCommand):
         access to Charmhub at the CLI (if `--export` option was not used
         otherwise it will only save the credentials in the indicated file).
 
+        If `--export <file>` option was used, a secret credentials file will
+        be created. And the file can be used to set "CHARMCRAFT_AUTH" 
+        environment variable.
+
+            e.g. `export CHARMCRAFT_AUTH=$(cat secret)`
```
mr-cal, line 94: pedantic: These indented lines get automatically formatted as markdown for the forums, so I would remove the `e.g.` and the backticks from this.

For example, you can see this automatic formatting with `charmcraft fetch-lib --help --format=markdown` and how it looks on the forums here: https://juju.is/docs/sdk/charmcraft-fetch-lib

charmcraft/commands/store/__init__.py
```diff
@@ -87,6 +87,17 @@ class LoginCommand(BaseCommand):
         access to Charmhub at the CLI (if `--export` option was not used
         otherwise it will only save the credentials in the indicated file).
 
+        If `--export <file>` option was used, a secret credentials file will
```
mr-cal, line 90: `was used` -> `is used`

mr-cal (changes requested):
Nice work.  The help message is well written and much more useful.

I added 2 minor nits, but other than that, we are ready to rock!

mr-cal (approved):
Rock and roll!

PR #1052, opened 2023-03-23
store: check the credentials validity before upload

syu-w:
Avoid wasting time uploading when credentials expire.

### DIFF
--- a/charmcraft/commands/store/store.py
+++ b/charmcraft/commands/store/store.py
@@ -251,6 +251,10 @@ class Store:
         )
         return result
 
+    def _check_authorized(self) -> None:
+        """Check if current credentials authenticated."""
+        self._client.whoami()
+
     @_store_client_wrapper()
     def register_name(self, name, entity_type):
         """Register the specified name for the authenticated user."""
@@ -280,6 +284,8 @@ class Store:
 
     def _upload(self, endpoint, filepath, *, extra_fields=None):
         """Upload for all charms, bundles and resources (generic process)."""
+        self._check_authorized()
+
         upload_id = self._client.push_file(filepath)
         payload = {"upload-id": upload_id}
         if extra_fields is not None:
--- a/tests/commands/test_store_api.py
+++ b/tests/commands/test_store_api.py
@@ -658,6 +658,7 @@ def test_upload_straightforward(client_mock, emitter, config):
 
     # check all client calls
     assert client_mock.mock_calls == [
+        call.whoami(),
         call.push_file(test_filepath),
         call.request_urlpath_json("POST", test_endpoint, json={"upload-id": test_upload_id}),
         call.request_urlpath_json("GET", test_status_url),
@@ -715,7 +716,7 @@ def test_upload_polls_status_ok(client_mock, emitter, config):
             result = store._upload("/test/endpoint/", "some-filepath")
 
     # check the status-checking client calls (kept going until third one)
-    assert client_mock.mock_calls[2:] == [
+    assert client_mock.mock_calls[3:] == [
         call.request_urlpath_json("GET", test_status_url),
         call.request_urlpath_json("GET", test_status_url),
         call.request_urlpath_json("GET", test_status_url),
@@ -885,6 +886,7 @@ def test_upload_including_extra_parameters(client_mock, emitter, config):
 
     # check all client calls
     assert client_mock.mock_calls == [
+        call.whoami(),
         call.push_file(test_filepath),
         call.request_urlpath_json(
             "POST",
--- a/tests/commands/test_store_commands.py
+++ b/tests/commands/test_store_commands.py
@@ -16,17 +16,18 @@
 
 """Tests for the Store commands (code in store/__init__.py)."""
 
+import base64
 import datetime
 import sys
 import zipfile
 from argparse import Namespace, ArgumentParser
-from unittest.mock import patch, call, MagicMock, ANY
+from unittest.mock import patch, call, MagicMock, Mock, ANY
 
 import dateutil.parser
 import pytest
 import yaml
 from craft_cli import CraftError
-from craft_store.errors import CredentialsUnavailable
+from craft_store.errors import CredentialsUnavailable, StoreServerError
 
 from charmcraft.cmdbase import JSON_FORMAT
 from charmcraft.config import CharmhubConfig
@@ -82,6 +83,16 @@ noargs = Namespace()
 
 # used to flag defaults when None is a real option
 DEFAULT = object()
+
+
+def _fake_response(status_code, reason=None, json=None):
+    response = Mock(spec="requests.Response")
+    response.status_code = status_code
+    response.ok = status_code == 200
+    response.reason = reason
+    if json is not None:
+        response.json = Mock(return_value=json)
+    return response
 
 
 @pytest.fixture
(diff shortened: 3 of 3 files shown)


charmcraft/commands/store/store.py
```diff
@@ -315,6 +315,9 @@ def _upload(self, endpoint, filepath, *, extra_fields=None):
     def upload(self, name, filepath):
         """Upload the content of filepath to the indicated charm."""
         endpoint = f"/v1/charm/{name}/revisions"
+        if not self.whoami():
+            raise CraftError("Existing credentials are no longer valid for Charmhub.")
+
```
lengau, line 320: Based on the tests, couldn't we just do:

```suggestion
        # This raises an exception if the user's credentials are invalid.
        self.whoami()
```

I can't see a case where `whoami` would return some falsey value rather than raising an exception.

lengau (changes requested):

charmcraft/commands/store/store.py
syu-w: I can't see it either, but I put it here just in case.

charmcraft/commands/store/store.py
```diff
@@ -315,6 +315,7 @@ def _upload(self, endpoint, filepath, *, extra_fields=None):
     def upload(self, name, filepath):
         """Upload the content of filepath to the indicated charm."""
         endpoint = f"/v1/charm/{name}/revisions"
+        self.whoami()
```
tigarmo, line 318: I think this should go at the start of `_upload()`, because it looks like `upload_resource()`, which also calls `_upload()`, would also benefit from this change.
And then please add a comment to remind us  that the whoami call is to verify that the user is logged in; otherwise we might be tempted to remove this "spurious" call in the future.

tests/commands/test_store_commands.py
```diff
@@ -955,6 +966,25 @@ def test_upload_call_error(emitter, store_mock, config, tmp_path, formatted):
...
+        mock_whoami.side_effect = StoreServerError(_fake_response(401, json={}))
+
+        test_charm = tmp_path / "mystuff.charm"
+        _build_zip_with_yaml(test_charm, "metadata.yaml", content={"name": "mycharm"})
+        args = Namespace(filepath=test_charm, release=[], name=None, format=formatted)
+
+        with pytest.raises(CraftError) as cm:
+            UploadCommand(config).run(args)
+        assert str(cm.value) == (
+            "Provided credentials are no longer valid for Charmhub. "
```
tigarmo, line 983: what was the error before this fix?

tigarmo (changes requested):

tests/commands/test_store_commands.py
syu-w: It should be a 401 unauthorized from here 
https://github.com/canonical/charmcraft/blob/9d0957ee4e7dda3327a93c79e0e0aff867c62498/charmcraft/commands/store/store.py#L152-L165

charmcraft/commands/store/store.py
syu-w: OK, will add.

tests/commands/test_store_commands.py
tigarmo: Right, I mean, the purpose of this test is to check the early-failure right? How do we know that this failure is being caused by the whoami, and not by the rest of the upload process?

syu-w: Do we want to check traceback for this?

tigarmo: Nah; can we mock the "heavy" call (looks like it might be `self._client.push_file()`?) and verify that it was _not_ called?

syu-w: OK, that works.

syu-w:
Updated

charmcraft/commands/store/store.py
```diff
@@ -280,6 +280,9 @@ def list_registered_names(self, include_collaborations: bool) -> List[Entity]:
 
     def _upload(self, endpoint, filepath, *, extra_fields=None):
         """Upload for all charms, bundles and resources (generic process)."""
+        # Check if the current credentials still vaild before upload
```
tigarmo, line 283:
```suggestion
        # Check if the current credentials are still valid before upload
```

tests/commands/test_store_api.py
```diff
@@ -656,8 +656,8 @@ def test_upload_straightforward(client_mock, emitter, config):
     with patch.dict("charmcraft.commands.store.store.UPLOAD_ENDING_STATUSES", fake_statuses):
         result = store._upload(test_endpoint, test_filepath)
 
-    # check all client calls
-    assert client_mock.mock_calls == [
+    # check all client calls, excpet whoami() in first 10 calls
```
tigarmo, line 659:
```suggestion
    # check all client calls, expect whoami() in first 10 calls
```

(please fix the other `expect` typos in this file too)

tigarmo, line 659: why are there ten (!) `whoami()` calls? That's a lot!

tests/commands/test_store_commands.py
```diff
@@ -955,6 +966,27 @@ def test_upload_call_error(emitter, store_mock, config, tmp_path, formatted):
...
+            _build_zip_with_yaml(test_charm, "metadata.yaml", content={"name": "mycharm"})
+            args = Namespace(filepath=test_charm, release=[], name=None, format=formatted)
+
+            with pytest.raises(CraftError) as cm:
+                UploadCommand(config).run(args)
+            assert str(cm.value) == (
+                "Provided credentials are no longer valid for Charmhub. "
+                "Regenerate them and try again."
+            )
+            assert not push_file_mock.called
```
tigarmo, line 987: :+1:

tigarmo (changes requested):

tests/commands/test_store_api.py
syu-w: No, it just one, but since `whoami()` in the mock itself, it become:
```
E         +  call.whoami(),
E         +  call.whoami().__getitem__('account'),
E         +  call.whoami().__getitem__().__getitem__('display-name'),
E         +  call.whoami().__getitem__().__getitem__('username'),
E         +  call.whoami().__getitem__().__getitem__('id'),
E         +  call.whoami().__getitem__('packages'),
E         +  call.whoami().__getitem__('packages'),
E         +  call.whoami().__getitem__().__iter__(),
E         +  call.whoami().__getitem__('channels'),
E         +  call.whoami().__getitem__('permissions'),
```

tigarmo: Oh I see. I think this shows that maybe we can improve the design a bit; the only thing we need is the `whoami()`  call on the client, right? All that other handling of packages and channels and permissions is unnecessary here. 
So what if we add a `_check_authorized()` method that just does `self._client_whoami()`? Then we can add the single `call.whoami()` call to these lists to document the fact that the upload methods call whoami.

What do you think?

syu-w: Hmmm, this also works.

charmcraft/commands/store/store.py
```diff
@@ -251,6 +251,11 @@ def whoami(self):
         )
         return result
 
+    @_store_client_wrapper(auto_login=False)
```
tigarmo, line 254: possibly we don't need this decorator, as this method is called internally by public methods that have the decorator already (looks like the other private methods don't use the decorator).

tigarmo (approved):
thanks! Please remove the unnecessary decorator and then we're good to go I think

charmcraft/commands/store/store.py
syu-w: Oh, it should be wrapped already.

lengau (approved):
Looks good. Thanks!

PR #1053, opened 2023-03-24
parts: CentOS 7 python 3.8 path workaround

syu-w:
The Python 3.8 in CentOS 7 from SCL repo not included in PATH.

### DIFF
--- a/charmcraft/parts.py
+++ b/charmcraft/parts.py
@@ -20,7 +20,8 @@ import os
 import pathlib
 import shlex
 import sys
-from typing import Any, Dict, List, Set, cast
+from typing import Any, Dict, List, Set, Optional, cast
+from contextlib import suppress
 
 import pydantic
 from craft_cli import emit, CraftError
@@ -184,6 +185,10 @@ class CharmPlugin(plugins.Plugin):
 
     def get_build_environment(self) -> Dict[str, str]:
         """Return a dictionary with the environment to use in the build step."""
+        os_special_paths = self._get_os_special_priority_paths()
+        if os_special_paths:
+            return {"PATH": os_special_paths + ":${PATH}"}
+
         return {}
 
     def get_build_commands(self) -> List[str]:
@@ -203,6 +208,10 @@ class CharmPlugin(plugins.Plugin):
         ]:
             if key in os.environ:
                 build_env[key] = os.environ[key]
+
+        os_special_paths = self._get_os_special_priority_paths()
+        if os_special_paths:
+            build_env["PATH"] = os_special_paths + ":" + build_env["PATH"]
 
         env_flags = [f"{key}={value}" for key, value in build_env.items()]
 
@@ -256,6 +265,16 @@ class CharmPlugin(plugins.Plugin):
     def post_build_callback(self, step_info):
         """Collect metrics left by charm_builder.py."""
         instrum.merge_from(env.get_charm_builder_metrics_path())
+
+    def _get_os_special_priority_paths(self) -> Optional[str]:
+        """Return a str of PATH for special OS."""
+        with suppress(OsReleaseIdError, OsReleaseVersionIdError):
+            os_release = os_utils.OsRelease()
+            if (os_release.id(), os_release.version_id()) in (("centos", "7"), ("rhel", "7")):
+                # CentOS 7 Python 3.8 from SCL repo
+                return "/opt/rh/rh-python38/root/usr/bin"
+
+        return None
 
 
 class BundlePluginProperties(plugins.PluginProperties, plugins.PluginModel):
--- a/tests/test_parts.py
+++ b/tests/test_parts.py
@@ -106,11 +106,26 @@ def test_charmplugin_get_build_snaps(charm_plugin):
     assert charm_plugin.get_build_snaps() == set()
 
 
-def test_charmplugin_get_build_environment(charm_plugin):
+def test_charmplugin_get_build_environment_ubuntu(charm_plugin, mocker):
+    mock_id = mocker.patch("craft_parts.utils.os_utils.OsRelease.id")
+    mock_version = mocker.patch("craft_parts.utils.os_utils.OsRelease.version_id")
+    mock_id.return_value = "ubuntu"
+    mock_version.return_value = "22.04"
     assert charm_plugin.get_build_environment() == {}
 
 
-def test_charmplugin_get_build_commands(charm_plugin, tmp_path, monkeypatch):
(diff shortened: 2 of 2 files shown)


charmcraft/parts.py
```diff
@@ -184,6 +184,15 @@ def get_build_packages(self) -> Set[str]:
...
     def get_build_environment(self) -> Dict[str, str]:
         """Return a dictionary with the environment to use in the build step."""
+        try:
+            os_release = os_utils.OsRelease()
+            if (os_release.id(), os_release.version_id()) in (("centos", "7"), ("rhel", "7")):
+                # CentOS 7 Python 3.8 from SCL repo
+                # get_build_commands will ignored build.sh env
+                os.environ["PATH"] = "/opt/rh/rh-python38/root/usr/bin:" + os.environ["PATH"]
+                return {"PATH": "/opt/rh/rh-python38/root/usr/bin:${PATH}"}
+        except (OsReleaseIdError, OsReleaseVersionIdError):
```
lengau, line 194: Since we're using `except/pass`, please use `contextlib.suppress`

charmcraft/parts.py
```diff
@@ -184,6 +184,15 @@ def get_build_packages(self) -> Set[str]:
 
     def get_build_environment(self) -> Dict[str, str]:
         """Return a dictionary with the environment to use in the build step."""
+        try:
+            os_release = os_utils.OsRelease()
```
lengau, line 188: Please move this outside the `try` since it doesn't throw any of the expected errors

charmcraft/parts.py
```diff
@@ -184,6 +184,15 @@ def get_build_packages(self) -> Set[str]:
...
+            if (os_release.id(), os_release.version_id()) in (("centos", "7"), ("rhel", "7")):
+                # CentOS 7 Python 3.8 from SCL repo
+                # get_build_commands will ignored build.sh env
+                os.environ["PATH"] = "/opt/rh/rh-python38/root/usr/bin:" + os.environ["PATH"]
```
tigarmo, line 192: I think it's a bad idea to "pollute" the "main" environment here; it's going to affect everything that comes after this.

If the need is to also use this path in `get_build_commands()` then I think a better way would be to also check for centos in that function and extend the PATH to use there; possibly with a helper function to isolate this try/except (or contextlib.supress).

charmcraft/parts.py
tigarmo, line 194: I'm curious about these exceptions; when did you get these errors?

tigarmo (changes requested):

charmcraft/parts.py
lengau: Actually, I'm retracting that.

tigarmo (changes requested):

tests/test_parts.py
```diff
@@ -106,11 +106,26 @@ def test_charmplugin_get_build_snaps(charm_plugin):
     assert charm_plugin.get_build_snaps() == set()
 
 
-def test_charmplugin_get_build_environment(charm_plugin):
-    assert charm_plugin.get_build_environment() == {}
+def test_charmplugin_get_build_environment_ubuntu(charm_plugin):
+    with patch("craft_parts.utils.os_utils.OsRelease.id") as mock_id:
```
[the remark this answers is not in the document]
syu-w: Yep, that's better.

charmcraft/parts.py
```diff
@@ -257,6 +266,17 @@ def post_build_callback(self, step_info):
...
         instrum.merge_from(env.get_charm_builder_metrics_path())
 
+    def _get_os_special_path(self, path: Optional[str] = None) -> Optional[str]:
+        with suppress(OsReleaseIdError, OsReleaseVersionIdError):
+            os_release = os_utils.OsRelease()
+            if (os_release.id(), os_release.version_id()) in (("centos", "7"), ("rhel", "7")):
+                # CentOS 7 Python 3.8 from SCL repo
+                if path:
+                    return f"/opt/rh/rh-python38/root/usr/bin:{path}"
+                return f"/opt/rh/rh-python38/root/usr/bin:{os.environ.get('PATH', '')}"
```
tigarmo, line 276: I don't think this is equivalent to the original code that you had in `get_build_environment()`; the returned dict should not use the value of the "current" environment variables, but placeholders like `${PATH}` (I think this is what you had originally) which are expanded later, when the build script is actually executed. Does that make sense?
From that point of view, it might be simpler to have this `_get_os_special_path` function receive no parameters and then have the callers append the correct strings.

tigarmo: @cmatsuoka can you please check my reasoning here?

tigarmo (changes requested):

charmcraft/parts.py
syu-w: Will this `PATH` only used in shell scripts? As you cannot set `PATH="${PATH}"` if it is not a shell.

tigarmo: Yes, this dict is saved to a file that is then run with bash
https://github.com/canonical/craft-parts/blob/d70898380720fe37f6cce3d71d0919188fd79bac/craft_parts/plugins/rust_plugin.py#L119

syu-w: OK, so I had to leave `${PATH}` in the unit test. Since it is not call shell to run it.

syu-w: Since this one use `shlex.quote` later, it is not possible to use `${PATH}`. It get escaped to `env -i LANG=C.UTF-8 LC_ALL=C.UTF-8 'PATH=/opt/rh/rh-python38/root/usr/bin:${PATH}' SNAP=snap_value ...`. Unless I create that command by hand.

syu-w: For what it looks like in the test case results, `get_build_environment()` does not affect `get_build_commands()`.

charmcraft/parts.py
```diff
@@ -257,6 +266,16 @@ def post_build_callback(self, step_info):
         """Collect metrics left by charm_builder.py."""
         instrum.merge_from(env.get_charm_builder_metrics_path())
 
+    def _get_os_special_priority_paths(self) -> list:
```
tigarmo, line 269: I really think we should just return a `Optional[str]` instead of a list here
if you're really set on returning a list, this annotation should be `List[str]`

tigarmo (changes requested):

charmcraft/parts.py
syu-w: I can go for str, don't seems use a lot atm.

tigarmo: alright!

PR #1055, opened 2023-03-28
Update charm templates to use ops 2.2.0 and new "import ops" style

benhoyt:
With https://github.com/canonical/operator/pull/910 (released in ops 2.2.0) the recommended way to import ops is simply `import ops`, so you don't have to worry about which sub-module the various names are in.

Also add type hints for the event types.

I've created a charm using `charmcraft init` with each of these 3 template types, and verified the code is correct and passes the unit tests.


jnsgruk (approved):

lengau (approved):
Do we have any tests that initialise and run the templates? (It'd be nice to have automated testing to guarantee that each of the templates work OOTB)

PR #1060, opened 2023-04-04
pack: add recursive bundle packing

lengau:


tests/commands/test_pack.py
```diff
@@ -71,7 +78,10 @@ def bundle_yaml(tmp_path):
     bundle_path.write_text("{}")
     content = {}
 
-    def func(*, name):
+    def func(*, name, base_content: Optional[Dict[str, Any]] = None):
+        if base_content:
+            for key, value in base_content.items():
```
tigarmo, line 83: minor but can you do `content.update(base_content)` here?

charmcraft/commands/pack.py
```diff
@@ -253,3 +329,81 @@ def _pack_bundle(self, parsed_args) -> List[pathlib.Path]:
...
 
         return [zipname]
+
+    def _recursive_pack_bundle(
+        self,
+        parsed_args: argparse.Namespace,
+        bundle_config: Optional[Dict[str, Any]],
+        charms: Mapping[str, pathlib.Path],
+    ) -> List[pathlib.Path]:
+        bundle_charms = bundle_config.get("applications", {})
```
tigarmo, line 339: curious that this typechecked, can't `bundle_config` be `None` here?

charmcraft/commands/pack.py
```diff
@@ -253,3 +329,81 @@ def _pack_bundle(self, parsed_args) -> List[pathlib.Path]:
...
+                    duplicate_charms[charm_name] = temp_dir.glob(f"{charm_name}_*.charm")
+                continue
+            generated_charms[charm_name] = charm_file
+        if duplicate_charms:
+            raise DuplicateCharmsError(duplicate_charms)
+        for charm, charm_file in generated_charms.items():
+            destination = cwd / charm_file.name
+            destination.unlink(missing_ok=True)
+            generated_charms[charm] = shutil.move(charm_file, destination)
+    os.chdir(cwd)
```
tigarmo, line 408: might want to put this in a `finally` block just in case

tests/spread/charms/bundle/task.yaml
```diff
@@ -0,0 +1,23 @@
...
+  pushd bundle
+  charmcraft clean
+  popd
+
+  rm -rf bundle
+
+execute: |
+  cd bundle
+  charmcraft pack --verbose --include-all-charms
+  juju deploy --dry-run *.zip
```
tigarmo, line 23: I don't know juju so how can we verify that the bundle contains the expected multiple charms?

tigarmo (commented):
thanks, looks like a lot of work!

tests/commands/test_pack.py
lengau: That's the very thing I was looking for, but last night it just *was not* coming to mind. Thank you!

charmcraft/commands/pack.py
lengau: Fair enough. I didn't because any error there will exit the application, but that may not be the case in future so I've fixed it.

charmcraft/commands/pack.py
lengau: Charmcraft's checks don't currently include type checking, though I'm working towards that in #1062 . Currently though, `pyright` produces 262 errors and 1 warning, and it'll be fairly heavy lifting to get things corrected. 

I've corrected the typing issue (made `bundle_config` non-optional and ensured it's always a dict in `run`).

tests/spread/charms/bundle/task.yaml
lengau: The bundle itself won't contain the charms — they'll be in the working directory and referenced (by their full path) from the bundle zip file. I've added some tests to ensure correctness. (Not sure how to make a parametrizable test like this with an explicit `--include-charms` so there's no actual e2e test for that at this point.)

lengau:
Thanks for the review @tigarmo ! I've fixed what you pointed out :-)

tigarmo (approved):
nice!

PR #1062, opened 2023-04-04
tools: import tox from starbase

lengau:
This adds the tox starbase config.

Note that several linters fail, so it's not appropriate to move this into CI yet.


mr-cal (approved):
Nice work!  I tested it and it works great.

PR #1064, opened 2023-04-05
deps: remove ops dependency

lengau:
Spread tests were failing due to this. We're going to have to keep this pre-2.0 until we can guarantee at least Python 3.8 in our spread tests, which will likely only happen when we drop support for 18.04.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -50,7 +50,7 @@ dev_requires = [
     "black",
     "coverage",
     "flake8",
-    "ops>=1.4.0",
+    "ops>=1.4.0,<2.0",  # *1
     "pydocstyle",
     "pytest",
     "pytest-cov",
@@ -59,6 +59,9 @@ dev_requires = [
     "responses",
     "tox",
 ]
+
+# *1: Ops 2.0 requires Python 3.8, so we can only upgrade once we can
+# guarantee we'll have it. Currently, spread tests on bionic use Python 3.6.
 
 extras_require = {
     "dev": dev_requires,
--- a/setup.py
+++ b/setup.py
@@ -50,7 +50,6 @@ dev_requires = [
     "black",
     "coverage",
     "flake8",
-    "ops>=1.4.0,<2.0",  # *1
     "pydocstyle",
     "pytest",
     "pytest-cov",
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -60,6 +60,9 @@
     "tox",
 ]
 
+# *1: Ops 2.0 requires Python 3.8, so we can only upgrade once we can
+# guarantee we'll have it. Currently, spread tests on bionic use Python 3.6.
```
tigarmo, line 64: if the issue is only spread tests on bionic, could we conceivably use the deadsnakes ppa to install 3.8 there?

tigarmo (approved):

tigarmo (approved):

jameinel:
Note that Juju 3.0+ and Ops 2.0+ no longer support Bionic. I don't know
what your use cases are for bionic, and how long you intend to support them
(juju 2.9 still supports it, and so does ops 1.5)
But you'll want some sort of strategy in another year as it gets harder to
support, do you want to fork the charm for a bionic vs post-bionic one, etc.

sergiusens:
we will drop 18.04 when we bump the major version

sergiusens:
@lengau why does charmcraft itself depend on ops?

we don't need to run charmcraft tests on 18.04. I would argue we only need to run the tests on the base we provide charmcraft as and potentially the latest version of python.

Charmcraft can then operate on any base (even 18.04) and will use the ops package defined by the charm project with the python version corresponding to that base.

lengau:
@sergiusens That's a good question that I have at best a guess for. After doing some digging rather than just reverting the commit that `git bisect` found to be the problem, ops is a dev requirement dating back to at least 2021. I presume it was done that way to make it easy to test charms, but I'm not sure.

The tests that are failing are spread tests that do use the snap, which deepens the mystery in a way.

I'm doing some local checking with ops simply removed from even dev dependencies — if that works I'll change the PR to do that. If not, I'll re-evaluate from that point and update here.

lengau:
@sergiusens Removing the dependency worked too, so that's probably the better way to go.

sergiusens (approved):
Event better!

PR #1065, opened 2023-04-05
bases: support creating charms use CentOS 7 as base

syu-w:
needs Ubuntu 18.04 for cgroup v1

needs craft-providers unreleased version.

CRAFT-1697

### DIFF
--- a/charmcraft/commands/build.py
+++ b/charmcraft/commands/build.py
@@ -25,6 +25,7 @@ import zipfile
 from typing import List, Optional
 
 from craft_cli import emit, CraftError
+from craft_providers.bases import get_base_alias
 
 from charmcraft import env, linters, parts, providers, instrum
 from charmcraft.charm_builder import DISPATCH_FILENAME, HOOKS_DIR
@@ -344,7 +345,7 @@ class Builder:
             "(may take a while the first time but it's reusable)"
         )
 
-        build_base = providers.BASE_CHANNEL_TO_PROVIDER_BASE[build_on.channel]
+        build_base_alias = get_base_alias((build_on.name, build_on.channel))
         instance_name = providers.get_instance_name(
             bases_index=bases_index,
             build_on_index=build_on_index,
@@ -353,16 +354,24 @@ class Builder:
             target_arch=get_host_architecture(),
         )
         base_configuration = providers.get_base_configuration(
-            alias=build_base,
+            alias=build_base_alias,
             instance_name=instance_name,
         )
+
+        if build_on.name != "ubuntu":
+            allow_unstable = True
+            emit.message(
+                f"Warning: Base {build_on.name} {build_on.channel} daily image may be unstable."
+            )
+        else:
+            allow_unstable = False
 
         with self.provider.launched_environment(
             project_name=self.metadata.name,
             project_path=self.charmdir,
             base_configuration=base_configuration,
-            build_base=build_base.value,
             instance_name=instance_name,
+            allow_unstable=allow_unstable,
         ) as instance:
             emit.debug("Mounting directory inside the instance")
             with instrum.Timer("Mounting directory"):
--- a/charmcraft/parts.py
+++ b/charmcraft/parts.py
@@ -20,7 +20,7 @@ import os
 import pathlib
 import shlex
 import sys
-from typing import Any, Dict, List, Set, Optional, cast
+from typing import Any, Dict, List, Optional, Set, cast
 from contextlib import suppress
 
 import pydantic
--- a/charmcraft/parts.py
+++ b/charmcraft/parts.py
@@ -166,6 +166,13 @@ class CharmPlugin(plugins.Plugin):
                 if (os_release.id(), os_release.version_id()) in (("centos", "7"), ("rhel", "7")):
                     # CentOS 7 Python 3.8 from SCL repo
                     return {
+                        "autoconf",
+                        "automake",
+                        "gcc",
+                        "gcc-c++",
+                        "git",
+                        "make",
+                        "patch",
                         "rh-python38-python-devel",
                         "rh-python38-python-pip",
                         "rh-python38-python-setuptools",
@@ -175,6 +182,13 @@ class CharmPlugin(plugins.Plugin):
                 pass
 
             return {
+                "autoconf",
+                "automake",
+                "gcc",
+                "gcc-c++",
+                "git",
+                "make",
+                "patch",
                 "python3-devel",
                 "python3-pip",
                 "python3-setuptools",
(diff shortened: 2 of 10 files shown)


charmcraft/providers.py
```diff
@@ -171,26 +173,37 @@ def get_instance_name(
...
 def get_base_configuration(
     *,
-    alias: bases.BuilddBaseAlias,
+    alias: enum.Enum,
     instance_name: str,
-) -> bases.BuilddBase:
-    """Create a BuilddBase configuration."""
-    environment = get_command_environment()
-
+) -> Union[bases.ubuntu.BuilddBase, bases.centos.CentOSBase]:
```
mr-cal, line 178: Can you return the parent class, `bases.Base`?

charmcraft/providers.py
```diff
@@ -171,26 +173,37 @@ def get_instance_name(
...
+        charmcraft_snap = bases.ubuntu.Snap(name="charmcraft", channel=snap_channel, classic=True)
+        return bases.ubuntu.BuilddBase(
+            alias=alias,
+            environment=environment,
+            hostname=instance_name,
+            snaps=[charmcraft_snap],
+            compatibility_tag=f"charmcraft-{bases.ubuntu.BuilddBase.compatibility_tag}.0",
+        )
+
+    if isinstance(alias, bases.centos.CentOSBaseAlias):
```
mr-cal, line 197: Can you add something to this function to handle when neither of the if statements are true?

Since it only returns inside `if` statements, perhaps a `raise Exception("base is not supported")` at the end of the function will do the trick.

In theory, this error case shouldn't occur with how this function is called, but it's still a good design practice.

charmcraft/providers.py
```diff
@@ -36,10 +37,11 @@
 from charmcraft.snap import get_snap_configuration
 
 
-BASE_CHANNEL_TO_PROVIDER_BASE = {
-    "18.04": bases.BuilddBaseAlias.BIONIC,
-    "20.04": bases.BuilddBaseAlias.FOCAL,
-    "22.04": bases.BuilddBaseAlias.JAMMY,
+BASE_NAME_VERSION_TO_PROVIDER_BASE = {
```
mr-cal, line 40: opinion: Consider the tuple a "charmcraft base" and call this `CHARMCRAFT_BASE_TO_PROVIDER_BASE`, which is more clear than using each part of the tuple itself (`NAME` and `VERSION`)

mr-cal (commented):
Looking good!  I added some comments while reviewing side-by-side with the craft-providers PR.

charmcraft/providers.py
syu-w: Moved to craft-providers.

tests/spread/charms/centos-7/task.yaml
```diff
@@ -0,0 +1,35 @@
+summary: pack default charm for centos 7
+manual: true
```
sergiusens, line 2: why is this manual?

syu-w: I see the tests under this category are all manual. Should it be `false`?

sergiusens: move it to smoketests, the "charms" suite was meant for actual production charms to be tested

@lengau perhaps move bundle from there too ^ ?

charmcraft/commands/build.py
```diff
@@ -353,16 +354,16 @@ def pack_charm_in_instance(
...
             instance_name=instance_name,
         )
 
         with self.provider.launched_environment(
             project_name=self.metadata.name,
             project_path=self.charmdir,
             base_configuration=base_configuration,
-            build_base=build_base.value,
             instance_name=instance_name,
+            allow_unstable=(build_on.name != "ubuntu"),
```
sergiusens, line 366: if centos is unstable, we probably want to emit a warning

charmcraft/parts.py
```diff
@@ -154,6 +154,7 @@ def get_build_packages(self) -> Set[str]:
         """Return a set of required packages to install in the build environment."""
         if platform.is_deb_based():
             return {
+                "build-essential",
```
sergiusens, line 157: why are we adding this here?

charmcraft/commands/build.py
syu-w: Added

charmcraft/commands/build.py
```diff
@@ -353,16 +354,22 @@ def pack_charm_in_instance(
...
         )
         base_configuration = providers.get_base_configuration(
-            alias=build_base,
+            alias=build_base_alias,
             instance_name=instance_name,
         )
 
+        if build_on.name != "ubuntu":
+            allow_unstable = True
+            emit.message(f"Will use {build_on.name} on unstable daily image for build.")
```
mr-cal, line 363: Can you more explicitly warn the user that CentOS is a "use at your own risk" base?

I'm thinking something like `Warning: Base {build_on.name} {build_on.channel} daily image may be unstable.`

charmcraft/commands/build.py
```diff
@@ -344,7 +345,7 @@ def pack_charm_in_instance(
             "(may take a while the first time but it's reusable)"
         )
 
-        build_base = providers.BASE_CHANNEL_TO_PROVIDER_BASE[build_on.channel]
+        build_base_alias = get_base_alias((build_on.name, build_on.channel))
```
mr-cal, line 348: Thinking back, I think we could have made this function take the name and channel as separate parameters, rather than a tuple.  Oh well.

tests/spread/smoketests/basic-centos-7/task.yaml
```diff
@@ -0,0 +1,34 @@
+summary: pack a simple init-created charm for CentOS 7
+systems:
+  - ubuntu-18.04-64
+kill-timeout: 30m
+
+environment:
```
mr-cal, line 6: Is this to clear the environment?  If not, you can remove this.

tests/test_providers.py
```diff
@@ -546,20 +547,65 @@ def test_get_base_configuration(
...
+    "platform, snap_channel, expected_snap_channel",
+    [
+        ("linux", None, None),
+        ("linux", "edge", "edge"),
+        ("darwin", "edge", "edge"),
+        # default to stable on non-linux system
+        ("darwin", None, "stable"),
+    ],
+)
+@pytest.mark.parametrize(
```
mr-cal, line 574: pedantic: premature abstraction: Unless you're going to add another CentOS base in the near future, I wouldn't parameterize this.

mr-cal (changes requested):
Looking great!  I only have minor feedback.

charmcraft/commands/build.py
syu-w: And convert to tuple in the function? Maybe more flexible.

charmcraft/commands/build.py
syu-w: OK

Issue #331, opened 2021-05-04
unregister command


jamesbeedy:
Hello,

I have registered a charm for the wrong username. I need to unregister the charm from my personal user and register it under a different username.

Is it possible for charmcraft to support some functionality that will allow a user to release the charm name registration?

Thanks

facundobatista:
Opened [this bug](https://bugs.launchpad.net/snapstore-server/+bug/1945220) requesting for the functionality in Charmhub.

facundobatista:
The API for this is [now in place](https://api.charmhub.io/docs/default.html#unregister_package)

lengau:
Sorry, we overlooked closing this, but this has been in charmcraft for about a year now :-)

PR #1069, opened 2023-04-13
store: add unregister command

lengau:
Fixes CRAFT-882 and #331

As soon as [craft-store v2.4.0 is released](https://github.com/canonical/craft-store/pull/85) the tests should succeed.


lengau:
Please note that the snap-store test failure is unrelated to this PR and I'm working on it separately.

charmcraft/commands/store/__init__.py
```diff
@@ -386,6 +386,40 @@ def run(self, parsed_args):
         emit.message(f"You are now the publisher of bundle {parsed_args.name!r} in Charmhub.")
 
 
+class UnregisterNameCommand(BaseCommand):
+    """Register a bundle name in the Store."""
```
tigarmo, line 390:
```suggestion
    """Unregister a name in the Store."""
```

charmcraft/commands/store/__init__.py
```diff
@@ -386,6 +386,40 @@ def run(self, parsed_args):
...
+
+    name = "unregister"
+    help_msg = "Unregister a name in the Store"
+    overview = textwrap.dedent(
+        """
+        Unregister a name in the Store.
+
+        Unregister a name from CharmHub if no revisions have been uploaded.
```
tigarmo, line 398:
```suggestion
        Unregister a name from Charmhub if no revisions have been uploaded.
```

setup.py
```diff
@@ -27,13 +27,13 @@
 
 install_requires = [
     "craft-cli",
-    "craft-parts",
+    "craft-parts>=1.18",
```
tigarmo, line 30: interesting; how come these are necessary?

tests/commands/test_store_api.py
```diff
@@ -527,6 +527,75 @@ def test_register_name_unauthorized_logs_in(client_mock, config):
...
+            id="exsting_revision",
+        ),
+        pytest.param(
+            FakeResponse("Charms can only be unregistered by their publisher", 401),
+            StoreServerError,
+            id="collaborator",
+        ),
+    ],
+)
+def test_unregister_name_errors(client_mock, config, http_response: FakeResponse, error_cls):
```
tigarmo, line 563: nice!

tigarmo (changes requested):
Nice! Some minor typos that we might as well handle now

setup.py
lengau: For craft-parts: Not entirely sure, but one of my environments had craft-parts 1.17 and stuff broke

For craft-store: 2.4 introduces the `unregister_name` method on the store client

For pydantic: Something got me 1.8.2 for some reason in an environment and it broke things. *technically* we need pydantic 1.10+ since craft-store depends on it, but if we fall back to an earlier craft-store (understanding that unregistering won't work) charmcraft still works fine.

charmcraft/commands/store/__init__.py
lengau: Thanks!

PR #1073, opened 2023-04-17
version: import logic from snapcraft

sergiusens:
Define version at package build time instead of at runtime. Import the version from package resources at runtime.

This change requires Windows builds to set the version prior to running.


mr-cal (approved):
Looks great and is aligned with snapcraft!

PR #1075, opened 2023-04-17
ci: only require CHARMCRAFT_AUTH for store spread

sergiusens:
Separate the logic into two in the Github Actions Workflow to always run spread for the general tests, while limiting store tests to having CHARMCRAFT_AUTH in GHA Secrets.

The required environment has been moved to the store suite in spread. Also making it run less test repetition by limiting them to Ubuntu 22.04 and avoiding it being run in the main testing phase for spread by marking them as manual.


mr-cal (approved):
Looks great!

PR #1080, opened 2023-04-18
store: ensure promote-bundle succeeds if there are no applications

lengau:
Depends on #1075.

All changes for this PR are in charmcraft/commands/store/__init__.py


lengau:
Closing for #1081 which will run the store tests

PR #1082, opened 2023-04-18
tools: update renovate config

lengau:


syu-w:
What is this regex supposed to match? I try use it to match `tox.ini` and got weird results.

sergiusens (approved):

lengau:
@syu-w it matches line pairs for dependencies, but it's also weirdly escaped to deal with JSON. The regex is also an RE2 regex, which is slightly different from Python's regexes.

In the end, it matches line pairs like:

```
    # renovate: datasource=pypi
    codespell[tomli]==2.2.4
```

The first matching group becomes becomes the data source in renovate (pypi above). The second matching group matches the package name, with the change being that it won't match extensions. extensions. Previously it was matching `codespell[tomli]`, but now it just matches `codespell`. I also changed it to be more flexible about equalities on versions (allowing `>=` and `<=` in addition to `==`). The final matching group there gets the version number of the package. This tells renovate how to parse the package and how to substitute the values in order to update the version number.

PR #1087, opened 2023-04-19
snap: pick the version from setup.py

sergiusens:


sergiusens:
Merging to get the version back from builds

PR #1089, opened 2023-04-19
store: fix issues with promote-bundle command

lengau:
No new tests because this fixes breakage of a spread test that wasn't running properly


charmcraft/commands/store/store.py
```diff
@@ -337,7 +337,7 @@ def list_revisions(self, name):
         return result
 
     @_store_client_wrapper()
-    def release(self, name, revision, channels, resources):
+    def release(self, name: str, revision: int, channels: List[str], resources):
```
lengau, line 340: Yes this is very incomplete, but it improved it when I was finding issues.

This is probably going away in favour of using craft_store's upstream `release` eventually anyway.

sergiusens:
This seems to still present the issue I mentioned yesterday https://github.com/canonical/charmcraft/actions/runs/4739483864/jobs/8419892970?pr=1089#step:6:89

syu-w (approved):
Looks good

PR #1091, opened 2023-04-20
snap: ensure cryptography builds correctly on all architectures

lengau:
This is a quick and dirty workaround to make cryptography build properly.

I'll follow up with a more robust fix, but this should at least get the snap building on all architectures.

Remote build: https://launchpad.net/~lengau/+snap/snapcraft-charmcraft-1bebeb44e32b4aa985026c6f977fd0b8

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -73,10 +73,10 @@ confinement: classic
 parts:
   # The cryptography library (indirect dependency via craft-store) requires rust, but only ships wheels for amd64 and arm64.
   # As of cryptography 39, it requires rust v1.48.0 or newer, which is available from focal-updates.
-  # This part contains the full set of build dependencies for cryptography.
-  # NOTE: cryptography requires at least pip version 22 for building.
+  # This part builds the wheel for cryptography.
   cryptography-deps:
-    plugin: nil
+    plugin: dump
+    source: .
     build-packages:
       - cargo
       - rustc
@@ -89,10 +89,9 @@ parts:
       - python3-pip
       - python3-setuptools-scm
     override-build: |
-      pip install 'pip>=22.0' 'setuptools>=61.0.0'
       # Dirty workaround - builds the cryptography wheel for later installation.
       # This must be exactly the version referenced in requirements.txt, so this is fragile.
-      pip wheel cryptography==40.0.2
+      pip wheel $(grep '^cryptography[<=>]' requirements.txt)
 
   # Classic core20 snaps require staged python.
   python3:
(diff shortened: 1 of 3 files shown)


tigarmo (approved):

snap/snapcraft.yaml
```diff
@@ -71,8 +71,33 @@ grade: stable
...
+      - pkg-config
+      - python3.8-dev
+      - libpython3.8-dev
+      - libssl-dev
+      - libffi-dev
+      - python3-wheel
+      - python3-pip
+      - python3-setuptools-scm
+    override-build: |
+      pip install 'pip>=22.0' 'setuptools>=61.0.0'
```
sergiusens, line 92: you can probably install this on https://github.com/canonical/charmcraft/pull/1091/files#diff-56759910381a014fecfd7556dd72ddd68c747d922a5b7df2044b9ce7c552f5f5R138

snap/snapcraft.yaml
```diff
@@ -71,8 +71,33 @@ grade: stable
...
+      # Dirty workaround - builds the cryptography wheel for later installation.
+      # This must be exactly the version referenced in requirements.txt, so this is fragile.
+      pip wheel cryptography==40.0.2
+      snapcraftctl build
```
sergiusens, line 96: this is not needed for nil

snap/snapcraft.yaml
```diff
@@ -71,8 +71,33 @@ grade: stable
...
+      - libssl-dev
+      - libffi-dev
+      - python3-wheel
+      - python3-pip
+      - python3-setuptools-scm
+    override-build: |
+      pip install 'pip>=22.0' 'setuptools>=61.0.0'
+      # Dirty workaround - builds the cryptography wheel for later installation.
+      # This must be exactly the version referenced in requirements.txt, so this is fragile.
+      pip wheel cryptography==40.0.2
```
sergiusens, line 95: how about (but in the charmcraft part, with the other suggestions)

```
parts:
  charmcraft:
    ..
    python-packages: [wheel, setuptools>=61.0.0, pip>=22.0]
    override-build: |
      pip wheel `$(grep` ^cryptography= requirements.txt)
      snapcraftctl build
```

This might work and be more straightforward

lengau: Whenever I try to build the cryptography wheel from another part it fails. At first I thought this was because `python-packages` was only getting installed during the build step, but even installing setuptools and pip beforehand was messing with it.

sergiusens: ah, it might be the python plugin installing it's dependencies

lengau: I riffed on what you did and came up with something much better than what I'd had. 

I've also introduced a workflow test for remote builds, but commented it out until we can get launchpad credentials as appropriate for the build.

PR #1114, opened 2023-05-30
config: support use actions in charmcraft.yaml

syu-w:
This allow we define actions in the same file rather than separated actions.yaml.

This also included some refactor to avoid circular imports.

CRAFT-1746

### DIFF
--- a/charmcraft/config.py
+++ b/charmcraft/config.py
@@ -45,6 +45,10 @@ analysis:
     attributes: [list of attribute names to ignore]
     linting: [list of linter names to ignore]
 
+actions:
+  my-action:
+    description: Action as defined at https://juju.is/docs/sdk/actions
+
 
 Object Definitions
 ==================
--- a/charmcraft/metafiles/__init__.py
+++ b/charmcraft/metafiles/__init__.py
@@ -16,4 +16,4 @@
 
 """Submodule for handling the all extra yaml files."""
 
-__all__ = ["manifest", "metadata"]
+__all__ = ["actions", "manifest", "metadata"]
--- a/charmcraft/metafiles/actions.py
+++ b/charmcraft/metafiles/actions.py
@@ -1,0 +1,44 @@
+# Copyright 2023 Canonical Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# For further info, check https://github.com/canonical/charmcraft
+
+"""Charmcraft project handle actions.yaml file."""
+
+import pathlib
+import logging
+from typing import Any, Dict, Optional
+
+import yaml
+
+logger = logging.getLogger(__name__)
+
+
+def create_actions(
+    basedir: pathlib.Path,
+    actions: Optional[Dict[str, Any]] = None,
+) -> Optional[pathlib.Path]:
+    """Create actions.yaml in basedir for given project configuration.
+
+    :param basedir: Directory to create Charm in.
+    :param actions: Relevant bases configuration, if any.
+
+    :returns: Path to created actions.yaml.
+    """
+    if actions is None:
+        return None
+
+    filepath = basedir / "actions.yaml"
+    filepath.write_text(yaml.dump(actions))
+    return filepath
--- a/charmcraft/models/config.py
+++ b/charmcraft/models/config.py
@@ -18,6 +18,9 @@
 
 import datetime
 import pathlib
+import os
+import re
+import keyword
 from typing import Optional, List, Dict, Any
 
 import pydantic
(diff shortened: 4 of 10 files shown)


charmcraft/metafiles/actions.py
```diff
@@ -0,0 +1,44 @@
...
+import pathlib
+import logging
+from typing import Any, Dict, Optional
+
+import yaml
+
+logger = logging.getLogger(__name__)
+
+
+def create_actions(
```
lengau, line 28: Are we going to have to do more with actions later? (e.g. validate them, lint them, etc.?) If so, we should probably go ahead and make a model for this, even if it's very basic right now.

Even if we don't make a full model, the `actions` dict is probably a `Dict[ActionName, pydantic.json_schema.JsonSchemaValue]`, with `ActionName` being either just `str` or a constrained string that does some basic validation (the regex `[A-z_-][A-z0-9_-]*` meets the documentation's requirements).

charmcraft/metafiles/actions.py
```diff
@@ -0,0 +1,44 @@
...
+) -> Optional[pathlib.Path]:
+    """Create actions.yaml in basedir for given project configuration.
+
+    :param basedir: Directory to create Charm in.
+    :param actions: Relevant bases configuration, if any.
+
+    :returns: Path to created actions.yaml.
+    """
+    if actions is None:
+        return
```
lengau, line 40: Please explicitly return `None` here — we're starting to standardise on that IIRC.

As a side note: Is this the right place to put the logic of it being optional? (That question doesn't have any implications - it's a genuine yes/no question. On the one hand we could keep it as is but potentially have to wrap logic around the result of this method. OTOH, we could make this method always write the file and have to write `if self.config.actions:` in the existing spots.)

tests/spread/smoketests/metafiles/actions/task.yaml
```diff
@@ -0,0 +1,31 @@
...
+
+include: 
+  - tests/
+
+prepare: |
+  tests.pkgs install unzip
+  charmcraft init --project-dir=charm
+  cat <<- EOF >> charmcraft.yaml
+  actions:
+    validate-configuration:
```
lengau, line 11: Here's an example of where validation for `actions` could come in handy. I'm guessing eventually we're going to want to ensure that the charm has an `_on_validate_configuration_action` method based on this? (That feels like something we should be catching before it becomes Somebody Else's Problem™.)

charmcraft/config.py
```diff
@@ -45,6 +45,9 @@
     attributes: [list of attribute names to ignore]
     linting: [list of linter names to ignore]
 
+actions:
+  whatever:
```
lengau, line 49:
```suggestion
actions:
  my-action:
    description: Action as defined at https://juju.is/docs/sdk/actions
```

charmcraft/format.py
```diff
@@ -0,0 +1,105 @@
...
+            raise RuntimeError(f"unhandled loc: {loc_part}")
+
+    loc = ".".join(loc_parts)
+
+    # Filter out internal __root__ detail.
+    loc = loc.replace(".__root__", "")
+    return loc
+
+
+def format_pydantic_error_message(msg):
```
lengau, line 44:
```suggestion
def format_pydantic_error_message(msg: str) -> str:
```

Since this is moving anyway :smiley:

lengau (commented):
Thanks for doing this! I know I've put a lot of comments here, but most of it is just requesttn for minor "additional cleanup" from the refactoring that I probably wouldn't have noticed had I been looking at the source files rather than the diff.

charmcraft/metafiles/__init__.py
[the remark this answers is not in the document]
syu-w: I'm curious why we're currently doing this. We don't use `from X import *`, and it seems others also not use that way.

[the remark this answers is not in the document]
syu-w: I don't see anything `__all__` related to the import order that are not `form A import *`, unless it isn't documented.

For import orders are most likely done by:
```
form a.d import d
from a.b import b
```
in `__init__.py`

tests/test_actions.py
```diff
@@ -0,0 +1,38 @@
...
+# For further info, check https://github.com/canonical/charmcraft
+
+import os
+
+import yaml
+
+from charmcraft.metafiles.actions import create_actions
+
+
+def test_create_actions_yaml(tmp_path):
```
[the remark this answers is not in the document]
syu-w: Added complex example form sdk docs.

charmcraft/metafiles/metadata.py
```diff
@@ -14,18 +14,20 @@
...
 
-from charmcraft.config import format_pydantic_errors
+from charmcraft.format import format_pydantic_errors
+from charmcraft.const import CHARM_METADATA
 
-CHARM_METADATA = "metadata.yaml"
+logger = logging.getLogger(__name__)
 
 
 class CharmMetadata(pydantic.BaseModel, frozen=True, validate_all=True):
```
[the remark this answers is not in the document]
syu-w: I moved it to the models, and we probably need to refactor it later to move all other models.

charmcraft/metafiles/actions.py
syu-w: If you only want to do sanity check that can be done. As for now it will check if it is a valid yaml.

syu-w: I added some sanity check for the actions name, but not full model.

charmcraft/models/config.py
```diff
@@ -225,6 +228,35 @@ def validate_bases_presence(cls, bases, values):
...
+        And individual "actions.yaml" should not exists when actions
+        is defined in charmcraft.yaml.
+        """
+        if actions is None:
+            return None
+
+        actions_yaml_file_path = values["project"].dirpath / "actions.yaml"
+        if os.path.isfile(actions_yaml_file_path):
+            raise ValueError(
+                "'actions.yaml' file should not be used when 'actions' section is defined"
```
lengau, line 245:
```suggestion
                "'actions.yaml' file should not be used when an 'actions' section is defined in 'charmcraft.yaml'"
```

tests/spread/smoketests/metafiles/actions/task.yaml
```diff
@@ -0,0 +1,49 @@
+summary: create a charm that use actions in charmcraft.yaml
+
+include: 
+  - tests/
+
+prepare: |
+  tests.pkgs install unzip
+  charmcraft init --project-dir=charm
+  cat <<- EOF >> charmcraft.yaml
+  pause:
```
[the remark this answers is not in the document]
syu-w: Yes

lengau (approved):
Looks great, thank you!

PR #1117, opened 2023-06-01
refactor: move some pydantic models and metafiles

syu-w:
This should reduce some of the circular imports.

CRAFT-1746


lengau (approved):

charmcraft/metafiles/__init__.py
```diff
@@ -0,0 +1,19 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# For further info, check https://github.com/canonical/charmcraft
+
+"""Submodule for handling the all extra yaml files."""
+
+__all__ = ["manifest", "metadata"]
```
lengau, line 19:
```suggestion
from charmcraft.metafiles import manifest, metadata

__all__ = ["manifest", "metadata"]
```

Upcoming linter changes are probably going to include complaints about exports that aren't explicitly imported :-)

charmcraft/metafiles/metadata.py
```diff
@@ -14,47 +14,27 @@
 #
 # For further info, check https://github.com/canonical/charmcraft
 
-"""Logic related to metadata.yaml."""
+"""Charmcraft project handle metadata.yaml file."""
```
lengau, line 17:
```suggestion
"""Handlers for metadata.yaml file."""
```

charmcraft/models/__init__.py
```diff
@@ -0,0 +1,19 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# For further info, check https://github.com/canonical/charmcraft
+
+"""Charmcraft pydantic models."""
+
+__all__ = ["config", "metadata"]
```
lengau, line 19:
```suggestion
from charmcraft.models import config, metadata

__all__ = ["config", "metadata"]
```

charmcraft/models/config.py
```diff
@@ -0,0 +1,286 @@
...
+
+
+class CustomStrictStr(pydantic.StrictStr):
+    """Generic class to create custom strict strings validated by pydantic."""
+
+    @classmethod
+    def __get_validators__(cls):
+        """Yield the relevant validators."""
+        yield from super().__get_validators__()
+        yield cls.custom_validate
```
lengau, line 44: Did this get moved from elsewhere or added?

If added: This technically makes the class abstract, right? (Since there's no `custom_validate` implementation?) If so, please add an abstract method for `custom_validate` that we override below.

(If it's just been moved, I'll make an issue for it so we can do that cleanup separately)

lengau (changes requested):
Thanks for bearing with me! Two minor things for upcoming linter changes (not actually functional issues), one grammatical suggestion and one question :-)

charmcraft/models/config.py
syu-w: This one is moved, I did not touch the code.

lengau: Thanks! I've made an issue for it. https://github.com/canonical/charmcraft/issues/1118

lengau (approved):
Looks great, thank you!

PR #1126, opened 2023-06-16
Allow use actions/config/metadata in charmcraft.yaml

syu-w:
Allow use actions/config/metadata in `charmcraft.yaml` when corresponding yaml file does not exist.

Metadata in `charmcraft.yaml` also must be new keys defined in spec ST087.

CRAFT-1747
CRAFT-1788
CRAFT-1789


charmcraft/models/actions.py
```diff
@@ -0,0 +1,58 @@
...
+
+    See also: https://juju.is/docs/sdk/actions
+    """
+
+    _action_name_regex = re.compile(r"^[a-zA-Z_][a-zA-Z0-9-_]*$")
+    actions: Optional[Dict[str, Dict]]
+    legacy: bool = False
+
+    @pydantic.validator("actions", pre=True)
+    def validate_actions(cls, actions, values):
```
lengau, line 40: Let's break this into two validators: One that validates the actions names and one that validates each action item.

charmcraft/models/actions.py
```diff
@@ -0,0 +1,58 @@
...
+    """Juju actions for charms.
+
+    See also: https://juju.is/docs/sdk/actions
+    """
+
+    _action_name_regex = re.compile(r"^[a-zA-Z_][a-zA-Z0-9-_]*$")
+    actions: Optional[Dict[str, Dict]]
+    legacy: bool = False
+
+    @pydantic.validator("actions", pre=True)
```
lengau, line 39: Why does this need to be a pre validator? (Please include the reasoning in the docstring or a comment so it doesn't get lost to time)

charmcraft/models/charmcraft.py
```diff
@@ -0,0 +1,365 @@
...
+    storage: Optional[Dict[str, Any]]
+    subordinate: Optional[bool]
+    terms: Optional[List[str]]
+    links: Optional[Links]
+    config: Optional[JujuConfig]
+
+    @pydantic.validator("type")
+    def validate_charm_type(cls, charm_type):
+        """Verify charm type is valid with exception when instantiated without YAML."""
+        if charm_type not in ["bundle", "charm"]:
```
lengau, line 146: Can't you just make `type` be `Literal["bundle", "charm"]` to get this same effect?

charmcraft/models/charmcraft.py
```diff
@@ -0,0 +1,365 @@
...
+            raise ValueError("must be either 'charm' or 'bundle'")
+        return charm_type
+
+    @pydantic.validator("name", pre=True, always=True)
+    def validate_name(cls, name, values):
+        """Verify charm name is valid with exception when instantiated without YAML."""
+        if values.get("type") == "charm" and not name:
```
lengau, line 153: Given the number of these, could we make separate models for if it's a charm or a bundle and a factory for generating the correct model?

tests/spread/smoketests/metafiles/actions/task.yaml
lengau, line 1: Which test replaces this?

tests/spread/smoketests/metafiles/charmcraft_yaml/task.yaml
```diff
@@ -0,0 +1,102 @@
...
+  
+  rm -rf charm
+
+execute: |
+  cd charm
+  charmcraft pack --verbose
+  test -f charm*.charm
+  unzip -l charm*.charm | grep "actions.yaml"
+  unzip -l charm*.charm | grep "config.yaml"
+  unzip -l charm*.charm | grep "metadata.yaml"
```
lengau, line 98:
```suggestion
  unzip -l charm*.charm | MATCH "actions.yaml"
  unzip -l charm*.charm | MATCH "config.yaml"
  unzip -l charm*.charm | MATCH "metadata.yaml"
```

Uses spread's [built-in MATCH function](https://github.com/snapcore/spread#functions)

tests/test_main.py
```diff
@@ -206,7 +216,17 @@ def run(self, parsed_args):
                 """
             ),
         ],
-        ["type: charm", None],
+        [
+            dedent(
+                """\
+                type: charm
+                name: test-charm-name-from-metadata-yaml
```
lengau, line 223:
```suggestion
                name: test-charm-name-from-charmcraft-yaml
```

if I'm reading this correctly?

tests/test_main.py
```diff
@@ -206,7 +216,17 @@ def run(self, parsed_args):
                 """
             ),
         ],
-        ["type: charm", None],
+        [
```
lengau, line 219: The old version test should still work, right?

lengau (changes requested):
Wow, this is a lot! :slightly_smiling_face: 

I've done a first pass here — mostly asking questions, but there are a couple of things that should go into their own refactors.

charmcraft/models/metadata.py
```diff
@@ -25,18 +25,88 @@
 from charmcraft.const import METADATA_FILENAME
 
 
-class CharmMetadata(pydantic.BaseModel, frozen=True, validate_all=True):
-    """Object representing metadata.yaml contents."""
+class CharmMetadataLegacy(
+    pydantic.BaseModel,
+    extra=pydantic.Extra.forbid,
```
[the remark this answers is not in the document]
syu-w: OK, didn't notice it allow arbitrary keys.

charmcraft/models/metadata.py
```diff
@@ -25,18 +25,88 @@
...
+            obj["maintainers"] = [obj["maintainer"]]
+            del obj["maintainer"]
+
+        try:
+            return cls.parse_obj(obj)
+        except pydantic.error_wrappers.ValidationError as error:
+            raise CraftError(format_pydantic_errors(error.errors(), file_name=METADATA_FILENAME))
+
+
+class BundleMetadataLegacy(
```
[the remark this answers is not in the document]
syu-w: They are suppose to be the same by the docs, but in reality they are different. Docs says it at least needs `name`, `summary`, and `description`, but I found most bundles only have `name`.

charmcraft/linters.py
```diff
@@ -184,7 +184,7 @@ def _check_operator(self, basedir: pathlib.Path) -> bool:
     def _check_reactive(self, basedir: pathlib.Path) -> bool:
         """Detect if the Reactive Framework is used."""
         try:
-            metadata = parse_metadata_yaml(basedir)
+            metadata = parse_charm_metadata_yaml(basedir).dict()
```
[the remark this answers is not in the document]
syu-w: Because pydantic model is read only after this.

charmcraft/models/actions.py
syu-w: OK

charmcraft/models/actions.py
syu-w: I don't think it need `pre` in this case.

charmcraft/models/charmcraft.py
syu-w: That's possible, just another refactor.

charmcraft/models/config.py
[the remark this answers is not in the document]
syu-w: This is not a refactor, but a file move and reuse the same file name.

tests/spread/smoketests/metafiles/actions/task.yaml
syu-w: `combine_yaml` test

tests/test_main.py
syu-w: No, "type: charm" only, without metadata will now fail pydantic model.

tests/test_main.py
syu-w: Yes

charmcraft/models/config.py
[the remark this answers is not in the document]
lengau: That file move (and its associated import changes) are a refactor :-)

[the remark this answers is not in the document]
syu-w: That part is -_-

tests/test_main.py
lengau: Ahh, okay. Makes sense.

I'm trying to determine whether there are any use cases here that we're breaking that we care not to break. For example, if a user wants to run a store command on a custom store they might create a `charmcraft.yaml` file that has just `type: charm` and a `charmhub` section. I'm not sure we care about backwards compatibility with that case though.

charmcraft/linters.py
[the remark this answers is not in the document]
lengau: I'm not sure I understand. Doesn't `parse_charm_metadata_yaml` already parse the YAML?

tests/test_main.py
syu-w: I don't think that a charm without a name is allowed in any case.

charmcraft/linters.py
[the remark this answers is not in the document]
syu-w: Oh, I see what you mean. That `dict()` is no longer needed. It was a result of some refactoring I did halfway through.

tests/test_main.py
lengau: It's not — the idea is that this would be a file that just contains charmcraft configuration for use with commands that don't require a charm. However, in the last day I haven't come up with a reason to support this use case, so I'm going to let it rest.

charmcraft/commands/store/__init__.py
```diff
@@ -999,10 +999,10 @@ def run(self, parsed_args: "Namespace") -> None:
                 emit.debug(f"Error when running PRIME step: {error}")
                 raise
 
-            from charmcraft.metafiles.actions import create_actions
+            from charmcraft.metafiles.metadata import create_metadata_yaml
```
lengau, line 1002: Oh my, I don't remember why I did these imports here :flushed: 

I sure hope they were because importing them at the module level was bad

tests/test_models.py
```diff
@@ -0,0 +1,1194 @@
...
+                        architectures=[get_host_architecture()],
+                    )
+                ],
+            }
+        )
+    ]
+    assert config.metadata_legacy
+
+
+def test_load_minimal_metadata_from_metadata_yaml_missing_name(
```
lengau, line 222: These could be parametrized, but I'm not going to hold up the PR on that — we can do it another time.

tests/test_models.py
```diff
@@ -0,0 +1,1194 @@
...
+        CraftError, match="field 'description' required in top-level configuration"
+    ):
+        load(tmp_path)
+
+
+def test_load_full_metadata_from_charmcraft_yaml(tmp_path, prepare_charmcraft_yaml):
+    """Load a charmcraft.yaml with full metadata. (Spec ST087)"""
+    prepare_charmcraft_yaml(
+        dedent(
+            """
```
lengau, line 338: @dboddie this might be worth bookmarking for when we tackle improvements to charmcraft's docs. This (and its companion at ST087) might be the most complete description of charms that we have.

tests/spread/smoketests/metafiles/combine_yaml/task.yaml
```diff
@@ -0,0 +1,125 @@
...
+  cd charm
+  charmcraft pack --verbose
+  test -f charm*.charm
+
+  unzip -l charm*.charm | MATCH "actions.yaml"
+  unzip -l charm*.charm | MATCH "config.yaml"
+  unzip -l charm*.charm | MATCH "metadata.yaml"
+
+  unzip -p actions.yaml charm*.charm | MATCH "description: The type of compression to use."
+  unzip -p actions.yaml charm*.charm | MATCH "#### TEST-COPY ####"
```
lengau, line 116: This is really clever!

charmcraft/models/config.py
```diff
@@ -0,0 +1,51 @@
...
+    See also: https://juju.is/docs/sdk/config
+    """
+
+    options: Optional[Dict[str, Dict]]
+    legacy: bool = False
+
+    @pydantic.validator("options", pre=True)
+    def validate_actions(cls, options, values):
+        """Verify options section."""
+        for option in options:
```
lengau, line 38: No strong preferences, but I'm used to seeing this as something like:

```
for name, option in options.items():
```

charmcraft/models/config.py
```diff
@@ -0,0 +1,51 @@
...
+            if not isinstance(options[option], dict):
+                raise ValueError(f"'{option}' is not a dictionary")
+
+            if not set(options[option].keys()).issubset({"description", "type", "default"}):
+                raise ValueError(f"'{option}' has an invalid key")
```
lengau, line 43: We should list the invalid keys in that option.

charmcraft/models/config.py
```diff
@@ -0,0 +1,51 @@
...
+
+            if "type" not in options[option]:
+                raise ValueError(f"'{option}' is missing a type")
+
+            if options[option]["type"] not in ["string", "int", "float", "boolean"]:
+                raise ValueError(f"'{option}' has an invalid type")
```
lengau, line 49:
```suggestion
                invalid_type = options[option]["type"]
                raise ValueError(f"'{option}' type {invalid_type!r} is invalid. Should be one of ['string', 'int', 'float', 'boolean']")
```

or something like that :-)

lengau (approved):
Please consider my suggestions in `charmcraft/models/config.py`, but whether you implement them or not is up to you. I'm fine either way!

Thanks for all of this! That was a LOT of work!

charmcraft/const.py
```diff
@@ -40,34 +43,98 @@
...
+MANDATORY_HOOK_NAMES = frozenset(("install", "start", "upgrade-charm"))
 HOOKS_DIRNAME = "hooks"
 
 # The minimum set of files for a charm to be considered valid
-CHARM_FILES = [
-    METADATA_FILENAME,
-    DISPATCH_FILENAME,
-    HOOKS_DIRNAME,
-]
+CHARM_FILES = frozenset(
```
lengau, line 50: :heart:

lengau (approved):
I hereby reaffirm my approval! :tada:

charmcraft/const.py
syu-w: The only downside is some test is order sensitive when this changed to unordered, had to add a workaround.

PR #1130, opened 2023-06-22
bases: allow builds with ubuntu non-lts

syu-w:
Allow user create charms with non-LTS Ubuntu versions.
User will get warning when building on those versions.

CRAFT-1759

### DIFF
--- a/charmcraft/commands/build.py
+++ b/charmcraft/commands/build.py
@@ -35,7 +35,13 @@ import charmcraft.instrum
 from charmcraft.metafiles.metadata import parse_metadata_yaml
 from charmcraft.metafiles.actions import create_actions
 from charmcraft.metafiles.manifest import create_manifest
-from charmcraft.const import BUILD_DIRNAME, CHARM_FILES, CHARM_OPTIONAL, VENV_DIRNAME
+from charmcraft.const import (
+    BUILD_DIRNAME,
+    CHARM_FILES,
+    CHARM_OPTIONAL,
+    VENV_DIRNAME,
+    UBUNTU_LTS_STABLE,
+)
 from charmcraft.commands.store.charmlibs import collect_charmlib_pydeps
 from charmcraft.models.config import Base, BasesConfiguration
 from charmcraft.parts import Step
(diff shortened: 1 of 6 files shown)


charmcraft/commands/build.py
```diff
@@ -341,13 +347,20 @@ def pack_charm_in_instance(
...
 
-        if build_on.name != "ubuntu":
+        if build_on.name == "ubuntu":
+            if build_on.channel in UBUNTU_LTS_STABLE:
+                allow_unstable = False
+            else:
+                allow_unstable = True
+                emit.message(
+                    f"Warning: non-LTS Ubuntu {build_on.channel} are intended for "
+                    "experimental use only."
```
lengau, line 357:
```suggestion
                    f"Warning: non-LTS Ubuntu releases {build_on.channel} are "
                    "intended for experimental use only."
```

tests/test_providers.py
```diff
@@ -711,7 +709,11 @@ def test_is_base_available(
     base = Base(name=name, channel=channel, architectures=architectures)
     valid, reason = providers.is_base_available(base)
 
-    assert (valid, reason) == (expected_valid, expected_reason)
+    assert valid == expected_valid
+    if reason is None:
+        assert expected_reason == reason
+    else:
+        assert re.match(expected_reason, reason)
```
lengau, line 716:
```suggestion
        assert re.fullmatch(expected_reason, reason)
```

This will ensure that the whole string matches the regex rather than a substring

requirements.txt
```diff
@@ -4,7 +4,8 @@ cffi==1.15.1
 charset-normalizer==3.1.0
 craft-cli==1.2.0
 craft-parts==1.20.0
-craft-providers==1.12.0
+#craft-providers==1.12.0
+git+https://github.com/syu-w/craft-providers.git@support-23.04#egg=craft-providers
```
lengau, line 8: I'm guessing we're on merge hold for this until we get craft-providers 1.13 for https://github.com/canonical/craft-providers/pull/330

lengau (approved):
Looks good, thank you! Just a couple of very minor suggestions

requirements.txt
syu-w: We don't need that PR for 23.04, but need craft-providers 1.13.

PR #1131, opened 2023-06-22
tests: refactor for variable multi-file support

syu-w:
Allow test cases to create multiple config yaml files.

### DIFF
--- a/tests/commands/test_build.py
+++ b/tests/commands/test_build.py
@@ -35,7 +35,6 @@ from charmcraft.bases import get_host_as_base
 from charmcraft.commands.build import BUILD_DIRNAME, Builder, format_charm_file_name, launch_shell
 from charmcraft.models.config import Base, BasesConfiguration
 from charmcraft.config import load
-from charmcraft.const import METADATA_FILENAME
 from charmcraft.providers import get_base_configuration
 from charmcraft.utils import get_host_architecture
 
@@ -64,20 +63,10 @@ def get_builder(
 
 
 @pytest.fixture
-def basic_project(tmp_path, monkeypatch, create_config):
+def basic_project(tmp_path, monkeypatch, prepare_charmcraft_yaml, prepare_metadata_yaml):
     """Create a basic Charmcraft project."""
     build_dir = tmp_path / BUILD_DIRNAME
     build_dir.mkdir()
-
-    # the metadata
-    metadata_data = {
-        "name": "name-from-metadata",
-        "summary": "test-summ",
-        "description": "text",
-    }
-    metadata_file = tmp_path / "metadata.yaml"
-    metadata_raw = yaml.dump(metadata_data).encode("ascii")
-    metadata_file.write_bytes(metadata_raw)
 
     # a lib dir
     lib_dir = tmp_path / "lib"
@@ -108,19 +97,29 @@ def basic_project(tmp_path, monkeypatch, create_config):
 
     # the config
     host_base = get_host_as_base()
-    create_config(
-        f"""
-        type: charm
-        bases:
-          - name: {host_base.name}
-            channel: "{host_base.channel}"
-            architectures: {host_base.architectures!r}
-        parts:  # just to avoid "default charm parts" sneaking in
-          foo:
-            plugin: nil
-        """
+    prepare_charmcraft_yaml(
+        dedent(
+            f"""
+            type: charm
+            bases:
+              - name: {host_base.name}
+                channel: "{host_base.channel}"
+                architectures: {host_base.architectures!r}
+            parts:  # just to avoid "default charm parts" sneaking in
+              foo:
+                plugin: nil
+            """
+        )
     )
-
+    prepare_metadata_yaml(
+        dedent(
+            """\
+            name: test-charm-name-from-metadata-yaml
+            summary: test summary
+            description: test description
+            """
+        ),
+    )
     # paths are relative, make all tests to run in the project's directory
     monkeypatch.chdir(tmp_path)
 
@@ -128,27 +127,40 @@ def basic_project(tmp_path, monkeypatch, create_config):
 
 
 @pytest.fixture
-def basic_project_builder(basic_project):
+def basic_project_builder(basic_project, prepare_charmcraft_yaml):
     def _basic_project_builder(bases_configs: List[BasesConfiguration], **builder_kwargs):
-        charmcraft_file = basic_project / "charmcraft.yaml"
-        with charmcraft_file.open("w") as f:
-            print("type: charm", file=f)
-            print("bases:", file=f)
-            for bases_config in bases_configs:
-                print("- build-on:", file=f)
-                for base in bases_config.build_on:
-                    print(f"  - name: {base.name!r}", file=f)
-                    print(f"    channel: {base.channel!r}", file=f)
-                    print(f"    architectures: {base.architectures!r}", file=f)
+        charmcraft_yaml = dedent(
+            """
+            type: charm
+            bases:
+            """
+        )
 
-                print("  run-on:", file=f)
-                for base in bases_config.run_on:
-                    print(f"  - name: {base.name!r}", file=f)
-                    print(f"    channel: {base.channel!r}", file=f)
-                    print(f"    architectures: {base.architectures!r}", file=f)
-            print("parts:", file=f)
-            print("  foo:", file=f)
-            print("    plugin: nil", file=f)
+        for bases_config in bases_configs:
+            charmcraft_yaml += "  - build-on:\n"
+            for base in bases_config.build_on:
+                charmcraft_yaml += (
+                    f"    - name: {base.name!r}\n"
+                    f"      channel: {base.channel!r}\n"
+                    f"      architectures: {base.architectures!r}\n"
+                )
+
+            charmcraft_yaml += "    run-on:\n"
+            for base in bases_config.run_on:
+                charmcraft_yaml += (
+                    f"    - name: {base.name!r}\n"
+                    f"      channel: {base.channel!r}\n"
+                    f"      architectures: {base.architectures!r}\n"
+                )
+        charmcraft_yaml += dedent(
+            """
+            parts:
+              foo:
+                plugin: nil
+            """
+        )
+
+        prepare_charmcraft_yaml(charmcraft_yaml)
 
         config = load(basic_project)
         return get_builder(config, **builder_kwargs)
@@ -222,16 +234,17 @@ def mock_is_base_available():
 
 def test_build_error_without_metadata_yaml(basic_project):
     """Validate error if trying to build project without metadata.yaml."""
-    metadata = basic_project / METADATA_FILENAME
+    metadata = basic_project / "metadata.yaml"
     metadata.unlink()
 
-    config = load(basic_project)
+    with pytest.raises(CraftError) as exc_info:
+        config = load(basic_project)
 
-    msg = re.escape(
+        get_builder(config)
+
+    assert exc_info.value.args[0] == (
         "Cannot read the metadata.yaml file: FileNotFoundError(2, 'No such file or directory')"
     )
-    with pytest.raises(CraftError, match=msg):
-        get_builder(config)
 
 
 @pytest.mark.skipif(sys.platform == "win32", reason="Windows not [yet] supported")
@@ -246,7 +259,8 @@ def test_build_with_charmcraft_yaml_destructive_mode(basic_project_builder, emit
 
     host_arch = host_base.architectures[0]
     assert zipnames == [
-        f"name-from-metadata_{host_base.name}-{host_base.channel}-{host_arch}.charm"
+        "test-charm-name-from-metadata-yaml_"
+        f"{host_base.name}-{host_base.channel}-{host_arch}.charm"
     ]
 
     emitter.assert_debug("Building for 'bases[0]' as host matches 'build-on[0]'.")
@@ -268,7 +282,8 @@ def test_build_with_charmcraft_yaml_managed_mode(
 
     host_arch = host_base.architectures[0]
     assert zipnames == [
-        f"name-from-metadata_{host_base.name}-{host_base.channel}-{host_arch}.charm"
+        "test-charm-name-from-metadata-yaml_"
+        f"{host_base.name}-{host_base.channel}-{host_arch}.charm"
     ]
 
     emitter.assert_debug("Building for 'bases[0]' as host matches 'build-on[0]'.")
@@ -393,8 +408,9 @@ def test_build_multiple_with_charmcraft_yaml_destructive_mode(basic_project_buil
 
     host_arch = host_base.architectures[0]
     assert zipnames == [
-        f"name-from-metadata_{host_base.name}-{host_base.channel}-{host_arch}.charm",
-        "name-from-metadata_cross-name-cross-channel-cross-arch1.charm",
+        "test-charm-name-from-metadata-yaml_"
+        f"{host_base.name}-{host_base.channel}-{host_arch}.charm",
+        "test-charm-name-from-metadata-yaml_" "cross-name-cross-channel-cross-arch1.charm",
     ]
 
     reason = f"name 'unmatched-name' does not match host {host_base.name!r}."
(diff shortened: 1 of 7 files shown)


sergiusens (changes requested):
this does not look like a refactor, there are functional changes here

syu-w:
This is an intermediate process, for a separated PR I need to undo some changes.

tests/conftest.py
```diff
@@ -184,26 +193,69 @@ def is_provider_installed(cls) -> bool:
 
 
 @pytest.fixture
-def create_config(tmp_path):
+def create_config(tmp_path: pathlib.Path):
```
lengau, line 196: This is starting to get quite complex, to the point that it would be easy to modify and accidentally break some tests.

At this point I think I'd prefer if we create a data directory with various fake charms and use `shutil.copytree` to copy the relevant charm into `tmp_path`

tests/test_config.py
```diff
@@ -120,346 +144,382 @@ def check_schema_error(expected_msg):
         """The real checker."""
         with pytest.raises(CraftError) as cm:
             load(tmp_path)
-        assert str(cm.value) == dedent(expected_msg)
+        assert dedent(expected_msg) in str(cm.value)
```
lengau, line 147: This is a major behaviour change for any tests that use this fixture.

tests/test_config.py
```diff
@@ -120,346 +144,382 @@ def check_schema_error(expected_msg):
...
 
     return check_schema_error
 
 
-def test_schema_top_level_no_extra_properties(create_config, check_schema_error):
+@pytest.mark.parametrize(
```
lengau, line 152: Please make sure each of these tests also tests the version of the charm from the old version. In this case, an additional parameter set that contains the `charmcraft.yaml` from before without having a `metadata.yaml`

tests/test_linters.py
```diff
@@ -341,15 +342,23 @@ def test_framework_operator_no_ops_imported(tmp_path, monkeypatch, import_line):
...
+def test_framework_reactive_used_ok(create_config, import_line, charm_name, charm_module):
     """The reactive framework was used.
 
     Parametrized args:
     - import_line: different ways to express the import
     """
-    # metadata file with needed name field
-    metadata_file = tmp_path / "metadata.yaml"
-    metadata_file.write_text(f"name: {charm_name}")
+    tmp_path = create_config(
```
lengau, line 351: This is counterintuitive. Pytest will use the same `tmp_path` for all fixtures passed to a single test run, so it would be better to keep `tmp_path` as a direct fixture used by this test than to have `create_config` return the path.

tests/test_linters.py
```diff
@@ -341,15 +342,23 @@ def test_framework_operator_no_ops_imported(tmp_path, monkeypatch, import_line):
...
+        {
+            "metadata.yaml": dedent(
+                f"""
+                name: {charm_name}
```
lengau, line 355: Please ensure you're keeping the same charm as before. If you want a version that contains these extra fields, add a parametrize fixture to provide both.

tests/test_main.py
```diff
@@ -42,12 +42,17 @@
 @pytest.fixture
 def base_config_present(create_config, monkeypatch):
     tmp_path = create_config(
```
lengau, line 44: Like in the other files, please ensure you keep versions of these tests that test the same charm directory.

tests/test_metadata.py
```diff
@@ -47,25 +47,33 @@ def test_parse_metadata_yaml_complete(tmp_path):
...
-    metadata_file.write_text(f"name: {name}")
-
-    expected_error_msg = dedent(
-        """\
-        Bad metadata.yaml content:
-        - string type expected in field 'name'"""
-    )
-    with pytest.raises(CraftError, match=re.escape(expected_error_msg)):
-        parse_metadata_yaml(tmp_path)
+# @pytest.mark.parametrize("name", [1, "false", "[]"])
```
lengau, line 65: Work in progress?

tests/test_providers.py
```diff
@@ -711,7 +708,11 @@ def test_is_base_available(
     base = Base(name=name, channel=channel, architectures=architectures)
     valid, reason = providers.is_base_available(base)
 
-    assert (valid, reason) == (expected_valid, expected_reason)
+    assert valid == expected_valid
+    if reason is None:
```
lengau, line 712: This implies that we probably have two tests here at this point: One for valid bases and one for invalid bases, with a reason check.

At that point it may also be worth changing the reason check into a regex check. For example, the modified test case above could be the following regex if you use `fullmatch`:

```
base 'ubuntu' channel '10.04' is not yet supported \(must be '.+'\)
```

(Try to keep the regexes simple as with the above example — `.+` and `.*` are generally fine to use since they signal to the reader tha we don't care about the specific contents of that particular substring.)

lengau (changes requested):

syu-w:
@lengau just a quick clarification, this PR is reverse generation BEFORE doing `metadata.yaml` merge. All the parametrized that will test in both cases has been removed to pass the CI.

tests/commands/test_build.py
```diff
@@ -109,16 +98,21 @@ def basic_project(tmp_path, monkeypatch, create_config):
...
-        type: charm
-        bases:
-          - name: {host_base.name}
-            channel: "{host_base.channel}"
-            architectures: {host_base.architectures!r}
-        parts:  # just to avoid "default charm parts" sneaking in
-          foo:
-            plugin: nil
-        """
+        {
```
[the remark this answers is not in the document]
syu-w: This change result is identical before and after. It just moved create `metadata.yaml` part from this fixture to the `create_config()`.

parametrized is used for some cases, but not suitable for this one. e.g. error handle will need to remove some or all predefined values, include any parametrized values.

tests/conftest.py
syu-w: I though about that, but I don't think it will reduced those logic. The new `create_config()` is designed to work with/without parametrized fixtures. And if the value is not given, it will use the default value. Basically all related test cases will change some of the values that make copy and use cannot save much.

tests/test_linters.py
syu-w: OK

tests/test_linters.py
syu-w: That's not possible. The old test case used an invalid `metadata.yaml` by specs. It's just because old charmcraft doesn't have these validations cause it can pass a single line of `name: {charm_name}`

tests/test_main.py
syu-w: OK, I will need to test this.

tests/test_metadata.py
syu-w: This one going to be moved to another test file.

tests/test_providers.py
syu-w: OK, I will use regex to check substring.

tests/test_config.py
syu-w: Maybe regex? It has huge amount of prefix line and variable part (file name / version) that will make any modification would require at least dozens of texts to be changed.

tests/test_config.py
syu-w: This is test the old version. The new version is not added in this PR.

tests/test_linters.py
lengau: Ahh, I see — so if you put that charm into the old charmcraft it would still (correctly) error out? In that case I'm fine with this change :smile:

tests/test_config.py
syu-w: OK, I found a better way to do this, and completely removed this function.

tests/commands/test_build.py
```diff
@@ -109,16 +98,21 @@ def basic_project(tmp_path, monkeypatch, create_config):
     # the config
     host_base = get_host_as_base()
     create_config(
```
sergiusens, line 100: rename to create_charmcraft.yaml
and create a discrete create_metadata_yaml

tests/conftest.py
```diff
@@ -184,29 +194,171 @@ def is_provider_installed(cls) -> bool:
...
+
+
+@pytest.fixture
+def create_metadata_yaml(tmp_path: pathlib.Path):
+    """Helper to create a metadata.yaml file in disk.
+
+    If content is not given, create a minimum valid file.
+    """
+
+    def create_metadata_yaml(content="LEGACY", content_extend=None):
```
sergiusens, line 239: content="LEGACY" makes a little noise for me, 

can we make the default content value None or "" and have a different bool like create_metadata_file=False

@lengau ?

syu-w: I was plan to add the `content="COMBINED"` for the merged `charmcraft.yaml`.
I am not sure the best practices here, since we need 2 default values, and a way to not create / delete it.

lengau: I agree. The more flexible we're trying to make this the more we seem to be building business logic into the test setups.

I think the amount of logic that was in the original is about the maximum we want to do: either write the full file or have a sentinel value to write a default file. Having content and a content extension can make a failing test harder to reason about. The more locally we can keep things scoped, the better, even if it means we're repeating the name/summary/description lines a dozen times.

I also keep coming back to the thought of putting the "charm bases" in their own data directories. While charmcraft doesn't currently distinguish unit tests from integration tests, these feel pretty integration-y to me. Perhaps a better helper to have would be a `prepare_config` that receives a directory containing the relevant YAML files and just copies the contents of that directory into `tmp_dir`?

syu-w: Well, in that case we may just duplicate the full yaml files without reuse any common parts. That will make less business logic, but a lot more yaml lines in the test cases.

I still don't think we need to copy the contents of an directory, it usually just 1 or 2 files.

sergiusens: yes, this type of logic makes it really hard on future maintenance

tests/test_main.py
```diff
@@ -283,8 +404,32 @@ def test_main_crash(base_config_present):
...
+                """
+            ),
+            dedent(
+                """\
+                name: test-charm-name-from-metadata-yaml
+                summary: test summary
+                description: test description
+                """
+            ),
+        ],
```
lengau, line 423:
```suggestion
        ],
        ["type: charm", None],
        [None, None],
        [
            dedent(
                """\
                type: charm
                bases:
                  - name: ubuntu
                    channel: "20.04
                """
            ),
            None
        ],
        ["invalid file", "charmcraft shouldn't even read the files in this test."],
```

tests/test_main.py
```diff
@@ -299,8 +444,32 @@ def test_main_interrupted(base_config_present):
...
```
lengau, line 463:
```suggestion
        ],
        ["type: charm", None],
        [None, None],
        [
            dedent(
                """\
                type: charm
                bases:
                  - name: ubuntu
                    channel: "20.04
                """
            ),
            None
        ],
        ["invalid file", "charmcraft shouldn't even read the files in this test."],
```

tests/test_main.py
```diff
@@ -314,8 +483,32 @@ def test_main_controlled_arguments_error(capsys, base_config_present):
...
```
lengau, line 502:
```suggestion
        ],
        ["type: charm", None],
        [None, None],
        [
            dedent(
                """\
                type: charm
                bases:
                  - name: ubuntu
                    channel: "20.04
                """
            ),
            None
        ],
        ["invalid file", "charmcraft shouldn't even read the files in this test."],
```

tests/test_main.py
```diff
@@ -244,8 +293,32 @@ def test_main_no_args():
...
```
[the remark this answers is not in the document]
lengau: Oh, and I suppose a third option would be to put the entire set of fixtures into a constant and just reference the constant for each of these tests.

tests/test_metadata.py
```diff
@@ -45,27 +45,34 @@ def test_parse_metadata_yaml_complete(tmp_path):
...
+def test_parse_metadata_yaml_valid_names(tmp_path, name, prepare_metadata_yaml):
+    prepare_metadata_yaml(
+        dedent(
+            f"""\
+            name: {name}
+            summary: Test summary
+            description: Lot of text.
+            """
+        )
+    )
```
lengau, line 57:
```suggestion
@pytest.mark.parametrize(
    "metadata_yaml_template",
    [
        "name: {name}",
        dedent(
            """\
            name: {name}
            summary: Test summary
            description: Lot of text.
            """
        ),
    ],
)
def test_parse_metadata_yaml_valid_names(
    tmp_path, name, prepare_metadata_yaml, metadata_yaml_template
):
    prepare_metadata_yaml(metadata_yaml_template.format(name=name))
```

tests/test_metadata.py
lengau: Please uncomment it for now

tests/commands/test_build.py
```diff
@@ -222,15 +234,11 @@ def mock_is_base_available():
...
 
-    config = load(basic_project)
-
-    msg = re.escape(
-        "Cannot read the metadata.yaml file: FileNotFoundError(2, 'No such file or directory')"
-    )
-    with pytest.raises(CraftError, match=msg):
+    with pytest.raises(CraftError):
+        config = load(basic_project)
         get_builder(config)
```
lengau, line 242: Please ensure we still check the error message.

```suggestion
    with pytest.raises(CraftError) as exc_info:
        config = load(basic_project)
        get_builder(config)
        
    assert exc_info.value.args[0] == (
        "Cannot read the metadata.yaml file: FileNotFoundError(2, 'No such file or directory')"
    )
```

tests/commands/test_build.py
```diff
@@ -462,8 +474,31 @@ def test_build_multiple_with_charmcraft_yaml_managed_mode(
...
+@pytest.mark.parametrize(
+    "charmcraft_yaml, metadata_yaml",
+    [
+        [
+            dedent(
+                """\
+                type: charm
+                """
+            ),
+            dedent(
```
lengau, line 486: Since `basic_project` has this metadata.yaml, do we need to include it?

tests/commands/test_build.py
```diff
@@ -462,8 +474,31 @@ def test_build_multiple_with_charmcraft_yaml_managed_mode(
...
 
 
+@pytest.mark.parametrize(
+    "charmcraft_yaml, metadata_yaml",
+    [
+        [
+            dedent(
+                """\
+                type: charm
+                """
```
lengau, line 484: This is a bit misleading. Please either include the whole thing here as a string template or don't parametrize it

tests/commands/test_build.py
```diff
@@ -611,10 +671,33 @@ def test_build_project_is_not_cwd(
...
+    ],
+)
 def test_build_bases_index_scenarios_provider(
     mode,
     cmd_flags,
     basic_project,
+    prepare_charmcraft_yaml,
+    prepare_metadata_yaml,
+    charmcraft_yaml,
+    metadata_yaml,
```
lengau, line 700:
```suggestion
@pytest.mark.parametrize("charmcraft_yaml", ["type: charm\n"],)
def test_build_bases_index_scenarios_provider(
    mode,
    cmd_flags,
    basic_project,
    prepare_charmcraft_yaml,
    charmcraft_yaml,
```

The metadata fixtures aren't getting used (and aren't needed).

tests/commands/test_build.py
```diff
@@ -931,36 +1014,63 @@ def test_build_bases_index_scenarios_managed_mode(basic_project_builder, monkeyp
...
+@pytest.mark.parametrize(
+    "charmcraft_yaml, metadata_yaml",
+    [
+        [
+            dedent(
+                """\
+                type: charm
+                """
+            ),
+            dedent(
```
lengau, line 1034: Same as above - we don't need this because it's provided by `basic_project`

lengau (changes requested):

tests/commands/test_build.py
syu-w: Following PR needs it.

tests/commands/test_build.py
syu-w: Following PR needs it.

tests/test_main.py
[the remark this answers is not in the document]
syu-w: Added.

tests/test_config.py
```diff
@@ -1095,252 +2042,241 @@ class FakeChecker:
...
+    with pytest.raises(CraftError) as cm:
+        load(tmp_path)
+    assert str(cm.value) == dedent(
         """\
         Bad charmcraft.yaml content:
         - Bad lint name 'check_missing' in field 'analysis.ignore.linters[1]'"""
     )
-
-
-def test_actions_defined_in_charmcraft_yaml(create_config, tmp_path):
```
lengau, line 1204: Sorry I missed this last night - I don't see equivalents of these 3 tests still here. Can you put some up?

syu-w: Added for now, it will be moved to another test files.

lengau (approved):
Looks good, thanks!

sergiusens (approved):

PR #1141, opened 2023-07-06
pydantic: split base and charmcraft config models

syu-w:
Common models moved to basic.py

Charmcraft Config model moved to charmcraft.py and renamed to CharmcraftConfig to avoid confuse with juju's config.

### DIFF
--- a/charmcraft/bases.py
+++ b/charmcraft/bases.py
@@ -18,7 +18,7 @@
 
 from typing import Tuple, Union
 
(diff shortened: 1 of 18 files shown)


lengau (approved):
Thanks!

lengau:
Spread test failure is unrelated (see: https://github.com/canonical/prometheus-k8s-operator/issues/493)

PR #1149, opened 2023-07-11
command: charmcraft init to use charmcraft.yaml only

syu-w:
Use unified charmcraft.yaml by default, that included `metadata.yaml`, `actions.yaml` and `config.yaml`.

Also added profiles descriptions and updated URL.

CRAFT-1795
CRAFT-1824


syu-w:
Blocked by:
https://github.com/canonical/craft-cli/pull/164
https://github.com/canonical/operator/issues/969

tests/commands/test_init.py
```diff
@@ -114,6 +111,75 @@ def test_all_the_files_kubernetes(tmp_path, config):
...
+        "requirements.txt",
+        "src",
+        os.path.join("src", "charm.py"),
+        "tests",
+        os.path.join("tests", "integration"),
+        os.path.join("tests", "integration", "test_charm.py"),
+        os.path.join("tests", "unit"),
+        os.path.join("tests", "unit", "test_charm.py"),
+        "tox.ini",
+    }
```
lengau, line 130: Do we have a spot already where we're testing that it gets templated correctly? (Other than the spread tests that use the templates)

If not, could we test for one of the files that the template gets correctly filled?

charmcraft/templates/init-machine-legacy/.gitignore.j2
lengau, line 1: These have historically been symlinked around to reduce duplication. Could we do the same here?

(Unless you've already done so and github is just not showing it in the UI, in which case ignore me :slightly_smiling_face: )

syu-w: I specifically de-symlinked them to prevent confusion when edit one profile, others changed. I don't think it's worth symlink them together just to save 100kb of data that can be highly compressed when distribution.

tests/commands/test_init.py
syu-w: Added.

jnsgruk:
I think we're better off just removing the `*-legacy` templates - I can't see a particularly compelling reason to keep them.

lengau (approved):
As long as @jnsgruk and @sergiusens are on the same page as each other about removing the `-legacy` templates I'm happy with that. Let's check in stand-up today and then you can merge either this as-is or without the legacy directories.

Thanks for all this work!

mr-cal (approved):
:+1:

Issue #1193, opened 2023-07-26
Fix broken spread tests


lengau:
### What needs to get done

The following spread tests are broken and need fixing:

* ~~Notebook_operators bundle does not build (Probably related to a dependency update, as it fails on stable versions too)~~
* ~~Prometheus k8s operator fails due to a dependency issue~~
* ~~Store tests need to be decoupled from an account~~ These are okay for now with #1195
* magma pack test (see #1215 )

### Why it needs to get done

CI always looks like it's failing because of these and we're spending a lot of time reviewing the failures and determining that they're just the same thing.

Issue #1179, opened 2023-07-25
Hashes not supported in requirements.txt


carlcsaposs-canonical:
### Bug Description

#1157 introduced a regression—hashes in requirements.txt cannot be used since PYDEPS requirements do not use hashes, and "hash-checking is an all-or-nothing proposition" (from https://pip.pypa.io/en/stable/topics/secure-installs/#hash-checking-mode)

### To Reproduce

Create requirements.txt file with --hash and install a charm lib with PYDEPS

Example: https://github.com/canonical/zookeeper-operator/actions/runs/5651535680

### Environment

Ubuntu 22.04

### charmcraft.yaml

```shell
# Copyright 2023 Canonical Ltd.
# See LICENSE file for licensing details.

type: charm
parts:
  charm:
    charm-binary-python-packages:
      - setuptools
    build-packages:
      # The following are needed for tls-certificates-interface
      - build-essential
      - python3-dev
      - pkg-config
      - libffi-dev
      - libssl-dev
      - rustc
      - cargo
bases:
  - build-on:
      - name: "ubuntu"
        channel: "22.04"
    run-on:
      - name: "ubuntu"
        channel: "22.04"
```

### Relevant log output

```shell
2023-07-25 01:24:55.067 :: 2023-07-25 01:24:48.341 :: Installing dependencies
2023-07-25 01:24:55.067 :: 2023-07-25 01:24:48.341 :: Running external command ['python3', '-m', 'venv', '/root/parts/charm/build/staging-venv']
2023-07-25 01:24:55.067 :: 2023-07-25 01:24:51.875 :: Running external command ['/root/parts/charm/build/staging-venv/bin/pip3', '--version']
2023-07-25 01:24:55.067 :: 2023-07-25 01:24:52.200 ::    :: pip 22.0.2 from /root/parts/charm/build/staging-venv/lib/python3.10/site-packages/pip (python 3.10)
2023-07-25 01:24:55.067 :: 2023-07-25 01:24:52.240 :: Running external command ['/root/parts/charm/build/staging-venv/bin/pip3', 'install', '--upgrade', 'setuptools']
2023-07-25 01:24:55.067 :: 2023-07-25 01:24:52.679 ::    :: Requirement already satisfied: setuptools in ./staging-venv/lib/python3.10/site-packages (59.6.0)
2023-07-25 01:24:55.067 :: 2023-07-25 01:24:53.117 ::    :: Collecting setuptools
[...]

Issue #1178, opened 2023-07-19
spread tests: notebook-operators failing to build


lengau:
### Bug Description

https://github.com/canonical/charmcraft/actions/runs/5593782587

### To Reproduce

1. `git clone https://github.com/canonical/notebook-operators`
2. `cd notebook-operators`
3. `charmcraft pack --include-all-charms`

This doesn't seem to be an issue on 2.3.1, so it could be fixed by the unmerged fixes at https://github.com/canonical/charmcraft/pull/1165

### Environment

CI

### charmcraft.yaml

```shell
Happens for both the jupyter-controller and jupyter-ui charms: https://github.com/canonical/notebook-operators
```

### Relevant log output

```shell
2023-07-19 14:49:28.659 :: ::    :: Collecting pyyaml==5.4 (from -r requirements.txt (line 56))
2023-07-19 14:49:28.690 :: ::    ::   Downloading PyYAML-5.4.tar.gz (174 kB)
2023-07-19 14:49:28.700 :: ::    ::      ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 174.8/174.8 kB 24.3 MB/s eta 0:00:00
2023-07-19 14:49:28.822 :: ::    ::   Installing build dependencies: started
2023-07-19 14:51:25.544 :: ::    ::   Installing build dependencies: still running...
2023-07-19 14:52:33.307 :: ::    ::   Installing build dependencies: still running...
2023-07-19 14:53:33.207 :: ::    ::   Installing build dependencies: finished with status 'done'
2023-07-19 14:53:33.207 :: ::    ::   Getting requirements to build wheel: started
2023-07-19 14:53:33.460 :: ::    ::   Getting requirements to build wheel: finished with status 'error'
2023-07-19 14:53:33.470 :: ::    ::   error: subprocess-exited-with-error
2023-07-19 14:53:33.470 :: ::    ::
2023-07-19 14:53:33.470 :: ::    ::   × Getting requirements to build wheel did not run successfully.
2023-07-19 14:53:33.475 :: ::    ::   │ exit code: 1
2023-07-19 14:53:33.475 :: ::    ::   ╰─> [62 lines of output]
2023-07-19 14:53:33.475 :: ::    ::       running egg_info
2023-07-19 14:53:33.475 :: ::    ::       writing lib3/PyYAML.egg-info/PKG-INFO
[...]

Issue #1135, opened 2023-07-03
PYDEPS overrides pinned dependencies in requirements.txt


carlcsaposs-canonical:
### Bug Description

Example:
pgbouncer-operator has `pydantic==1.10.9` in its [requirements.txt](https://github.com/canonical/pgbouncer-operator/blob/7cb23b6f616f54cbb5d1284883ddc35a29a5368e/requirements.txt#L10C1-L10C17)
pgbouncer-operator also uses a [charm lib](https://github.com/canonical/pgbouncer-operator/blob/7cb23b6f616f54cbb5d1284883ddc35a29a5368e/lib/charms/grafana_agent/v0/cos_agent.py#L194) with `PYDEPS = ["cosl", "pydantic"]`

During `charmcraft pack`, the unpinned `pydantic` charm lib dependency overrides the pinned `pydantic` dependency and installs pydantic version 2

### To Reproduce

```
git clone https://github.com/canonical/pgbouncer-operator
cd pgbouncer-operator
charmcraft pack
```

### Environment

Ubuntu 22.04 LTS

### charmcraft.yaml

```shell
# Copyright 2022 Canonical Ltd.
# See LICENSE file for licensing details.

type: charm
bases:
  - build-on:
      - name: "ubuntu"
        channel: "20.04"
    run-on:
      - name: "ubuntu"
        channel: "20.04"
  - build-on:
      - name: "ubuntu"
        channel: "22.04"
    run-on:
      - name: "ubuntu"
        channel: "22.04"

parts:
  charm:
    charm-binary-python-packages:
      - psycopg2-binary==2.9.6  # renovate
      - psycopg[binary]==3.1.9  # renovate
```

### Relevant log output

```shell
2023-07-03 01:24:02.790 Starting charmcraft version 2.3.0
2023-07-03 01:24:02.795 System details: OSPlatform(system='ubuntu', release='22.04', machine='x86_64'); Environment: None
2023-07-03 01:24:02.795 Packing the charm.
2023-07-03 01:24:02.795 Reading '/home/runner/work/pgbouncer-operator/pgbouncer-operator/metadata.yaml'
2023-07-03 01:24:02.801 Validating metadata format
2023-07-03 01:24:02.952 Executing on host: lxc --project default profile show local:default
2023-07-03 01:24:03.017 Building for 'bases[0]' as host matches 'build-on[0]'.
2023-07-03 01:24:03.017 Skipping 'bases[1]' due to --base-index usage.
[...]

PR #1197, opened 2023-07-26
ci: temporarily disable externally broken spread tests

lengau:
Related to #1193, this should make our CI succeed. We will need to re-enable these tests once the issues with those charms are resolved.

### DIFF
--- a/.github/workflows/spread.yaml
+++ b/.github/workflows/spread.yaml
@@ -32,7 +32,7 @@ jobs:
       matrix:
         spread:
           - "google:"
-          - "google:ubuntu-22.04-64:tests/spread/charms/k8s-operator:prometheus"
+#          - "google:ubuntu-22.04-64:tests/spread/charms/k8s-operator:prometheus"
 
     steps:
       - name: Cleanup job workspace
--- a/tests/spread/charms/bundle/task.yaml
+++ b/tests/spread/charms/bundle/task.yaml
@@ -1,4 +1,4 @@
-#summary: Recursively pack a bundle with all charms on
+summary: Recursively pack a bundle with all charms
 #systems:
 #  - ubuntu-22.04-64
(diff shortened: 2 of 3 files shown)


sergiusens:
do we have an issue link for this one?

lengau:
@sergiusens #1193

jnsgruk:
@simskij any thoughts on the prometheus charm not building correctly? @lengau can you provide some example of where it's broken?

lengau:
@jnsgruk These breakages are because of dependency issues related to a couple of third-party packages (jsonschema being one of them). Here's an example breakage: https://pipelines.actions.githubusercontent.com/serviceHosts/dd6d691f-d1ef-4869-a054-5c766a163a63/_apis/pipelines/1/runs/7391/signedlogcontent/11?urlExpires=2023-07-26T18%3A38%3A36.4349221Z&urlSigningMethod=HMACV1&urlSignature=7SeUO9O9yUGl2poXZMyfw52%2B6AO87TnO3jxhE5gSWJs%3D

I'm hoping to come up with a workaround/fix and fairly quickly release a version with that after I release 2.4.0 this week with Alma Linux support and such, and to include a restructure of our requirements handling in 3.0.

The relevant open issues are:

* #1178
* #1179 (a regression in my attempt to solve #1135 for a use case I wasn't aware of)

The root of the issue is an interaction with the current way we run pip to get dependencies, the `--no-binary` option and the fact that a lot of packages are starting to bring in rust dependencies. Unfortunately, there's no clear "good" solution.

syu-w (approved):

sergiusens:
@lengau thanks, when I asked for the issue, I meant something to track to re-enable with later, seems you provided it in a different response

lengau:
@sergiusens ope, sorry! In retrospect that makes more sense :-)

PR #1207, opened 2023-08-11
packaging: update apt requirements, remove bionic

sergiusens:
Bionic has entered ESM, we also no longer support the Python included in 18.04.


lengau (approved):
Thanks!

I've marked the store spread test failure in bug #1209 since it's unrelated

sergiusens:
thanks!

cmatsuoka (approved):

PR #1216, opened 2023-08-16
tests: disable broken magma test

lengau:
Temporary until we can complete #1193


sergiusens (changes requested):
Can we leave the task untouched, it is already set to manual

sergiusens (approved):

cmatsuoka (approved):

PR #1223, opened 2023-08-18
tests: fix resource spread test

lengau:
progress toward #1193 

Libraries are fixed in #1234

### DIFF
--- a/spread.yaml
+++ b/spread.yaml
@@ -152,6 +152,28 @@ suites:
       # https://linuxcontainers.org/lxd/docs/master/howto/network_bridge_firewalld/#prevent-issues-with-lxd-and-docker
       # https://github.com/canonical/lxd-cloud/blob/f20a64a8af42485440dcbfd370faf14137d2f349/test/includes/lxd.sh#L13-L23
       iptables -P FORWARD ACCEPT
+
+      # Ensure that the reused charms and bundles are registered if necessary.
+      cat <<- EOF > charmcraft.yaml
+      name: store_config_charm
+      type: charm
+      summary: Charmcraft.yaml file for using the staging store.
+      description: Charmcraft.yaml file for using the staging store.
+      charmhub:
+        api-url: https://api.staging.charmhub.io
+        storage-url: https://storage.staging.snapcraftcontent.com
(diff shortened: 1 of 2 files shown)


lengau:
The remaining broken store tests are fixed in other PRs (e.g. #1226)

spread.yaml
```diff
@@ -151,6 +151,28 @@ suites:
...
+      EOF
+      
+      if ! charmcraft status $CHARM_DEFAULT_NAME; then
+        charmcraft register $CHARM_DEFAULT_NAME || ERROR Charm $CHARM_DEFAULT_NAME cannot be registered to this account.
+      fi
+      if ! charmcraft status $BUNDLE_DEFAULT_NAME; then
+        charmcraft register-bundle $BUNDLE_DEFAULT_NAME || ERROR Charm $BUNDLE_DEFAULT_NAME cannot be registered to this account.
+      fi
+
+      rm -f chramcraft.yaml
```
mr-cal, line 174: chram -> charm

mr-cal (approved):

spread.yaml
```diff
@@ -136,6 +136,28 @@ suites:
...
+      type: charm
+      summary: Charmcraft.yaml file for using the staging store.
+      description: Charmcraft.yaml file for using the staging store.
+      charmhub:
+        api-url: https://api.staging.charmhub.io
+        storage-url: https://storage.staging.snapcraftcontent.com
+        registry-url: https://registry.staging.jujucharms.com
+      EOF
+
+      if ! charmcraft status $CHARM_DEFAULT_NAME; then
```
cmatsuoka, line 152: Shellcheck would complain about unquoted variable arguments, but since charm and bundle names can't have spaces in them it should be ok.

cmatsuoka (approved):
Can't see obvious problems, but I have little experience working with Charmcraft resources.

spread.yaml
lengau, line 152:
```suggestion
      if ! charmcraft status "${CHARM_DEFAULT_NAME}"; then
```

lengau: Good catch, thanks! I'm working towards enabling shellcheck so it's good to get that taken care of now rather than later.

PR #1225, opened 2023-08-18
tests: improve multipass spread backend

lengau:
This allows multiple simultaneous workers with multipass using the same naming scheme the built-in lxd backend uses.

### DIFF
--- a/spread.yaml
+++ b/spread.yaml
@@ -30,60 +30,44 @@ backends:
   multipass:
     type: adhoc
     allocate: |
-      if [ "$SPREAD_SYSTEM" = "ubuntu-18.04-64" ]; then
-        image="18.04"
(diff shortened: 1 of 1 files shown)


mr-cal:
What do you think about making this a configurable feature?  8 multipass instances building charms simultaneously is too much for my laptop to handle

lengau:
@mr-cal I'm happy to set the default number of workers to whatever. What about 2x 22.04 ones and 1x on the others?

mr-cal:
Sounds like a good balance to me!

mr-cal (approved):

syu-w (approved):

lengau:
@syu-w alas, spread (currently) only supports integer literals in this space. Might be worth a future feature request though.

PR #1226, opened 2023-08-18
templates: remove k8s info from machine charm template

lengau:
This change should make the `charm-upload-and-release` store task pass.

Progress on #1193

### DIFF
--- a/charmcraft/templates/init-machine/charmcraft.yaml.j2
+++ b/charmcraft/templates/init-machine/charmcraft.yaml.j2
@@ -36,24 +36,3 @@ bases:
     run-on:
     - name: ubuntu
       channel: "22.04"
-
-# The containers and resources metadata apply to Kubernetes charms only.
-# See https://juju.is/docs/sdk/metadata-reference for a checklist and guidance.
-
-# Your workload’s containers.
-containers:
-  some-container:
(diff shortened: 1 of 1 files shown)


lengau:
The broken store tests are part of #1193, but I'm fixing them independently.

syu-w:
That part include its own tests, temporary remove it?

lengau:
@syu-w — looking at what `charmcraft init` generated before we switched to only `charmcraft.yaml`, it probably shouldn't have that section anyway (we want `charmcraft init` to create as close as possible to a working charm, and machine charms don't use containers or resources). I also don't see any tests that are affected by this

syu-w:
I think that part is to let people fill in some base info if they don't know how to use it. So maybe comment out them better than remove.

lengau:
If someone wants to use those values they should use the `kubernetes` profile rather than the `machine` profile though?

syu-w:
I am not sure, since it added as that.

syu-w:
OK, I found that one is an error cause by symlink...
Old one do not have that.

https://github.com/canonical/template-operator/tree/charmcraft-templates/machine

syu-w (approved):

cmatsuoka (approved):

PR #1230, opened 2023-08-19
pack: refactor business logic out of pack command class 

### LINKED ISSUE #1147: Remove business logic from pack command class
### What needs to get done

The pack command should have its business logic moved into a service class.

Ideally we'd do this with other commands too, but it's absolutely necessary for the pack command.

### Why it needs to get done

Precursor for #1133

lengau:
Builds on #1229 

Fixes #1147

### DIFF
--- a/charmcraft/commands/pack.py
+++ b/charmcraft/commands/pack.py
@@ -16,29 +16,15 @@
 
 """Infrastructure for the 'pack' command."""
 import argparse
-import functools
-import os
 import pathlib
-import shutil
-import subprocess
-import tempfile
-from typing import Any, Collection, Dict, List, Mapping
+from typing import Dict, List
 
 import yaml
 from craft_cli import ArgumentParsingError, CraftError, emit
 
-from charmcraft import const, env, instrum, package, parts
+from charmcraft import env, instrum, package
 from charmcraft.cmdbase import BaseCommand
-from charmcraft.errors import DuplicateCharmsError
-from charmcraft.metafiles.manifest import create_manifest
-from charmcraft.parts import Step
-from charmcraft.utils import (
-    build_zip,
-    find_charm_sources,
-    get_charm_name_from_path,
-    humanize_list,
-    load_yaml,
-)
+from charmcraft.utils import find_charm_sources, get_charm_name_from_path, load_yaml
 
 # the minimum set of files in a bundle
 MANDATORY_FILES = ["bundle.yaml", "README.md"]
@@ -135,6 +121,16 @@ class PackCommand(BaseCommand):
     def run(self, parsed_args: argparse.Namespace) -> None:
         """Run the command."""
         self._check_config(config_file=True)
+
+        builder = package.Builder(
+            config=self.config,
+            force=parsed_args.force,
+            debug=parsed_args.debug,
+            shell=parsed_args.shell,
+            shell_after=parsed_args.shell_after,
+            measure=parsed_args.measure,
+        )
+
         # decide if this will work on a charm or a bundle
         if self.config.type == "charm":
             if parsed_args.include_all_charms:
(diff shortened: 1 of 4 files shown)


tests/commands/test_pack.py
```diff
@@ -1024,108 +1022,3 @@ def test_validator_bases_index_invalid(bases_indices, bad_index, config):
             f"Bases index '{bad_index}' is invalid (must be >= 0 and fit in configured bases)."
         )
         assert str(exc_cm.value) == expected_msg
-
-
-# region Unit tests for private functions
```
lengau, line 1029: For clarity — these got moved to `test_package.py`

charmcraft/package.py
```diff
@@ -418,3 +423,123 @@ def handle_package(self, prime_dir, bases_config: BasesConfiguration):
...
+            linting_results=[],
+        )
+        zipname = self.config.project.dirpath / (bundle_name + ".zip")
+        if overwrite:
+            primed_bundle_path = lifecycle.prime_dir / "bundle.yaml"
+            with primed_bundle_path.open("w") as bundle_file:
+                yaml.safe_dump(bundle, bundle_file)
+        build_zip(zipname, lifecycle.prime_dir)
+
+        return [*charms.values(), zipname]
```
syu-w, line 499: What do you think if we `return zipname, [*charms.values(), zipname]`? Get a [-1] as filename feels unintuitive.

syu-w (approved):

charmcraft/package.py
lengau: What about returning (as pseudocode) `(charm_paths, [bundle_path])`? (Probably as a data structure)

Then in the outer command we can add the list of generated charms to the JSON?

syu-w: We don't seems use the `output_files` except the zipname? `_pack_bundle` return the list but nobody use it?

charmcraft/commands/pack.py
```diff
@@ -153,8 +149,12 @@ def run(self, parsed_args: argparse.Namespace) -> None:
                     f"Currently trying to pack: {self.config.project.dirpath}"
                 )
             self._check_config(bases=True)
-            pack_method = self._pack_charm
+            with instrum.Timer("Whole pack run"):
+                self._pack_charm(parsed_args, builder)
         elif self.config.type == "bundle":
+            if parsed_args.shell:
+                package.launch_shell()
+                return
```
cmatsuoka, line 157: Why is this defined only for bundles instead of having common shell handling for both types?

charmcraft/commands/pack.py
```diff
@@ -259,150 +240,41 @@ def _pack_bundle(
...
-            config_parts,
-            work_dir=work_dir,
-            project_dir=project.dirpath,
-            project_name=bundle_name,
-            ignore_local_sources=[bundle_name + ".zip"],
-        )
+
         try:
-            lifecycle.run(Step.PRIME)
+            output_files = builder.pack_bundle(
```
cmatsuoka, line 259: Maybe we could have the zip name returned separately to avoid giving special meaning to the last element?

charmcraft/package.py
```diff
@@ -89,7 +91,10 @@ def __init__(self, *, config, force, debug, shell, shell_after, measure):
         self.charmdir = config.project.dirpath
         self.buildpath = self.charmdir / BUILD_DIRNAME
         self.config = config
-        self._parts = self.config.parts.copy()
+        if self.config.parts:
+            self._parts = self.config.parts.copy()
+        else:
+            self._parts = None
```
cmatsuoka, line 97: Is `config.parts` undefined or defined as an empty dictionary if no parts are defined by the user? In the former case, how was this working before?

cmatsuoka (commented):
Added a few questions.

charmcraft/package.py
lengau: I've changed this around a bit — you and @cmatsuoka tell me what you think

charmcraft/commands/pack.py
lengau: Merging this conversation with https://github.com/canonical/charmcraft/pull/1230#discussion_r1303513129

charmcraft/package.py
syu-w: LGTM

PR #1233, opened 2023-08-22
feat(dependencies): new Python dependency management

### LINKED ISSUE #1140: Reconsider python package dependency management
### What needs to get done

Currently we run pip (up to) 4 times, and in a way that can be counterintuitive to the user. Package sets can override each other, and explicitly frozen package versions can be ignored.

Ideally, we'd have a way to specify this all to be calculated in a single `pip` command that:

- Only installs binary packages for those specified as binaries (and their dependencies?)
- Fails on package conflicts
- Doesn't overwrite pinned versions

### Why it needs to get done

The current method has evolved over years to become counterintuitive and sometimes problematic (see #1135). However, changes related to this could cause compatibility issues, so they should be done carefully and deliberately. As such, a rethink of this would need to come with a major version release if we decide to do it.

lengau:
This has new dependency management in a single step, falling back to what charmcraft 2.3.0 did if it fails.

Future steps: with future platforms we're not going to fall back to classic behaviour.

Fixes #1140


charmcraft/charm_builder.py
```diff
@@ -233,31 +239,48 @@ def _install_dependencies(self, staging_venv_dir):
...
+                    )
+                )
+            except RuntimeError:
+                # Fall back to old behaviour if pip install fails.
+                # This will go away with a future base.
+                print(
+                    "WARNING: Initial package installation failed. "
+                    "Falling back to older method, which may leave your charm "
+                    "in an un-runnable state."
+                )
```
cmatsuoka, line 263: Will this only show in verbose mode?

charmcraft/utils.py
```diff
@@ -405,3 +409,67 @@ def get_charm_name_from_path(path: pathlib.Path) -> str:
...
+    exclusions = set()
+    for requirement in requirements:
+        if match := PACKAGE_LINE_REGEX.match(requirement):
+            if match.group(0) in excluded:
+                exclusions.add(requirement)
+
+    return requirements - exclusions
+
+
+def get_pip_command(
```
mr-cal, line 437: How does this work with hashed requirement files? Or with a mix of hashed and unhashed requirement files?

lengau: It doesn't work with a mix (current behaviour falls back to the charmcraft 2.3 behaviour), but it does work with hashed files.

charmcraft/charm_builder.py
lengau: The charm builder's content ends up getting output directly to the console regardless, but with `stdout` and `stderr` collapsed (so there's no point in setting `file=sys.stderr`).

cmatsuoka (approved):
Thanks. This seems to match the previous discussion. I think we can land this, and be prepared to fix any corner case bug should any appear.

charmcraft/charm_builder.py
```diff
@@ -228,33 +232,50 @@ def _install_dependencies(self, staging_venv_dir):
         # create virtualenv using the host environment python
         with instrum.Timer("Creating venv"):
             _process_run(["python3", "-m", "venv", str(staging_venv_dir)])
-        pip_cmd = str(_find_venv_bin(staging_venv_dir, "pip3"))
+        pip_cmd = str(_find_venv_bin(staging_venv_dir, "pip"))
 
-        with instrum.Timer("Getting pip version"):
-            _process_run([pip_cmd, "--version"])
+        with instrum.Timer("Ensuring a recent enough pip..."):
+            if get_pip_version(pip_cmd) < (22, 0):
```
mr-cal, line 238: Can you add a comment on why `22.0` is the minimum?

charmcraft/utils.py
```diff
@@ -404,3 +419,107 @@ def get_charm_name_from_path(path: pathlib.Path) -> str:
...
+
+    :param prefix: The pip command and any earlier arguments.
+    :param requirements_files: Paths to the requirements files to include.
+    :param source_deps: Additional dependencies that can only be installed from source.
+    :param binary_deps: Dependencies (including from requirements files) allowed for binary install.
+    :returns: A full pip command line
+    """
+    charm_packages = get_pypi_packages(source_deps)
+    binary_packages = get_pypi_packages(binary_deps)
+    requirements_packages = get_pypi_packages(
```
mr-cal, line 489: Charmcraft only supports pypi packages?  If you have `my-library @ git+https://github.com/...` in a requirements file, it will be silently excluded, right?  

Was it allowed in the old behavior?

lengau: If a requirements file contains values like that, it'll still be parsed, but it won't be included in consideration of whether the package needs to be included in the `--no-binary` option. Likewise, a requirements file that links directly to a wheel file will still work just as before.

charmcraft/charm_builder.py
lengau, line 238:
```suggestion
            # Get at least a known-working minimum pip version if needed.
            if get_pip_version(pip_cmd) < (22, 0):
```

charmcraft/charm_builder.py
```diff
@@ -228,33 +232,52 @@ def _install_dependencies(self, staging_venv_dir):
...
-        pip_cmd = str(_find_venv_bin(staging_venv_dir, "pip3"))
+        pip_cmd = str(_find_venv_bin(staging_venv_dir, "pip"))
 
-        with instrum.Timer("Getting pip version"):
-            _process_run([pip_cmd, "--version"])
+        with instrum.Timer("Ensuring a recent enough pip..."):
+            # Get at least a known-working minimum pip version if needed.
+            # This becomes unnecessary once all supported bases include pip >= 22.0
+            if get_pip_version(pip_cmd) < (22, 0):
+                _process_run([pip_cmd, "install", "-U", "pip"])
```
lengau, line 241:
```suggestion
        with instrum.Timer("Ensuring a recent enough pip..."):
            # pip 20 (included with focal) has dependency resolution issues related to
            # common charm dependencies (e.g. ops). Resolve this by updating to a
            # known working version of pip.
            if get_pip_version(pip_cmd) < (22, 0):
                pip_url = "https://files.pythonhosted.org/packages/ba/19/e63fb4e0d20e48bd2167bb7e857abc0e21679e24805ba921a224df8977c0/pip-23.2.1.tar.gz"
                _process_run([pip_cmd, "install", f"pip@{pip_url}"])
```

charmcraft/charm_builder.py
lengau: Strictly speaking I don't know whether that's the actual minimum version needed (it could be lower), but the version of pip included with `focal` has dependency resolution issues. Comment added.

mr-cal (approved):

charmcraft/utils.py
```diff
@@ -57,6 +70,8 @@
 
 PathOrString = Union[os.PathLike, str]
 
+PACKAGE_LINE_REGEX = re.compile(r"^([A-Za-z0-9_.-]+)([<>=!]=[0-9.]+)?")
```
syu-w, line 73: There are also `~=` and `===` operators
https://peps.python.org/pep-0440/#version-specifiers

Also space should be ignored?

tests/test_charm_builder.py
```diff
@@ -604,18 +604,17 @@ def test_build_dependencies_virtualenv_simple(tmp_path, assert_output):
...
-    pip_cmd = str(charm_builder._find_venv_bin(tmp_path / STAGING_VENV_DIRNAME, "pip3"))
+    pip_cmd = str(charm_builder._find_venv_bin(tmp_path / STAGING_VENV_DIRNAME, "pip"))
 
     assert mock.mock_calls == [
         call(["python3", "-m", "venv", str(tmp_path / STAGING_VENV_DIRNAME)]),
-        call([pip_cmd, "--version"]),
-        call(
-            [pip_cmd, "install", "--upgrade", "--no-binary", ":all:", f"--requirement={reqs_file}"]
-        ),
+        call([pip_cmd, "install", f"--requirement={reqs_file}"]),
```
syu-w, line 617: So we dropped no-binary?

tests/test_utils.py
```diff
@@ -694,4 +700,95 @@ def test_get_charm_name_from_path_wrong_name(tmp_path, build_charm_directory, na
...
+        pytest.param(["-e ."], set(), id="editable-ignored"),
+    ],
+)
+def test_get_pypi_packages(requirements, expected):
+    assert get_pypi_packages(requirements) == expected
+
+
+@pytest.mark.parametrize(("packages", "expected"), [({"abc"}, {"abc"}), ({"abc==1.0.0"}, {"abc"})])
+def test_get_package_names(packages, expected):
+    assert get_package_names(packages) == expected
```
syu-w, line 721: We may want to test all operators: `~=`, `==`, `!=`, `>=`, `<=`, `>`, `<`, `===`

syu-w (changes requested):

charmcraft/charm_builder.py
```diff
@@ -228,33 +232,54 @@ def _install_dependencies(self, staging_venv_dir):
...
+        pip_cmd = str(_find_venv_bin(staging_venv_dir, "pip"))
 
-        with instrum.Timer("Getting pip version"):
-            _process_run([pip_cmd, "--version"])
+        with instrum.Timer("Ensuring a recent enough pip..."):
+            # pip 20 (included with focal) has dependency resolution issues related to
+            # common charm dependencies (e.g. ops). Resolve this by updating to a
+            # known working version of pip.
+            if get_pip_version(pip_cmd) < (22, 0):
+                pip_url = "https://files.pythonhosted.org/packages/ba/19/e63fb4e0d20e48bd2167bb7e857abc0e21679e24805ba921a224df8977c0/pip-23.2.1.tar.gz"
```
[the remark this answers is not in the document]
lengau: Yes, I decided to do that to ensure that we get a specific version (known-good version) rather than the latest.

tests/test_charm_builder.py
lengau: The new version specifies the full `--no-binary` list by extracting it from all the requirements. Since the requirements file is empty there aren't any packages to enforce source install for, so no `--no-binary` expression gets added.

tests/test_utils.py
lengau: Agreed, thanks! Updated.

charmcraft/utils.py
lengau: You're right. I've changed this to cover all those cases and included tests for them.

charmcraft/charm_builder.py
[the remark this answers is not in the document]
lengau: I want to ensure that we're using a specific known-good wheel.

syu-w (approved):
Thanks!

PR #1234, opened 2023-08-22
store: allow publish-lib to work without a metadata.yaml

lengau:
Resource test is fixed in #1223

### DIFF
--- a/charmcraft/commands/store/__init__.py
+++ b/charmcraft/commands/store/__init__.py
@@ -1373,11 +1373,11 @@ class PublishLibCommand(BaseCommand):
 
     def run(self, parsed_args):
         """Run the command."""
-        charm_name = get_name_from_metadata()
+        charm_name = self.config.name or get_name_from_metadata()
         if charm_name is None:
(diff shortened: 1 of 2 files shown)


charmcraft/commands/store/__init__.py
```diff
@@ -1373,11 +1373,11 @@ def fill_parser(self, parser):
 
     def run(self, parsed_args):
         """Run the command."""
-        charm_name = get_name_from_metadata()
+        charm_name = self.config.name or get_name_from_metadata()
```
tigarmo, line 1376: if `metadata.yaml` exists, should it have preference over `self.config.name`?

lengau: It doesn't really matter — `self.config.name` gets the info from `charmcraft.yaml`, and loading the config gives an error if there's a name defined in both, so they're mutually exclusive.

Good question though!

tigarmo (approved):

mr-cal (approved):

PR #1240, opened 2023-08-24
ci: update snap build action to use main

lengau:


lengau:
Closing since these builds did prove what was needed (https://github.com/snapcore/action-build/issues/62)

Issue #1127, opened 2023-06-20
Fix "released" message for bundle


sed-i:
### What needs to get done

After a bundle is released successfully,
```
$ charmcraft release cos-lite --channel=edge --revision=18
Revision 18 of charm 'cos-lite' released to edge
```
The output message should say "bundle" instead of "charm".

### Why it needs to get done

For accuracy :)

PR #1244, opened 2023-08-30
fix(store): temporary hide package type

syu-w:
Fix: https://github.com/canonical/charmcraft/issues/1127

### DIFF
--- a/charmcraft/commands/store/__init__.py
+++ b/charmcraft/commands/store/__init__.py
@@ -776,7 +776,7 @@ class ReleaseCommand(BaseCommand):
             parsed_args.resource,
         )
 
-        msg = "Revision {:d} of charm {!r} released to {}"
+        msg = "Revision {:d} of {!r} released to {}"
         args = [parsed_args.revision, parsed_args.name, ", ".join(parsed_args.channel)]
         if parsed_args.resource:
             msg += " (attaching resources: {})"
--- a/tests/commands/test_store_commands.py
+++ b/tests/commands/test_store_commands.py
@@ -1381,7 +1381,7 @@ def test_release_simple_ok(emitter, store_mock, config):
         call.release("testcharm", 7, channels, []),
     ]
 
-    expected = "Revision 7 of charm 'testcharm' released to somechannel"
+    expected = "Revision 7 of 'testcharm' released to somechannel"
     emitter.assert_message(expected)
 
 
@@ -1395,7 +1395,7 @@ def test_release_simple_multiple_channels(emitter, store_mock, config):
     )
     ReleaseCommand(config).run(args)
 
-    expected = "Revision 7 of charm 'testcharm' released to channel1, channel2, channel3"
+    expected = "Revision 7 of 'testcharm' released to channel1, channel2, channel3"
     emitter.assert_message(expected)
 
 
@@ -1411,7 +1411,7 @@ def test_release_including_resources(emitter, store_mock, config):
     ]
 
     expected = (
-        "Revision 7 of charm 'testcharm' released to testchannel "
+        "Revision 7 of 'testcharm' released to testchannel "
         "(attaching resources: 'foo' r3, 'bar' r17)"
     )
     emitter.assert_message(expected)


tests/commands/test_store_commands.py
mr-cal, line 1: Is there a test that tests the original bug report? (`Revision 18 of bundle 'cos-lite' released to edge` displayed for bundles)

May be worthwhile to have a regression test here

syu-w: OK, I will add one,

charmcraft/commands/store/__init__.py
```diff
@@ -769,15 +769,23 @@ def fill_parser(self, parser):
     def run(self, parsed_args):
         """Run the command."""
         store = Store(self.config.charmhub)
+        packages_list = store.list_registered_names(include_collaborations=True)
+        packages_type_map = {p.name: p.entity_type for p in packages_list}
```
cmatsuoka, line 773: Since we're iterating over packages here we could just cherry-pick the one we want instead of building the full map. (Is there a way to query a specific package instead of downloading the full list?)

syu-w: I don't see any API can do that (https://api.charmhub.io/docs/default.html), but we can only keep the one that matched name.

syu-w: Updated.

charmcraft/commands/store/__init__.py
```diff
@@ -769,15 +769,25 @@ def fill_parser(self, parser):
     def run(self, parsed_args):
         """Run the command."""
         store = Store(self.config.charmhub)
+        packages_list = store.list_registered_names(include_collaborations=True)
+        packages_type_map = {
+            p.name: p.entity_type for p in packages_list if p.name == parsed_args.name
+        }
```
cmatsuoka, line 775: We could obtain the package type directly without using a map (which now would have only one entry?)

syu-w: Updated.

charmcraft/commands/store/__init__.py
```diff
@@ -769,15 +769,27 @@ def fill_parser(self, parser):
     def run(self, parsed_args):
         """Run the command."""
         store = Store(self.config.charmhub)
+        packages_list = store.list_registered_names(include_collaborations=True)
+        package_type: str = "UNKNOWN"
```
cmatsuoka, line 773: One last question: in which situation a valid package would not be listed as registered? Should we go ahead with an unknown type, or stop with an error?

syu-w: In no situation that should happened. `store.release` should raise error so the `UNKNOWN` may not able to be used at all. Just put it there in case.

cmatsuoka: So I think we should raise an error (e.g. in an`else` in the `for` below) if the name is not found instead of letting it move forward should the "impossible situation" happen.

cmatsuoka: Another possibility would be to just not retrieve the source and put what we think is the type, and let the store error if the type doesn't match. Then we wouldn't need to download the large list at all.

syu-w: I let it pass because if we added new type, old charmcraft should still able to work. Also the store does not care the type at all, as it is not used in the release API. This is pure for the end user to see what type it is, not for store to verify it, as charmcraft does not know it either.

tests/commands/test_store_commands.py
```diff
@@ -1371,22 +1371,65 @@ def test_revisions_errors_multiple(emitter, store_mock, config, formatted):
...
+    channels = ["somechannel"]
+    args = Namespace(name="testbundle", revision=8, channel=channels, resource=[])
+    ReleaseCommand(config).run(args)
+
+    assert store_mock.mock_calls == [
+        call.list_registered_names(include_collaborations=True),
+        call.release("testbundle", 8, channels, []),
+    ]
+
+    expected = "Revision 8 of bundle 'testbundle' released to somechannel"
```
sergiusens, line 1418: Revision 8 of 'testbundle' released to somechannel

And request the store through a bug to return the package type in the release response

sergiusens (approved):

PR #1247, opened 2023-09-06
ci: reduce the python/OS matrix size

lengau:
This will also need adjustments to our required checks.

* Remove Python 3.9 tests (no longer widely used in any of our use cases)
* Use only built-in Python versions on Ubuntu releases (but ready to add 3.12 for 24.04)
* Add 3.11 (current latest) on Windows/mac

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -44,8 +44,13 @@ jobs:
   run-tests:
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


syu-w (approved):

syu-w:
We may want to test it in 3.11 anyways since 3.11 used in 23.04.

cmatsuoka (approved):

sergiusens (approved):

Issue #1238, opened 2023-08-24
Support for extensions (basic support)


sergiusens:
### What needs to get done

Extension framework as done in Snapcraft and Rockcraft

### Why it needs to get done

To support easy creation of charms without the complexities of each framework

tigarmo:
Reference PRs in rockcraft, since we did this recently:
https://github.com/canonical/rockcraft/pull/281
https://github.com/canonical/rockcraft/pull/292

PR #1248, opened 2023-09-06
feat(extensions): add basic support

syu-w:
Fix: https://github.com/canonical/charmcraft/issues/1238

### DIFF
--- a/charmcraft/errors.py
+++ b/charmcraft/errors.py
@@ -20,8 +20,6 @@ from typing import Iterable, Mapping
 
 from craft_cli import CraftError
 
-from charmcraft import utils
-
 
 class BadLibraryPathError(CraftError):
     """Subclass to provide a specific error for a bad library path."""
@@ -71,7 +69,9 @@ class DuplicateCharmsError(CraftError):
     )
 
     def __init__(self, charms: Mapping[str, Iterable[pathlib.Path]], source: bool = True):
-        charm_names = utils.humanize_list(charms.keys(), "and")
+        import charmcraft.utils
+
+        charm_names = charmcraft.utils.humanize_list(charms.keys(), "and")
         super().__init__(
             f"Duplicate charms found: {charm_names}",
             details=self._format_details(charms),
@@ -94,3 +94,7 @@ class DuplicateCharmsError(CraftError):
             for path in path_iter:
                 print(path_tree_line_format.format(name="", path=path), file=details)
         return details.getvalue()
+
+
+class ExtensionError(CraftError):
+    """Error related to extension handling."""
--- a/charmcraft/extensions/_utils.py
+++ b/charmcraft/extensions/_utils.py
@@ -1,0 +1,115 @@
+# Copyright 2023 Canonical Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# For further info, check https://github.com/canonical/charmcraft
+
+"""Extension application helpers."""
+
+import copy
+from pathlib import Path
+from typing import Any, Dict, List, Set, Union, cast
+
+from charmcraft.extensions.extension import Extension
+from charmcraft.extensions.registry import get_extension_class
+
+
+def apply_extensions(project_root: Path, yaml_data: Dict[str, Any]) -> Dict[str, Any]:
+    """Apply all extensions.
+
+    :param dict yaml_data: Loaded, unprocessed charmcraft.yaml
+    :returns: Modified charmcraft.yaml data with extensions applied
+    """
+    # Don't modify the dict passed in
+    yaml_data = copy.deepcopy(yaml_data)
+    declared_extensions: List[str] = cast(List[str], yaml_data.get("extensions", []))
+    if not declared_extensions:
+        return yaml_data
+
+    del yaml_data["extensions"]
+
+    # Process extensions in a consistent order
+    for extension_name in sorted(declared_extensions):
+        extension_class = get_extension_class(extension_name)
+        extension = extension_class(project_root=project_root, yaml_data=copy.deepcopy(yaml_data))
+        extension.validate(extension_name=extension_name)
+        _apply_extension(yaml_data, extension)
+    return yaml_data
+
+
+def _apply_extension(
+    yaml_data: Dict[str, Any],
+    extension: Extension,
+) -> None:
+    # Apply the root components of the extension (if any)
+    root_extension = extension.get_root_snippet()
+    for property_name, property_value in root_extension.items():
+        yaml_data[property_name] = _apply_extension_property(
+            yaml_data.get(property_name), property_value
+        )
+
+    # Next, apply the part-specific components
+    part_extension = extension.get_part_snippet()
+    if "parts" not in yaml_data:
+        yaml_data["parts"] = {}
+
+    parts = yaml_data["parts"]
+    for _, part_definition in parts.items():
+        for property_name, property_value in part_extension.items():
+            part_definition[property_name] = _apply_extension_property(
+                part_definition.get(property_name), property_value
+            )
+
+    # Finally, add any parts specified in the extension
+    parts_snippet = extension.get_parts_snippet()
+    parts_names = (pn for pn in parts_snippet if pn not in yaml_data["parts"])
+    for part_name in parts_names:
+        parts[part_name] = parts_snippet[part_name]
+
+
+def _apply_extension_property(
+    existing_property: Union[Dict, List], extension_property: Union[Dict, List]
+) -> Union[Dict, List]:
+    if existing_property:
+        # If the property is not scalar, merge them
+        if isinstance(existing_property, list) and isinstance(extension_property, list):
+            merged = extension_property + existing_property
+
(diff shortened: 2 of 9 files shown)


syu-w:
2 things:
1. It only works with unified `charmcraft.yaml`, otherwise it violated the pydantic model since it added other keys in the yaml dict.
2. How details we want to ask extensions the supported base? OS/channel/arch?

charmcraft/extensions/_utils.py
```diff
@@ -0,0 +1,113 @@
...
+    declared_extensions: List[str] = cast(List[str], yaml_data.get("extensions", []))
+    if not declared_extensions:
+        return yaml_data
+
+    del yaml_data["extensions"]
+
+    # Process extensions in a consistent order
+    for extension_name in sorted(declared_extensions):
+        extension_class = get_extension_class(extension_name)
+        extension = extension_class(project_root=project_root, yaml_data=copy.deepcopy(yaml_data))
```
lengau, line 42: This question may be more appropriate for @sergiusens since he wrote the snapcraft extensions implementation, but why do we do a deep copy here too? Do we expect the extensions to modify the dictionary or are we just taking precautions in case they do even though they're not supposed to?

Based on the extensions in snapcraft I presume it's the latter. How strict do we want to be about this? (e.g. if we're okay with using static analysis to ensure this, could we decorate `Extension.__init__` with `typing.final`?  Or do we want this to be a strict runtime guarantee?)

charmcraft/extensions/extension.py
```diff
@@ -0,0 +1,143 @@
...
+
+from craft_cli import emit
+
+from charmcraft import errors
+
+
+class Extension(abc.ABC):
+    """Extension is the class from which all extensions inherit.
+
+    Extensions have the ability to add snippets to apps, parts, and indeed add new parts
```
lengau, line 31:
```suggestion
    Extensions have the ability to add snippets to charms, parts, and indeed add new parts
```

charmcraft/extensions/extension.py
```diff
@@ -0,0 +1,143 @@
...
+        project_root: Path,
+        yaml_data: Dict[str, Any],
+    ) -> None:
+        """Create a new Extension."""
+        self.project_root = project_root
+        self.yaml_data = yaml_data
+
+    @staticmethod
+    @abc.abstractmethod
+    def get_supported_bases() -> List[Tuple[str, ...]]:
```
lengau, line 49: What about

```suggestion
    def get_supported_bases() -> Tuple[charmcraft.bases.Base, ...]:
```

or possibly a `Tuple[Dict[str, str], ...]`

charmcraft/extensions/extension.py
```diff
@@ -0,0 +1,143 @@
...
+        """Validate that the extension can be used with the current project.
+
+        :param extension_name: the name of the extension being parsed.
+        :raises errors.ExtensionError: if the extension is incompatible with the project.
+        """
+        if "bases" not in self.yaml_data:
+            # There is nothing to validate, the extension will set the preferred base.
+            return
+
+        bases: str = self.yaml_data["bases"]
```
lengau, line 80: This isn't a string in charmcraft.

tests/conftest.py
```diff
@@ -344,3 +344,13 @@ def helper(tmp_path, fake_charms, file_type="charm"):
         return expected
 
     return helper
+
+
+@pytest.fixture()
+def mock_extensions(monkeypatch):
```
lengau, line 350: This is a stub, not a mock.

lengau (changes requested):

charmcraft/extensions/registry.py
```diff
@@ -0,0 +1,66 @@
...
+    """Obtain a extension class given the name.
+
+    :param name: The extension name.
+    :return: The extension class.
+    :raises ExtensionError: If the extension name is invalid.
+    """
+    try:
+        return _EXTENSIONS[extension_name]
+    except KeyError as key_error:
+        raise errors.ExtensionError(f"Extension {extension_name!r} does not exist") from key_error
```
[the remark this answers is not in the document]
sergiusens: aren't tracebacks more thorough though?

charmcraft/extensions/extension.py
sergiusens: we should just remove apps, it refers to the apps entry in snapcraft.yaml; perhaps replaced with config, actions, ....

charmcraft/errors.py
```diff
@@ -20,8 +20,6 @@
 
 from craft_cli import CraftError
 
-from charmcraft import utils
```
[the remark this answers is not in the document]
syu-w: Yes

charmcraft/extensions/extension.py
syu-w: The base is not exist at that moment.

syu-w:
I moved `apply_extensions` into `unmarshal` after `expand_short_form_bases`, so it does not need to handle 2 different formats. The down side is `expand_short_form_bases` need to be called twice in case it got changed in the extension.

charmcraft/extensions/registry.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+def get_extension_names() -> List[str]:
+    """Obtain a extension class given the name.
+
+    :param name: The extension name.
+    :return: The list of available extensions.
+    :raises ExtensionError: If the extension name is invalid.
+    """
+    return list(_EXTENSIONS.keys())
```
lengau, line 34: I wonder if it wouldn't be more user-friendly to alphabetise this list?
```suggestion
    return sorted(_EXTENSIONS.keys())
```

charmcraft/extensions/registry.py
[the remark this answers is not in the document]
lengau: If the traceback adds useful information, definitely. But I think this is one of the rare cases where it doesn't. This snippet is functionally equivalent to

```python
if extension_name in _EXTENSIONS:
    return _EXTENSIONS[extension_name]
raise errors.ExtensionError(...)
```

and I doubt we'd ask for more context if it were written using `if` rather than `try`.

It's a minimal difference since it only affects charmcraft with `--verbosity=debug`, but here's an example script:

```python
import sys
import craft_cli

craft_cli.emit.init(craft_cli.EmitterMode.DEBUG, "testcraft", "hi")


def emit_ext_err(extension):
    try:
        {}[extension]
    except KeyError as e:
        if "--no-cause" in sys.argv:
            e = None
        raise craft_cli.CraftError(f"Extension {extension!r} does not exist") from e


try:
    emit_ext_err("invalid_extension")
except craft_cli.CraftError as e:
    craft_cli.emit.error(e)
```

The output in the two cases looks like:

```
(charmcraft) lengau@hyperion:~/Work/Code/charmcraft$ python test.py 
2023-09-07 13:44:12.119 hi                                                                                                                                                                                                                                           
2023-09-07 13:44:12.119 Logging execution to '/home/lengau/.local/state/testcraft/log/testcraft-20230907-134412.119016.log'                                                                                                                                          
2023-09-07 13:44:12.119 Extension 'invalid_extension' does not exist                                                                                                                                                                                                 
2023-09-07 13:44:12.120 Traceback (most recent call last):                                                                                                                                                                                                           
2023-09-07 13:44:12.120   File "/home/lengau/Work/Code/charmcraft/test.py", line 9, in emit_ext_err                                                                                                                                                                  
2023-09-07 13:44:12.120     {}[extension]                                                                                                                                                                                                                            
2023-09-07 13:44:12.120 KeyError: 'invalid_extension'                                                                                                                                                                                                                
2023-09-07 13:44:12.120 Full execution log: '/home/lengau/.local/state/testcraft/log/testcraft-20230907-134412.119016.log'                                                                                                                                           
(charmcraft) lengau@hyperion:~/Work/Code/charmcraft$ python test.py --no-cause
2023-09-07 13:44:16.709 hi                                                                                                                                                                                                                                           
2023-09-07 13:44:16.709 Logging execution to '/home/lengau/.local/state/testcraft/log/testcraft-20230907-134416.708554.log'                                                                                                                                          
2023-09-07 13:44:16.709 Extension 'invalid_extension' does not exist                                                                                                                                                                                                 
2023-09-07 13:44:16.709 Full execution log: '/home/lengau/.local/state/testcraft/log/testcraft-20230907-134416.708554.log' 
```

Thinking through UX now, maybe we should put a `details=f"Registered extensions: {get_extension_names()}"`

charmcraft/extensions/extension.py
lengau: Then it's probably best with the dict option, since our charmcraft bases naturally contain a `name` and a `channel`

lengau: Ahh ok. I misread this, never mind :-)

charmcraft/extensions/_utils.py
sergiusens: this is probably legacy at this point.

charmcraft/extensions/registry.py
sergiusens: extensions are not going to be registered externally

charmcraft/extensions/registry.py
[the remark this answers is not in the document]
syu-w: OK, I changed it to raise from None with details.

charmcraft/extensions/_utils.py
syu-w: I added `self.yaml_data: Final[Dict[str, Any]] = yaml_data`, but I kept deep copy for now.

sergiusens (approved):

charmcraft/errors.py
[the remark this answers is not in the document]
lengau: I'd like to tackle these issues at some point, but it's out of scope for now.

lengau (approved):
Looks good, thank you!

charmcraft/errors.py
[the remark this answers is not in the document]
syu-w: yeah, I don't think we should try import something in the error file.

PR #1264, opened 2023-09-19
fix(logging): print details errors from craft-providers

syu-w:


charmcraft/main.py
```diff
@@ -30,7 +30,8 @@
     ProvideHelpException,
     emit,
 )
-from craft_store import errors
+import craft_store.errors
+import craft_providers.errors
```
lengau, line 34:
```suggestion
import craft_providers.errors
import craft_store.errors
```

This is what ruff's complaining about I believe. (A `tox run -e format-ruff` should also fix it)

lengau (approved):

Issue #1242, opened 2023-08-28
Support for extensions (command line)


tigarmo:
### What needs to get done

Extension framework as done in Snapcraft and Rockcraft

### Why it needs to get done

To support easy creation of charms without the complexities of each framework

PR #1268, opened 2023-09-20
feat(extensions): add cli commands

syu-w:
Add new commands expand-extensions and list-extensions

Fix: https://github.com/canonical/charmcraft/issues/1242

### DIFF
--- a/charmcraft/commands/extensions.py
+++ b/charmcraft/commands/extensions.py
@@ -1,0 +1,100 @@
+# Copyright 2023 Canonical Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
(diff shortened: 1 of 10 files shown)


tests/test_main.py
```diff
@@ -704,7 +705,7 @@ def test_systemdetails_hidden_auth():
 
 # -- generic tests for all Charmcraft commands
 
-all_commands = list.__add__(*[cgroup.commands for cgroup in COMMAND_GROUPS])
+all_commands = list(itertools.chain(*(cgroup.commands for cgroup in COMMAND_GROUPS)))
```
lengau, line 708: Did this come from a linter issue? If not, please remove this change here (though a separate PR with this fix would be much appreciated!)

syu-w: It is a invalid code when `COMMAND_GROUPS` has more than 2 items.
`TypeError: expected 1 argument, got 2`

charmcraft/commands/extensions.py
```diff
@@ -0,0 +1,90 @@
...
+_overview = """
+Extensions allow user use pre-defined configuration in their charmcraft.yaml.
+"""
+
+
+class ListExtensionsCommand(BaseCommand):
+    """List available extensions for all supported bases."""
+
+    common = True
+    name = "list-extensions"
```
lengau, line 40: Please include a hidden `extensions` version of this command (see the plurals section at https://discourse.ubuntu.com/t/command-structure/18556).

lengau (changes requested):

tests/test_main.py
lengau: Ahh, I see. All good then, thanks!

sergiusens (approved):

Issue #1256, opened 2023-09-13
Support global cache for pip in charmcraft


mr-cal:
### What needs to get done

Integrate https://github.com/canonical/craft-providers/issues/397 into charmcraft via https://github.com/canonical/charmcraft/pull/1249

### Why it needs to get done

This will allow for pip packages (and especially wheels) to be reused between containers and reduce the time needed to build a charm.

PR #1270, opened 2023-09-21
feat(pack): add shared cache support

syu-w:
Default enabled.
Currently only support pip.
Default cache storage at `~/.cache/charmcraft`

Path could be changed use env `CRAFT_SHARED_CACHE=XXX`

Fix: https://github.com/canonical/charmcraft/issues/1256


syu-w:
Not fully finished yet, but should working.

charmcraft/commands/pack.py
```diff
@@ -100,6 +100,11 @@ def fill_parser(self, parser):
             type=pathlib.Path,
             help="Dump measurements to the specified file",
         )
+        parser.add_argument(
+            "--no-use-shared-cache",
```
cmatsuoka, line 104: `--disable-shared-cache` can be an alternative if we're not using the `--no` pattern elsewhere.

sergiusens: we agreed to remove this for now

sergiusens: in the future, in should be `enable/disable-shared-cache`

syu-w: I was thinking use `BooleanOptionalAction`, but that one only added in Python 3.9.
So now I just change it to `--disable-shared-cache`

charmcraft/commands/pack.py
```diff
@@ -100,6 +100,11 @@ def fill_parser(self, parser):
             type=pathlib.Path,
             help="Dump measurements to the specified file",
         )
+        parser.add_argument(
+            "--disable-shared-cache",
+            action="store_true",
```
sergiusens, line 105: I thought we said not to do this now, in any case, if we are keeping it,  we should also use `dest=shared_cache_enabled` and make the action store False, and the default to be True to avoid negative positive checks which play tricks when reading code.

@lengau ?

lengau: Definitely agree that we should prefer a `store_false` here if we're keeping the argument. 

I'm okay to drop the ability to disable the shared cache too — that originally came through my intention to start with it being an optional enable that we'd eventually turn on by default.

syu-w: OK, then I remove that part.

charmcraft/env.py
```diff
@@ -22,6 +22,13 @@
 import pathlib
 from typing import Optional
 
+from charmcraft.const import DEFAULT_SHARED_CACHE_DIR, SHARED_CACHE_ENV_VAR
+
+
+def get_host_shared_cache_path():
+    """Path for host shared cache."""
+    return pathlib.Path(os.getenv(SHARED_CACHE_ENV_VAR, DEFAULT_SHARED_CACHE_DIR)).expanduser()
```
lengau, line 30: Please use the system shared cache directory and ensure the cache directory exists.

We already have `platformdirs` as an indirect dependency, so we can make it a direct dependency without much harm. Platformdirs has a [user_cache_path](https://platformdirs.readthedocs.io/en/latest/api.html#cache-directory) that can be filled in like:

```suggestion
    shared_cache_env = os.getenv(SHARED_CACHE_ENV_VAR)
    if shared_cache_env is not None:
        cache_path = pathlib.Path(shared_cache).expanduser().resolve()
        cache_path.mkdir(parents=True, exist_ok=True)
        return cache_path
    return platformdirs.user_cache_path(appname="charmcraft", ensure_exists=True)
```

charmcraft/package.py
```diff
@@ -89,7 +89,17 @@ def launch_shell(*, cwd: Optional[pathlib.Path] = None) -> None:
...
+    def __init__(
+        self,
+        *,
+        config,
+        force,
+        debug,
+        shell,
+        shell_after,
+        measure,
+        use_shared_cache=True,
```
lengau, line 101: Let's rather push the shared cache path into the builder on creation and make it the creator's task to determine the cache path.

tests/test_providers.py
```diff
@@ -515,6 +515,13 @@ def test_get_instance_name(
     ) == expected.format(inode=tmp_path.stat().st_ino)
 
 
+@pytest.mark.parametrize(
+    "cache_path",
+    [
+        None,
+        pathlib.Path("/tmp/charmcraft-cache"),
```
cmatsuoka, line 522: This global cache may affect other test runs in the same machine, shouldn't we remove it after running this specific test with all parametrized test cases?

cmatsuoka:
Thanks! It seems that we have a glob expansion problem in a test however (or rather a non-expansion?)

charmcraft/const.py
```diff
@@ -22,6 +22,9 @@
 
 IMAGE_INFO_ENV_VAR = "CHARMCRAFT_IMAGE_INFO"
 
+SHARED_CACHE_ENV_VAR = "CRAFT_SHARED_CACHE"
+DEFAULT_SHARED_CACHE_DIR = "~/.cache/charmcraft"
```
lengau, line 26:
```suggestion
```

charmcraft/main.py
```diff
@@ -105,7 +105,7 @@
 ]
 
 # non-charmcraft useful environment variables to log
-EXTRA_ENVIRONMENT = ("DESKTOP_SESSION", "XDG_CURRENT_DESKTOP")
+EXTRA_ENVIRONMENT = ("DESKTOP_SESSION", "XDG_CURRENT_DESKTOP", "CRAFT_SHARED_CACHE")
```
lengau, line 108:
```suggestion
EXTRA_ENVIRONMENT = ("DESKTOP_SESSION", "XDG_CURRENT_DESKTOP", SHARED_CACHE_ENV_VAR)
```

tests/test_providers.py
lengau, line 522:
```suggestion
        pathlib.Path("charmcraft-cache"),
```

lengau (approved):
Looks good after committing the left-out change I marked as a suggestion :smile:

sergiusens:
The snaps tests failure is unrelated

PR #1275, opened 2023-09-22
feat: use craft-application for pack command

lengau:


syu-w (approved):

sergiusens:
charmcraft internal error: AttributeError("'CharmcraftConfig' object has no attribute 'get_build_plan'")

charmcraft/application/main.py
```diff
@@ -19,42 +19,49 @@
...
-        self.services.set_kwargs(
-            "package",
-            work_dir=self._work_dir,
-            prime_dir=self.services.lifecycle.prime_dir,
-        )
+        return [craft_cli.CommandGroup("Lifecycle", commands=[pack.PackCommand])]
+
+    # def _configure_services(self) -> None:
+    #     self.services.set_kwargs(
+    #         "package",
```
cmatsuoka, line 58: Leftover comment?

charmcraft/application/main.py
```diff
@@ -65,7 +72,7 @@ def _get_dispatcher(self) -> craft_cli.Dispatcher:
             log_filepath=self.log_path,
         )
 
-        dispatcher = craft_cli.Dispatcher(
+        dispatcher = craft_application.application._Dispatcher(
```
cmatsuoka, line 75: Shouldn't Craft Application make `_Dispatcher` public? (also no linter complaints?)

charmcraft/cmdbase.py
```diff
@@ -60,15 +60,29 @@ def _check_config(self, config_file: bool = False, bases: bool = False) -> None:
...
+
+        if isinstance(self.config, CharmcraftConfig):
+            # Classic, pre craft-application run
+            project = self.config.project
+            project_bases = self.config.bases
+        else:
+            # Using craft-application
+            project = self._services.project.project
+            project_bases = self._services.project.bases
+        if config_file and not project.config_provided:
```
cmatsuoka, line 73: Is Charmcraft able to run with and without c-a? How does that work?

charmcraft/commands/pack.py
```diff
@@ -122,51 +124,54 @@ def run(self, parsed_args: argparse.Namespace) -> None:
...
             debug=parsed_args.debug,
             shell=parsed_args.shell,
             shell_after=parsed_args.shell_after,
             measure=parsed_args.measure,
+            package_service=self._services.package,
         )
 
         # decide if this will work on a charm or a bundle
-        if self.config.type == "charm":
+        if config.type == "charm":
```
cmatsuoka, line 140: I suppose the plans are to eviscerate the charm builder at a later point and split its functionality so that the lifecycle run is handled by the lifecycle service, is that correct?

mr-cal (commented):
Let me know when you want a review from me (I missed the original request and since then, it looks like there's been reviews and some tests to fix, so my review may not be very valuable now).

lengau:
superseded

PR #1289, opened 2023-09-28
chore(deps): update dependencies

lengau:


syu-w (approved):

setup.py
```diff
@@ -47,10 +47,10 @@
...
 
 lint_requires = [
-    "black==23.7.0",
-    "codespell[tomli]==2.2.5",
-    "ruff==0.0.280",
-    "yamllint==1.32.0",
+    "black>=23.7.0,<24.0.0",
+    "codespell[tomli]>=2.2.5,<3.0.0",
+    "ruff==0.0.282",
+    "yamllint>=1.32.0,<2.0.0",
```
syu-w, line 53: Maybe we can use `~=` for those deps? Not sure if it is supported in the `setup.py`

mr-cal (approved):

setup.py
lengau: `~=` sometimes gets weird with renovate, so I'm preferring not right now until they fix it.

PR #1294, opened 2023-09-28
chore(deps): craft-providers to v1.18.0

syu-w:


mr-cal (approved):

lengau (approved):
Thanks! I've rebased which should get all the tests passing now too I believe.

PR #1296, opened 2023-09-29
ci: enable codespell

lengau:
This is broken into two commits, first fixing spelling and then enabling codespell.

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -1316,7 +1316,7 @@ def test_schema_other_bundle_part_no_source(
 
 
 # -- tests to check the double layer schema loading; using the 'charm' plugin
-#    because it is the default (and has good default properties to be overriden and )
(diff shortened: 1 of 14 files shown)


syu-w (approved):

tests/test_config.py
```diff
@@ -1747,7 +1747,7 @@ def test_schema_doublelayer_parts_with_charm_overriding_properties(
                     channel: "30.04"
                 parts:
                   charm:
-                    charm-point: different.py  # mispelled!
+                    charm-point: different.py  # misspelled!
```
tigarmo, line 1750: :laughing:

tigarmo (approved):

tests/test_config.py
lengau: I was almost tempted to leave it!

PR #1299, opened 2023-09-29
refactor(store): use dataclasses for store models

lengau:


charmcraft/commands/store/charmlibs.py
```diff
@@ -29,10 +28,20 @@
 
 from charmcraft.errors import BadLibraryNameError, BadLibraryPathError
 
-LibData = namedtuple(
-    "LibData",
-    "lib_id api patch content content_hash full_name path lib_name charm_name",
-)
+
+@dataclass(frozen=True)
+class LibData:
```
tigarmo, line 33: do you think there's any chance that this change breaks someone somehow?

charmcraft/commands/store/charmlibs.py
```diff
@@ -42,9 +51,9 @@ class LibInternals:
     lib_id: str
     api: int
     patch: int
-    pydeps: list
+    pydeps: List[str]
     content_hash: str
-    content: bytes
+    content: str
```
tigarmo, line 56: is the type change here on purpose? why?

lengau: It is — when I was making these changes I double-checked it with mypy and discovered it was always supposed to be a `str`

charmcraft/commands/store/charmlibs.py
lengau: I shouldn't think so. This file contains helpers for the various `*-lib` store commands, so it would have to be someone who's using our internal data structures and trying to iterate over a LibData object (which one shouldn't be doing anyway). The fields are kept in the same order so the constructors are the same.

If we do desire to keep it a NamedTuple still I could convert it into a typed NamedTuple, but I would be surprised if there's any direct external use of this class.

tigarmo (approved):

mr-cal (approved):

charmcraft/commands/store/store.py
```diff
@@ -35,24 +35,193 @@
...
+
+    name: str
+    fallback: str
+    track: str
+    risk: str
+    branch: str
+
+
+@dataclasses.dataclass(frozen=True)
+class Library:
```
mr-cal, line 199: I guess the order doesn't matter?  This one is the only one I noticed that is in a different order.

lengau: No, these are just internal data structures

PR #1304, opened 2023-10-02
test: make parts test self-contained

lengau:
The test in question has to run in a directory with a requirements.txt file - this environmental requirement should be created during setup.


lengau:
Lint issue is taken care of in #1305

lengau:
Redoing in #1311

PR #1306, opened 2023-10-03
feat: strict dependency management

### LINKED ISSUE #1269: Packing method is inherently susceptible to failure even with the most strict version pinning
### What needs to get done
Guarantee the "exact same charm" would keep packing fine.

### Why it needs to get done

Recently, following up a new release of `setuptools-scm` (8), all 20.04 charms failed to pack:
- https://github.com/pypa/setuptools_scm/issues/918
- https://chat.charmhub.io/charmhub/pl/bkagbm1i87f6ijjt5dw8tubtpc

Various types of version pinning did not work for some reason.
This is concerning because it means version pinning isn't enough to guarantee the "exact same charm" would keep packing/working fine.

Eventually two solutions surfaced:
- Bump charm OS to 22.04, but that is just avoiding the problem (also not currently viable due to https://bugs.launchpad.net/juju/+bug/2030110).
- Use `charm-binary-python-packages` for the toplevel package that introduced the `setuptools-scm` dependency, but that is also avoiding the problem.

lengau:
Fixes #1269


tests/spread/dependencies/strict-dependencies-errors/extra-packages/charmcraft.yaml
```diff
@@ -0,0 +1,13 @@
+name: charm
+type: charm
+title: Strict dependencies test charm
+summary: a test charm for checking that strict dependencies are used.
+description: a test charm for checking that strict dependencies are used.
```
tigarmo, line 5: suggestion: use the description to record why exactly this charm should fail to pack

tests/spread/dependencies/strict-dependencies-errors/task.yaml
```diff
@@ -0,0 +1,10 @@
+summary: test charm build failures with strict dependencies
+
+environment:
+  CHARM/no_requirements: no-requirements/
+  CHARM/extra_packages: extra-packages/
+
+execute: |
+  cd $CHARM
+  # Charmcraft should fail for these charms.
+  ! charmcraft pack --verbose
```
tigarmo, line 10: is there an error message that you can match against, to protect against this packing failing for reasons _other_ than strict package validation?

tests/spread/dependencies/strict-dependencies/charm/requirements.txt
```diff
@@ -0,0 +1 @@
+ops==2.5.1
```
tigarmo, line 1: so just to clarify, this requirements file has ops at 2.5.1, but the charm-binary-python-packages property is ops>=2.6, and that's accepted? How come?

tigarmo: I see from the task.yaml comments that the version installed is 2.5.1, but isn't that very surprising to users that explicitly requested >2.6?

tigarmo (commented):
thanks!

tests/spread/dependencies/strict-dependencies-errors/extra-packages/charmcraft.yaml
lengau: Good idea! I blindly copy/pasted these from the success case and didn't think about updating that metadata.

I've also added another variant of this test that checks for extra packages in `charm-python-binary-packages`, since that should also be a subset of the packages mentioned in requirements.txt.

tests/spread/dependencies/strict-dependencies/charm/requirements.txt
lengau: Hmmm good point. I didn't consider them "requesting" ops >= 2.6 here, but rather allowing it as binary. However, for someone used to the old dependency management the principle of least surprise says I should probably enforce than strict dependencies only allow a package name in this field.

tests/spread/dependencies/strict-dependencies-errors/extra-binary-packages/charmcraft.yaml
```diff
@@ -0,0 +1,17 @@
+name: charm
+type: charm
+title: Strict dependencies test charm
+summary: a test charm for checking strict dependency build failures.
+description: |
+  This charm fails to build because it has strict dependencies enabled and
```
tigarmo, line 6: :star_struck:

tigarmo (approved):

syu-w (approved):

cmatsuoka (approved):
Looks good. Can you update the date in the copyright notices in the changed files if it's not already 2023 before landing?

tests/spread/dependencies/strict-dependencies-errors/task.yaml
lengau: Any relevant error messages should warn about `charm-strict-dependencies` so I've MATCHed for that

tests/spread/dependencies/strict-dependencies/charm/requirements.txt
lengau: Strict mode now only allows package names in `charm-binary-python-packages`

PR #1312, opened 2023-10-03
style(ci): enable mypy

lengau:


charmcraft/__init__.py
```diff
@@ -21,9 +21,9 @@
 import pkg_resources
 
 
-def _get_version():
+def _get_version() -> str:
     if os.getenv("SNAP_NAME") == "charmcraft":
-        return os.getenv("SNAP_VERSION")
+        return os.getenv("SNAP_VERSION", "")
```
tigarmo, line 26: a possibly pedantic point: if `SNAP_NAME` is set then `SNAP_VERSION` should always be set too, right? In that case, how about raising an error instead if it's not set, instead of returning an empty string?

pyproject.toml
```diff
@@ -31,30 +31,58 @@ skip_empty = true
 #fail_under = 80
 
 [tool.pyright]
-strict = ["starcraft"]
+strict = ["charmcraft"]
```
tigarmo, line 34: :sweat_smile:

pyproject.toml
```diff
@@ -31,30 +31,58 @@ skip_empty = true
...
     "results",
+    "tests",
+    "tools",
 ]
 warn_unused_configs = true
 warn_redundant_casts = true
 strict_equality = true
-strict_concatenate = true
-warn_return_any = true
+#warn_return_any = true
```
tigarmo, line 54: should these stay commented?

pyproject.toml
```diff
@@ -31,30 +31,58 @@ skip_empty = true
...
 disallow_subclassing_any = true
 disallow_untyped_decorators = true
-disallow_any_generics = true
+#disallow_any_generics = true
+
+[[tool.mypy.overrides]]
+# Ignore typing errors in most legacy packages until we fix them.
```
tigarmo, line 60: do we have an issue to fix these?

tigarmo (approved):

pyproject.toml
lengau: For now yes, as these cause CI failures that'll be fixed with craft-application update. I've put together an issue and premade a PR for it though so the change is trivial once it's ready. https://github.com/canonical/charmcraft/issues/1314

pyproject.toml
lengau: #1316

I also removed a couple of these now as changes since I created this PR fixed them.

charmcraft/__init__.py
lengau: Pedantically correct :-)

I'm choosing not to take action now though since the craft-application rebase will change this function anyway.

lengau:
Spread test is fixed in https://github.com/canonical/charmcraft/pull/1322

PR #1321, opened 2023-10-10
chore(snap): base migration from core20 to core22

### LINKED ISSUE #1320: Migrate to core22
### What needs to get done

Migrate to core22

### Why it needs to get done

To use a newer stack

syu-w:
Fixes #1320 
CRAFT-2117

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,4 +1,4 @@
-# Copyright 2020 Canonical Ltd.
+# Copyright 2020-2023 Canonical Ltd.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -14,7 +14,7 @@
 #
 # For further info, check https://github.com/canonical/charmcraft
 name: charmcraft
-base: core20
+base: core22
 summary: Charmcraft supports Kubernetes operator development and collaboration
 license: Apache-2.0
 description: |
@@ -39,7 +39,7 @@ description: |
   Windows apps. The universal operator pattern is very exciting for
   multi-cloud application management.
 
-adopt-info: charmcraft  # look for 'snapcraftctl set-*' in the charmcraft part
+adopt-info: charmcraft  # look for 'craftctl set *' in the charmcraft part
 architectures:
   - build-on: s390x
   - build-on: ppc64el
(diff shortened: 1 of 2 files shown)


charmcraft/charm_builder.py
```diff
@@ -240,6 +240,28 @@ def _install_dependencies(self, staging_venv_dir):
         # create virtualenv using the host environment python
         with instrum.Timer("Creating venv"):
             _process_run(["python3", "-m", "venv", str(staging_venv_dir)])
+
+        python_cmd = str(_find_venv_bin(staging_venv_dir, "python3"))
```
tigarmo, line 244: can you explain why this new bit is necessary?

snap/snapcraft.yaml
```diff
@@ -149,28 +161,27 @@ parts:
...
     python-packages: [wheel]
     build-environment:
-      - LDFLAGS: -L/usr/lib/python3.8
-      - CPPFLAGS: -I/usr/include/python3.8
+      - LDFLAGS: -L/usr/lib/python3.10
+      - CPPFLAGS: -I/usr/include/python3.10
     override-pull: |
-      # do the usual pull stuff
-      snapcraftctl pull
+      ${SNAP}/libexec/snapcraft/craftctl default
```
tigarmo, line 167: it's odd that you'd have to spell out the full path to `craftctl` like this, how come?

sergiusens: oh, interesting, all crafts might suffer from this, it might be to avoid using the craftctl from this build itself

syu-w:
It seems there are multiple bugs:

snap/snapcraft.yaml
syu-w: In the `PRIME` step, the `PATH` changed to `PATH=/root/parts/python3/install/usr/bin:/root/stage/usr/bin:/root/stage/bin:/snap/snapcraft/x1/libexec/snapcraft:/snap/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin`, where there is another `craftctl` located at `/root/stage/bin/craftctl`. And that one doesn't work because `craft-parts` not exist.

tigarmo: Maybe you need to `organize` the `craftctl` into `libexec`, like you did in rockcraft?
```
    organize:
        bin/craftctl: libexec/rockcraft/craftctl

File: charmcraft/charm_builder.py (hunk 1 above)
Comment (syu-w) [reply charmcraft/charm_builder.py:244]: This is another bug:
When it creating the `venv`, `pip` think the `pip` is already installed, and point it to the snapcraft pip. So the pip commands missing in the venv. And It has to be force reinstall to make it work.

File: snap/snapcraft.yaml (hunk 2 above)
Comment (syu-w) [reply snap/snapcraft.yaml:144]: > Maybe you need to `organize` the `craftctl` into `libexec`, like you did in rockcraft?
> 
> ```
>     organize:
>         bin/craftctl: libexec/rockcraft/craftctl
> ```

Yeah, it work without need the full path, since it is not in the path.

File: charmcraft/charm_builder.py (hunk 1 above)
Comment (tigarmo) [reply charmcraft/charm_builder.py:244]: you mean that after that `["python3", "-m", "venv", str(staging_venv_dir)]` command, `pip` does *not* exist in `staging_venv_dir`?

File: snap/snapcraft.yaml (hunk 2 above)
Comment (syu-w) [reply snap/snapcraft.yaml:144]: But the actually issue is `craftctl default` result is different than the non-override result.

The linking error not happened if `override-prime` not exist. But if you put:
```
    override-prime: |
      craftctl default
```
The `enable-patchelf` not work for some binaries.

File: charmcraft/charm_builder.py (hunk 1 above)
Comment (syu-w) [reply charmcraft/charm_builder.py:244]: Yes

Comment (tigarmo) [reply charmcraft/charm_builder.py:244]: This is super weird, that's pretty much what [craft-parts does](https://github.com/canonical/craft-parts/blob/main/craft_parts/plugins/python_plugin.py#L95) and it worked fine for snapcraft and rockcraft. How can pip not be there??

Comment (syu-w) [reply charmcraft/charm_builder.py:244]: The process are, create the basic python link, then use `ensurepip` to call pip to install another pip from wheel. However, due to how `snapcraft`(?) set the env, the pip found the pip in the snapcraft's venv, and skipped the install.

File: snap/snapcraft.yaml [hunk 3]
```diff
@@ -148,28 +138,27 @@ parts:
...
-      snapcraftctl set-version "$version"
+      ${SNAP}/libexec/snapcraft/craftctl set version="${version}"
     override-build: |
-      snapcraftctl build
+      ${SNAP}/libexec/snapcraft/craftctl default
       # why is this needed?
       cp -v completion.bash ../install
       # python3 fixup symlink (snapcraft bug)
-      ln -sf ../usr/bin/python3.8 $SNAPCRAFT_PART_INSTALL/bin/python3
+      ln -sf ../usr/bin/python3.10 $CRAFT_PART_INSTALL/bin/python3
```
tigarmo, line 155: You can remove this (the plugin in core22 already does this correctly)

charmcraft/charm_builder.py
tigarmo: weird. As an experiment, you can try "unstaging" `bin/pip` from the charmcraft snap, and then there wouldn't be a `pip`in the snapcraft venv

snap/snapcraft.yaml
syu-w: OK, removed

snap/snapcraft.yaml
tigarmo: Yes I can confirm that if you don't override-prime, the "patchelfing" works.
From what I can tell, the patchelf feature in Snapcraft is implemented as a prime callback, so maybe there's a bug in the interaction between callbacks and `craftctl`
ping @cmatsuoka

tigarmo: @syu-w I think the cleanest course of action here is to actually tear most of this `snapcraft.yaml` down and rewrite it in a way that's similar to Rockcraft and Snapcraft, like Sergio suggested. I have the feeling that there's a lot of cruft here that is no longer necessary and is making our understanding of what's going on impossible

snap/snapcraft.yaml
```diff
@@ -70,71 +74,57 @@ apps:
...
-      - LDFLAGS: -L/usr/lib/python3.8
-      - CPPFLAGS: -I/usr/include/python3.8
-    override-build: |
-      pip wheel $(cat requirements-focal.txt)
+      ${SNAP}/libexec/snapcraft/craftctl default
+      install -D -m 0755 $CRAFT_PROJECT_DIR/snap/local/sitecustomize.py $CRAFT_PART_INSTALL/usr/lib/python3.10/sitecustomize.py
+    override-prime: |
+      ${SNAP}/libexec/snapcraft/craftctl default
+      # override python default venv wheels dir
+      sed -i "s/sysconfig.get_config_var('WHEEL_PKG_DIR')/'\/snap\/charmcraft\/current\/usr\/share\/python-wheels\/'/" ${CRAFT_PRIME}/usr/lib/python3.10/ensurepip/__init__.py
```
sergiusens, line 117: why do we need this?

snap/snapcraft.yaml
sergiusens: yes, @syu-w , please as stated 3 days ago, copy what is in rockcraft and snapcraft's snapcraft.yaml, even the part names, and add the special things needed for charmcraft in a separate commit. There are things that do not make sense here.

tigarmo (approved):
Great! Glad you got this working

snap/snapcraft.yaml
```diff
@@ -48,93 +48,69 @@ architectures:
...
   # - build-on: i386  # there is no i386 core20
 
+# https://github.com/snapcore/snapcraft/issues/4187
+environment:
+  PATH: "$SNAP/libexec/charmcraft:/snap/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
+  LD_LIBRARY_PATH: "$SNAP/none"
+
 apps:
   charmcraft:
-    # run Python in unbuffered mode so when it's running internally in the instance it will output
```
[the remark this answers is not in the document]
sergiusens: now that we have a working baseline, can you create a commit adding this unbuffered option back in, easy to revert if it does not work

snap/snapcraft.yaml
syu-w: We no longer need this as it fixed in the new version

sergiusens (approved):

PR #1323, opened 2023-10-11
perf(snap): speed up repeat snap builds

lengau:
This change caches the python-apt wheel, speeding up repeated builds of the snap in the same provider.


lengau:
Strict dependencies spread failure is fixed in #1322

syu-w (approved):

cmatsuoka (approved):

PR #1325, opened 2023-10-12
test(spread): consolidate basic init/build tests

lengau:
CRAFT-2127


tests/spread/smoketests/basic-centos-7/task.yaml
```diff
@@ -1,38 +1,34 @@
...
-      - name: centos
-        channel: "7"
-  EOF
+prepare: |
+  tests.pkgs install unzip
+  charmcraft init --project-dir=charm
 
-  popd
+  sed -i "s/- name: ubuntu/- name: ${BASE_NAME}/g" charm/charmcraft.yaml
+  sed -i "s/channel:.*/channel: '${BASE_CHANNEL}'/g" charm/charmcraft.yaml
```
cmatsuoka, line 20: Is it safe to replace inline in this case, instead of using a template file as source?

syu-w: It is safe, if the channel misplaced, pydantic should given an error.

syu-w (approved):

mr-cal (approved):

PR #1328, opened 2023-10-14
feat(craft-app): Add craft-application compatible models

lengau:


charmcraft/const.py
```diff
@@ -48,6 +49,31 @@
 MANDATORY_HOOK_NAMES = frozenset(("install", "start", "upgrade-charm"))
 HOOKS_DIRNAME = "hooks"
 
+SUPPORTED_BASES = frozenset(
+    (
+        BaseName("ubuntu", "18.04"),
```
tigarmo, line 54: out of curiosity, is there a plan to drop bionic?

charmcraft/models/manifest.py
```diff
@@ -0,0 +1,76 @@
...
+class Manifest(models.BaseMetadata):
+    """A manifest.yaml file for Juju.
+
+    See: https://juju.is/docs/sdk/manifest-yaml
+    """
+
+    charmcraft_version: str = charmcraft.__version__
+    charmcraft_started_at: str
+    bases: Optional[List[Base]]
+    analysis: Dict[Literal["attributes"], List[Attribute]] = {"attributes": []}
```
tigarmo, line 51: that's a curious structure

charmcraft/models/project.py
```diff
@@ -0,0 +1,480 @@
...
+        """Create a Charmcraft project from a dictionary of data."""
+        project_type = data.get("type")
+        if project_type == "charm":
+            return Charm.parse_obj(data)
+        if project_type == "bundle":
+            return Bundle.parse_obj(data)
+        raise ValueError(f"field type cannot be {project_type!r}")
+
+    @classmethod
+    def from_yaml_file(cls, path: pathlib.Path) -> Self:
```
tigarmo, line 238: Is the git revision of craft-application pinned?

charmcraft/models/project.py
```diff
@@ -0,0 +1,480 @@
...
+                "Invalid 'charmcraft.yaml' file",
+                details=f"File generated a {type(data)} object, expected a dictionary",
+                resolution="Ensure 'charmcraft.yaml' is valid",
+                reportable=False,
+                docs_url="https://juju.is/docs/sdk/charmcraft-yaml",
+            )
+
+        charm_dir = path.parent
+
+        metadata_file = charm_dir / METADATA_FILENAME
```
tigarmo, line 262: I suppose in the future this code could go in https://github.com/canonical/craft-application/blob/main/craft_application/application.py#L415

mr-cal (approved):
I don't have a strong understanding of the data structures and functions needed for craft-application, so I trust you're creating them correctly.  Overall, I don't see any code smells and the tests look comprehensive.

tigarmo (approved):
oh man this is a lot of work! well done ;)

charmcraft/models/manifest.py
```diff
@@ -0,0 +1,76 @@
...
+# For further info, check https://github.com/canonical/charmcraft
+"""Model for output charm's manifest.yaml file."""
+import itertools
+import json
+import os
+from typing import Any, Dict, Iterable, List, Literal, Optional
+
+from craft_application import models
+from craft_cli import CraftError
+from pydantic import dataclasses
```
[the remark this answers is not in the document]
lengau: It's intentionally written as a drop-in replacement for built-in dataclasses, but it provides pydantic validation. So in this case, `Attribute(name=1, result=None)` would raise a ValidationError

charmcraft/models/manifest.py
lengau: It is, but when I looked at the spec, the purpose is to allow more analysis output in the manifest in future.

charmcraft/const.py
lengau: I'm sure there is, but I'm not aware of it right now. I think we have a few big customers using ESR for their charms, so we'd have to be careful.

charmcraft/models/project.py
```diff
@@ -0,0 +1,480 @@
...
+    This is equivalent to the short form base definition.
+    """
+
+    name: str
+    channel: str
+    architectures: List[str]
+
+
+LongFormBasesDict = TypedDict(
+    "LongFormBasesDict", {"build-on": List[BaseDict], "run-on": List[BaseDict]}
```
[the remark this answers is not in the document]
lengau: Unfortunately yes. My (very informal) plan right now is to wait for us to standardise platforms and then allow either a `bases` key or a `platforms` key, with `platforms` using the current preferred terminology.

charmcraft/models/project.py
```diff
@@ -0,0 +1,480 @@
...
+                build_on_index=1
+            ),
+        ]
+
+        Here the string "multi" defines a destination platform that has multiple architectures.
+        """
+        return list(CharmBuildInfo.gen_from_bases_configurations(*self.bases))
+
+
+class Bundle(CharmcraftProject):
```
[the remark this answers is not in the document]
lengau: Bundles don't even start up any providers. A bundle is basically just a zip file with a bundle.yaml that references the charms and their relations. When I implement bundle packing I might make a static build plan, but I don't even think I'll need to do that.

PR #1330, opened 2023-10-16
feat(craft-app): Add craft-application services

lengau:
CRAFT-2142

### DIFF
--- a/charmcraft/application/commands/lifecycle.py
+++ b/charmcraft/application/commands/lifecycle.py
@@ -1,0 +1,421 @@
+# Copyright 2023 Canonical Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# For further info, check https://github.com/canonical/charmcraft
+"""craft-application based lifecycle commands."""
+from __future__ import annotations
+
+import os
+import pathlib
+import subprocess
+import textwrap
+from typing import TYPE_CHECKING
+
+from craft_cli import ArgumentParsingError, CommandGroup, CraftError, emit
+from typing_extensions import override
+
+from charmcraft.application.commands.base import CharmcraftCommand
+
+if TYPE_CHECKING:  # pragma: no cover
+    import argparse
+
+BUNDLE_MANDATORY_FILES = ["bundle.yaml", "README.md"]
+
+
+def get_lifecycle_command_group() -> CommandGroup:
+    """Return the lifecycle related command group."""
+    commands: list[type[_LifecycleCommand]] = [
+        CleanCommand,
+        PullCommand,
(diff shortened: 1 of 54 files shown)


charmcraft/application/commands/lifecycle.py
lengau, line 1: This file is mostly a copy of lifecycle.py in craft-application. I'll try to point out the important pieces.

charmcraft/application/commands/lifecycle.py
```diff
@@ -0,0 +1,421 @@
...
+        usually `src/charm.py`.  See `charmcraft init` to create a
+        template charm directory structure.
+
+        For the bundle you must already have a `bundle.yaml` (can be
+        generated by Juju) and a README.md file.
+        """
+    )
+
+    @override
+    def fill_parser(self, parser: argparse.ArgumentParser) -> None:
```
lengau, line 272: These additional parser fills are to make `charmcraft pack` backwards compatible. The rest of the commands don't need this since charmcraft stable doesn't have any lifecycle commands except `pack` and `clean`

mr-cal (commented):
Overall it looks good, but again this is a naive review as I'm not very comfortable with craft-application.

Dumb question - Is this branch supposed to work? When building a charm, I get the same error as the one in CI.

tigarmo (approved):
thanks! This is a lot but looks good

charmcraft/services/package.py
```diff
@@ -0,0 +1,138 @@
...
+    CharmcraftServiceFactory = "CharmcraftServiceFactory"
+
+
+class PackageService(services.PackageService):
+    """Business logic for creating packages."""
+
+    _project: models.CharmcraftProject  # type: ignore[assignment]
+    _services: CharmcraftServiceFactory
+
+    def __init__(  # (too many arguments)
```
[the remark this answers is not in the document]
lengau: Ruff auto-formatter removing the noqa but not the rest of the comment 😆

charmcraft/services/provider.py
```diff
@@ -0,0 +1,38 @@
...
+    """Business logic for creating packages."""
+
+    def get_base(
+        self,
+        base_name: bases.BaseName | tuple[str, str],
+        *,
+        instance_name: str,
+        **kwargs: bool | str | None,
+    ) -> bases.Base:
+        """Get."""
```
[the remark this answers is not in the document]
lengau: "Okay, I'm going to describe this method in a docstring. It's to get the... wait why am I out of tea?"

lengau:
@mr-cal to answer your question, the init command doesn't work currently, which is why the spread tests were failing. Your note about it failing for you made me realise I had a couple of things on the wrong changelist, so I've added those. I'm getting a weird hanging with providers right now that I think might be my environment rather than the code (it goes into a provider and then hangs on the pull), but you should be able to build charms just fine now.

charmcraft/services/provider.py
```diff
@@ -0,0 +1,38 @@
...
+        alias = bases.get_base_alias(base_name)
+        base_class = bases.get_base_from_alias(alias)
+        kwargs.setdefault("compatibility_tag", f"charmcraft-{base_class.compatibility_tag}.0")
```
lengau, line 37: This is the only change, needed for backwards compatibility. (We can make the configure hook remove all the old ones too)

lengau: (note to self: discuss compatibility tags in providers)

tests/commands/regression/test_help.py
lengau, line 1: This is just to ensure backwards compatibility during the migration. Goes away once we strip the legacy CLI

PR #1331, opened 2023-10-17
test: remove black linting from test_infra

lengau:
We're using tox for black linting now.

CRAFT-2143


tigarmo (approved):
this one was an easy review ;)

mr-cal (approved):

PR #1334, opened 2023-10-17
refactor(parts): move craft-parts related things into their own package

lengau:
Organised into a few sensible commits


tigarmo (approved):

sergiusens (approved):
Nice clean up
For symmetry with rockcraft we might want to reconsider the location of plugins themselves. Thoughts @tigarmo

PR #1340, opened 2023-10-20
feat(craft-app): move init command

lengau:
includes lint fixes needed for this update


charmcraft/application/commands/base.py
```diff
@@ -26,6 +29,14 @@ class CharmcraftCommand(craft_application.commands.AppCommand):
     format_option: bool = False
     _services: services.CharmcraftServiceFactory
 
+    def __init__(self, config: dict[str, Any] | None) -> None:
+        super().__init__(config)
+
+        if config is None:
```
tigarmo, line 35: when is config `None`?

charmcraft/application/main.py
```diff
@@ -42,15 +47,24 @@
...
+        services: CharmcraftServiceFactory,
+    ) -> None:
+        super().__init__(app=app, services=services)
+        self._global_args: dict[str, Any] = {}
+
     @property
     def command_groups(self) -> list[craft_cli.CommandGroup]:
         """Return command groups."""
-        lifecycle_commands = get_lifecycle_command_group()
+        lifecycle_commands = commands.get_lifecycle_command_group()
```
tigarmo, line 61: for craft-application 1.1 we could see about maybe adding a `get_lifecycle_command_group()` method that this class can then override

charmcraft/application/main.py
```diff
@@ -135,6 +157,67 @@ def _get_dispatcher(self) -> craft_application.application._Dispatcher:  # type:
 
         return dispatcher
 
+    def run(self) -> int:  # (too many branches due to error handling)
```
tigarmo, line 160: leftover lint disable

charmcraft/application/main.py
```diff
@@ -135,6 +157,67 @@ def _get_dispatcher(self) -> craft_application.application._Dispatcher:  # type:
...
+        """Bootstrap and run the application."""
+        dispatcher = self._get_dispatcher()
+        craft_cli.emit.trace("Preparing application...")
+
+        return_code = 1  # General error
+        try:
+            command = cast(
+                CharmcraftCommand,
+                dispatcher.load_command(
+                    {"app": self.app, "services": self.services, "global_args": self._global_args}
```
tigarmo, line 170: another candidate for customization for 1.1
is this the only change in this run() method?

tests/test_infra.py
```diff
@@ -67,6 +67,7 @@ def test_setup_version():
     assert output.strip() == __version__
 
 
+@pytest.mark.xfail(reason="Should fail while migrating commands")
```
tigarmo, line 70: consider using `strict=True` so we don't forget to revisit this when migration is done

tigarmo (approved):
looking good! Left some minor comments for possible clarifications
thanksss

charmcraft/application/commands/base.py
syu-w: I guess no `charmcraft.yaml`?

syu-w (approved):

charmcraft/utils/project.py
```diff
@@ -24,7 +24,6 @@
 from craft_cli import emit
 from jinja2 import Environment, FileSystemLoader, PackageLoader, StrictUndefined
 
-import charmcraft
```
mr-cal: I almost left a comment from an [earlier commit](https://github.com/canonical/charmcraft/pull/1340/commits/ddd361738b46a28329361a478a496ac262365c0e#diff-dd1802bfa45942a4028f22581575ecce25ece667160712c4ce35bc723a00688bR27) but decided to wait to see if a subsequent commit would use this import

mr-cal (approved):
:+1:

tests/test_infra.py
lengau: Ope, yeah that was an oversight on my part. I've [requested a lint rule](https://github.com/m-burst/flake8-pytest-style/issues/267) for it since it makes sense for us to enforce a value here.

charmcraft/application/commands/base.py
lengau: Not sure TBH, but the craft-cli base accepts a `None` so I presumed it was for getting help or something.

charmcraft/application/main.py
lengau: Something like that, yeah. Another option would be to provide the full set of command groups to `__init__` with a fallback to generating the default groups.

lengau: https://github.com/canonical/craft-application/issues/124

charmcraft/application/main.py
lengau: It is, and definitely.

lengau: https://github.com/canonical/craft-application/issues/125

PR #1341, opened 2023-10-20
Update the README to simplify and better align with docs

tmihoc:
This is part of a bigger initiative to update the READMEs for all the Juju-universe projects. The goal is to have a simpler, more to the point intro that provides better context and a better integration with the docs.


README.md
```diff
@@ -1,155 +1,66 @@
...
-Now, if you have a Kubernetes cluster with the Juju OLM accessible you can issue
-`juju deploy ./my-new-charm.charm --resource httpbin-image=kennethreitz/httpbin`.
-For classic machine deployments you can issue `juju deploy ./my-new-charm.charm`.
-You do not need to publish your operator on Charmhub, you can pass the charmed
-operator file around directly to users, or for CI/CD purposes.
-
-## Charmhub login and charm name reservations
+## Contribute to Charmcraft
+Read our [Code of conduct](https://ubuntu.com/community/code-of-conduct) and start chatting with us:
+- Join our chat: [Mattermost](https://chat.charmhub.io/charmhub/channels/charm-dev)
```
sergiusens, line 56: we have a specific charmcraft channel

sergiusens (commented):
this is a welcome facelift, but doesn't this look more like a general ecosystem documentation more than a specific Charmcraft one (perhaps linking back to the generality), the thing that made it more glaring to me at least was the mention of installing multipass

README.md
tmihoc: I know. I chose the Charm Development channel because it seemed to be the more popular entry point. As in, a user consumes Charmcraft through the SDK, whose main Mattermost channel is the Charm development channel. But I agree -- we either update that general channel to provide links to the more specific channels, including Charmcraft, or we link directly to the Charmcraft channel. What do you think?

tmihoc:
> this is a welcome facelift, but doesn't this look more like a general ecosystem documentation more than a specific Charmcraft one (perhaps linking back to the generality)

The idea here was that Charmcraft only makes sense in the context of the Charm SDK and the Charm SDK makes more sense after you've seen Juju and Charmhub. The table is supposed to quickly give all this context; we repeat it on the README of Juju (https://github.com/juju/juju/pull/16478/files) and of Ops (https://github.com/canonical/operator/pull/1052), varying only the "You are here" pointer. The "Use Charmcraft" section is supposed to act as a teaser, with any further information provided in the Charm SDK docs.

> the thing that made it more glaring to me at least was the mention of installing multipass

I did wonder about whether to go through Multipass. 

One reason I hesitated is because it hides the install Charmcraft step. Another reason is because it makes the success of engaging with Charmcraft dependent on the success of the engagement with Multipass. 

On the other hand, from the point of view of a user, the Multipass `charm-dev` VM route offers significant advantages:
- It's more versatile. It'll work for someone on a Windows machine too. (Unless https://juju.is/docs/sdk/install-charmcraft#heading--windows is out-of-date, Multipass is currently the recommended way to install Charmcraft on Windows.)
- It's faster and more focused on the relevant bits: They can quickly experience not just `charmcraft init` and `charmcraft pack` but also the ultimate point of all of that, namely, `juju deploy`, without all the hassle of setting up Juju and a cloud manually. (Sure, we can tell them how to set up Juju and MicroK8s / LXD on their own, but why bother when the whole point of the `charm-dev` blueprint was to spare charmers the trouble?)
- It feels safer -- if anything goes wrong they can just delete the VM.

Overall, imo, the pros outweigh the cons -- and I've heard the same from some of our engineering onboarders using our tutorials (the ones who decided to skip the Multipass step are the ones who end up running into issues related to matching the correct version of Juju to the correct version of MicroK8s etc.). But, you decide!

README.md
tmihoc: I changed it to the charmcraft channel.

README.md
```diff
@@ -1,155 +1,91 @@
...
-name.
+Read our [user documentation](https://juju.is/docs/sdk/charmcraft), which also includes other guides showing Charmcraft in action
 
-Operators in Charmhub can be renamed as needed, so feel free to register a
-temporary name, such as <username>-<charmname> as a placeholder.
+### Chat with us
 
-## Operator upload and release
+Read our [Code of conduct](https://ubuntu.com/community/code-of-conduct) and:
+- Join our chat: [Mattermost](https://chat.charmhub.io/charmhub/channels/charmcraft)
```
sergiusens, line 80:
```suggestion
- Join our chat: [Matrix](https://matrix.to/#/#charmhub-charmcraft:ubuntu.com)
```

PR #1370, opened 2023-11-03
build(deps): bump craft-providers to 1.19.2

mr-cal:
#### 1.19.2 (2023-11-02)
- Update base compatibility tag from ``base-v2`` to ``base-v3``
  This fixes an issue where LXD instances created with
  ``craft-providers==1.16.0`` may fail to start with
  ``craft-providers>=1.17.0``.

#### 1.19.1 (2023-10-26)
- Require a disk device in the default LXD profile

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -11,7 +11,7 @@ colorama==0.4.6
 coverage==7.3.2
 craft-cli==2.4.0
 craft-parts==1.25.1
-craft-providers==1.19.0
+craft-providers==1.19.2
 craft-store==2.4.0
 cryptography==41.0.4
 Deprecated==1.2.14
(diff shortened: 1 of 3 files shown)


syu-w (approved):

lengau:
@syu-w is there something we need to update in the snap hooks for the base change?

syu-w:
> @syu-w is there something we need to update in the snap hooks for the base change?

It is something not needed but can do if we want. Bump `MINIMAL_BASE_VERISON` in the hook will delete all old images during update, not at runtime.

mr-cal:
Ah, you're talking about bumping this? https://github.com/canonical/charmcraft/blob/a14d6d91711743e3940edd1cc279dac95089416a/snap/hooks/configure#L29

syu-w:
> Ah, you're talking about bumping this?
> 
> https://github.com/canonical/charmcraft/blob/a14d6d91711743e3940edd1cc279dac95089416a/snap/hooks/configure#L29

Yes

cmatsuoka (approved):

lengau (approved):

PR #1378, opened 2023-11-16
build: replace `black` with `ruff format` in charmcraft templates

jnsgruk:
As the title suggests! :)

Driveby to bump the default `ops` version to `2.8`


tigarmo:
hey @jnsgruk is this (moving from black to ruff) a general guidance for our repos? If so, do we have any concerns wrt to Astral's possible monetization story moving forward?

jnsgruk:
To be honest this is mostly because fewer deps = good imo. Astral may have a story, but this feels like core functionality.

lengau (approved):
We are astral of borg

mr-cal (approved):
:+1:

PR #1381, opened 2023-11-18
fix: fix most spread tests on craft-application

lengau:
Still broken after this:

- tests/spread/smoketests/different-dir
- tests/spread/smoketests/full-lifecycle
---
CRAFT-2239


charmcraft/application/main.py
```diff
@@ -63,6 +64,37 @@ def _project_vars(self, yaml_data: dict[str, Any]) -> dict[str, str]:
...
+
+        metadata_path = pathlib.Path(self._work_dir / "metadata.yaml")
+        if metadata_path.exists():
+
+            with metadata_path.open() as file:
+                metadata_yaml = util.safe_yaml_load(file)
+            if not isinstance(metadata_yaml, dict):
+                raise errors.CraftError(
+                    "Invalid file: 'metadata.yaml'",
+                    resolution="Ensure metadata.yaml is valid.",
```
mr-cal, line 78: I could see this confusing a user.  Is it invalid because they added a field they weren't supposed to?  Or is the file not valid yaml? 

```suggestion
                    resolution="Ensure metadata.yaml is a valid yaml file.",
```

mr-cal (approved):

mr-cal:
Looks good, linting and unit test failures look minor.

charmcraft/application/main.py
lengau:
```suggestion
                    resolution="Ensure metadata.yaml is meets the juju metadata.yaml specification.",
```

^ How about this instead? This comes from when metadata.yaml is a valid yaml file, but not valid for metadata.yaml.

mr-cal: Oh, I misunderstood the cause of the error.  Then I see your suggestion and raise you another suggestion:

```suggestion
                    resolution="Ensure metadata.yaml meets the juju metadata.yaml specification.",
                    docs_url="https://juju.is/docs/sdk/metadata-yaml",
```

lengau: Didn't your parents teach you not to place string bets? ;-)

(Seriously though thanks, and good catch!)

cmatsuoka (approved):
Looks good.

PR #1386, opened 2023-11-21
fix: final spread test fixes for craft-application

lengau:


tigarmo (approved):

setup.py
```diff
@@ -26,7 +26,8 @@
     long_description = fh.read()
 
 install_requires = [
-    "craft-application~=1.0",
+    # craft-application~=1.1",
+    "craft-application@git+https://github.com/canonical/craft-application.git@aml/1.1-candidate",
```
syu-w, line 30: We are going to use the git version?

lengau: We need to release craft-application 1.1 and then change that over before merging to main.

syu-w (approved):

PR #1387, opened 2023-11-22
build: use ruff snap

lengau:
Uses the version of ruff from snap in CI and tox.

Testing this in charmcraft as a canary before changing all of our repos over.


syu-w (approved):

mr-cal (approved):
Nice!

Can you update https://github.com/canonical/charmcraft/blob/main/CONTRIBUTING.md to install the ruff snap?

PR #1397, opened 2023-11-29
chore: merge main to craft-application feature branch

lengau:


lengau:
@tigarmo just for the record, there was one manual change, setting the craft-application dependency as 1.1.0. I probably should've done that separately

syu-w (approved):

PR #1400, opened 2023-11-29
fix: always include architectures in charm manifest

lengau:


tigarmo (approved):

spread.yaml
```diff
@@ -6,6 +6,7 @@ environment:
   SNAPD_TESTING_TOOLS: $PROJECT_PATH/tools/external/tools
   PATH: /snap/bin:$PATH:$SNAPD_TESTING_TOOLS
   CRAFT_DEBUG: 1  # Show exception tracebacks.
+  CHARMCRAFT_STORE_CONFIG: staging
```
sergiusens, line 9: you probably want two environment variables for this, one for the api endpoint and another for storage, they can be different and free form. This of future onprem stores

sergiusens: I would go as far as saying that you should use the same environment variables snapcraft uses, but with the charmcraft name

lengau: The store config can be "staging" or a JSON string that provides `api_url`, `storage_url` and `registry_url`. If any are missing from that JSON string it defaults to the prod store.

sergiusens: great then!

sergiusens (approved):

PR #1404, opened 2023-11-30
fix: set store config in separate environment variables.

lengau:


syu-w (approved):

syu-w:
So we no longer use `CHARMCRAFT_STORE_CONFIG`?

lengau:
@syu-w nope - Sergio had a good point that we should match Snapcraft: https://github.com/canonical/charmcraft/pull/1400/commits/fbc5554623b4da2a7a17d79fca984b618c8f00cb#r1411113839

PR #1419, opened 2023-12-04
refactor: remove classic analyze command

lengau:


syu-w (approved):

sergiusens (changes requested):
we should keep analyze (if you want, we can make it an alias to lint), this will need documentation updates

lengau:
@sergiusens this doesn't remove the command, it just removes the now-superfluous non-craft-application version I accidentally left in :-)

sergiusens (approved):

PR #1423, opened 2023-12-05
build: use tox for tests

lengau:


lengau:
@sergiusens please update the charmcraft required tests and merge this. I changed the matrix so it's just per-OS, using tox to handle Python versions.

PR #1428, opened 2023-12-07
Add init template for flask-framework

weiiwang01:
Introduce a new profile called `flask-framework` to the charmcraft init command to create basic projects for 12-factor Flask charms


charmcraft/templates/init-flask-framework/src/charm.py.j2
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env python3
+# Copyright 2023 Canonical Ltd.
```
jnsgruk, line 2: The author/year needs to be templated

charmcraft/templates/init-flask-framework/requirements.txt.j2
```diff
@@ -0,0 +1 @@
+https://github.com/canonical/xiilib/archive/v0.1.0.tar.gz
```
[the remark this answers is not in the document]
weiiwang01: Yes, it was originally going to be `12f`, but names starting with a number caused some problems, so I substituted it with number `XII`.

charmcraft/templates/init-flask-framework/src/charm.py.j2
weiiwang01: Updated, thanks!

charmcraft/templates/init-flask-framework/charmcraft.yaml.j2
```diff
@@ -0,0 +1,31 @@
...
+  - flask-framework
+
+# Uncomment the integrations used by your application
+# requires:
+#   mysql:
+#     interface: mysql_client
+#     limit: 1
+#   postgresql:
+#     interface: postgresql_client
+#     limit: 1
```
jdkandersson, line 31: Should this include all of the integrations we support? E.g., ingress seems to be missing?

charmcraft/templates/init-flask-framework/src/charm.py.j2
```diff
@@ -0,0 +1,30 @@
...
+    def __init__(self, *args: typing.Any) -> None:
+        """Initialize the instance.
+
+        Args:
+            args: passthrough to CharmBase.
+        """
+        super().__init__(*args)
+
+
+if __name__ == "__main__":  # pragma: nocover
```
jdkandersson, line 29: Should we remove the `pragma: nocover`? It might not be relevant for all users

charmcraft/templates/init-flask-framework/requirements.txt.j2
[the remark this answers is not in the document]
jdkandersson: Any great ideas for better names are more than welcome 🙂

charmcraft/templates/init-flask-framework/charmcraft.yaml.j2
weiiwang01: Ingress and other relations are default and static for now, so user can't change them.

jdkandersson: Ah I see

charmcraft/commands/init.py
```diff
@@ -35,6 +35,7 @@
     "simple": "init-simple",
     "kubernetes": "init-kubernetes",
     "machine": "init-machine",
+    "flask-framework": "init-flask-framework",
```
lengau, line 38: Please update `_overview` below to include a description of this template.

tests/spread/commands/init-flask-framework/task.yaml
```diff
@@ -0,0 +1,23 @@
+summary: test charmcraft init with flask-framework profile
+
+execute: |
+  mkdir -p test-init
+  cd test-init
+  charmcraft init --profile flask-framework
+  export CHARMCRAFT_ENABLE_EXPERIMENTAL_EXTENSIONS=true
```
lengau, line 7: This can be set in an `environment` keyword for the task. [Here's an example](https://github.com/canonical/charmcraft/blob/main/tests/spread/smoketests/basic-centos-7/task.yaml#L8-L10)

charmcraft/templates/init-flask-framework/charmcraft.yaml.j2
```diff
@@ -0,0 +1,31 @@
+# This file configures Charmcraft.
+# See https://juju.is/docs/sdk/charmcraft-config for guidance.
+
```
lengau, line 3: While the extension does fill in the `type` field, I'd prefer if we keep it in the template too, since a user might see from the documentation that it's required and be confused by it. This confusion will become more obvious when we make a jsonschema for `charmcraft.yaml`, since that will mark `type` as required (and thus IDEs will start complaining about its nonexistence).

lengau (approved):
Looks good. I shared some non-blocker opinions too

charmcraft/commands/init.py
weiiwang01: I have added an one sentence summary for now, but I will add a link to the document once we have published the document for the 12-factor project.

charmcraft/templates/init-flask-framework/charmcraft.yaml.j2
weiiwang01: Gotcha, added, thanks!

tests/spread/commands/init-flask-framework/task.yaml
weiiwang01: Updated, thanks!

charmcraft/templates/init-flask-framework/requirements.txt.j2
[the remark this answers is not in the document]
tigarmo: how about `twelve-f`?  Looks awkward at first glance but reads the same as `12f`

tigarmo (approved):

charmcraft/templates/init-flask-framework/requirements.txt.j2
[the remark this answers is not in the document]
lengau: I'll throw my hat in the ring with `zwolf`. Advantages:

* Is the German word for twelve
* Isn't yet registered on pypi
* Includes the substring "wolf" 🐺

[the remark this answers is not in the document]
weiiwang01: How about `ffffffffffff`, and it's pronounced as `/ɛfɛfɛfɛfɛfɛfɛfɛfɛfɛfɛfɛf/` 🤣

PR #1453, opened 2023-12-15
ci(spread): install spread

lengau:


lengau:
This isn't necessary, see #1463

PR #1455, opened 2023-12-15
build: replace setup.py with pyproject.toml

lengau:


syu-w (approved):

pyproject.toml
```diff
@@ -1,3 +1,107 @@
...
+    "Programming Language :: Python :: 3.10",
+]
+requires-python = ">=3.10"
+
+[project.scripts]
+
+[project.optional-dependencies]
+dev = [
+    "coverage",
+    "flake8",
```
syu-w, line 45: Should we move this to lint?

lengau: Thanks for bringing this up! We don't use flake8 or pydocstyle for linting charmcraft, but they're currently still used for testing the templates from `charmcraft init`, so this is the right place for them.

Given that the templates are [moving to using ruff](https://github.com/canonical/charmcraft/pull/1378), we should replace those tests with a test that runs ruff on them instead.

I've created an enhancement request for that: https://github.com/canonical/charmcraft/issues/1461

mr-cal (approved):

PR #1459, opened 2024-01-03
feat: basic store service

lengau:
This creates a craft-application style `StoreService` for running store commands.

The service is mostly written to be application agnostic so we could later upstream it if we want, but the default values are the ones that are convenient for Charmcraft.


charmcraft/services/store.py
```diff
@@ -0,0 +1,153 @@
...
+                description=description,
+                ttl=ttl,
+                packages=packages,
+                channels=channels,
+            )
+        except craft_store.errors.CredentialsAlreadyAvailable as exc:
+            raise errors.CraftError(
+                "Cannot login because credentials were found in your system "
+                "(which may be no longer valid, though).",
+                resolution="Please logout first, then login again.",
```
mr-cal, line 110: pedantic: 

```suggestion
                resolution="Please log out first, then log in again.",
```

mr-cal (approved):
LGTM

syu-w (approved):

PR #1462, opened 2024-01-03
feat: multi-architecture resource uploads

lengau:
This implements the client side of uploading architecture-dependent resources to Charmhub.

It follows the spec in ST104: https://docs.google.com/document/d/178mR7suCbC12Ou0W-XnwTs7JKXnGCCh_5Rd2Axh1_d8/edit


charmcraft/application/commands/store.py
```diff
@@ -1897,11 +1912,16 @@ def run(self, parsed_args):
             resource_filepath_is_temp = True
             resource_type = ResourceType.oci_image
 
+            image_arch = image_info.get("Architecture", "all")
+            image_arch = utils.ARCH_TRANSLATIONS.get(image_arch, image_arch)
+            bases = [{"name": "all", "channel": "all", "architectures": [image_arch]}]
```
syu-w, line 1917: I am not sure the `"name": "all", "channel": "all"` referred to. Should it be the charm name and channel?

lengau: The store uses bases that are roughly the same as charmcraft's bases, but we currently only care about the architecture. So setting name and channel to "all" allow it to not worry about, for example, whether it's running on Ubuntu 22.04.

charmcraft/application/commands/store.py
```diff
@@ -1826,16 +1827,30 @@ def fill_parser(self, parser):
...
     def run(self, parsed_args):
         """Run the command."""
         store = Store(env.get_store_config())
 
+        if parsed_args.arch:
+            if parsed_args.image:
+                raise ArgumentParsingError(
+                    "Cannot specify an architecture for an OCI image. OCI images contain architecture metadata that is used."
+                )
+            architectures = parsed_args.arch or ["all"]
```
mr-cal, line 1845: This parameter has to exist because of the `if parsed_args.arch:`, right?

```suggestion
            architectures = parsed_args.arch
```

mr-cal (approved):
LGTM (although I did not try uploading a resource)

syu-w (approved):

charmcraft/application/commands/store.py
lengau: I originally had that and it broke because if `--arch` isn't called it's set to `None`. But setting a default value fixed that. Thanks for making me think about it again!

lengau: Oh wait... I need the default to be `None` so I know it's not set  on the CLI

PR #1464, opened 2024-01-03
ci: split spread tests by system

lengau:


.github/workflows/spread.yaml
```diff
@@ -31,7 +31,9 @@ jobs:
       fail-fast: false
       matrix:
         spread:
```
mr-cal, line 33: Would a reference in the other direction also be useful?
```suggestion
        # from the backends defined in the top-level `spread.yaml`
        spread:
```

mr-cal (approved):

.github/workflows/spread.yaml
lengau: Good idea!

syu-w (approved):

PR #1468, opened 2024-01-05
replaced metadata.yaml with charmcraft.yaml in integration test templates

PietroPasotti:
The integration test jinja templates used for `charmcraft init --<profile>` still contain some references to outdated metadata.yaml. Updated them to point to the new charmcraft.yaml.

### DIFF
--- a/charmcraft/templates/init-kubernetes/tests/integration/test_charm.py.j2
+++ b/charmcraft/templates/init-kubernetes/tests/integration/test_charm.py.j2
@@ -12,7 +12,7 @@ from pytest_operator.plugin import OpsTest
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


lengau:
Hi Pietro,

Thanks for this! Not only did it fix this, it also uncovered that we aren't testing this aspect of `init`, so I added an issue: https://github.com/canonical/charmcraft/issues/1471

I'm not sure why the CLA checker failed. I'm rerunning it now, but we may need to do something further.

Thanks,

Alex

lengau:
Hey @PietroPasotti - I'm not sure why the CLA checker's being mean. The only known fix I have would be for you to re-commit this change your @canonical.com email address, but perhaps you know another way?

lengau:
I've created #1540 to fix the cla-check weirdness

PR #1474, opened 2024-01-10
[EXPERIMENTAL: DO NOT MERGE] feat(store): set-resource-architectures command

lengau:
This is a prototype for implementing `fetch-libs` using the proposed logic in ST101.

It doesn't include tests, etc.

### DIFF
--- a/charmcraft/application/commands/lifecycle.py
+++ b/charmcraft/application/commands/lifecycle.py
@@ -28,6 +28,7 @@ from typing_extensions import override
 
 from charmcraft import utils
 from charmcraft.application.commands.base import CharmcraftCommand
+from charmcraft.utils import charmlibs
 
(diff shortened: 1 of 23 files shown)


charmcraft/application/commands/lifecycle.py
```diff
@@ -148,6 +149,68 @@ def get_managed_cmd(self, parsed_args: argparse.Namespace) -> list[str]:
     @override
     def run(self, parsed_args: argparse.Namespace, step_name: str | None = None) -> None:
         """Run a lifecycle step command."""
+        # TODO: When lifecycle commands are replaced with craft-application ExtensibleCommands
```
lengau, line 152: @cmatsuoka would this be better off as a craft-parts prologue? (I would attach it as a method on the `LifecycleService` in that case.)

PR #1476, opened 2024-01-12
fix(templates): put example config sections on all templates

lengau:


syu-w:
The `config` part is not commented out?

lengau:
@syu-w yeah I decided not to comment it out since the `simple` template already has it without comments

syu-w (approved):

mr-cal (approved):

PR #1477, opened 2024-01-15
fix(templates): use more explicit __init__ param types in charm templates

benhoyt:
This is the style we've been advocating recently, so update the charm templates to this. It's more explicit and type-safe (and we can't change the signature of `CharmBase.__init__` without breaking things anyway, so `*args` is no safer), and allows the use of `framework` directly in the observe calls.


lengau (approved):
Looks good, thanks!

syu-w (approved):

PR #1489, opened 2024-01-25
fix: load metadata from charmcraft.yaml in profile integration tests

tonyandrewmeyer:
Since `charmcraft init` doesn't create a `metadata.yaml` file any more, the integration tests need to load the metadata from the `charmcraft.yaml` file, since they typically run from the source folder rather than within a deployed environment.


lengau:
Hi, and thanks for this!

This is the same as #1468 so I'm going to close this in favour of it

PR #1505, opened 2024-02-02
docs: make badges only use what's actually on main

lengau:
Currently PRs to main are included in those badges, meaning that if tests fail on those PRs it shows failing.


syu-w:
Should it be `?branch=main`? `?event=push` seems also include other branch?

lengau:
@syu-w if you don't specify a branch it uses the default branch (not all branches).

syu-w (approved):

mr-cal (approved):

mr-cal:
Nice, this reminds me that we should upstream badges to starbase: https://github.com/canonical/starbase/issues/180

tigarmo (approved):

PR #1523, opened 2024-02-15
ci: group dev dependencies in renovate

lengau:


syu-w (approved):

mr-cal (commented):
What will this change?  Were each of these dependencies coming in as separate PRs?

lengau:
@mr-cal essentially yes. Because we use `requirements.txt` stuff wasn't getting grouped. E.g.: https://github.com/canonical/charmcraft/pulls?q=is%3Apr+sort%3Aupdated-desc+is%3Aclosed+label%3Adependencies

All of the dependencies that match those expressions will be grouped into one PR rather than being split across many PRs, both making renovate less noisy and making our updates go quicker.

mr-cal (approved):
Sounds good, it's more verbose than I would like but the combined PRs are more important!

Issue #1488, opened 2024-01-23
Regression: 3.x places incorrect metadata.yaml in built charm (2.5.3 works)


ajkavanagh:
### Bug Description

This is a **regression**.

`charmcraft pack` for a reactive charm using 2.x/edge (2.5.3.post59+gb6e26ca) doesn't put the correct `metadata.yaml` into the built charm.  It should be the `metadata.yaml` that the reactive plugin builds, but it's actually just the `metadata.yaml` from the top most reactive layer.

This means that the `metadata.yaml` is wrong and thus the charm won't actually work; e.g. in manila's case, the hacluster relation is missing, etc.

### To Reproduce

```bash
sudo snap refresh charmcraft --channel=2.x/edge
git clone "https://review.opendev.org/openstack/charm-manila"
charmcraft pack --bases-index=0
unzip -p manila_ubuntu-22.04-amd64.charm metadata.yaml > wrong.yaml
```

Save the `metadata.yaml`.

```bash
sudo snap refresh charmcraft --channel=2.x/stable
git clone "https://review.opendev.org/openstack/charm-manila"
charmcraft pack --bases-index=0
unzip -p manila_ubuntu-22.04-amd64.charm metadata.yaml > right.yaml
```

Compare two files.

### Environment

* Ubuntu 22.04 LTS
* LXD

### charmcraft.yaml

```shell
type: charm

parts:
  charm:
    plugin: reactive
    reactive-charm-build-arguments:
      - --binary-wheels-from-source
      - --verbose
      - --verbose
    build-packages:
      - tox
      - git
      - python3-dev
      - libffi-dev
      - libssl-dev
      - rustc
      - cargo
      - pkg-config
    source: src/
    build-snaps:
      - charm/latest/edge
    build-environment:
      - CHARM_INTERFACES_DIR: $CRAFT_PROJECT_DIR/interfaces/
      - CHARM_LAYERS_DIR: $CRAFT_PROJECT_DIR/layers/
bases:
  - build-on:
      - name: ubuntu
        channel: "22.04"
        architectures: [amd64]
    run-on:
      - name: ubuntu
        channel: "22.04"
        architectures: [amd64]
  - build-on:
      - name: ubuntu
        channel: "22.04"
        architectures: [s390x]
    run-on:
      - name: ubuntu
        channel: "22.04"
        architectures: [s390x]
  - build-on:
      - name: ubuntu
        channel: "22.04"
        architectures: [ppc64el]
    run-on:
[...]

PR #1524, opened 2024-02-15
fix(package): don't write a metadata.yaml if it already exists.

lengau:


syu-w (approved):

syu-w:
Reactive charms?

mr-cal (approved):

lengau:
@syu-w yep, for reactive charms. This is work for #1488 but not the complete fix.

PR #1538, opened 2024-02-21
chore: replace most lifecycle commands with upstream ones

### LINKED ISSUE #1532: Use craft-application lifecycle commands
### What needs to get done

Except for `pack` and possibly `clean` (for backwards compatibility), we don't need to define our own lifecycle commands.

The commands we do need to modify for charmcraft should, if possible, use the lifecycle commands' extensible interface.

### Why it needs to get done

These commands are cruft and will begin to diverge from other craft applications.

lengau:
Work towards #1532

After this we still need to extend the pack command from upstream.


sergiusens (approved):
Looks good, please create an issue to fix the --force parameter, from the option alone it is not easy to see what is being forced and it is considered a bad practice

mr-cal (approved):
Nice work!
No unit tests to remove?

PR #1539, opened 2024-02-21
test: remove FS side-effects from package tests

lengau:


syu-w:
What was the side-effects this cause?

lengau:
> What was the side-effects this cause?

It was sticking `.charm` files in the pwd. It still does that, but now that directory is a temporary one.

sergiusens (approved):

PR #1544, opened 2024-02-22
feat: update charmcraft to use craft-application from main

lengau:


syu-w:
craft-application 2.0 already?

charmcraft/parts/charm.py
```diff
@@ -265,11 +265,16 @@ def get_build_packages(self) -> set[str]:
 
     def get_build_environment(self) -> dict[str, str]:
         """Return a dictionary with the environment to use in the build step."""
+        environment = {
+            # Cryptography fails to load OpenSSL legacy provider in some circumstances.
+            # Since we don't need the legacy provider, this works around that bug.
```
mr-cal, line 270: If you have a bug to link to, it would be nice to add here

mr-cal (approved):
LGTM

charmcraft/parts/charm.py
lengau: Sadly not. It's something that only appears once you snap the app (so it's possibly patchelf related). Problem is snapcraft uses a very old version of patchelf. I suppose I could make an issue on snapcraft about it?

lengau:
> craft-application 2.0 already?

Yeah this is for the `main` branch @syu-w

PR #1531, opened 2024-02-21
feat: add remote build command 

lengau:
Requires https://github.com/canonical/craft-application/pull/223

PR #1546, opened 2024-02-26
feat: remote build command

lengau:
This is a rebase of https://github.com/canonical/charmcraft/pull/1531 on the craft-application-on-main PR

Broken bundle build is fixed in https://github.com/canonical/charmcraft/pull/1550

CRAFT-2544

### DIFF
--- a/charmcraft/application/commands/__init__.py
+++ b/charmcraft/application/commands/__init__.py
@@ -68,7 +68,7 @@ from charmcraft.application.commands.version import Version
 def fill_command_groups(app: craft_application.Application) -> None:
     """Fill in all the command groups for Charmcraft."""
     app.add_command_group("Basic", [InitCommand])
-    app.add_command_group("Lifecycle", get_lifecycle_commands())
+    app.add_command_group("Lifecycle", [*get_lifecycle_commands(), RemoteBuild])
     app.add_command_group(
         "Store (account)",
         [
@@ -118,7 +118,6 @@ def fill_command_groups(app: craft_application.Application) -> None:
         [
             Analyse,
             Analyze,
-            RemoteBuild,
             Version,
         ],
     )
--- a/charmcraft/application/commands/__init__.py
+++ b/charmcraft/application/commands/__init__.py
@@ -32,6 +32,7 @@ from charmcraft.application.commands.lifecycle import (
     PrimeCommand,
     StageCommand,
 )
+from charmcraft.application.commands.remote import RemoteBuild
 from charmcraft.application.commands.store import (
     # auth
     LoginCommand,
@@ -117,6 +118,7 @@ def fill_command_groups(app: craft_application.Application) -> None:
         [
             Analyse,
             Analyze,
+            RemoteBuild,
             Version,
         ],
     )
--- a/charmcraft/application/commands/remote.py
+++ b/charmcraft/application/commands/remote.py
@@ -1,0 +1,196 @@
+# Copyright 2024 Canonical Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# For further info, check https://github.com/canonical/charmcraft
+"""Build a charm remotely on Launchpad."""
+import argparse
+import os
+import pathlib
+import time
+from collections.abc import Collection
+from typing import Any, cast
+
+from craft_application.commands import ExtensibleCommand
+from craft_application.launchpad.models import Build, BuildState
+from craft_application.remote.utils import get_build_id
+from craft_cli import emit
+from overrides import override  # pyright: ignore[reportUnknownVariableType]
+
+from charmcraft import models, utils
+
+OVERVIEW = """\
+Command remote-build sends the current project to be built
+remotely.  After the build is complete, packages for each
+architecture are retrieved and will be available in the
+local filesystem.
+
+Interrupted remote builds can be resumed using the --recover
+option, followed by the build number informed when the remote
+build was originally dispatched. The current state of the
+remote build for each architecture can be checked using the
+--status option.
+
+To set a timeout on the remote-build command, use the option
+``--launchpad-timeout=<seconds>``. The timeout is local, so
+the build on launchpad will continue even if the local instance
+is interrupted or times out.
+"""
+
+_CONFIRMATION_PROMPT = (
+    "All data sent to remote builders will be publicly available. "
+    "Are you sure you want to continue?"
+)
+
+
+class RemoteBuild(ExtensibleCommand):
+    """Run analysis on a built charm."""
+
+    name = "remote-build"
+    help_msg = "Build a charm remotely on Launchpad."
+    overview = OVERVIEW
+    always_load_project = True
+
+    @override
+    def _fill_parser(self, parser: argparse.ArgumentParser) -> None:
+        parser.add_argument("--recover", action="store_true", help="recover an interrupted build")
+        parser.add_argument(
+            "--launchpad-accept-public-upload",
+            action="store_true",
+            help="acknowledge that uploaded code will be publicly available.",
+        )
+        parser.add_argument(
+            "--launchpad-timeout",
+            type=int,
+            default=0,
+            metavar="<seconds>",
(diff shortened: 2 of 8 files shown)


lengau:
@sergiusens for this we'll need a `LAUNCHPAD_TOKEN` secret. I believe I've made it replicate snapcraft entirely

charmcraft/application/commands/__init__.py
```diff
@@ -117,6 +118,7 @@ def fill_command_groups(app: craft_application.Application) -> None:
         [
             Analyse,
             Analyze,
+            RemoteBuild,
```
syu-w, line 121: Should this count as Lifecycle?

charmcraft/application/commands/remote.py
```diff
@@ -0,0 +1,196 @@
...
+        :param parsed_args: parsed argument namespace from craft_cli.
+
+        :raises AcceptPublicUploadError: If the user does not agree to upload data.
+        """
+        if os.getenv("SUDO_USER") and os.geteuid() == 0:
+            emit.progress(
+                "Running with 'sudo' may cause permission errors and is discouraged.",
+                permanent=True,
+            )
+            time.sleep(1)
```
syu-w, line 98: Why sleep here

charmcraft/application/commands/remote.py
```diff
@@ -0,0 +1,196 @@
...
+
+        try:
+            returncode = self._monitor_and_complete(build_id, builds)
+        except KeyboardInterrupt:
+            if utils.confirm_with_user("Cancel builds?", default=True):
+                emit.progress("Cancelling builds.")
+                builder.cancel_builds()
+                emit.progress("Cleaning up")
+                builder.cleanup()
+            returncode = 0
```
syu-w, line 140: If user Ctrl+C multiple times, this going to skip. It shouldn't be a problem.

syu-w (approved):

charmcraft/application/commands/remote.py
lengau: Yep, that's intentional. If someone uses Ctrl+C multiple times they **really** want to quit and I don't want to get in their way. (Especially since the time between a keyboard interrupt and getting that confirmation prompt should be unnoticeable.)

mr-cal (commented):
The code looks excellent but I have the following feedback:

1. I would like launchpad credentials to be added to github before this lands, so I can see the spread test pass.
2. Something to do with craft-application is still failing for me :sob: 

```bash
~/dev/charms/simple-charm-from-init> charmcraft remote-build
charmcraft internal error: TypeError("Charmcraft._extra_yaml_transform() missing 2 required keyword-only arguments: 'build_on' and 'build_for'")
Full execution log: '/home/developer/.local/state/charmcraft/log/charmcraft-20240227-125134.189556.log'
```

<details>
  <summary>full logs</summary>

```bash
2024-02-27 12:51:34.197 Starting charmcraft, version 2.5.5.post103+g17a15974
2024-02-27 12:51:34.197 Configuring application...
2024-02-27 12:51:34.197 Preparing application...
2024-02-27 12:51:34.198 Build plan: platform=None, build_for=None
2024-02-27 12:51:34.198 Loading project file '/home/developer/dev/charms/simple-charm-from-init/charmcraft.yaml'
2024-02-27 12:51:34.201 Setting target machine to x86_64
2024-02-27 12:51:34.201 charmcraft internal error: TypeError("Charmcraft._extra_yaml_transform() missing 2 required keyword-only arguments: 'build_on' and 'build_for'")
2024-02-27 12:51:34.202 Traceback (most recent call last):
2024-02-27 12:51:34.202   File "/home/developer/dev/charms/zoneminder-charm/venv/lib/python3.11/site-packages/craft_application/application.py", line 478, in run
2024-02-27 12:51:34.202     self._try_load_project(command, platform, build_for)
2024-02-27 12:51:34.202   File "/home/developer/dev/charms/zoneminder-charm/venv/lib/python3.11/site-packages/craft_application/application.py", line 446, in _try_load_project
2024-02-27 12:51:34.202     self.services.project = self.get_project(
2024-02-27 12:51:34.202                             ^^^^^^^^^^^^^^^^^
2024-02-27 12:51:34.202   File "/home/developer/dev/charms/zoneminder-charm/venv/lib/python3.11/site-packages/craft_application/application.py", line 280, in get_project
2024-02-27 12:51:34.202     yaml_data = self._transform_project_yaml(yaml_data, build_on, build_for)
2024-02-27 12:51:34.202                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2024-02-27 12:51:34.202   File "/home/developer/dev/charms/zoneminder-charm/venv/lib/python3.11/site-packages/craft_application/application.py", line 570, in _transform_project_yaml
2024-02-27 12:51:34.202     return self._extra_yaml_transform(yaml_data)
2024-02-27 12:51:34.202            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2024-02-27 12:51:34.202 TypeError: Charmcraft._extra_yaml_transform() missing 2 required keyword-only arguments: 'build_on' and 'build_for'
2024-02-27 12:51:34.202 Full execution log: '/home/developer/.local/state/charmcraft/log/charmcraft-20240227-125134.189556.log'
```

</details>

charmcraft/application/commands/remote.py
lengau: To give a bit of a pause before moving on. I suppose we could do the same "confirm with user" dance with the Launchpad uploads as well and offer an obnoxiously long CLI parameter so people won't just add it every time they run. @mr-cal what do you think?

charmcraft/application/commands/__init__.py
lengau: It looks like that's where Snapcraft has it, so I've adjusted.

mr-cal:
I re-installed craft-application but perhaps it used a cached copy, let me try again

mr-cal (approved):
The simple case `charmcraft remote-build` works, which is what we discussed was the acceptance criteria of this ticket on 2024-Feb-23, so you have my approval.

The code here is good, but I do have notes:
1. We need credentials for the remote-build spread test
2. riscv doesn't build because charmcraft isn't available on `stable`, which means `charmcraft remote-build` will fail
3. Removing the build-id from `charmcraft remote-build --recover <build-id>` has some unintended consequences and bugs and breaks concurrent builds.
    - For example, I run remote-build, ctrl+c, modify charmcraft.yaml, and can no longer recover the build, because the build-id has changed
    - In another example, I run remote-build, ctrl+c, modify charmcraft.yaml, and cannot run remote-build until the previous build finishes.

syu-w:
> 3. Removing the build-id from `charmcraft remote-build --recover <build-id>` has some unintended consequences and bugs and breaks concurrent builds.
>    
>    * For example, I run remote-build, ctrl+c, modify charmcraft.yaml, and can no longer recover the build, because the build-id has changed
>    * In another example, I run remote-build, ctrl+c, modify charmcraft.yaml, and cannot run remote-build until the previous build finishes.

Seems the build hash is not account for some change?

mr-cal:
> > 3. Removing the build-id from `charmcraft remote-build --recover <build-id>` has some unintended consequences and bugs and breaks concurrent builds.
> >    
> >    * For example, I run remote-build, ctrl+c, modify charmcraft.yaml, and can no longer recover the build, because the build-id has changed
> >    * In another example, I run remote-build, ctrl+c, modify charmcraft.yaml, and cannot run remote-build until the previous build finishes.
> 
> Seems the build hash is not account for some change?

Yeah, I think that is the problem in itself - the build-id is generated from the contents of the files in the project directory to support concurrent builds.  Since `--recover <build-id>` is gone, I think we need to stop doing this and use the path to the project directory as the unique identifier.

lengau:
>     1. We need credentials for the remote-build spread test

I've got a note to ask tomorrow at stand-up, but if for some reason I don't please feel free to interject

>     2. riscv doesn't build because charmcraft isn't available on `stable`, which means `charmcraft remote-build` will fail

Yeah this is an unfortunate thing right now, because the default "all architectures" includes riscv64, so if you want to build your charm not for riscv64 you have to specify architectures in the charm. On the plus side, this is only going into charmcraft 3, which will have a riscv64 build

>     3. Removing the build-id from `charmcraft remote-build --recover <build-id>` has some unintended consequences and bugs and breaks concurrent builds.

Your notes on this have had me generate 3 more work items:

* https://github.com/canonical/craft-application/issues/246
* https://github.com/canonical/craft-application/issues/248
* https://github.com/canonical/craft-application/issues/249

Thank you for your in-depth and thoughtful feedback!

charmcraft/application/commands/remote.py
mr-cal: I imagine the UX is a little nicer with this pause and 1 second is small compared to the overall time for a remote build, so I'm fine with this.

A comment explaining the delay is for UX would be nice.

lengau: Comment added!

PR #1551, opened 2024-02-27
tests (templates): Use pytest fixtures and asserts instead of unittest

Batalex:
Related to https://github.com/canonical/operator/issues/1110.

I kept the changes as minimal as possible.

### DIFF
--- a/charmcraft/templates/init-kubernetes/tests/unit/test_charm.py.j2
+++ b/charmcraft/templates/init-kubernetes/tests/unit/test_charm.py.j2
@@ -6,7 +6,6 @@
 import ops
 import ops.testing
 import pytest
-
 from charm import {{ class_name }}
 
 
@@ -16,6 +15,7 @@ def harness():
     harness.begin()
     yield harness
     harness.cleanup()
+
 
 def test_pebble_ready(harness: ops.testing.Harness[{{ class_name }}]):
     # Simulate the container coming up and emission of pebble-ready event
--- a/charmcraft/templates/init-kubernetes/tests/unit/test_charm.py.j2
+++ b/charmcraft/templates/init-kubernetes/tests/unit/test_charm.py.j2
@@ -3,21 +3,22 @@
 #
 # Learn more about testing at: https://juju.is/docs/sdk/testing
 
-import unittest
-
 import ops
 import ops.testing
+import pytest
+
 from charm import {{ class_name }}
 
 
-class TestCharm(unittest.TestCase):
-    def setUp(self):
-        self.harness = ops.testing.Harness({{ class_name }})
-        self.addCleanup(self.harness.cleanup)
-        self.harness.begin()
+@pytest.fixture
+def harness():
+    harness = ops.testing.Harness({{ class_name }})
+    harness.begin()
+    yield harness
+    harness.cleanup()
 
-    def test_pebble_ready(self):
-        # Simulate the container coming up and emission of pebble-ready event
-        self.harness.container_pebble_ready("some-container")
-        # Ensure we set an ActiveStatus with no message
-        self.assertEqual(self.harness.model.unit.status, ops.ActiveStatus())
+def test_pebble_ready(harness: ops.testing.Harness[{{ class_name }}]):
+    # Simulate the container coming up and emission of pebble-ready event
+    harness.container_pebble_ready("some-container")
+    # Ensure we set an ActiveStatus with no message
+    assert harness.model.unit.status == ops.ActiveStatus()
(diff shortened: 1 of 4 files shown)


Batalex:
The launchpad issue should be good now

charmcraft/templates/init-kubernetes/tests/unit/test_charm.py.j2
```diff
@@ -3,21 +3,22 @@
...
-    def setUp(self):
-        self.harness = ops.testing.Harness({{ class_name }})
-        self.addCleanup(self.harness.cleanup)
-        self.harness.begin()
+@pytest.fixture
+def harness():
+    harness = ops.testing.Harness({{ class_name }})
+    harness.begin()
+    yield harness
+    harness.cleanup()
```
benhoyt, line 18: Have you generated each of these templates and run tox against the created charm? If not, that would be good to do. I'm wondering partly because I think the formatter would want two blank lines between top-level `def`s.

benhoyt (commented):
This looks good to me, thanks! Let's just ensure we've generated each of the templates with `charmcraft init` against this PR, and that the `tox` checks pass with flying colours for each.

I'd like to get @tonyandrewmeyer to review too.

tonyandrewmeyer (approved):
Looks good (but ditto on making sure that the formatting etc is done), thanks!

charmcraft/templates/init-kubernetes/tests/unit/test_charm.py.j2
Batalex: My apologies, I should have known better. All three profiles pass now

Batalex:
I took the liberty of formatting init-simple / charm.py as well so that the full charm pass the linter.

In addition, would you be interested in adding:

```toml
[tool.ruff.isort]
known-first-party = ["charm"]
```
to the `pyproject.toml`'s templates?

This would visually separate user code from `ops`/charmcraft code in the imports.

I do have a launchpad account with this email address, I do not understand why the job fails :thinking:

charmcraft/templates/init-kubernetes/tests/unit/test_charm.py.j2
lengau: Hmm... spread tests should have picked up any tox failures in the `full-lifecycle` test. Thanks for uncovering a testing flaw! https://github.com/canonical/charmcraft/issues/1553

charmcraft/templates/init-machine/tests/unit/test_charm.py.j2
```diff
@@ -3,21 +3,22 @@
...
 
-class TestCharm(unittest.TestCase):
-    def setUp(self):
-        self.harness = ops.testing.Harness({{ class_name }})
-        self.addCleanup(self.harness.cleanup)
+@pytest.fixture
+def harness():
+    harness = ops.testing.Harness({{ class_name }})
+    yield harness
+    harness.cleanup()
```
lengau, line 16: It would be nice to have this fixture be included upstream by making a pytest plugin as part of ops.

lengau (approved):
I'm strongly pro-pytest, so I'm very happy with this. Thanks!

Batalex:
> @Batalex
> 
> > I do have a launchpad account with this email address, I do not understand why the job fails 🤔
> 
> It was my understanding that the CLA checker didn't even check `@canonical.com` email addresses, so I'm not sure why it's getting to this point. I think it's not finding [your Launchpad account](https://launchpad.net/~abatisse) because your email address is private?

Ok, I made the address public, we'll see :crossed_fingers:

charmcraft/templates/init-machine/tests/unit/test_charm.py.j2
tonyandrewmeyer: You could indeed open a PR in pytest-operator and see if they are interested. Most charms currently use pytest-operator for integration tests so it wouldn't be an additional dependency for the project overall, although it probably would be for their unit tests.

I don't really see the value in putting a trivial 4-line fixture in a library personally. You're also likely going to need to provide custom metadata in some tests so will need alternative versions anyway.

benhoyt (approved):

syu-w (approved):

syu-w:
The CLA check still failed for some reasons.

Batalex:
Could someone please re-trigger the action? I finally signed the CLA. At first, I thought it was the Code of conduct on Launchpad...

syu-w:
> Could someone please re-trigger the action? I finally signed the CLA. At first, I thought it was the Code of conduct on Launchpad...

Now it pass.

PR #1555, opened 2024-02-28
refactor(tests/spread): stop use init in most spread tests

lengau:
This looks like a big PR, but it's mostly machine-generated and I'm including instructions to reproduce those changes so you can skip reviewing machine-generated files.

# Recommended review process:

0. Check this out locally
1. In each charm directory added, delete all files and run `charmcraft init --author=Canonical --profile=` with the appropriate profile (`simple` unless the charm name includes `kubernetes` or `machine`)
2. Compare the related `task.yaml` file to ensure that the changes correspond to the differences between what `charmcraft init` created and what exists in the directory. (NOTE: Most of these charms don't include the whole template, so `git status` will show a bunch of new files that you can discard.)


lengau:
Actually this is a bad way to do this.

PR #1554, opened 2024-02-28
feat: implement platforms for building charms

lengau:
This implements the Charmcraft portion of ST105 with the extra constraint that all newly added bases (including Ubuntu 24.04) require `platforms` rather than `bases` (per Gustavo comment on the spec). A future Charmcraft will drop support for `bases` entirely.

This PR has a hard dependency on #1556 for spread tests.

PR #1557, opened 2024-02-29
 feat: enable Ubuntu 24.04 as a development base

lengau:
Requires #1554 

Only the last commit is actually this PR


charmcraft/const.py
```diff
@@ -53,12 +55,26 @@
...
+    "ubuntu@20.04",
+    "ubuntu@22.04",
+    "ubuntu@23.10",
+    "centos@7",
+    "almalinux@9",
+]
+BaseStr = CommonBaseStr | Literal["ubuntu@24.04"]
+BuildBaseStr = CommonBaseStr | Literal["ubuntu@devel"]
+
+DEVEL_BASE_STRINGS = ("ubuntu@24.04",)  # Bases that require a specified build base.
```
sergiusens, line 69: the point was that from today we would not require build-base for 24.04

tigarmo (approved):

sergiusens (changes requested):
Nice work, while I did not review the entirety of this, I did spot that the build-base logic of devel for 24.04 is not a requirement starting today

charmcraft/const.py
sergiusens: new information makes my comment no longer valid

lengau: Ohh right! Fixed :-)

syu-w (approved):

Issue #1527, opened 2024-02-16
Remove `distutils` dependency from `env.py`


lengau:
### What needs to get done

Make `env.py` not need to import distutils

### Why it needs to get done

Needed for Python 3.12 support (for core24)

PR #1564, opened 2024-02-29
fix: remove distutils and use craft-application.util

syu-w:
`distutils` is removed in Python 3.12, use `craft-applicaion.util` replace it.

Fixes: #1527


lengau (approved):

mr-cal (approved):
:moneybag: :moneybag: :moneybag:

PR #1567, opened 2024-02-29
test: enable python 3.12 tests

syu-w:


mr-cal (approved):

tox.ini
```diff
@@ -51,7 +51,7 @@ deps =
     -rrequirements-jammy.txt
 description = Run tests with pytest
 labels =
-    py3.{10,11}: tests
```
lengau, line 54: We might as well remove 3.11 here

lengau (approved):
Could you add:

```
    "Programming Language :: Python :: 3.12",
```

to the relevant section of `pyproject.toml`?

Otherwise looks good, feel free to merge when done :-)

tox.ini
syu-w: I would like to keep it until 24.04 released

PR #1577, opened 2024-03-05
style(lint): enable some previously-ignored ruff rules

lengau:


tests/commands/test_store_registry.py
```diff
@@ -1000,7 +1001,7 @@ def test_imagehandler_extract_file_simple(tmp_path, emitter):
 
     assert size == len(test_content)
     assert digest == "sha256:" + hashlib.sha256(test_content).hexdigest()
-    assert open(tmp_filepath, "rb").read() == test_content
+    assert pathlib.Path(tmp_filepath).read_bytes() == test_content
```
bittner, line 1004: I would suggest to use `from pathlib import Path` above, which makes the code a bit more readable:

```suggestion
    assert Path(tmp_filepath).read_bytes() == test_content
```

All over the code base, obviously.

pyproject.toml
```diff
@@ -327,7 +325,6 @@ lint.ignore = [
     "RET505", "RET506", "RET508",
     "S108", "S113", "S324", "S701",
     "SIM102",
-    "TRY004", "TRY200",
 ]
```
bittner, line 328: You may want to use `select-extend = ["ALL"]` and just focus on maintaining the ignores, e.g.

```toml
[tool.ruff.lint]
extend-select = ["ALL"]
extend-ignore = ["ARG005", "ANN", "D"]

[tool.ruff.lint.per-file-ignores]
"charmcraft/foo.py" = ["INP"]
```

tests/commands/test_store_registry.py
lengau: I disagree here - I'd much rather have `import module ... module.Class()` than have dozens of `Class()`es all over that I don't know whether they're local or from somewhere else. I like this article for what I'm trying to achieve: https://benhoyt.com/writings/python-api-design/

syu-w (approved):

tests/commands/test_store_registry.py
lengau: I guess the question is where one draws the line. 

Personally I only really want to make an exception for stuff in `typing` and `collections.abc` that's there for improving annotations. I specify that because, for example, I don't think `typing` is actually the right place for `NamedTuple` (it should go in `collections` alongside `namedtuple`) and would rather inherit from `typing.NamedTuple` than from `NamedTuple`. However, typing elements like `Self`, `Any` and `Sequence` for use in annotations are fair game to me. (Though maybe that's just because I view them as abstractions of core types while I view `Path` as a constructed type.)

@mr-cal  @bittner I'd like to hear your opinions too on where that line should be drawn. I think any choice has its weaknesses and like to dive into stuff like this to better understand the various perspectives.

PR #1578, opened 2024-03-05
Add the `django-framework` extension and its corresponding initialization profile.

weiiwang01:
Add the `django-framework` extension, which is similar to the `flask-framework` but specifically designed for the Django framework.

The `flask-framework` has been refactored in hope to be more aligned with the extension system in charmcraft.

Due to the fact that we are still waiting for some changes in other libraries to land, one spread test might fail. Please review the current content, and I will update the pull request once the changes in other libraries are merged.


charmcraft/extensions/gunicorn.py
```diff
@@ -0,0 +1,241 @@
...
+                        "please rename or remove it"
+                    )
+
+    def _get_root_snippet(self) -> dict[str, Any]:
+        """Return the root snippet to be merged into the user charmcraft.yaml.
+
+        This method differs from get_root_snippet because it doesn't perform any check.
+        """
+        return {
+            "type": "charm",
```
lengau, line 183: I would rather keep this explicit in all `charmcraft.yaml` files unless we have direction from above otherwise.

charmcraft/extensions/gunicorn.py
```diff
@@ -0,0 +1,241 @@
...
+        return obj
+
+    def check_input(self) -> None:
+        """Check if the extension is applicable for user input charmcraft project file."""
+        for incompatible_field in ("devices", "extra-bindings", "storage"):
+            if incompatible_field in self.yaml_data:
+                raise ExtensionError(
+                    f"the {self.framework}-framework extension is not compatible with the "
+                    f"field {incompatible_field!r}"
+                )
```
lengau, line 146: I'd rather have these come up in a single error message. It's probably unwieldy to have this whole method produce a single exception, but each of these loops can probably raise a single extension that lists all of the relevant fields. Something like:

```suggestion
		incompatible_fields = {"devices", "extra-bindings", "storage"} & self.yaml_data.keys()
		if incompatible_fields:
			raise ExtensionError(
				f"the '{self.framework}-framework' extension is incompatible with the provided "
				f"field(s): {', '.join(sorted(incompatible_fields))}" 
			)
```

charmcraft/extensions/gunicorn.py
```diff
@@ -0,0 +1,241 @@
...
+                f"{self.framework}-app": {"resource": f"{self.framework}-app-image"},
+            },
+            "resources": {
+                f"{self.framework}-app-image": {
+                    "type": "oci-image",
+                    "description": f"{self.framework} application image.",
+                },
+            },
+            "peers": {"secret-storage": {"interface": "secret-storage"}},
+            "actions": _ACTIONS[self.framework],
```
lengau, line 195: What if we made `_ACTIONS` and `_OPTIONS` class-level constants instead? Then we could just use `self._ACTIONS` and have inheritance take care of the specifics for us.

charmcraft/extensions/gunicorn.py
```diff
@@ -0,0 +1,241 @@
...
+
+_ACTIONS = {
+    "flask": {
+        "rotate-secret-key": {
+            "description": "Rotate the flask secret key. Users will be forced to log in again. This might be useful if a security breach occurs."
+        }
+    },
+    "django": {
+        "rotate-secret-key": {
+            "description": "Rotate the flask secret key. Users will be forced to log in again. This might be useful if a security breach occurs."
```
lengau, line 34:
```suggestion
            "description": "Rotate the django secret key. Users will be forced to log in again. This might be useful if a security breach occurs."
```

I presume? (Or make it generic by removing the framework name)

charmcraft/extensions/gunicorn.py
```diff
@@ -0,0 +1,241 @@
+# Copyright 2023 Canonical Ltd.
```
lengau, line 1:
```suggestion
# Copyright 2023-2024 Canonical Ltd.
```

lengau (changes requested):
I like this overall, but I'm worried that a couple of the choices may become unwieldy if we're adding more gunicorn-based extensions. Some suggestions to hopefully prevent this later.

charmcraft/extensions/gunicorn.py
weiiwang01: Gotcha, remove the `type` from the root snippet and added a new check to make sure it's `charm`. Thanks

charmcraft/extensions/gunicorn.py
weiiwang01: Moved the constant variables into the class namespace, thanks!

charmcraft/extensions/gunicorn.py
```diff
@@ -0,0 +1,245 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# For further info, check https://github.com/canonical/charmcraft
+
+"""The flask extension."""
```
mthaddon, line 17: Should this be "Gunicorn extensions." or "The django and flask extensions."?

charmcraft/extensions/gunicorn.py
```diff
@@ -0,0 +1,245 @@
...
+            },
+            "flask-session-cookie-secure": {
+                "type": "boolean",
+                "description": "Set the secure attribute in the Flask application cookies. This configuration will set the FLASK_SESSION_COOKIE_SECURE environment variable. Run `app.config.from_prefixed_env()` in your Flask application in order to receive this configuration.",
+            },
+        },
+        "django": {
+            "django-debug": {
+                "type": "boolean",
+                "description": "Whether Django debug mode is enabled.",
```
mthaddon, line 83: Should we be explicit about the default for this?

charmcraft/extensions/gunicorn.py
weiiwang01: Yes, I have updated it to "Gunicorn-based extensions" as we might add more Gunicorn-based extensions (such as FastAPI, etc.). Is that okay?

charmcraft/extensions/gunicorn.py
weiiwang01: Yes, indeed, updated, thanks!

charmcraft/extensions/gunicorn.py
```diff
@@ -0,0 +1,248 @@
...
+    def framework(self) -> str:
+        """Return the name of the WSGI framework."""
+
+    def _get_nested(self, obj: dict, path: str) -> dict:
+        """Get a nested object using a path (a dot-separated list of keys)."""
+        for key in path.split("."):
+            obj = obj.get(key, {})
+        return obj
+
+    def check_input(self) -> None:
```
lengau, line 141:
```suggestion
    def _check_input(self) -> None:
```

charmcraft/templates/init-django-framework/requirements.txt.j2
```diff
@@ -0,0 +1 @@
+https://github.com/canonical/xiilib/archive/v0.2.0.tar.gz
```
lengau, line 1: This needs updating, right?

tests/integration/commands/test_init.py
```diff
@@ -240,6 +240,9 @@ def test_tox_success(new_path, init_command, profile):
 
     init_command.run(create_namespace(profile=profile))
 
+    if not (new_path / "tox.ini").exists():
+        pytest.skip("init template doesn't contain tox.ini file")
```
lengau, line 244: Since the user will still be able to write code for their charms, I'd rather include a tox.ini file in all the templates. I don't imagine it would be different from any of the others though?

tests/extensions/test_gunicorn.py
```diff
@@ -33,19 +33,20 @@ def input_yaml_fixture(monkeypatch, tmp_path):
     }
 
 
-def test_flask_extension(input_yaml, tmp_path):
-    applied = apply_extensions(tmp_path, input_yaml)
+def test_flask_extension(flask_input_yaml, tmp_path):
+    applied = apply_extensions(tmp_path, flask_input_yaml)
```
lengau, line 37: Please combine this and the `test_django_extension` test below into a single parametrised test. Something like:

```suggestion
@pytest.mark.parametrize(
    ("input_yaml", "expected"),
    [
        (
            {
                "type": "charm",
                "name": "test-flask",
                "summary": "test summary",
                "description": "test description",
                "bases": [{"name": "ubuntu", "channel": "22.04"}],
                "extensions": ["flask-framework"],
            },
            {
                "actions": FlaskFramework._ACTIONS,
                ...  # The rest of the expected output here
            },
        )
    ],
)
def test_apply_extensions_correct(tmp_path, extension_class, input_yaml, expected):
    monkeypatch.setenv("CHARMCRAFT_ENABLE_EXPERIMENTAL_EXTENSIONS", "1")
    
    assert apply_extensions(tmp_path, flask_input_yaml) == expected
```

lengau (changes requested):
Thanks for the updates! What's our documentation plan for these? (Obviously that's not part of these PRs, but I want to ensure @tmihoc doesn't get caught off guard)

charmcraft/extensions/gunicorn.py
```diff
@@ -0,0 +1,248 @@
...
+            "rotate-secret-key": {
+                "description": "Rotate the django secret key. Users will be forced to log in again. This might be useful if a security breach occurs."
+            },
+            "create-superuser": {
+                "description": "Create a new Django superuser account.",
+                "params": {"username": {"type": "string"}, "email": {"type": "string"}},
+                "required": ["username", "email"],
+            },
+        },
+    }
```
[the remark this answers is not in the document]
weiiwang01: Okay, sorry, I'm not sure why, but your awesome suggestions from the other pull request (https://github.com/canonical/charmcraft/pull/1582) were reverted here. I've reintroduced them. Thanks!

charmcraft/extensions/gunicorn.py
```diff
@@ -0,0 +1,248 @@
...
+
+    _OPTIONS = {
```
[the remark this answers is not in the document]
weiiwang01: Same as others, reintroduced, thanks!

charmcraft/extensions/gunicorn.py
```diff
@@ -0,0 +1,248 @@
...
+    @staticmethod
+    @override
+    def is_experimental(base: tuple[str, ...] | None) -> bool:  # noqa: ARG004
+        """Check if the extension is in an experimental state."""
+        return True
+
+    @property
+    @abc.abstractmethod
+    def framework(self) -> str:
+        """Return the name of the WSGI framework."""
```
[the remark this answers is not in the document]
weiiwang01: Same as others, reintroduced, thanks!

tests/extensions/test_gunicorn.py
lengau: You can make the `input_yaml` dict a constant and then have the fixture and this parametrization both return `FLASK_INPUT_YAML.copy()`

charmcraft/templates/init-django-framework/requirements.txt.j2
```diff
@@ -0,0 +1 @@
+paas-app-charmer~=0.2.1
```
lengau, line 1: Sorry to be a nag, but also congrats on the 1.0 release!
```suggestion
paas-app-charmer~=1.0
```

tests/extensions/test_gunicorn.py
weiiwang01: Got it, I've created a helper function called `make_flask_input_yaml` and used it in the fixture. Thanks!

tests/integration/commands/test_init.py
weiiwang01: In this context, `./charm/` is equivalent to `.snap/`. The workflow is that the user/developer developing their web application as usual and writing the `tox.ini` file for it. When they are ready for deployment, they will generate the "charm" directory, which contains the charm-related code, similar to creating the snap directory for an application.

lengau: Makes sense, but isn't there going to need to be separate testing for the code in `charm.py` as well then?

charmcraft/extensions/gunicorn.py
```diff
@@ -0,0 +1,238 @@
...
+        """Return the part snippet to apply to existing parts."""
+        return {}
+
+    @override
+    def get_parts_snippet(self) -> dict[str, Any]:
+        """Return the parts to add to parts."""
+        return {}
+
+
+class FlaskFramework(_GunicornBase):
```
jdkandersson, line 168: I think this will need to override the `is_experimental` and set it to `False` since flask already is no longer experimental 🙂

tests/integration/commands/test_init.py
weiiwang01: Yes, indeed, since the tests for charms are just a part of tests for the outer Flask application project, users/developers need to manually add the integration tests or unit tests for the charm in the Flask application project's test directory

charmcraft/extensions/gunicorn.py
weiiwang01: Okay, updated.

tests/integration/commands/test_init.py
lengau: I think I understand. Once we see some real-world use I might revisit the tox file, but it sounds like it's fine. Thanks for the explanation!

lengau (approved):
Thanks! I'm going to wait for approval from either @jdkandersson or @mthaddon for the PAAS side of things, but on the charmcraft side I'm happy with this.

Thank you for all the work on this! I know it's been rough.

jdkandersson (approved):
I'm ok with this to go into the feature/12f branch. Once merged I'll do a PR to `main` similar to what I did for Rockcraft just to double check what might change

Issue #1579, opened 2024-03-05
Publish all releases on PyPI


bittner:
### What needs to get done

There are releases of the charmcraft Python package on GitHub, which are not on PyPI.

### Why it needs to get done

It's more obvious, more natural to install the Python package from PyPI.

A release on GitHub suggests that there is also a release on the canonical Python package platform. Not keeping the releases in sync creates additional confusion.

tigarmo:
Hi @bittner. Out of curiosity, what is your use-case for consuming Charmcraft as a Python package (instead of the snap)?

bittner:
Most importantly, installing a snap in a `Dockerfile` is not trivial. I maintain [a charmcraft container image](https://gitlab.com/painless-software/docker/charmcraft/) for pipeline automation, and I'm happy it can be installed just as a normal Python package.

Apart from that, I want to work on Juju Charms preferably with a pure Python tool chain. Also, I think it's fair to assume that developers using a popular (Open Source) Python package expect that it can, primarily, be installed from PyPI.

As for another argument, _I_ use Ubuntu, and of course, _I_ can install a snap - at least on my development machine -, but I have other people in the teams that I work with that prefer to use macOS or Windows. How can I convince them that it's great to work with Juju Charms when it requires them to abandon their preferred environment?

tigarmo:
Thank you for the reply, this is very interesting. I definitely hear you on the Docker restriction on snaps, and I didn't know people were able to effectively use Charmcraft as a Python package in non-Linux environments.

bittner:
> I didn't know people were able to effectively use Charmcraft as a Python package in non-Linux environments.

TBH, I've not tried charmcraft on anything but Ubuntu, but I assumed that the OS-independent nature of Python is why charmcraft was implemented with it. Not sure what charmcraft features would require to be limited to Linux, or even Ubuntu. The snap packaging, OTOH, certainly is a limiting factor. A friend of mine who uses Arch Linux preferred to come up with [a Vagrant setup](https://gitlab.com/painless-software/cicd/examples/charmed-django-charm/-/blob/main/Vagrantfile) (with Ubuntu) for charm development.

As by the example in https://github.com/canonical/charmcraft/pull/1580#issuecomment-2030611216, the recent improvements on charmcraft were good to have released on PyPI, so that a few more CI use cases could be made possible. :100:

tigarmo:
> Not sure what charmcraft features would require to be limited to Linux, or even Ubuntu

[...]

PR #1580, opened 2024-03-05
Automate publishing Python package on PyPI

bittner:
Adds two Tox environments that help with publishing the Python package on PyPI.

- `tox -e package` can be used for linting to help detect packaging issues.
- `tox -e package -- upload` allows to publish the package on PyPI.
- `tox -e ensure_version_matches $GIT_TAG` verifies that the package version matches the semantic Git tag on the current commit. This is used by the GHA configuration to prevent a mismatch of tags and the version of the published package.

To make publishing work, you need to add the 2 secrets `PYPI_USERNAME` and `PYPI_PASSWORD` to [Settings ➜ Secrets ➜ Actions](https://github.com/canonical/charmcraft/settings/secrets/actions) with the PyPI access token for the project, as described in the [Python packaging user guide](https://packaging.python.org/en/latest/tutorials/packaging-projects/#uploading-the-distribution-archives).

## Related

- #1579
- [Is `setup.py` deprecated?](https://packaging.python.org/en/latest/discussions/setup-py-deprecated/) – _The use as a command line tool is deprecated_

### DIFF
--- a/.github/workflows/publish-pypi.yaml
+++ b/.github/workflows/publish-pypi.yaml
@@ -1,0 +1,30 @@
+name: Publish
+
+on:
+  push:
+    tags:
+    - '[1-9]+.[0-9]+.[0-9]+'
+
+env:
+  PIP_DISABLE_PIP_VERSION_CHECK: '1'
+  PY_COLORS: '1'
+
+jobs:
+  pypi:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - uses: actions/setup-python@v5
+      with:
+        python-version: '3.12'
+    - name: Install build tools
+      run: pip install tox
+    - name: Verify package version is same as Git tag
+      run: tox run -qe ensure_version_matches -- $GIT_TAG
+      env:
+        GIT_TAG: ${{ github.ref_name }}
+    - name: Build package and upload to PyPI
+      run: tox run -e package -- upload
+      env:
+        TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
+        TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,8 @@
 [project]
 name = "charmcraft"
+dynamic = ["version"]
 description = "The main tool to build, upload, and develop in general the Juju charms."
-dynamic = ["version", "readme"]
+readme = "README.md"
 dependencies = [
     "craft-application~=2.0",
     "craft-cli>=2.3.0",
@@ -82,9 +83,6 @@ requires = [
     "setuptools_scm[toml]>=7.1"
 ]
 build-backend = "setuptools.build_meta"
-
-[tool.setuptools.dynamic]
-readme = {file = "README.md"}
 
 [tool.setuptools_scm]
 write_to = "charmcraft/_version.py"
--- a/tox.ini
+++ b/tox.ini
@@ -115,3 +115,31 @@ env_dir = {work_dir}/pre-commit
 runner = ignore_env_name_mismatch
 description = Run pre-commit on staged files or arbitrary pre-commit commands (tox run -e pre-commit -- [args])
 commands = pre-commit {posargs:run}
+
+[testenv:clean]
+description = Clean up bytecode and build artifacts
+skip_install = true
+deps = pyclean
+commands = pyclean {posargs:. --debris --erase results/* results/ --yes}
+
+[testenv:ensure_version_matches]
+description = Verify package version is same as Git tag
+deps =
+commands = python -c 'import os; from importlib.metadata import version; pkg, tag = os.environ["PKG_NAME"], os.environ["GIT_TAG"]; ver = version(pkg); error = f"`{ver}` != `{tag}`"; abort = f"Package version does not match the Git tag ({error}). ABORTING."; raise SystemExit(0 if ver and tag and ver == tag else abort)'
+setenv =
+    PKG_NAME=charmcraft
+    GIT_TAG={posargs}
+
+[testenv:package]
+description = Build package and check metadata (or upload package)
+skip_install = true
+deps =
+    build
+    twine
+commands =
+    python -m build
+    twine {posargs:check --strict} dist/*
+passenv =
+    TWINE_USERNAME
+    TWINE_PASSWORD
+    TWINE_REPOSITORY_URL
--- a/.gitignore
+++ b/.gitignore
@@ -137,6 +137,7 @@ dmypy.json
 # backups
 *~
 /charmcraft/_version.py
+/results/
 
 # Spread files
 .spread-reuse*.yaml
--- a/HOWTO_RELEASE.md
+++ b/HOWTO_RELEASE.md
@@ -60,8 +60,8 @@ paste, but do please pay attention to details!
 
 - build a tarball to test
 
-    rm -rf dist/
-    ./setup.py sdist bdist_wheel
+    tox -e clean
+    tox -e package
 
 - try the tarball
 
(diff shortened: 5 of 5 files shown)


.github/workflows/publish-pypi.yaml
```diff
@@ -0,0 +1,30 @@
+name: Publish
+
+on:
+  push:
+    tags:
+    - '*'
```
lengau, line 6: Could we make this an `x.y.z` tag format instead? We sometimes use tags for other stuff (e.g. marking a specific snap revision) and I don't want that to trigger the workflow even if it won't publish.

lengau (commented):
Thanks for this!

I *think* snapcraft 8 lets us get rid of the `setup.py` file, but we'll find out when I run the CI

.github/workflows/publish-pypi.yaml
bittner: Good point. I modified the selector following GHA's [filter pattern cheat sheet](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet).

lengau: Thanks!

bittner:
> I _think_ snapcraft 8 lets us get rid of the `setup.py` file, but we'll find out when I run the CI

The `snap-build` job fails, because the charmcraft snap configuration takes for granted that `setup.py` is around:

```
:: ++ python3 setup.py --version
:: python3: can't open file '/root/parts/charmcraft/src/setup.py': [Errno 2] No such file or directory
```

From a Python (packaging) point of view the use of `setup.py` is legacy and every modern Python package should switch over to using (just) `pyproject.toml`. I guess it's okay if I try to fix the snap configuration with an additional commit to make the pipeline pass.

bittner:
The `snap-build` job fails as it can't find the `craft_cli` module.

```python
ModuleNotFoundError: No module named 'craft_cli'
```

I can't see that the `craft-cli` Python package is installed in the job, I can only see it as a dependency in [requirements.txt](https://github.com/canonical/charmcraft/blob/main/requirements.txt#L7).

Can anyone see what is the problem?

bittner:
> Can anyone see what is the problem?

Okay, looks like there is a problem with the `sed` command before that:

```
sed: -e expression #1, char 14: unterminated `s' command
```

The problem is that that with the new setup the returned version string also contains the package name. I'll try to find a solution. Suggestions welcome!

bittner:
The CI jobs all pass now (on my personal branch). I rebased the branch, and I made separate commits to document the specific changes that were necessary to work around a few build issues. :scroll: 

Hope that looks good enough now for the PR to be merged. :crossed_fingers:

bittner:
I have signed the [CLA](https://ubuntu.com/legal/contributors) now. I use a different email address on Launchpad than on GitHub, that's why the check probably didn't find my Launchpad account.

bittner:
The cla-check still fails. I checked my Launchpad account again; looks like there was another account with the email address I use on GitHub. I merged the accounts on Launchpad now.

bittner:
I received an email notification this morning saying, I was added as a member of [Canonical Contributor Agreement](https://launchpad.net/~contributor-agreement-canonical).

I assume, the cla-check job would now succeed.

pyproject.toml
```diff
@@ -1,8 +1,10 @@
 [project]
 name = "charmcraft"
+dynamic = ["version"]
 description = "The main tool to build, upload, and develop in general the Juju charms."
-dynamic = ["version", "readme"]
+readme = "README.md"
 dependencies = [
+    "attrs<23",
```
lengau, line 7: What drove this need? Please include a comment so we know when we can undo it

snap/snapcraft.yaml
```diff
@@ -138,7 +138,7 @@ parts:
       # Ensure we don't have a dubious ownership error from git with a remote build.
       git config --global --add safe.directory $CRAFT_PART_SRC
       # set the version
-      version="$(python3 setup.py --version)"
```
lengau, line 141: This change seems to be always making the charmcraft snap that gets built report the version `dev`, I'm guessing because it's not recognised as installed at this point, so there's not a `_version.py` file in the charmcraft directory from which to report.

lengau (changes requested):

pyproject.toml
bittner: The commit message says it all.

> Stick to older attr version to avoid hatchling UnknownPluginError (fancy-pypi-readme)

snap/snapcraft.yaml
bittner: Doesn't setuptools-scm take care of generating the `_version.py` file, transparently? Have you seen any issues in the build process? :astonished:

bittner:
@lengau Let me know what needs to be changed, if anything, so that I can help getting this merged. Thank you! :+1:

bittner:
Is there a reason why this PR isn't considered for merging? Any arguments against the changes? Anything that needs to be updated?

It would be very helpful if the recent release(s) of charmcraft were published also on PyPI. For example, I'm currently blocked by the [<q>No keyring found</q> error](https://discourse.charmhub.io/t/charmcraft-and-the-no-keyring-found-error/6144) that was [fixed in version 2.3.0](https://github.com/canonical/charmcraft/releases/tag/2.3.0) by allowing anonymous downloads of charm library modules. As explained in https://github.com/canonical/charmcraft/issues/1579#issuecomment-2001979224 a pure-Python installation is needed for the CI use case.

snap/snapcraft.yaml
lengau: The [snap artefact from the build](https://github.com/canonical/charmcraft/actions/runs/8500696494?pr=1580) is reporting that, and I've been unable to come up with something else that makes it work as intended. Could you revert the removal of `setup.py` for now, since having it there shouldn't break things?

requirements.txt
```diff
@@ -1,4 +1,4 @@
-attrs==23.2.0
+attrs==22.2.0
```
syu-w, line 1: Version downgrade?

lengau (commented):
Please re-add setup.py, and then I'm happy with this :-)

I've created https://github.com/canonical/charmcraft/issues/1635 for seeing if we can remove setup.py at a later date so we don't block this PR with it

snap/snapcraft.yaml
bittner: `setup.py` restored. The built `snap.zip` bears a version number again.

bittner:
***ping*** :bell: 

@lengau Alex, I think this PR is now ready for merging. Can you (or some other maintainer) take a look?

lengau (approved):

tox.ini
```diff
@@ -115,3 +115,31 @@ env_dir = {work_dir}/pre-commit
...
+[testenv:clean]
+description = Clean up bytecode and build artifacts
+skip_install = true
+deps = pyclean
+commands = pyclean {posargs:. --debris --erase results/* results/ --yes}
+
+[testenv:ensure_version_matches]
+description = Verify package version is same as Git tag
+deps =
+commands = python -c 'import os; from importlib.metadata import version; pkg, tag = os.environ["PKG_NAME"], os.environ["GIT_TAG"]; ver = version(pkg); error = f"`{ver}` != `{tag}`"; abort = f"Package version does not match the Git tag ({error}). ABORTING."; raise SystemExit(0 if ver and tag and ver == tag else abort)'
```
mattculler, line 128: I think it would be cleaner to do something like:
```
commands = python -c 'import os;' \
  'from importlib.metadata import version;' \
  'pkg, tag = os.environ["PKG_NAME"], os.environ["GIT_TAG"];' \
  ...
```
But if this doesn't work for some reason, or there's disagreement on this point, then I'm ok with it as-is.

mattculler (approved):
👍

tox.ini
lengau: This doesn't work in tox for some weird reasons :-)

bittner:
Thanks for merging! 💯 

Now don't forget to add the secrets to the project settings to make automatic publishing succeed! 🤖

PR #1581, opened 2024-03-05
feat(linter): add additional-files linter

syu-w:
Add extra files should use the dump plugin, not use the prime keywords.

This will errors any extra files added without using the proper ways.

### DIFF
--- a/charmcraft/linters.py
+++ b/charmcraft/linters.py
@@ -518,20 +518,20 @@ class Entrypoint(Linter):
 class AdditionalFiles(Linter):
     """Check that the charm does not contain any additional files in the prime directory.
 
-    A few generated files are ignored.
+    A few generated files and basic charm files are ignored.
     """
 
     name = "additional-files"
     text = "No additional files found in the charm."
     url = "https://juju.is/docs/sdk/include-extra-files-in-a-charm"
 
-    INGNORE_FILES: set[pathlib.Path] = {
-        pathlib.Path(const.BUNDLE_FILENAME),
-        pathlib.Path(const.CHARMCRAFT_FILENAME),
-        pathlib.Path(const.MANIFEST_FILENAME),
-        pathlib.Path(const.METADATA_FILENAME),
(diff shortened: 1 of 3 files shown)


syu-w:
This one is currently missing docs URL. Will be updated as soon as the docs become public.

charmcraft/linters.py
```diff
@@ -515,6 +515,57 @@ def run(self, basedir: pathlib.Path) -> str:
...
+    """Check that the charm does not contain any additional files in the prime directory.
+
+    A few generated files are ignored.
+    """
+
+    name = "additional-files"
+    text = "No additional files found in the charm."
+    url = "https://juju.is/docs/sdk/include-extra-files-in-a-charm"
+
+    INGNORE_FILES: set[pathlib.Path] = {
```
mr-cal, line 528:
```suggestion
    IGNORE_FILES: set[pathlib.Path] = {
```

mr-cal, line 528: Should `const.CHARM_OPTIONAL_FILES` be ignored?

syu-w: I think these files are automatically staged by the charm part? So they always exist in the `stage`. But we can ignore them just in case.

syu-w:
Updated

mr-cal (approved):
Thanks!

charmcraft/linters.py
lengau: It depends. If we change the plugins or something I don't want this being overlooked.

lengau (approved):
Looks good, thanks!

Before merging, can you pick a couple of real-world charms (e.g. some from our weekly spread test) and verify that this linter doesn't go off for them?

syu-w:
I do not see any linter errors but many other errors in the weekly tests: https://github.com/canonical/charmcraft/actions/runs/8193669673

lengau:
@syu-w yeah I'm working on triaging and categorising those :-)

PR #1582, opened 2024-03-06
Add the `flask-framework` extension

weiiwang01:
Bring the `flask-framework` extension from the `feature/12f` branch into the main branch.

The `flask-extension` is a part of the 12-factor toolchain that assists users in creating a charm to manage their Flask application within the Juju ecosystem.

As we are still awaiting some changes in other libraries to be finalized, please review the current content. I will update the pull request once the changes in the other libraries have been merged.


weiiwang01:
@lengau, here's another pull request that cherry-picks only the `flask-framework` into the main branch, as per @sergiusens's recommendation.

charmcraft/application/main.py
```diff
@@ -99,7 +100,7 @@ def _extra_yaml_transform(
             for field in const.METADATA_YAML_MIGRATE_FIELDS:
                 yaml_data.setdefault(field, metadata_yaml.get(field))
 
-        return yaml_data
+        return apply_extensions(self._work_dir, yaml_data)
```
lengau, line 103:
```suggestion
        return yaml_data
```

Merge conflict here :-)

The extensions get applied on line 76 so this is no longer needed.

charmcraft/extensions/__init__.py
```diff
@@ -39,3 +40,5 @@
     "register",
     "unregister",
 ]
+
+register("flask-framework", FlaskFramework)
```
lengau, line 44: Please do two things here:

1. Export the `FlaskFramework` extension so it's available as `charmcraft.extensions.FlaskFramework`
2. Move this next to the other `register`.

charmcraft/extensions/gunicorn.py
```diff
@@ -0,0 +1,211 @@
...
+class _GunicornBase(Extension):
+    """A base class for 12-factor WSGI applications."""
+
+    _ACTIONS = {
+        "flask": {
+            "rotate-secret-key": {
+                "description": "Rotate the flask secret key. Users will be forced to log in again. This might be useful if a security breach occurs."
+            }
+        },
+    }
```
lengau, line 36: What I'm looking for here is something like:

```python
class _GunicornBase(Extension):
    """A base class for 12-factor WSGI applications."""

    _ACTIONS: dict[str, dict[str, str]]

...

class FlaskFramework(_GunicornBase):
    """Extension for 12-factor Flask applications."""

    _ACTIONS = {
        "rotate-secret-key": {
            "description": "Rotate the flask secret key. Users will be forced to log in again. This might be useful if a security breach occurs."
        }
    }
```

That way all the flask-specific stuff will be together and the django-specific stuff will be together.

Since `_WEBSERVER_OPTIONS` is generic across all gunicorn extensions we can leave it in the gunicorn abstract base.

charmcraft/metafiles/actions.py
```diff
@@ -101,6 +101,7 @@ def create_actions_yaml(
             shutil.copyfile(original_file_path, target_file_path)
     else:
         if charmcraft_config.actions:
+            basedir.mkdir(exist_ok=True)
```
lengau, line 104: I gather you added this because something was failing - can you let me know what was failing, as that's probably a sign of a bigger bug.

tests/spread/commands/init-flask-framework/task.yaml
lengau, line 1: How different is this going to be from the django test? When we add the Django extension will this be able to be turned into two variants?

(No changes requested, just wondering.)

charmcraft/extensions/gunicorn.py
```diff
@@ -0,0 +1,211 @@
...
+            overlap = user_provided.keys() & self._get_nested(root_snippet, merging).keys()
+            if overlap:
+                raise ExtensionError(
+                    f"overlapping keys {overlap} in {merging} of charmcraft.yaml "
+                    f"which conflict with the {self.framework}-framework extension, "
+                    "please rename or remove it"
+                )
+        for config in self._get_nested(self.yaml_data, "config.options"):
+            for reserved_config_prefix in ("webserver-", f"{self.framework}-"):
+                if config[: len(reserved_config_prefix)] == reserved_config_prefix:
```
lengau, line 151:
```suggestion
                if config.startswith(reserved_config_prefix):
```

charmcraft/application/commands/init.py
```diff
@@ -58,6 +59,9 @@
         A basic charm but meant to be deployed in machine-based environments,
         without container requirements.
 
+    flask-framework:
+        A basic Flask application charm for the 12-factor charm project.
```
lengau, line 63:
```suggestion
        A basic Flask application charm using xiilib.
```

lengau (changes requested):
Thanks for this!

Looking back at the other PR, I wasn't clear in what I was asking for with reorganising the constants for the gunicorn extensions. Sorry about that!

I've made what should be a clearer example of the change I'm requesting.

charmcraft/extensions/gunicorn.py
weiiwang01: that's super clever and elegant and requires less indentation! updated, thank you!

tests/spread/commands/init-flask-framework/task.yaml
weiiwang01: I think they will be very similar, with only the `--profile` being different

charmcraft/metafiles/actions.py
weiiwang01: Sorry, this is too distant in the past, and I can't remember the context behind this change, and workflow logs have all been lost. Judging by this line, it seems to be caused by the prime directory not existing when creating the `action.yaml`?

Let me remove this line and run the test to see if that can trigger an error.

lengau (approved):
Looks good, thanks! 

It looks like you have a different version of `black` from what we're using in CI. Running `tox -f format` with a recent enough version of tox (3.8+, so you should be ok to install tox from apt all the way back to jammy) should be consistent.

syu-w (approved):

charmcraft/application/commands/init.py
weiiwang01: Regarding this change, I'm not sure if we should directly inform the user that we're using the `xiilib`, as this is somewhat of an internal implementation detail. I will ask the documentation team about it later, thanks!

charmcraft/extensions/gunicorn.py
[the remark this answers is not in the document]
weiiwang01: So, there will be another extension for Django that also uses `gunicorn`, see here: https://github.com/canonical/charmcraft/pull/1578, thanks

lengau:
@weiiwang01 I'm not sure why this PR is causing the tests to timeout on macos, but it seems like it's waiting for user input, I think when running the integration tests for the `init` command. Could you check the permissions on the files you added, especially the templates?

weiiwang01:
> @weiiwang01 I'm not sure why this PR is causing the tests to timeout on macos, but it seems like it's waiting for user input, I think when running the integration tests for the `init` command. Could you check the permissions on the files you added, especially the templates?

I check the init templates, and the file permissions should be consistent with the other templates. Let me temporarily change the test command to see if we can identify which test is blocking.

weiiwang01:
> > @weiiwang01 I'm not sure why this PR is causing the tests to timeout on macos, but it seems like it's waiting for user input, I think when running the integration tests for the `init` command. Could you check the permissions on the files you added, especially the templates?
> 
> I check the init templates, and the file permissions should be consistent with the other templates. Let me temporarily change the test command to see if we can identify which test is blocking.

Okay, I understand the problem: the `flask-framework` init profile does not generate a `tox.ini` file (in that use case, the tox configuration should be located in the project root, not in the `charm/` directory). So, the `test_tox_success` test ends up executing the tox environment in `charmcraft`, which causes an infinite loop. I have added a directive to skip the `test_tox_success` test if a `tox.ini` file is not generated.

PR #1589, opened 2024-03-06
feat: hide extensions with names prefixed with "_"

### LINKED ISSUE #1586: Hide extensions with names prefixed with "_"
### What needs to get done

If an extension's name is prefixed with an underscore, it should be hidden from the user.

Likewise, when initially loading a charm (but before adding these extensions to the yaml data), ensure that the user doesn't have any underscore-prefixed extensions listed.

(Add an underscore to the bundle extension while you're at it)

### Why it needs to get done

These extensions are not meant to be explicitly added, but are rather used to support YAML transforms we want to do under certain circumstances.

lengau:
fixes #1586


syu-w (approved):

charmcraft/application/main.py
```diff
@@ -68,9 +68,19 @@ def _extra_yaml_transform(
     ) -> dict[str, Any]:
         yaml_data = yaml_data.copy()
 
+        ext_names = yaml_data.setdefault("extensions", [])
+        invalid_extensions = [ext for ext in ext_names if ext.startswith("_")]
```
tigarmo, line 72: is this PR still needed? I think we discussed doing the transformations in extra_yaml_transforms

PR #1597, opened 2024-03-12
fix: adjust max summary length

lengau:


mr-cal (approved):
Any value in adding a comment on why this deviates from the other applications?

I remember you discussing this with Sergio during planning and it could be useful to document that somewhere, in case someone wants to change this in the future.

lengau:
Good idea @mr-cal! I've updated the docstring to imply that the user should make it significantly shorter than the 78 character length, added a comment that explains this change, and created (and linked to) #1598 so we can nudge people towards shorter summary fields.

Issue #1332, opened 2023-10-17
(CI) Parallel build test is flaky


lengau:
### Bug Description

The parallel build smoke test doesn't always succeed.

### To Reproduce

Run the smoke tests in Google	

### Environment

Google environment (cannot replicate on multipass on my machine

### charmcraft.yaml

```shell
N/A
```

### Relevant log output

```shell
2023-10-17 01:04:32 Error executing google:ubuntu-20.04-64:tests/spread/smoketests/parallel-build (oct170052-258541) : 
-----
+ sleep 60
+ cd charm2
+ charmcraft pack --verbose
+ cd charm1
+ charmcraft pack --verbose
Starting charmcraft version 2.4.1.post77+git81e1ff9
Logging execution to '/root/.local/state/charmcraft/log/charmcraft-20231017-005949.512686.log'
Packing the charm.
Starting charmcraft version 2.4.1.post77+git81e1ff9
Logging execution to '/root/.local/state/charmcraft/log/charmcraft-20231017-005949.511956.log'
Packing the charm.
Launching environment to pack for base name='ubuntu' channel='22.04' architectures=['amd64'] (may take a while the first time but it's reusable)
Launching environment to pack for base name='ubuntu' channel='22.04' architectures=['amd64'] (may take a while the first time but it's reusable)
Creating new instance from remote
Creating new base instance from remote
Creating new instance from remote
Creating new base instance from remote
Failed to launch instance 'base-instance-charmcraft-buildd-base-v20-2046f14fb4a00aa9ce50'.
Full execution log: '/root/.local/state/charmcraft/log/charmcraft-20231017-005949.511956.log'
+ wait
+ cd charm3
+ charmcraft pack --verbose
Starting charmcraft version 2.4.1.post77+git81e1ff9
Logging execution to '/root/.local/state/charmcraft/log/charmcraft-20231017-010051.612722.log'
Packing the charm.
Launching environment to pack for base name='ubuntu' channel='22.04' architectures=['amd64'] (may take a while the first time but it's reusable)
Creating new instance from base instance
Creating instance from base instance
Starting instance
Starting instance
Packing the charm
Packing the charm
Starting charmcraft version 2.4.1.post77+git81e1ff9
Logging execution to '/tmp/charmcraft.log'
Packing the charm.
Building charm in '/root'
Installing build-packages
Starting charmcraft version 2.4.1.post77+git81e1ff9
Logging execution to '/tmp/charmcraft.log'
Packing the charm.
[...]

PR #1600, opened 2024-03-12
test(spread): move flaky parallel-build test to weeklies

lengau:
Workaround for #1332


syu-w (approved):

.github/workflows/spread-large.yaml
```diff
@@ -6,6 +6,11 @@ on:
 
   workflow_dispatch:
 
+# Only ever run one of this test at a time.
+concurrency:
+  group: ${{ github.workflow }}
+  cancel-in-progress: true
```
syu-w, line 12: Not sure if we want the `cancel-in-progress: true`

tigarmo (approved):

PR #1614, opened 2024-03-22
style(typing): enable most disabled pyright checks

lengau:
It might be easier to review this commit by commit, which goes rule by rule


charmcraft/parts/reactive.py
```diff
@@ -107,7 +108,7 @@ class ReactivePlugin(plugins.Plugin):
     properties_class = ReactivePluginProperties
     validator_class = ReactivePluginEnvironmentValidator
 
-    @override
+    @overrides.override
```
sergiusens, line 111: why did you use typing override and then switch to this?

pyproject.toml
```diff
@@ -150,6 +150,7 @@ ignore = [
 include = [
     "charmcraft",
     "tests/unit",
+    "tests/integration",
     # "tests",  # TODO: enable tests
```
sergiusens, line 154: is this TODO still valid, can we get rid of it now?

lengau: There are legacy tests outside of those structures, but they can probably be left. TODO removed.

charmcraft/parts/reactive.py
lengau: Working towards consistency with snapcraft/rockcraft, and this provides more/better functionality than the implementation in `typing`

Issue #1616, opened 2024-03-25
Default `charm` or `bundle` part is not added if `parts` is defined but empty in `charmcraft.yaml`.


weiiwang01:
### Bug Description

Currently, the default charm part is added to the charmcraft project configuration using the Pydantic model validator `charmcraft.models.project.CharmcraftProject.preprocess_parts`. However, because Pydantic validators are executed in the order they are defined, and the `parts` field is defined in the parent class `craft_application.models.project.Project` while the `type` field is defined in the child class, the validator for `parts` field will always execute before validator for `type` field. 

This means the `type` attributes will never exist in the `values` parameter of `preprocess_parts` validator function. This causes the `parts` not to be added if the `parts` in the charmcraft project file are defined as `{}`. However, when the `parts` field is not defined in `charmcraft.yaml`, its value will be `None`, and the default value of `charmcraft.models.project.BasesCharm.parts` will take effect, setting the parts to `{"charm": {"plugin": "charm", "source": "."}}`.

### To Reproduce

To reproduce this issue, create a `charmcraft.yaml` with the following parts. It will not execute the `charm` part when packed:

```
parts: {}
```

### Environment

Ubuntu 22.04 LTS

### charmcraft.yaml

```shell
name: charm
type: charm
bases:
  - build-on:
    - name: ubuntu
      channel: "22.04"
    run-on:
    - name: ubuntu
      channel: "22.04"
summary: A test charm.
description: A test charm.

# important
parts: {}
```

### Relevant log output

```shell
n/a
```

sergiusens:
ah, this was meant to be the original design. @lengau we need to do some design work here to see if we finally make charm and bundle actual extensions

weiiwang01:
> ah, this was meant to be the original design. @lengau we need to do some design work here to see if we finally make charm and bundle actual extensions

Gotcha, thanks, the issue now is that when extensions are applied, the parts is set to `{}` even if the extension does not modify the parts. So, maybe we can find a workaround to set the parts to `None` if they are empty after the extension has been applied?

PR #1620, opened 2024-03-26
fix: add default charm part in flask-framework

weiiwang01:
As charmcraft will not automatically add the default `charm` part if extensions are in use, the `flask-framework` extension should include the charm part. Furthermore, as the `flask-framework` extension is designed exclusively for `ops` based charms, it should incorporate a check to ensure that there is no customized charm part in `charmcraft.yaml` that could potentially conflict with the extension.

Some miscellaneous changes include the use of new library packages in the template `requirements.txt`.

Fix: https://github.com/canonical/charmcraft/issues/1616

### DIFF
--- a/charmcraft/extensions/gunicorn.py
+++ b/charmcraft/extensions/gunicorn.py
@@ -75,6 +75,12 @@ class _GunicornBase(Extension):
                 f"the '{self.framework}-framework' extension is incompatible with "
                 f"type {charm_type!r}"
             )
+        parts = self.yaml_data.get("parts")
+        if parts and "charm" in parts:
+            raise ExtensionError(
+                f"the '{self.framework}-framework' extension is incompatible with "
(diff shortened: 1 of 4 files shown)


lengau (approved):
Looks good, thanks!

syu-w (approved):

lengau:
@weiiwang01 before we merge this, could you update the `paas-app-charmer` requirement (and the package) so the sdist works?

weiiwang01:
> @weiiwang01 before we merge this, could you update the `paas-app-charmer` requirement (and the package) so the sdist works?
> 
> I messed around with it a little bit and ended up with [this change](https://github.com/canonical/paas-app-charmer/pull/2) that makes it work.

Super thanks for the fix, I have yanked the old version and updated the requirements in the template.

PR #1623, opened 2024-03-28
feat(config): support `type:secret` in config.options

tonyandrewmeyer:
The `config.options` section of `charmcraft.yaml` supports, from Juju 3.1, a type `secret` (this is currently missing from the documentation, but the Juju team are addressing that now).

This adds support for that, so that Charmcraft 2.5 can be used to pack charms that make use of this functionality.


tonyandrewmeyer:
cf. canonical/operator#1166 where this was raised in ops.

jameinel:
I can't say that I know the charmcraft bits well enough to concretely give an approval, but thank you for putting this together.

syu-w (approved):

lengau (approved):
Thanks for doing this! The code looks good, but I'm thinking about how (and whether) we want to do this in 2.x considering that 3.0 is around the corner and will include this from your other PR.

I'll sleep on this and talk with some folks, and if you, @jameinel or anyone else would like to weigh in on it I'm open to hearing from you. (User stories that include staying on the 2.x branch after 3.0 will be most convincing - in-progress release notes for 3.0 are here: https://docs.google.com/document/d/1CPo8-dwlBU3Nd1dXUaZ1Rdfo3jPXvrkZs-ulg-DS_e0/edit)

lengau:
We're doing a 2.6 release to fix #1640 , so I'm going to retarget this and include it in that release.

Thanks again for your patience folks!

PR #1625, opened 2024-04-02
feat(config): support `type:secret` in config.options

tonyandrewmeyer:
Since Juju 3.1, the `config.options` section of `charmcraft.yaml` (or the older `config.yaml`) has supported a `type` value of `secret` (for user secrets). This is missing from the documentation right now, but the Juju team are working on fixing that.

This PR adds support for `type: secret` in config options.


charmcraft/models/config.py
```diff
@@ -58,8 +58,19 @@ class JujuBooleanOption(_BaseJujuOption, frozen=True):
...
 
 
+class JujuSecretOption(_BaseJujuOption, frozen=True):
+    """A Juju option field containing a secret ID."""
+
+    type: Literal["secret"]
+    # A secret doesn't really make sense, since it's unlikely
+    # that anyone would know what the secret ID (specific to
+    # the deployment in a model) is at the time that they are
+    # writing the config, but included for completeness.
```
lengau, line 68: Thanks for this comment! I presume Juju does support setting the default like this? (If not you can just set it to `None`)

lengau (approved):

tests/test_models.py
```diff
@@ -1135,6 +1135,9 @@ def test_load_config_in_charmcraft_yaml(tmp_path, prepare_charmcraft_yaml):
                 test-bool:
                   default: true
                   type: boolean
+                test-secret:
+                  default: secret:co1s9mnmp25c762drvtg
```
syu-w, line 1139: So the secret string begins with `secret:`?

syu-w (approved):

tests/test_models.py
tonyandrewmeyer: Yes, e.g.:

```
$ juju add-secret example foo=bar
secret:co6tjmvmp25c77dhf980
```

```
$ juju config secretconfig secretopt
secret:co6tjmvmp25c77dhf980
```

```
$ juju config secretconfig secretopt=co6tjmvmp25c77dhf980
{}
ERROR updating charm config settings: option "secretopt" expected secret, got "co6tjmvmp25c77dhf980"
```

charmcraft/models/config.py
tonyandrewmeyer: It does. Maybe there's some use-case that I'm not seeing but I imagine Juju supports it just because it's handling default values consistently across types, rather than because it's of any use here. I guess you could use it in dev maybe?

lengau: Awesome, thanks!

tonyandrewmeyer:
I (foolishly!) made the same mistake here as on the earlier PR and did the GitHub UI "update" without picking the address to associate that with. I've now signed the CLA with my personal address also, so I think if that action is re-run it should pass?

lengau:
@tonyandrewmeyer I'm not sure why it's still not liking it. If you want you can rebase and force-push; otherwise I'll check tomorrow and see what we can do.

lengau:
Thanks again @tonyandrewmeyer ! This is merged

PR #1626, opened 2024-04-03
add jammy and x64 labels for self-hosted runners

jdkandersson:
Adds the jammy and x64 labels for the self-hosted runner workflows to ensure the workflows don't break as we add more architectures and Ubuntu OS versions.


lengau:
Thanks for this! This particular job only actually cares that the runner has spread installed and has access to the correct Google Cloud project. Should we be adding these other tags as well, or should the tag adjustment in https://github.com/canonical/charmcraft/pull/1628 be sufficient?

jdkandersson:
Yes that is sufficient 🙂

PR #1629, opened 2024-04-03
remove experimental flag for flask-framework extension

jdkandersson:
We have now tested the extension with the web team and they are happy with. This means we're ready to move the `flask-framework` extension to be stable.


lengau (commented):
LGTM, thanks

lengau (approved):

tigarmo (approved):

PR #1663, opened 2024-04-22
refactor: remove unused fetch-lib command

lengau:
This removes the unused pre-craft-application `fetch-lib` command.


mattculler (approved):

tigarmo (approved):

maurorodrigues:
hey @lengau, I'm no charm expert, but I've been using fetch-lib. Is there an alternative to fetch-lib? Maybe something automatically done within the charmcraft process?

lengau:
@maurorodrigues this only applies to an implementation that's no longer used, not to the command itself.

We're moving towards defining the libraries in `charmcraft.yaml` and using the newer `fetch-libs` command with that, but we currently have no plans to remove `fetch-lib` as a charmcraft command available to those who need it.

PR #1683, opened 2024-05-16
chore: explicitly use the 'check' command to avoid ruff warning about it

tonyandrewmeyer:
Modern `ruff` requires explicitly saying `ruff check` for checks (since there's `format` now). Not doing this outputs a warning explaining that it should be done.

This adjusts the templates to include `check` explicitly (both in lint and format).


cmatsuoka (approved):

cmatsuoka:
Thanks!

lengau (approved):
Thanks!

PR #1691, opened 2024-05-29
feat: flask and django, update to paas-app-charmer 1.0.3

javierdelapuente:
Version 1.0.3 of [paas-app-charmer](https://github.com/canonical/paas-app-charmer) has been released, providing the new following functionality:

- Two new integrations, SAML and S3.
- Blocking on non optional integrations. If the optional attribute in the `charmcraft.yaml` file is set to false and the integration is not active, paas-app-charmer will block the charm (and stop all the services).

This PR adapts charmcraft to get advantage of that functionality.

Besides those changes, `charm-libs` has been added to the extension (including the new two new libraries for SAML and S3). That way the user can do `charmcraft fetch-libs` instead of one `fetch-lib` per library.


lengau:
@javierdelapuente please ping me when this is ready to review

jdkandersson (approved):

javierdelapuente:
Hi @lengau , can you have a look at this PR? Thanks!

charmcraft/extensions/gunicorn.py
```diff
@@ -134,6 +134,17 @@ def _get_root_snippet(self) -> dict[str, Any]:
                     "description": f"{self.framework} application image.",
                 },
             },
+            "charm-libs": [
+                {"lib": "traefik_k8s.ingress", "version": "2"},
+                {"lib": "observability_libs.juju_topology", "version": "0"},
```
lengau, line 139: Is there any worry with version 0 libs here that they'll make breaking changes?

lengau (approved):
Looks good, thanks! I have one question but it's for my own understanding.

charmcraft/extensions/gunicorn.py
jdkandersson: I think for some reason it is common to start the versions on version 0 for charm libraries, I have seen it a lot. There generally shouldn't be breaking changes within any major version

javierdelapuente: Hopefully LIBPATCH will not introduce non-breaking changes. https://juju.is/docs/sdk/document-your-charm-library.

I do not think the LIBAPI 0 has the usual semantics of semver with major number 0 (that any breaking change can happen). In the wild, we haven't have problems with that either.

mr-cal (approved):
Approving with the understanding that:
- charms using these extensions with `charm-libs` defined in a `charmcraft.yaml` will now get an error
- there is not a mechanism to override or modify the charm-libs defined by the extension
    - this can be added later as a feature
- these extensions are experimental, so this is an OK change

javierdelapuente:
> Approving with the understanding that:
> 
> * charms using these extensions with `charm-libs` defined in a `charmcraft.yaml` will now get an error
> * there is not a mechanism to override or modify the charm-libs defined by the extension
>   
>   * this can be added later as a feature
> * these extensions are experimental, so this is an OK change

I will add a new PR with a fix and a test. As you poing, the error is not nice and no merging is possible. Thanks.

Issue #1685, opened 2024-05-16
Multi-arch OCI image incorrectly uploaded to Charmhub registry


carlcsaposs-canonical:
### Bug Description

`charmcraft upload-resource` with a multi-arch image (e.g. https://github.com/canonical/charmed-postgresql-rock/pkgs/container/charmed-postgresql/208426154?tag=14.11-22.04_edge) uploads the amd64 image (if running charmcraft on amd64 machine) to the Charmhub registry

Context:
https://warthogs.atlassian.net/browse/DPE-4280
https://matrix.to/#/!LhFxJIPEcCacgdMghH:ubuntu.com/$JMmA26BWFkqAKsCJ0tfAx4fpVOSndhej3-CYHwHyBpc?via=ubuntu.com&via=matrix.org

### To Reproduce

1. Release charm with multi-arch image using charmcraft on amd64 machine (using upload-resource with `ghcr.io/canonical/charmed-postgresql:14.11-22.04_edge@sha256:31cf150b4523481202c1ff9b7b5d7f0b36729edad89d61242d8f1eb56b2912c0`)
2. `juju deploy` charm on arm64 machine
3. Inspect juju debug-log
```
ops.pebble.ChangeError: cannot perform the following tasks:
- Start service "postgresql" (cannot start service: fork/exec /usr/bin/patroni: exec format error)
```

Note: if `juju deploy ./postgresql-k8s.charm --resource postgresql-image=ghcr.io/canonical/charmed-postgresql:14.11-22.04_edge@sha256:31cf150b4523481202c1ff9b7b5d7f0b36729edad89d61242d8f1eb56b2912c0` is used on amd64 or arm64, the charm deploys correctly

### Environment

Ubuntu 22.04 amd64

### charmcraft.yaml

```shell
type: "charm"
bases:
  - name: ubuntu
    channel: "22.04"
    architectures: [amd64]
  - name: ubuntu
    channel: "22.04"
    architectures: [arm64]
parts:
  charm:
    override-pull: |
      craftctl default
      if [[ ! -f requirements.txt ]]
      then
          echo 'ERROR: Use "tox run -e build-dev" instead of calling "charmcraft pack" directly' >&2
          exit 1
      fi
    build-packages:
      - libffi-dev
      - libssl-dev
      - rustc
      - cargo
      - pkg-config
      - libpq-dev
    charm-strict-dependencies: true
  libpq:
    build-packages:
      - libpq-dev
    plugin: dump
[...]

PR #1696, opened 2024-06-07
feat: upload OCI image resources with skopeo

### LINKED ISSUE #1684: Allow Charmcraft to upload a rock directly from the file or directly between registries
### What needs to get done

Currently charmcraft requires an image digest on the local docker registry to upload an OCI image. When it does that, it extracts the OCI image to a tarball (equivalent to `docker image save`) and then uploads that.

Charmcraft should be able to take a rock file directly. For example:

```
$ charmcraft upload-resource my-charm my-resource ./my-rock_1.0.rock
```

This would do the equivalent of `skopeo copy oci-archive:my-rock_1.0.rock docker://registry.charmhub.io/my-rock:latest`

For copying between registries, Charmcraft should take something like:

```
$ charmcraft upload-resource my-charm my-resource docker://ghcr.io/canonical/charmed-postgresql:14.11-22.04_edge
```
This would do the equivalent of `skopeo copy docker://ghcr.io/canonical/charmed-postgresql:14.11-22.04_edge docker://registry.charmhub.io/my-rock:latest`

We have permission from @sergiusens to replace the current Python-based implementation with `skopeo`, but we must still be able to take:

```
$ charmcraft upload-resource my-charm my-resource  --filepath some-file
$ charmcraft upload-resource my-charm my-resource --image 31cf150b4523481202c1ff9b7b5d7f0b36729edad89d61242d8f1eb56b2912c0
```

cc @carlcsaposs-canonical 

### Why it needs to get done

Rockcraft outputs files directly

lengau:
Fixes #1684,  partial for #1685

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -67,6 +67,8 @@ jobs:
           export $(cat /etc/os-release | grep VERSION_CODENAME)
           pip install -U -r "requirements-${VERSION_CODENAME}.txt"
       - name: Install external dependencies with homebrew
+        # This is only necessary for Linux until skopeo >= 1.11 is in repos.
+        # Once we're running on Noble, we can get skopeo from apt.
         if: ${{ runner.os == 'Linux' || runner.os == 'macOS' }}
         run: |
           if [[ $(uname --kernel-name) == "Linux" ]]; then
@@ -82,6 +84,7 @@ jobs:
         run: |
           if [[ $(uname --kernel-name) == "Linux" ]]; then
             # Ensure the version of skopeo comes from homebrew
+            # This is only necessary until we move to noble.
             eval "$(/home/linuxbrew/.linuxbrew/bin/brew shellenv)"
             # Allow skopeo to access the contents of /run/containers
             sudo chmod 777 /run/containers
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -63,9 +63,16 @@ jobs:
         if: ${{ startsWith(matrix.os, 'ubuntu') }}
         run: |
           sudo apt update
-          sudo apt install -y python3-pip python3-setuptools python3-wheel python3-venv libapt-pkg-dev skopeo
+          sudo apt install -y python3-pip python3-setuptools python3-wheel python3-venv libapt-pkg-dev
           export $(cat /etc/os-release | grep VERSION_CODENAME)
           pip install -U -r "requirements-${VERSION_CODENAME}.txt"
+      - name: Install external dependencies with homebrew
+        if: ${{ runner.os == 'Linux' || runner.os == 'macOS' }}
+        run: |
+          if [[ $(uname --kernel-name) == "Linux" ]]; then
+            eval "$(/home/linuxbrew/.linuxbrew/bin/brew shellenv)"
+          fi
+          brew install skopeo
       - name: Configure environment
         run: |
           python -m pip install tox
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -63,7 +63,7 @@ jobs:
         if: ${{ startsWith(matrix.os, 'ubuntu') }}
         run: |
           sudo apt update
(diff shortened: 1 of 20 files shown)


.github/workflows/tests.yaml
```diff
@@ -66,12 +66,30 @@ jobs:
...
+          brew install skopeo
       - name: Configure environment
         run: |
           python -m pip install tox
           tox run --colored yes -m tests --notest
       - name: Run tests
+        shell: bash
         run: |
+          if [[ $(uname --kernel-name) == "Linux" ]]; then
+            # Ensure the version of skopeo comes from homebrew
```
tigarmo, line 84: how come skopeo needs to be from homebrew?

charmcraft/application/commands/store.py
```diff
@@ -1936,87 +1937,95 @@ def run(self, parsed_args):
...
+                bases=bases,
+            )
         elif parsed_args.image:
+            emit.progress("Getting image")
+            emit.debug("Trying to get image from Docker")
+            image_service = self._services.image
+            if digest := image_service.get_maybe_id_from_docker(parsed_args.image):
+                emit.debug("Image is available via the local Docker daemon.")
+                source_path = f"docker-daemon:{digest}"
+            elif (image_path := pathlib.Path(parsed_args.image)).exists():
```
tigarmo, line 1956: should you do this check first? It seems cheaper than checking docker

charmcraft/errors.py
```diff
@@ -171,3 +175,22 @@ def __init__(self, extra_dependencies: Iterable[str]):
...
+class SubprocessError(CraftError):
+    """A craft-cli friendly subprocess error."""
+
+    @classmethod
+    def from_subprocess(cls, error: subprocess.CalledProcessError) -> Self:
+        """Convert a CalledProcessError to a craft-cli error."""
+        error_details = f"Full command: {shlex.join(error.cmd)}\nError text:\n"
+        if isinstance(error.stderr, str):
+            error_details += textwrap.indent(error.stderr, "  ")
+        else:
```
tigarmo, line 189: `stderr` might also be `bytes` right?

tigarmo (approved):
nice!

charmcraft/errors.py
lengau: Yes, but the use case we have for this is always with `text=True`, so it shouldn't be an issue.

charmcraft/application/commands/store.py
lengau: It is cheaper, but I don't want to break backwards compatibility here.

mr-cal (approved):
Looks good!

charmcraft/services/image.py
```diff
@@ -0,0 +1,158 @@
...
+        """Inspect an image with Skopeo and return the relevant metadata.
+
+        :param image: A skopeo-friendly image name
+        :returns: An OCIMetadata object containing metadata about the image.
+        """
+        try:
+            raw_manifest = self._skopeo.inspect(image, raw=True)
+            image_info = self._skopeo.inspect(image)
+        except errors.SubprocessError as exc:
+            raise errors.CraftError(
```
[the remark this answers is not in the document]
lengau: This is catching the error and replacing the error message with something more specific to the use case.

charmcraft/services/image.py
```diff
@@ -0,0 +1,158 @@
...
+    @override
+    def setup(self) -> None:
+        """Set up the image service."""
+        super().setup()
+        if not self._skopeo_path:
+            raise errors.CraftError(
+                "Could not find skopeo on the machine.",
+            )
+        logger.debug("skopeo path: %s", self._skopeo_path)
+        self._skopeo = utils.Skopeo(skopeo_path=self._skopeo_path)
```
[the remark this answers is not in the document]
lengau: Good catch - this is duplicated work.

charmcraft/application/commands/store.py
tigarmo: makes sense! Should we add a note about backwards-compat here?

lengau: Done!

PR #1718, opened 2024-07-04
Flask extension. Allow charm-libs to be extended by the user

javierdelapuente:
As pointed by @mr-cal in https://github.com/canonical/charmcraft/pull/1691#pullrequestreview-2156902965, the Flask extension fails when a `charm-libs` is already defined in a `charmcraft.yaml`, with a not very clear message because of an `AttributeError` exception.

This PR allows having a `charm-libs`  in the original `charmcraft.yaml`, and will combine both `charm-libs` lists (if there are duplicated elements they will be only downloaded once).


lengau (approved):

lengau:
Looks good to me, thanks!

mattculler (approved):

Issue #1664, opened 2024-04-23
`charm-python-packages` does not guarantee these packages are installed before packages in `requirements.txt`


DnPlas:
### Bug Description

I have noticed that the list of packages in `charm-python-packages` may not guarantee they will be installed before charm requirements, as the [docs suggest](https://juju.is/docs/sdk/charmcraft-yaml#heading--the-charm-plugin).

The behaviour I am getting is that the packages listed in `charm-python-packages` get collected and it looks like they are being installed as the very first thing, for example:

```
:: Running external command ['/root/parts/charm/build/staging-venv/bin/pip', 'install', '--no-binary=anyio,attrs,certifi,charmed-kubeflow-chisme,charset-normalizer,deepdiff,exceptiongroup,h11,httpcore,httpx,idna,importlib-resources,jinja2,jsonschema,lightkube,lightkube-models,markupsafe,oci-image,ops,ordered-set,pip,pkgutil-resolve-name,pyrsistent,pyyaml,requests,ruamel-yaml,ruamel-yaml-clib,serialized-data-interface,setuptools,sniffio,tenacity,urllib3,websocket-client,wheel,zipp', '--requirement=requirements.txt', 'pip>23.0', 'setuptools>69.0', 'wheel']    
::    :: Collecting pip>23.0                                                                                                                                                                 
::    ::   Using cached pip-24.0.tar.gz (2.1 MB)                                                                                                                                             
::    ::   Installing build dependencies: started                                                                                                                                            
::    ::   Installing build dependencies: finished with status 'done'                                                                                                                        
::    ::   Getting requirements to build wheel: started                                                                                                                                      
::    ::   Getting requirements to build wheel: finished with status 'done'                                                                                                                  
::    ::     Preparing wheel metadata: started                                                                                                                                               
::    ::     Preparing wheel metadata: finished with status 'done'                                                                                                                           
::    :: Collecting setuptools>69.0                                                                                                                                                          
::    ::   Using cached setuptools-69.5.1.tar.gz (2.3 MB)                                                                                                                                    
::    ::   Getting requirements to build wheel: started                                                                                                                                      
::    ::   Getting requirements to build wheel: finished with status 'done'                                                                                                                  
::    ::   Installing backend dependencies: started                                                                                                                                          
::    ::   Installing backend dependencies: finished with status 'done'                                                                                                                      
::    ::     Preparing wheel metadata: started                                                                                                                                               
[...]

PR #1727, opened 2024-07-15
fix(builder)!: remove single-step non-strict install

lengau:
This separates out the install step for non strict-dependencies builds as follows:

1. Install from `python-binary-packages`
2. Install (source-only) from `python-packages`
3. Install (source-only) from requirements files and charm libs.


charmcraft/charm_builder.py
```diff
@@ -251,44 +251,36 @@ def _install_dependencies(self, staging_venv_dir):
...
-            # This method is not valid for any bases added after 2024-01-01 or for DEVEL bases.
-            try:
+            # Non-strict dependency resolution:
+            # 1. Install binary-allowed packages
+            # 2. Install source packages
+            # 3. Install from requirements files and charm libs dependencies
+            if self.binary_python_packages:
+                print(
+                    "Installing binary-allowed packages and their dependencies.\n"
+                    "WARNING: dependencies may also be installed from binary wheels."
```
mr-cal, line 261: It would be useful to suggest using strict mode to resolve this warning

charmcraft/charm_builder.py
```diff
@@ -251,44 +251,36 @@ def _install_dependencies(self, staging_venv_dir):
...
+                )
                 _process_run(
                     get_pip_command(
                         [pip_cmd, "install"],
-                        self.requirement_paths,
-                        source_deps=[*self.python_packages, *self.charmlib_deps],
+                        [],
```
mr-cal, line 266: pedantic: the previous usage hinted at what this parameter is.  Now that it is an empty list, it is not obvious.
```suggestion
                        requirement_files=[],
```

mr-cal (approved):

charmcraft/charm_builder.py
lengau: Great idea! Added.

charmcraft/charm_builder.py
lengau: Good point. Added!

dariuszd21 (approved):

lengau:
Bypassing protections to merge - test failure is due to: https://github.com/canonical/charmcraft/issues/1743

kimwnasptd:
The PR seems to not fully resolve the issue in https://github.com/canonical/charmcraft/issues/1664#issuecomment-2242468045

PR #1736, opened 2024-07-16
fix(tests): make resources test more resilient

lengau:
It currently breaks because it exceeds the possible length of a bash var


charmcraft/models/project.py
```diff
@@ -1043,9 +1043,7 @@ def _check_base_is_legacy(base: BaseDict) -> bool:
             and base["channel"] < "24.04"  # pyright: ignore[reportTypedDictNotRequiredAccess]
         ):
             return True
-        if base in ({"name": "centos", "channel": "7"}, {"name": "almalinux", "channel": "9"}):
-            return True
-        return False
+        return base in ({"name": "centos", "channel": "7"}, {"name": "almalinux", "channel": "9"})
```
lengau, line 1046: https://github.com/canonical/charmcraft/pull/1737

charmcraft/application/commands/store.py
```diff
@@ -1992,7 +1992,7 @@ def run(self, parsed_args: argparse.Namespace) -> int:
             if const.STORE_REGISTRY_ENV_VAR in os.environ:
                 # If the user has specified a registry to use, replace what the store
                 # gives with that registry.
-                registry_url = os.environ[const.STORE_REGISTRY_ENV_VAR][7:]
+                registry_url = os.environ[const.STORE_REGISTRY_ENV_VAR][8:]
```
mattculler, line 1995: This looks fragile.  Maybe add a comment explaining what the list looks like that we're slicing here, or why the first n elements are skipped, etc.

tests/spread/store/resources/task.yaml
```diff
@@ -64,11 +64,11 @@ execute: |
 
   # upload and verify the file resource
   upload_revision_number=$(charmcraft upload-resource $CHARM_DEFAULT_NAME example-file --format=json --file=resourcefile.txt --arch=amd64,riscv64 | jq -r '.revision')
-  revisions_json=$(charmcraft resource-revisions $CHARM_DEFAULT_NAME example-file --format=json)
-  upload_revision=$(echo "$revisions_json" | jq -c ".[] | select(.revision == $upload_revision_number)" )
+  charmcraft resource-revisions $CHARM_DEFAULT_NAME example-file --format=json > file_revisions.json
```
mattculler, line 67: Do you need to delete file_revisions.json when you're done?

mattculler (approved):

tests/spread/store/resources/task.yaml
lengau: Nah, the directory gets deleted when the test is cleaned up

charmcraft/application/commands/store.py
lengau: Good point. I've made this more obvious.

tests/spread/store/resources/task.yaml
mattculler: Much better, thanks!

charmcraft/application/commands/store.py
mattculler: Much better, thanks!

charmcraft/application/commands/store.py
```diff
@@ -1992,9 +1992,10 @@ def run(self, parsed_args: argparse.Namespace) -> int:
             if const.STORE_REGISTRY_ENV_VAR in os.environ:
                 # If the user has specified a registry to use, replace what the store
                 # gives with that registry.
-                registry_url = os.environ[const.STORE_REGISTRY_ENV_VAR][7:]
+                registry_url = os.environ[const.STORE_REGISTRY_ENV_VAR]
+                registry_url_without_scheme = registry_url.partition("://")[2]
```
dariuszd21, line 1996: Is there a particular reason for not using `urlparse` for that ?

tests/spread/store/resources/task.yaml
```diff
@@ -64,11 +64,11 @@ execute: |
...
+  upload_revision=$(cat file_revisions.json | jq -c ".[] | select(.revision == $upload_revision_number)" )
   upload_revision_architectures=$(echo "$upload_revision" | jq -c '.bases[0].architectures')
   [[ "$upload_revision_architectures" = '["amd64","riscv64"]' ]]
-  last_revision=$(echo "$revisions_json" | jq -r .[-1])
+  last_revision=$(cat file_revisions.json | jq -r .[-1])
```
dariuszd21, line 71: I believe you can use jq directly on file without piping
```suggestion
  last_revision=$(jq -r .[-1] file_revisions.json)
```

dariuszd21 (approved):

charmcraft/application/commands/store.py
lengau: `urlparse` doesn't have an easy way to present just the sections I need, so I'd still have to do some manual string manipulation after parsing it.

tests/spread/store/resources/task.yaml
lengau: Thanks! Add me here: https://porkmail.org/era/unix/award

Issue #1473, opened 2024-01-10
Strict dependencies allows binary packages for indirect dependencies


carlcsaposs-canonical:
### Bug Description

`charm-strict-dependencies: true` changes the behavior of `charm-binary-python-packages` ([source](https://discourse.charmhub.io/t/upcoming-charmcraft-changes-and-a-call-for-testers/12175))

If requirements.txt does not contain all dependencies (e.g. it contains `ops` but not `pyyaml`, which is a dependency of `ops`), indirect dependencies (e.g. `pyyaml`) will be installed from binary wheels instead of from source—even if they are not specified in `charm-binary-python-packages`

This is because charmcraft uses `pip install --no-binary` with a list of package names in requirements.txt

To solve this, charmcraft should use `pip install --no-binary :all: --only-binary` with a list of package names in `charm-binary-python-packages` (allowlist binary packages instead of blocklist source packages, since not all packages are necessarily in requirements.txt). Note that **order matters**

### To Reproduce

```
python3 -m venv venv3
source venv3/bin/activate
pip install ops --no-binary ops
```

### Environment

N/A

### charmcraft.yaml

```shell
N/A
```

### Relevant log output

```shell
ubuntu@j31-3:~$ python3 -m venv venv3
ubuntu@j31-3:~$ source venv3/bin/activate
(venv3) ubuntu@j31-3:~$ pip install ops --no-binary ops
Collecting ops
  Downloading ops-2.9.0.tar.gz (255 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 255.8/255.8 KB 3.1 MB/s eta 0:00:00
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
Collecting PyYAML==6.*
  Downloading PyYAML-6.0.1-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (705 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 705.5/705.5 KB 38.4 MB/s eta 0:00:00
Collecting websocket-client==1.*
  Downloading websocket_client-1.7.0-py3-none-any.whl (58 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 58.5/58.5 KB 11.2 MB/s eta 0:00:00
[...]

PR #1740, opened 2024-07-18
fix(pip): set `--no-binary=:all:` if possible

lengau:
This reduces the chances of having pip install binary packages as indirect dependencies.

Partial fix for #1473

### DIFF
--- a/charmcraft/utils/package.py
+++ b/charmcraft/utils/package.py
@@ -103,10 +103,6 @@ def get_pip_command(
     binary_packages = get_pypi_packages(binary_deps)
     requirements_packages = get_requirements_file_package_names(*requirements_files)
     all_packages = charm_packages | binary_packages | requirements_packages
-    source_only_packages = sorted(
-        get_package_names(all_packages) - get_package_names(binary_packages)
-    )
-
     non_requirements_packages = sorted(
         exclude_packages(
             set(source_deps) | set(binary_deps),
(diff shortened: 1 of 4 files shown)


charmcraft/utils/package.py
```diff
@@ -103,21 +103,25 @@ def get_pip_command(
...
-    if source_only_packages:
-        no_binary = [f"--no-binary={','.join(source_only_packages)}"]
-    else:
-        no_binary = []
+    if not binary_packages:
+        return [
+            *prefix,
+            "--no-binary=:all:",
+            *(f"--requirement={path}" for path in requirements_files),
+            *non_requirements_packages,
```
dariuszd21, line 118: `non_requirements_package` may include binary dependencies. 
Will it work in such scenario ?

lengau: `non_requirements_packages` only contains binary dependencies if `binary_packages` is not empty, as `binary_packages` is a 1:1 map of `binary_deps`

dariuszd21: So you'll have always a combination of `--no-binary` with explicitly allowed binary_packages ?

lengau: No, `--no-binary=:all` only gets set if there aren't any explictly allowed binary packages.

If any explicitly-allowed binary packages are set, the code path below that selects that list of binary packages gets used instead.

dariuszd21 (approved):

charmcraft/utils/package.py
dariuszd21: Got it, thanks!

mr-cal (approved):

PR #1744, opened 2024-07-19
fix(flask): use rustup snap

lengau:
A new update to a flask dependency requires a more recent version of rustc than is available in the jammy repos.
This works around that by using the same rustup snap as the craft-parts rust plugin

Fixes #1743

Spread test failure is fixed in https://github.com/canonical/charmcraft/pull/1749

PR #1750, opened 2024-07-22
fix(charm_builder): de-duplicate installed pkgs

lengau:
Ensure that charmlibs that are already installed using requirements.txt don't get additionally listed on the command line, as this can break pip.

Spread test failures are fixed in #1744 and #1749


dariuszd21 (approved):
Looks good!

mr-cal (approved):

PR #1757, opened 2024-07-23
Minor cleanup

lengau:
To be rebased - check the commits individually


tigarmo (approved):

charmcraft/models/project.py
```diff
@@ -377,7 +377,7 @@ def get_build_plan(self) -> list[models.BuildInfo]:
         """
         if self.type == "bundle":
             # A bundle can build anywhere, so just present the current system.
-            current_arch = utils.get_host_architecture()
+            current_arch = util.get_host_architecture()
             current_base = utils.get_os_platform()
```
mr-cal, line 381: I don't like how `util` and `utils` are used in the same module but I don't have a good alternative.

Some of the other modules do `from charmcraft.utils import ...`, which is a bit more readable.

lengau: Eventually we're going to get rid of this in `utils` too, but also I've raised this: https://github.com/canonical/craft-application/issues/395

mr-cal (approved):

PR #1765, opened 2024-07-25
chore: remove classic entrypoint

lengau:
This was leftover from the craft-application changeover and is no longer necessary.


mr-cal (approved):
Nice!

tigarmo (approved):

PR #1779, opened 2024-08-01
build(deps): use uv for freeze-requirements

lengau:
This speeds up the `freeze-requirements.sh` script quite substantially while still having equivalent output.


mr-cal (approved):
What do you think about updating the HACKING guide to advise installing `uv`?

lengau:
@mr-cal good idea. We don't have  a HACKING file but I did the next best thing.

dimaqq (approved):
😎 UV 😎

CONTRIBUTING.md
```diff
@@ -1,5 +1,10 @@
 ## Development environment
 
+It's recommended to use `uv` for setting up your development environment, but
+this is not required. `uv` is available [as a snap](https://snapcraft.io/astral-uv)
+and from [official sources](https://docs.astral.sh/uv/getting-started/installation/).
+If using `uv`, substitute `virtualenv` with `uv venv` and `pip` with `uv pip`.
```
sergiusens, line 6: if `uv` is recommended, these instructions should be along the lines of _if not using ..._

lengau: Easier to just not explain how to use something else anyway :-)

CONTRIBUTING.md
```diff
@@ -15,7 +19,7 @@ can install ruff with:
 
 Otherwise, you can install ruff in your virtual environment with:
 
-    pip install ruff
+    uv pip install ruff
```
dimaqq, line 22: If we're going to use the latest version, then I think the uv way would be:

```
uv tool install ruff
# now it's available globally

uvx
# ruff v0.6.5

uvx ruff --help
# blah blah blah
```

lengau: I'd rather keep the snap as the preferred way since it'll do auto-updates (which `uv tool` and `uvx` don't), but I agree that we should be using

```suggestion
    uv tool install ruff
```

for someone who'd rather use `uv` to install ruff.

PR #1795, opened 2024-08-06
fix(package): Limit the bases in manifest.yaml

lengau:
For a `bases` charm, this was previously writing out all run bases rather than the specific bases for this build.

### DIFF
--- a/charmcraft/services/package.py
+++ b/charmcraft/services/package.py
@@ -17,7 +17,6 @@
 """Service class for packing."""
 from __future__ import annotations
 
-import itertools
 import json
 import os
 import pathlib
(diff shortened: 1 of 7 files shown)


tests/integration/test_charm_builder.py
```diff
@@ -35,7 +35,7 @@
     ],
 )
 def test_install_strict_dependencies_pip_check_error(
-    new_path: pathlib.Path, requirements: list[str]
+    monkeypatch, new_path: pathlib.Path, requirements: list[str]
```
lengau, line 38: this is a side-effect of fixing a merge issue

mr-cal (approved):

PR #1800, opened 2024-08-08
feat: go-framework init profile and extension

javierdelapuente:
This PR implements the spec ISD153, "12-Factor Go Support" for charmcraft.

For that, a new init profile called "go-framework" and a new go extension, called "go-framework" are needed. These additions are similar to the flask and django init profiles and extensions.

For the Go profile, the  base 24.04 is used.

 Code in extensions is refactored, so all 12 factor extensions, that are similar, inherit from _TwelveFactorBase.


charmcraft/extensions/twelvefactor.py
```diff
@@ -262,3 +273,45 @@ class DjangoFramework(_GunicornBase):
...
         },
     }
+
+
+class GoFramework(_TwelveFactorBase):
+    """Extension for 12-factor Go applications."""
+
+    framework = "go"
+    actions = {
+        "rotate-secret-key": {
```
javierdelapuente, line 283: make this generic as a property

charmcraft/templates/init-go-framework/charmcraft.yaml.j2
```diff
@@ -0,0 +1,55 @@
...
+
+base: ubuntu@24.04
+
+platforms:
+  amd64:
+  arm64:
+  armhf:
+  ppc64el:
+  riscv64:
+  s390x:
```
lengau, line 16: OK so I personally love this, but does it accurately represent what's going to work?

While Charmcraft supports building on/for `armhf` and `riscv64`, the Juju snap isn't available for either and IDK whether Juju supports machines running on either. I could potentially see riscv64 as an "upcoming platform", but if there's not a directive to support `armhf` we might want to drop that (or both) for now.

charmcraft/templates/init-go-framework/src/charm.py.j2
```diff
@@ -0,0 +1,30 @@
...
+import typing
+
+import ops
+
+import paas_app_charmer.go
+
+logger = logging.getLogger(__name__)
+
+
+class GoCharm(paas_app_charmer.go.Charm):
```
lengau, line 17: I know this is the same in the Django and Flask extensions, but we might want to go back to using the same as our other templates:

```suggestion
class {{ GoCHarm }}(paas_app_charmer.go.Charm):
```

charmcraft/extensions/twelvefactor.py
```diff
@@ -184,17 +170,41 @@ def get_parts_snippet(self) -> dict[str, Any]:
         """Return the parts to add to parts."""
         return {}
 
+    def get_container_name(self) -> str:
+        """Return name of the container for the app image."""
+        return f"{self.framework}-app"
```
lengau, line 175: Should this maybe be the charm's name? (Likewise `{charm_name}-image` below?)

charmcraft/templates/init-go-framework/charmcraft.yaml.j2
javierdelapuente: Thanks for the info (I just learned that). Dropped both of them.

charmcraft/templates/init-go-framework/src/charm.py.j2
javierdelapuente: I have updated the three templates to use this code. Thanks!

charmcraft/extensions/twelvefactor.py
javierdelapuente: This name is used in other places (like the paas-app-charmer library), and changing it will be a backward incompatible change.

In the new framework (Go), the name is just app.

charmcraft/extensions/__init__.py
```diff
@@ -18,7 +18,7 @@
 
 from charmcraft.extensions._utils import apply_extensions
 from charmcraft.extensions.extension import Extension
-from charmcraft.extensions.gunicorn import DjangoFramework, FlaskFramework
+from charmcraft.extensions.twelvefactor import DjangoFramework, FlaskFramework, GoFramework
```
jdkandersson, line 21: let's change it to `app` with a backup option of `app_framework`

charmcraft/extensions/twelvefactor.py
```diff
@@ -24,27 +24,8 @@
...
-        "webserver-timeout": {
-            "type": "int",
-            "description": "Time in seconds to kill and restart silent webserver workers.",
-        },
-        "webserver-workers": {
-            "type": "int",
-            "description": "The number of webserver worker processes for handling requests.",
-        },
-    }
+class _TwelveFactorBase(Extension):
```
jdkandersson, line 27: `_AppBase`?

charmcraft/extensions/twelvefactor.py
```diff
@@ -262,3 +271,40 @@ class DjangoFramework(_GunicornBase):
...
+        },
+        "metrics-path": {
+            "type": "string",
+            "default": "/metrics",
+            "description": "Path where the prometheus metrics will be scraped.",
+        },
+        "secret-key": {
+            "type": "string",
+            "description": "Long secret you can use for sessions, csrf or any other thing where you need a random secret shared by all units",
+        },
```
jdkandersson, line 299: Update names please

charmcraft/templates/init-go-framework/.gitignore.j2
```diff
@@ -0,0 +1,9 @@
+venv/
+build/
+*.charm
+.tox/
```
jdkandersson, line 4: Perhaps this isn't needed?

charmcraft/templates/init-go-framework/charmcraft.yaml.j2
```diff
@@ -0,0 +1,53 @@
...
+
+type: charm
+
+base: ubuntu@24.04
+
+platforms:
+  amd64:
+  arm64:
+  ppc64el:
+  s390x:
```
jdkandersson, line 14: Let's do this for the rest of the charms and rocks as well please

charmcraft/templates/init-go-framework/.gitignore.j2
javierdelapuente: After reviewing, I'd rather leave it there, in case there it a tox file for things like tests or linting.

charmcraft/extensions/__init__.py
javierdelapuente: Renamed to app

charmcraft/extensions/twelvefactor.py
javierdelapuente: Renamed to _AppBase

charmcraft/extensions/twelvefactor.py
javierdelapuente: Done

lengau (approved):
Looks good, thanks!

tigarmo (approved):

tigarmo (changes requested):
removing my approval so we fix the supported bases

javierdelapuente:
> removing my approval so we fix the supported bases

Corrected to base 24.04 for go and commented all platforms except amd64

charmcraft/templates/init-go-framework/charmcraft.yaml.j2
```diff
@@ -9,9 +9,9 @@ base: ubuntu@24.04
 
 platforms:
   amd64:
-  arm64:
-  ppc64el:
-  s390x:
+  # arm64:
```
jdkandersson, line 12: Perhaps include a note explaining what should be done including which command to run to check the architecture, e.g., `uname -m` and how to interpret the command output?

tigarmo (approved):

jdkandersson (approved):

PR #1815, opened 2024-08-14
fix(yaml): correctly add default parts

lengau:
The pydantic 2 change caused some weirdness in adding default parts, so this fixes it. After this, pydantic 2 should be ready to merge to main.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -80,15 +80,28 @@ jobs:
         run: |
           sudo apt update
           sudo apt install -y python3-pip python3-setuptools python3-wheel python3-venv libapt-pkg-dev
-      - name: Install external dependencies with homebrew
+      - name: Install skopeo (mac)
         # This is only necessary for Linux until skopeo >= 1.11 is in repos.
         # Once we're running on Noble, we can get skopeo from apt.
-        if: ${{ runner.os == 'Linux' || runner.os == 'macOS' }}
+        if: ${{ runner.os == 'macOS' }}
         run: |
-          if [[ $(uname --kernel-name) == "Linux" ]]; then
-            eval "$(/home/linuxbrew/.linuxbrew/bin/brew shellenv)"
-          fi
           brew install skopeo
+      - name: Install skopeo (Linux)
+        if: ${{ runner.os == 'Linux' }}
+        run: |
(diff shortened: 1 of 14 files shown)


.github/workflows/tests.yaml
```diff
@@ -80,15 +80,28 @@ jobs:
         run: |
           sudo apt update
           sudo apt install -y python3-pip python3-setuptools python3-wheel python3-venv libapt-pkg-dev
-      - name: Install external dependencies with homebrew
+      - name: Install skopeo (mac)
```
lengau, line 83: These changes prevent the Jammy builders from getting a Homebrew-installed version of Python 3.12 that has weird behaviour.

tigarmo (approved):

.github/workflows/tests.yaml
```diff
@@ -80,15 +80,28 @@ jobs:
...
         # Once we're running on Noble, we can get skopeo from apt.
-        if: ${{ runner.os == 'Linux' || runner.os == 'macOS' }}
+        if: ${{ runner.os == 'macOS' }}
         run: |
-          if [[ $(uname --kernel-name) == "Linux" ]]; then
+          brew install skopeo
+      - name: Install skopeo (Linux)
+        if: ${{ runner.os == 'Linux' }}
+        run: |
+          if [[ $(cat /etc/os-release | grep VERSION_CODENAME) == 'VERSION_CODENAME=jammy' ]]; then
```
dariuszd21, line 92: I'm kind of lost in here, cause it checks if it's jammy and it installs `skopeo` through homebrew and then deletes the binary ?
It's quite different from the comment above which states that it should prevent Jammy from getting a homebrew installed version of python3.12 :D

You're installing it to remove it and to link it to the version I have no idea how was installed :smile:

lengau: I'm installing it from homebrew, which puts it in `~/.linuxbrew/bin`. There's an (old) skopeo executable in `/bin`, which I'm deleting and then replacing with this linuxbrew installed version.

dariuszd21: Got it! Makes sense. I was confused by the above comment.
One additional small question, is the below script required, also when you use apt version of skopeo?:
```
# Allow skopeo to access the contents of /run/containers
          sudo chmod 777 /run/containers
          # Add an xdg runtime dir for skopeo to look into for an auth.json file
          sudo mkdir -p /run/user/$(id -u)
          sudo chown $USER /run/user/$(id -u)
```

PR #1817, opened 2024-08-15
chore: merge pydantic 2 support to main

lengau:
Everything here has already been reviewed.


pyproject.toml
```diff
@@ -4,20 +4,20 @@ dynamic = ["version"]
 description = "The main tool to build, upload, and develop in general the Juju charms."
 readme = "README.md"
 dependencies = [
-    "craft-application~=3.0",
+    "craft-application>=4.0.0",
```
dariuszd21, line 7: This will also pick up 5.0, is it intended ?

pyproject.toml
```diff
@@ -4,20 +4,20 @@ dynamic = ["version"]
...
+    "craft-store>=3.0.0",
     "distro>=1.3.0",
     "docker>=7.0.0",
     "humanize>=2.6.0",
     "jsonschema",
     "jinja2",
-    # Pydantic will need to be updated all at once with our libraries.
-    # When you update it, remove the pydantic constraints from renovate.
-    "pydantic>=1.10,<2.0",
+    "pydantic~=2.0",
```
dariuszd21, line 20:
```suggestion
    "pydantic~=2.8",
```

requirements-dev.lock
dariuszd21, line 1: Isn't it rye byproduct and it should be included in this PR ?

requirements.lock
dariuszd21, line 1: Same here

snap/snapcraft.yaml
```diff
@@ -145,6 +145,8 @@ parts:
       craftctl set version="${version}"
       [ -n "$(echo $version | grep "post")" ] && grade=devel || grade=stable
       craftctl set grade="${grade}"
+      # Set up the rust compiler
+      rustup default 1.79
```
dariuszd21, line 149: Let's go with the latest :smile: 
```suggestion
      rustup default 1.80
```

dariuszd21 (approved):

snap/snapcraft.yaml
lengau: I'm intentionally going with the minimum version to support the libraries - otherwise I'd just put in `stable` :-)

pyproject.toml
lengau: For now, yes. This is changed separately.

requirements-dev.lock
lengau: It's not currently used, but I'm experimenting with it.

PR #1828, opened 2024-08-16
feat: mark charmhub parameter as deprecated

lengau:


mattculler (approved):

mr-cal (approved):
Approving without testing - I assume the `repeat` logic works as-expected.

PR #1847, opened 2024-08-26
chore(renovate): ignore tox-gh updates

lengau:
This is just until we no longer need Windows CI

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -5,7 +5,9 @@
     // Each ignore is probably connected with an ignore in pyproject.toml.
     // Ensure you change this and those simultaneously.
     "urllib3",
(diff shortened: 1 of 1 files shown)


.github/renovate.json5
```diff
@@ -5,7 +5,9 @@
     // Each ignore is probably connected with an ignore in pyproject.toml.
```
mattculler, line 5: Per this comment, did you need to also add an ignore in pyproject.toml?

lengau: In this case no, because we don't even have tox-gh in pyproject.toml

tigarmo (approved):

mr-cal (approved):

PR #1861, opened 2024-08-27
docs(changelog): add 3.2.0 release

lengau:


docs/reference/changelog.rst
```diff
@@ -77,6 +77,40 @@ Changelog
...
+
+go-framework
+""""""""""""
+
+New ``go-framework`` extension for easily charming go applications
+
+Documentation
+=============
+
+Documentation now contains a changelog, and you're reading it right now.
```
medubelko, line 111:
```suggestion
The changelog is now included in the Charmcraft documentation. For completeness, we've back-filled the log with all the important changes from previous releases documented on GitHub.
```

Cheeky, but I think we should be careful with the continuous "you're". It can sound accusatory. It's the same reason the style guide advises us not to write "allow" or "let" because it speaks as though the software has power over the user.

medubelko (changes requested):

docs/reference/changelog.rst
```diff
@@ -77,6 +77,42 @@ Changelog
 
   For a complete list of commits, see the `X.Y.Z`_ release on GitHub.
 
+3.2.0 (2024-08-27)
```
tigarmo, line 80: might want to bump the date (up to you though)

tigarmo (approved):

PR #1863, opened 2024-08-27
Use the name variable for placeholder container, resource names

sed-i:
Currently, the k8s template uses `some-container` as a placeholder for container, resource names.
This PR changes the template to use the `name` variable instead.
This follows a convention many charmers happen to practice:
- container name matches the main workload's name
- resource name is the main workload's name + `-image` suffix.


lengau:
Thanks for the PR, @sed-i ! Could you re-author with an email address that meets the CLA requirement and fix the unit test breakage? I'd be happy to accept this PR with that 😄

PR #1894, opened 2024-09-11
refactor: use craft-platforms for build plans

lengau:
Only applies to `platforms` charms.


charmcraft/models/project.py
```diff
@@ -345,38 +347,28 @@ def get_build_plan(self) -> list[models.BuildInfo]:
...
+                for name, platform in self.platforms.items()
+            },
+        )
+        build_infos = charm.get_platforms_charm_build_plan(
+            base=self.base,
+            build_base=self.build_base,
+            platforms=platforms,
+        )
+        return [
+            models.BuildInfo(
```
tigarmo, line 364: suggestion: maybe craft-application's BuildInfo should have a staticmethod to convert a craft-platform's BuildInfo

charmcraft/models/project.py
```diff
@@ -345,38 +347,28 @@ def get_build_plan(self) -> list[models.BuildInfo]:
...
-                            )
-                            for build_for in platform.build_for  # type: ignore[union-attr]
-                        ]
-                    )
-        return build_infos
+        platforms = cast(
+            # https://github.com/canonical/craft-platforms/issues/43
+            craft_platforms.Platforms,  # pyright: ignore[reportPrivateImportUsage]
+            {
+                name: (platform.marshal() if platform else None)
```
tigarmo, line 354: this is just me rambling but it bothers me a bit that we have to do this 'round-trip' where we unmarshal the project, which includes validation (necessary) and possibly transformation (dangerous), and then re-marshal to pass to craft-platforms. This makes the assembling of the build plan a *little* different between the application and systems that just use craft-platforms (like launchpad), and makes us still need "all" of this code here.

I don't have any good suggestions to address this though; maybe the base BuildPlanner class should retain the original values of base, build-base and platforms somewhere?

tigarmo (approved):

charmcraft/models/project.py
lengau: Both of your comments here are things I'd like to address in https://github.com/canonical/craft-application/issues/391 :-)

mr-cal (approved):

PR #1898, opened 2024-09-11
feat(package): create a dispatch file if none exists

### LINKED ISSUE #1813: Add a dispatch script in the package service if it doesn't exist
### What needs to get done

Create a plugin for charmcraft that can take an arbitrary virtualenv + charm.py file and add the relevant hooks, etc.

### Why it needs to get done

The current `charm` plugin is currently way too tightly coupled.

lengau:
This creates a dispatch file during the post-prime step if there is no existing dispatch file and there isn't a hooks directory.

Progress towards #1813


charmcraft/dispatch.py
```diff
@@ -0,0 +1,60 @@
...
+    dispatch_path = prime_dir / const.DISPATCH_FILENAME
+    hooks_path = prime_dir / const.HOOKS_DIRNAME
+
+    if hooks_path.is_dir() or dispatch_path.is_file():
+        return False
+
+    if not (prime_dir / entrypoint).exists():
+        return False
+
+    dispatch_path.write_text(DISPATCH_SCRIPT_TEMPLATE.format(entrypoint=entrypoint))
```
mr-cal, line 57: It could be useful to log a debug message that this file was created, if a user is trying to figure out where this file is coming from.

mr-cal (approved):
Looks good, sorry for the delay in reviewing

dariuszd21 (approved):
Looks good, thank you!

charmcraft/dispatch.py
lengau: Great idea! I did one better and made it a progress message. It'll probably only flash up on the screen for under a second, but there's no reason that a user in brief mode wouldn't want to theoretically see it.

lengau:
Force-merging, as test failures are due to: https://github.com/RobertCraigie/pyright-python/issues/304

PR #1899, opened 2024-09-12
feat: add Python and Poetry plugins

lengau:
This adds Charmcraft-specific versions of the craft-parts `python` and `poetry` plugins. Notable differences:

- The venv is in a subdirectory
- Uses the system's `pip` so it doesn't have to install `pip` in the charm venv.
- Neither plugin tries to install the current directory as a module
- Defaults to not using binaries
- Adds the `src` and `lib` directories to the charm if they exist.


charmcraft/parts/plugins/_poetry.py
```diff
@@ -0,0 +1,97 @@
...
+        return self._part_info.part_install_dir / "venv"
+
+    def _get_pip(self) -> str:
+        """Get the pip command to use."""
+        return f"{self._get_system_python_interpreter()} -m pip --python=${{PARTS_PYTHON_VENV_INTERP_PATH}}"
+
+    def _get_pip_install_commands(self, requirements_path: pathlib.Path) -> list[str]:
+        """Get the commands for installing with pip.
+
+        This only installs the dependencies from requirements, unlike the upstream version.
```
tigarmo, line 52: how come?

charmcraft/parts/plugins/_python.py
```diff
@@ -0,0 +1,100 @@
...
+
+from craft_parts.plugins import python_plugin
+from overrides import override
+
+from charmcraft import utils
+
+
+class PythonPluginProperties(python_plugin.PythonPluginProperties, frozen=True):
+
+    python_requirements: list[str] = ["requirements.txt"]  # By default use requirements.txt
```
tigarmo, line 29: seems like an awkward default, what if there isn't a 'requirements.txt' file?

tigarmo (approved):

charmcraft/parts/plugins/_poetry.py
lengau: Because charms are not Python packages, so `pip install .` won't work.

charmcraft/parts/plugins/_python.py
lengau: If there's not a requirements.txt file and the user uses the default it'll fail with that error.

The reason I chose this is that most existing charms (as well as the templates) use a `requirements.txt` file. The ones that don't are using poetry, so we'd want to move them to the poetry module instead. However, this is probably worth discussing in video call today since I don't have a strong reason to do it, so making this change from the default may be the wrong choice.

lengau: OK based on direct chat I'm fully convinced that we should match `craft-parts` rather than the `charm` plugin here. Changing...

charmcraft/parts/plugins/_poetry.py
lengau: I've made the relevant docstrings both here and in the python module more explicit to explain this.

tigarmo (approved):

mr-cal (approved):

PR #1868, opened 2024-08-30
feat: fastapi-framework init profile and extension

javierdelapuente:
This PR implements the spec ISD160 - 12-Factor FastAPI Support" for charmcraft.

For that, a new init profile called "fastapi-framework" and a new fastapi extension, called "fastapi-framework" are needed. These additions are similar to the flask and django init profiles and extensions.

For the fastapi profile, the base 24.04 is used.

PR #1905, opened 2024-09-17
Add rabbitmq to init templates

javierdelapuente:
RabbitMQ is already in paas-app-charmer (see https://github.com/canonical/paas-app-charmer/pull/39).

Add RabbitMQ requires to init templates (commented out) so it is easier for the user to be able to uncomment it if needed.


lengau (approved):
Looks good to me! Thanks for also putting this in #1868 so we don't have to update that.

PR #1907, opened 2024-09-17
chore(cleanup): remove unused code

lengau:
The cmdbase module was only being used for a constant that said "json", and the registry module has been replaced with skopeo and the docker library.


mr-cal (approved):

mattculler (approved):
-1857, love it

PR #1915, opened 2024-09-23
ci: add security scans

lengau:
Adds two security scanning items:

1. A local dependencies security scan using trivy and osv
2. A (currently disabled) remote security scan using the Canonical security scanning API

Snap build error is fixed in #1916

This is the applications version. Compare to the libraries version at https://github.com/canonical/craft-platforms/pull/52

Once approval is provided here, I will upstream to Starbase and reproduce on our other applications.

NOTE: The Starbase version will be commented out below the libraries version, with a note that applications need to switch which version is commented.


tests/spread/dependencies/strict-dependencies-errors/extra-binary-packages/requirements.txt
lengau, line 1: This is to handle the fact that ops < 2.15.0 has a known CVE

.github/workflows/local-security-scan.yaml
```diff
@@ -0,0 +1,50 @@
+name: Local security scan
+on:
+  pull_request:
+    paths:
+      - "requiremests*.txt"
```
tigarmo, line 5:
```suggestion
      - "requirements*.txt"
```

mr-cal (approved):

.github/workflows/local-security-scan.yaml
lengau: Thanks!

lengau:
Closing this as these are going to go into https://github.com/canonical/starflow. I'll open a separate one that uses the starflow version.

PR #1918, opened 2024-09-24
feat(pack): automatically install non-matching charmlibs

lengau:
If `charmcraft.yaml` specifies charmlibs, pack will now automatically upgrade libraries on disk that don't match the specifications in the yaml file.


sergiusens (changes requested):
I copied verbatim from the spec, which believe we are not following with the change as described

---
The library fetching code can now always be run as part of charmcraft pack. There are three possible options during the pack process:

- The required libraries are not in the source tree, and are fetched as part of the packing process  
- There exist libraries in the source tree that satisfy the requirements  
- There exist libraries in the source tree that conflict with charmcraft.yaml requirements (or are not specified at all)

It is important to note that while charmcraft pack will now be looking at the store, the charmcraft pack should **never** mutate the libraries on the filesystem, except in the first case. The interaction with the store during the pack process is only to provide warnings or information about the libraries in use (such as the availability of updates). The inability to reach the store should not make the pack process fail (except in the first case).

In the second and third cases, the fetch code is still run \- if there are newer versions of the library available in the store, or the libraries do not match the constraints specified, the user should be notified as part of the packing process, but it should only be a warning/info notice, and the pack process should carry on using the in-tree libraries.

charmcraft/services/charmlibs.py
```diff
@@ -0,0 +1,95 @@
...
+        app: craft_application.AppMetadata,
+        services: craft_application.ServiceFactory,
+        *,
+        project: models.CharmcraftProject,
+        project_dir: pathlib.Path,
+    ) -> None:
+        super().__init__(app, services, project=project)
+        self._project_dir = project_dir
+
+    def is_lib_downloaded(
```
sergiusens, line 43: is the `_lib` superfluous in these three methods?

At the end of the day you will be doing `self._services.charm_libs.[get_version,is_downloaded,write]`

charmcraft/application/commands/lifecycle.py
```diff
@@ -173,13 +173,46 @@ def run_managed(self, parsed_args: argparse.Namespace) -> bool:
...
         return super().run_managed(parsed_args)
 
+    def _update_charm_libs(self) -> None:
+        """Update charm libs attached to the project."""
+        craft_cli.emit.progress("Checking that charmlibs match 'charmcraft.yaml' values")
+        project = cast(models.CharmcraftProject, self._services.project)
+        libs_svc = cast(services.CharmLibsService, self._services.charm_libs)
+        installable_libs: list[models.CharmLib] = []
+        for lib in project.charm_libs:
+            charm_name, _, lib_name = lib.lib.partition(".")
```
sergiusens, line 186: what is the library? `<charm-name>.<lib-name>` or just `<lib-name>`, `project.charm_libs` implies the former, `get_lib_version` the latter. I would probably encapsulate the details of splitting (as it seems technical) to the service itself

charmcraft/application/commands/lifecycle.py
```diff
@@ -173,13 +173,46 @@ def run_managed(self, parsed_args: argparse.Namespace) -> bool:
...
+            if not libs_svc.get_lib_version(charm_name=charm_name, lib_name=lib_name):
+                installable_libs.append(lib)
+        if installable_libs:
+            store = cast(services.StoreService, self._services.store)
+            libraries_md = store.get_libraries_metadata(installable_libs)
```
sergiusens, line 191: note: ensure auth is not required

charmcraft/services/charmlibs.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def write_lib(self, library: Library) -> None:
+        """Write the given library to disk.
+
+        :param library: A store library object with valid content.
+        """
+        if library.content is None:
+            # This should be considered an internal error.
+            raise ValueError("Library has no content.")
+        lib_path = self._project_dir / utils.get_lib_path(
```
sergiusens, line 91: the library needs to be fetched, and stored in the lifecycle directories, not the project. If not the library will persist (which we have fetch-libs for).

sergiusens (changes requested):
Looks good, have some minor comments and perhaps an expectation mismatch on fetching on pack (I thought it would be ephemeral and not persisted if I ran `charmcraft clean`)

charmcraft/application/commands/lifecycle.py
lengau: The convention (easiest shown in the [LibData class](https://github.com/canonical/charmcraft/blob/main/charmcraft/utils/charmlibs.py#L33-L44)) doesn't really have a name for what `<charm-name>.<lib-name>` is - `full_name` is being used for `charms.<charm-name>.v<api-version>.<lib-name>`.

charmcraft/services/charmlibs.py
lengau: (Marking as resolved from external conversation. The spec is unclear, but it was determined that we do need to mutate the project directory here.)

charmcraft/services/charmlibs.py
lengau: Agreed! One other change - I replaced `get_lib_version` with `get_local_version` to make it more obvious that this is checking only locally and not making any requests to the store.

charmcraft/application/commands/lifecycle.py
lengau: This was already done in the spread test, but I've ensured that `CHARMCRAFT_AUTH` is unset in the spread test now with a comment as to why so we don't accidentally delete it.

charmcraft/services/charmlibs.py
```diff
@@ -0,0 +1,95 @@
...
+        :param lib_name: The name of the lib itself.
+        :param api: The api version of the lib
+        :param patch: If given, the specific patch version of the lib.
+        """
+        lib_path = utils.get_lib_path(charm_name, lib_name, api)
+        if not (self._project_dir / lib_path).exists():
+            return False
+
+        if patch is None:
+            return True
```
mr-cal, line 58: Can you add a comment for this?  I can't figure out why this is needed

Also as the least expensive check, this can be done first.

docs/reference/changelog.rst
```diff
@@ -77,6 +77,16 @@ Changelog
 
   For a complete list of commits, see the `X.Y.Z`_ release on GitHub.
 
+X.Y.Z (2024-MM-DD)
```
mr-cal, line 80: Asking because this is a new process. Should changes from bumping craft-application in this PR be included now or later?

mr-cal (approved):

charmcraft/services/charmlibs.py
lengau: This is already as high as it can go - it just short-circuits reading the file if the caller doesn't care about the patch version.

docs/reference/changelog.rst
lengau: My thought is that if there were new features that affected charmcraft users we'd include them, but the features are kinda under-the-hood so I didn't include them.

sergiusens (approved):

PR #1925, opened 2024-09-26
tests(spread): install node for changed pyright

lengau:


lengau:
This was a pyright packaging bug and no longer needed

PR #1927, opened 2024-09-26
fix: update the simple profile to avoid Container.can_connect

tonyandrewmeyer:
Using `Container.can_connect` as a guard for Pebble is a poor pattern, because it introduces a race condition. We recommend that charms use the "easier to ask forgiveness than permission" pattern by executing the container calls and catching any errors that occur. This PR updates the `simple` profile to use that pattern.

There are two other small changes:
 * The message in the blocked status was intending to use an f-string, but was just a regular string, so would have always shown the literal text "{log_level}" rather than the value of that variable.
 * `WaitingStatus` should be used when waiting on an integrated charm, not the local charm. `MaintenanceStatus` is the correct class to use when waiting on the local charm.


lengau (approved):
Always great to see improvements like this. Thanks!

tigarmo (approved):

PR #1933, opened 2024-10-04
feat: use the newer ops.testing classes for the charm unit tests

tonyandrewmeyer:
Updates the `simple`, `machine`, and `kubernetes` profiles to use the ops.testing state transition classes (previously: Scenario) rather than the legacy Harness API.

The content of what each set of tests is testing remains unchanged, it's just using the new API to do the tests.

`ops[testing]` is installed via the tox dependencies (but should likely be in an optional dependencies group in the future, once CC005 is finalised). I've included it in the `static` dependencies as well, because as the charm gains additional tests those should also be statically checked and that will require the `ops[testing]` classes.


lengau (approved):
Looks great. Thanks!

mr-cal (approved):

PR #1939, opened 2024-10-07
feat(docs): add craft-parts common documentation

lengau:
This brings in the common documentation sets from craft-parts and adds some of the pages. It mostly exists to enable my next PR, where I'll document the `python` and `poetry` plugins.

### DIFF
--- a/docs/reference/parts.rst
+++ b/docs/reference/parts.rst
@@ -1,0 +1,14 @@
+.. _parts:
+
+Parts
+*****
+
+Parts, powered by :external+craft-parts:ref:`craft-parts <reference>`, power the build
+system that charmcraft uses.
+
+.. toctree::
+   :maxdepth: 1
+
+   /common/craft-parts/reference/part_properties
+   /common/craft-parts/reference/parts_steps
+   /common/craft-parts/reference/step_execution_environment
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -58,6 +58,7 @@ extensions = [
 
 extensions.extend(
     [
+        "sphinx.ext.ifconfig",
         "sphinx.ext.intersphinx",
         "sphinx.ext.viewcode",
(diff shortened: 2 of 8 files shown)


docs/reference/parts.rst
```diff
@@ -0,0 +1,14 @@
+.. _parts:
```
medubelko, line 1: I ran into an oddity. The first link in [Filesets](https://canonical-charmcraft--1939.com.readthedocs.build/en/1939/common/craft-parts/explanation/filesets/) works but there's a missing second link that should be pointing to `lifecycle.rst` in Craft Parts. I believe the first link works because the reference is colliding with reference on this line here.

The missing link was causing a local build error for me. By adding `.. _lifecycle:` to a random file, the build succeeded.

docs/reference/parts.rst
```diff
@@ -0,0 +1,14 @@
...
+
+Parts
+*****
+
+Parts, powered by :external+craft-parts:ref:`craft-parts <reference>`, power the build
```
medubelko, line 6: I'm not sure whether this is new behaviour, but some of the `:ref:` and `:external+` directives are appending a fragment to the URL when none was specified. In this case the link is:

```
https://canonical-craft-parts.readthedocs-hosted.com/en/latest/reference/index.html#reference
```

medubelko (changes requested):
I ran into two possible bugs in the implementation. One of them is breaking the build on my end. Let me know what you think.

docs/reference/parts.rst
lengau: It wasn't erroring out for me, but I've added some brief lifecycle docs that mostly just call out the craft-parts docs.

medubelko (approved):
Build is working on my end now, thanks!

tigarmo (approved):
🚀

PR #1941, opened 2024-10-07
fix(templates): replace flake8 and pydocstyle with ruff

### LINKED ISSUE #1461: Replace flake8 and pydocstyle template tests with ruff
### What needs to get done

We currently have tests that run each profile for `charmcraft init` and then run them through pydocstyle etc.

Replace these with tests that run `ruff` on them the way the linter says to do so.

### Why it needs to get done

We moved everything to ruff but forgot these tests. It could become an issue if the tools diverge.

lengau:
Fixes #1461


mr-cal (approved):

charmcraft/templates/init-django-framework/pyproject.toml.j2
```diff
@@ -0,0 +1,41 @@
...
+    "D204",
+    "D213",
+    "D215",
+    "D400",
+    "D404",
+    "D406",
+    "D407",
+    "D408",
+    "D409",
+    "D413",
```
sergiusens, line 29: maybe add a comment for what each these are as this is a template

requirements-dev.txt
```diff
@@ -18,7 +18,6 @@ cryptography==43.0.1
 dill==0.3.9
 distro==1.9.0
 docker==7.1.0
-flake8==7.1.1
```
dariuszd21, line 21: Those are `flake8` install_requirements, are other tools still rely on them if you delete it ?
```
    mccabe>=0.7.0,<0.8.0
    pycodestyle>=2.12.0,<2.13.0
    pyflakes>=3.2.0,<3.3.0
```

charmcraft/templates/init-fastapi-framework/tox.ini.j2
```diff
@@ -0,0 +1,84 @@
+# Copyright {{ year }} {{ author }}
+# See LICENSE file for licensing details.
```
dariuszd21, line 2: Shouldn't be worth to move this `tox` configuration to the `pyproject.toml` as you require `4.0.0` minimal version anyway? 

https://tox.wiki/en/latest/config.html#pyproject-toml-native

dariuszd21 (approved):

charmcraft/templates/init-django-framework/pyproject.toml.j2
lengau: These came across from the other templates, so I'd have to research them and probably update the other templates too. I've created https://github.com/canonical/charmcraft/issues/1964 for this.

requirements-dev.txt
lengau: Not entirely sure, but once I merge in some uv-based changes those will be automatically recalculated.

charmcraft/templates/init-fastapi-framework/tox.ini.j2
lengau: Possibly, but out of scope as this is just copying the tox files from the other templates.

sergiusens (approved):

PR #1949, opened 2024-10-08
chore: add oracular as a valid base

lengau:


charmcraft/const.py
```diff
@@ -73,7 +74,6 @@
         BaseName("ubuntu", "18.04"),
         BaseName("ubuntu", "20.04"),
         BaseName("ubuntu", "22.04"),
-        BaseName("ubuntu", "23.10"),
         BaseName("ubuntu", "24.04"),
```
mr-cal, line 77: Are you intentionally not adding 24.10 as a supported base here?

lengau: Yeah - this should use devel build-base for now.

tigarmo (approved):

mr-cal (approved):

Issue #1810, opened 2024-08-14
Remove Windows support


lengau:
### What needs to get done

- Remove Windows tests and other support items.
- Remove `tox-gh` and `windows` ignoreDeps items from renovate config.

### Why it needs to get done

Windows support is only partially tested and it's unclear how well it actually works.

jnsgruk:
Pretty sure that Juju deprecated Windows charms when they released 3.x, so there shouldn't be any issues.

PR #1951, opened 2024-10-09
feat(linter): add 'pip check' linter

lengau:
This linter runs 'pip check' on the venv of the resulting charm and warns if the virtual environment is inconsistent.


pyproject.toml
```diff
@@ -24,6 +24,11 @@ dependencies = [
     "requests-toolbelt",
     "snap-helpers",
     "tabulate",
+    # Needed until requests-unixsocket supports urllib3 v2
+    # https://github.com/msabramo/requests-unixsocket/pull/69
+    # When updating, remove the urllib3 constraint from renovate config.
+    "urllib3<2.0",
```
lengau, line 30:
```suggestion
```

I think this was just a merge conflict that got auto-resolved the wrong way

tests/unit/test_linters.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+def test_pip_invalid_venv(fake_path: pathlib.Path):
+    (fake_path / "venv").mkdir()
+    lint = linters.PipCheck()
+    assert lint.run(fake_path) == LintResult.NONAPPLICABLE
+    assert lint.text == "Python venv is not valid."
+
+
+@pytest.mark.skipif(sys.platform == "win32", reason="Windows not [yet] supported.")
```
mr-cal, line 40: Is this because a fixture or preliminary step fails?  Or should there be a test that mocks `os.platform` to exercise these lines?

```python
        if sys.platform == "win32":
            self.text = "Linter does not work on Windows."
            return self.Result.NONAPPLICABLE
```

charmcraft/linters.py
```diff
@@ -689,6 +691,56 @@ def run(self, basedir: pathlib.Path) -> str:
...
+        python_exe = venv_dir / "bin" / "python"
+        delete_parent = False
+        if not python_exe.parent.exists():
+            delete_parent = True
+            python_exe.parent.mkdir()
+        if not python_exe.exists():
+            delete_python_exe = True
+            python_exe.symlink_to(sys.executable)
+        else:
+            delete_python_exe = False
```
mr-cal, line 722: If I'm understanding this correctly, these are temporary changes that are undone?

I think there should be some unit tests that ensure these changes are undone.

charmcraft/linters.py
```diff
@@ -689,6 +691,56 @@ def run(self, basedir: pathlib.Path) -> str:
...
+
+        pip_cmd = [sys.executable, "-m", "pip", "--python", str(python_exe), "check"]
+        check = subprocess.run(
+            pip_cmd,
+            text=True,
+            capture_output=True,
+            check=False,
+        )
```
mr-cal, line 730: If this fails, then the venv changes above won't be undone.  Should this be in a `try/finally` block?

mr-cal (commented):
Looks good so far!

Some of these open-ended questions may need @lengau's input, so don't feel like you must have an answer :)

tests/unit/test_linters.py
bepri: I think this might actually be removable. This was a skip I originally put in when I misdiagnosed an issue with the test passing on Windows, but the problem ended up being something else. Considering #1810 though, is this something I should fix?

mr-cal: Yeah, I guess it doesn't really matter if it's tested or not since windows is on the chopping block

charmcraft/linters.py
```diff
@@ -689,6 +691,63 @@ def run(self, basedir: pathlib.Path) -> str:
...
+
+        pip_cmd = [sys.executable, "-m", "pip", "--python", str(python_exe), "check"]
+        try:
+            check = subprocess.run(
+                pip_cmd,
+                text=True,
+                capture_output=True,
+                check=False,
+            )
+            if check.returncode == 0:
```
mr-cal, line 732: pedantic: 0 is unlikely to be misunderstood for return codes, but consts are cleaner.  Ruff complains about this for most return codes anyways.

```suggestion
            if check.returncode == os.EX_OK:
```

mr-cal (approved):
Nice work!

If I ever leave a pedantic/opinion comment, it's your call to take or leave it.

charmcraft/linters.py
lengau: Good catch!

tigarmo (approved):
nice :)

PR #1967, opened 2024-10-17
docs: add howto guides for switching from the charm plugin

lengau:
This adds howto guides to switch from the charm plugin to the python and poetry plugins.

### DIFF
--- a/docs/howto/charm-to-poetry.rst
+++ b/docs/howto/charm-to-poetry.rst
@@ -28,7 +28,8 @@ can be created as follows:
 Select compatible versions of ``pip`` and ``poetry``
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-The ``poetry`` plugin requires at least pip 22.3, released in October 2022. If the
+The Poetry plugin requires at least `pip 22.3
+<https://pypi.org/project/pip/22.3>`_, released in October 2022. If the
 charm's base uses an older version of pip, a newer version can be installed in the
 build environment using a dependency part. Likewise, a charm may require a newer
 version of Poetry than is available in the distribution's repositories. The following
@@ -83,7 +84,7 @@ from all loaded charm libs.
 Include extra files
 -------------------
 
-A ``poetry`` plugin only includes the contents of the ``src`` and ``lib`` directories
+A Poetry plugin only includes the contents of the ``src`` and ``lib`` directories
 as well as the generated virtual environment. If other files were previously included
 from the main directory, they can be included again using the
 :ref:`craft_parts_dump_plugin`:
--- a/docs/howto/charm-to-poetry.rst
+++ b/docs/howto/charm-to-poetry.rst
@@ -25,13 +25,13 @@ can be created as follows:
         plugin: poetry
         source: .
 
-Ensure a recent enough version of ``pip`` and ``poetry``
+Select compatible versions of ``pip`` and ``poetry``
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 The ``poetry`` plugin requires at least pip 22.3, released in October 2022. If the
 charm's base uses an older version of pip, a newer version can be installed in the
 build environment using a dependency part. Likewise, a charm may require a newer
-version of poetry than is available in the distribution's repositories. The following
+version of Poetry than is available in the distribution's repositories. The following
 ``parts`` section can be used in place of the section above to upgrade pip and Poetry
 for charms that build on Ubuntu 22.04 or earlier:
 
@@ -61,7 +61,7 @@ when creating the virtual environment.
 
 .. note::
     This is useful and encouraged, though not mandatory, for keeping track of
-    library dependencies (see below). For an example, see
+    library dependencies, as covered in the next section. For an example, see
     `postgresql-operator`_.
 
 Include charm library dependencies
--- a/docs/howto/charm-to-poetry.rst
+++ b/docs/howto/charm-to-poetry.rst
@@ -1,0 +1,107 @@
+.. _howto-migrate-to-poetry:
+
+Migrate from the ``charm`` plugin to Poetry
+===========================================
+
+Many charms use `Poetry`_ to manage their Python projects. For these charms, Charmcraft
+has a :ref:`craft_parts_poetry_plugin`. Migrating from the charm plugin provides some
+benefits, such as no longer having to maintain a ``requirements.txt`` file. If the
+charm to be migrated does not currently use poetry, refer to the
+`Poetry documentation <https://python-poetry.org/docs/basic-usage/>`_ for instructions
+on how to use poetry for a Python project.
+
+Update ``charmcraft.yaml``
+--------------------------
+
+The first step is to update ``charmcraft.yaml`` to include the correct parts definition.
+Depending on the history of a specific charm, it may not have an explicitly-included
+``parts`` section determining how to build the charm. In this case, a ``parts`` section
+can be created as follows:
+
+.. code-block:: yaml
+
+    parts:
+      my-charm:  # This can be named anything you want
+        plugin: poetry
+        source: .
+
+Ensure a recent enough version of ``pip`` and ``poetry``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The ``poetry`` plugin requires at least pip 22.3, released in October 2022. If the
+charm's base uses an older version of pip, a newer version can be installed in the
+build environment using a dependency part. Likewise, a charm may require a newer
+version of poetry than is available in the distribution's repositories. The following
+``parts`` section can be used in place of the section above to upgrade pip and Poetry
+for charms that build on Ubuntu 22.04 or earlier:
+
+.. code-block:: yaml
+    :emphasize-lines: 2-9,11
+
+    parts:
+      poetry-deps:
+        plugin: nil
+        build-packages:
+          - curl
+        override-build: |
+          /usr/bin/python3 -m pip install pip==24.2
+          curl -sSL https://install.python-poetry.org | python3 -
+          ln -sf $HOME/.local/bin/poetry /usr/local/bin/poetry
+      my-charm:  # This can be named anything you want
+        after: [poetry-deps]
+        plugin: poetry
+        source: .
+
+Add optional dependency groups
+------------------------------
+
+If the charm has optional `dependency groups`_ that should be included when creating
+the virtual environment, the ``poetry-with`` key can be used to include those groups
+when creating the virtual environment.
+
+.. note::
+    This is useful and encouraged, though not mandatory, for keeping track of
+    library dependencies (see below). For an example, see
+    `postgresql-operator`_.
+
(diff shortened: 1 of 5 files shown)


mattculler (approved):
Very clear!

jnsgruk:
Gave this a go here: https://github.com/jnsgruk/zinc-k8s-operator/pull/288

Might be worth including a section about using `override-pull` as a place where people can generate `requirements'txt` if they need to.

I guess they could also use `override-build` and just place their generation code before `craftctl default`.

jnsgruk:
Heh. The python plugin doesn't seem to have included any `dispatch` file...

```
ubuntu@ubuntu-noble-2e39:~/zinc_charm_unzipped$ ls -la
total 39480
drwxrwxr-x 5 ubuntu ubuntu     4096 Oct 18 10:30 .
drwxr-xr-x 9 ubuntu ubuntu     4096 Oct 18 10:30 ..
drwxrwxr-x 2 ubuntu ubuntu     4096 Oct 18 10:30 bin
drwxrwxr-x 3 ubuntu ubuntu     4096 Oct 18 10:30 lib
drwxrwxr-x 3 ubuntu ubuntu     4096 Oct 18 10:30 lib64
-rw-r--r-- 1 ubuntu ubuntu      255 Oct 18 10:21 manifest.yaml
-rw-r--r-- 1 ubuntu ubuntu     1402 Oct 18 10:21 metadata.yaml
-rw-r--r-- 1 ubuntu ubuntu      169 Oct 18 10:21 pyvenv.cfg
```

lengau:
Thanks @jnsgruk! Looks like I forgot to test with `python-keep-bins: false` (the default). I've made [a fix PR](https://github.com/canonical/charmcraft/pull/1972), but as a temporary workaround you should also be able to set `python-keep-bins: true` in the part definition.

lengau:
@jnsgruk re: the `override-pull` - while people could do that, I'd prefer not to document it here as I'd like to encourage people to check in some sort of lock file. For the `uv` case, @bepri is heading up our addition of uv plugins for 25.04

docs/howto/charm-to-python.rst
```diff
@@ -0,0 +1,126 @@
...
+      my-charm:  # This can be named anything you want
+        plugin: python
+        source: .
+        python-requirements:
+          - requirements.txt  # Or whatever your requirements file is called.
+
+Ensure a recent enough version of ``pip``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The ``python`` plugin requires at least pip 22.3, released in October 2022. If the
```
medubelko, line 30: > pip 22.3, released in October 2022

Rather than mentioning the date, can we link to https://pypi.org/project/pip/22.3 instead?

docs/howto/charm-to-poetry.rst
```diff
@@ -0,0 +1,107 @@
+.. _howto-migrate-to-poetry:
+
+Migrate from the ``charm`` plugin to Poetry
```
medubelko, line 3: For consistency with existing docs and docs best practices, we should use the form `Charm plugin` and `Python plugin`.

docs/howto/charm-to-poetry.rst
```diff
@@ -0,0 +1,107 @@
...
+
+The first step is to update ``charmcraft.yaml`` to include the correct parts definition.
+Depending on the history of a specific charm, it may not have an explicitly-included
+``parts`` section determining how to build the charm. In this case, a ``parts`` section
+can be created as follows:
+
+.. code-block:: yaml
+
+    parts:
+      my-charm:  # This can be named anything you want
```
medubelko, line 24: A morsel for thought: we generally [discourage the use of comments](https://docs.ubuntu.com/styleguide/en/#code-examples-in-documentation) in bash, but I'm not sure where configs fit in, since we encourage users to copy them. As a general principle it's better to use the doc to explain things.

docs/howto/charm-to-poetry.rst
```diff
@@ -0,0 +1,107 @@
...
+        plugin: poetry
+        source: .
+
+Ensure a recent enough version of ``pip`` and ``poetry``
```
medubelko, line 28:
```suggestion
Select compatible versions of ``pip`` and ``poetry``
```

docs/howto/charm-to-poetry.rst
```diff
@@ -0,0 +1,107 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The ``poetry`` plugin requires at least pip 22.3, released in October 2022. If the
+charm's base uses an older version of pip, a newer version can be installed in the
+build environment using a dependency part. Likewise, a charm may require a newer
+version of poetry than is available in the distribution's repositories. The following
```
medubelko, line 34:
```suggestion
version of Poetry than is available in the distribution's repositories. The following
```

docs/howto/charm-to-poetry.rst
```diff
@@ -0,0 +1,107 @@
...
+      my-charm:  # This can be named anything you want
+        plugin: poetry
+        source: .
+
+Ensure a recent enough version of ``pip`` and ``poetry``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The ``poetry`` plugin requires at least pip 22.3, released in October 2022. If the
+charm's base uses an older version of pip, a newer version can be installed in the
+build environment using a dependency part. Likewise, a charm may require a newer
```
medubelko, line 33: > newer

If I'm reading the code right, wouldn't it grab the latest?

docs/howto/charm-to-poetry.rst
```diff
@@ -0,0 +1,107 @@
...
+Add optional dependency groups
+------------------------------
+
+If the charm has optional `dependency groups`_ that should be included when creating
+the virtual environment, the ``poetry-with`` key can be used to include those groups
+when creating the virtual environment.
+
+.. note::
+    This is useful and encouraged, though not mandatory, for keeping track of
+    library dependencies (see below). For an example, see
```
medubelko, line 64:
```suggestion
    library dependencies, as covered in the next section. For an example, see
```

docs/howto/charm-to-poetry.rst
```diff
@@ -0,0 +1,107 @@
...
+    `postgresql-operator`_.
+
+Include charm library dependencies
+----------------------------------
+
+Unlike the ``charm`` plugin, the ``poetry`` plugin does not install the dependencies
+for included charmlibs. If any of the charm libraries used have PYDEPS, these will
+need to be added to the charm's dependencies, potentially as their own
+`dependency group <dependency groups>`_.
```
medubelko, line 73: Unspecified link.

docs/howto/charm-to-python.rst
```diff
@@ -0,0 +1,126 @@
+.. _howto-migrate-to-python:
+
+Migrate from the ``charm`` plugin to the ``python`` plugin
+==========================================================
+
+The ``python`` Charmcraft plugin offers a faster, stricter means of packing an operator
```
medubelko, line 6: > Charmcraft plugin

Do we need to be this specific here? "Python Charmcraft plugin" tripped me up as a phrase.

docs/howto/charm-to-python.rst
```diff
@@ -0,0 +1,126 @@
...
+.. code-block:: yaml
+
+    parts:
+      my-charm:  # This can be named anything you want
+        plugin: python
+        source: .
+        python-requirements:
+          - requirements.txt  # Or whatever your requirements file is called.
+
+Ensure a recent enough version of ``pip``
```
medubelko, line 27:
```suggestion
Select a compatible version of ``pip``
```

docs/howto/charm-to-python.rst
```diff
@@ -0,0 +1,126 @@
+.. _howto-migrate-to-python:
+
+Migrate from the ``charm`` plugin to the ``python`` plugin
+==========================================================
+
+The ``python`` Charmcraft plugin offers a faster, stricter means of packing an operator
+charm with a virtual environment. This guide shows how to migrate from a charm using
```
medubelko, line 7: "packing with" is slightly ambiguous, so we might want to take special care with how we express this. If I'm understanding this correctly, could we say "packing a virtual environment inside an operator charm"?

medubelko (changes requested):
Thanks for putting these together!

docs/howto/charm-to-python.rst
lengau: We can link there as well, but the date is also useful because it gives them a good idea as to whether their base distribution is going to have a new enough pip out of the box. For example, CentOS 7 and Ubuntu Jammy won't because they were all released before then (so they'll need to follow this), but Ubuntu Noble does. I don't want to list specific bases because that just gives us a treadmill of bases to update.

docs/howto/charm-to-poetry.rst
lengau: We use comments pretty heavily in our [config file templates](https://github.com/canonical/charmcraft/blob/main/charmcraft/templates/init-simple/charmcraft.yaml.j2). In this case, the goal of the comment is to counter something that's a fairly common misconception amongst charm authors (that the part name needs to be named the plugin name or the actual charm name) without having an aside about it, since that's out of scope for the document.

docs/howto/charm-to-poetry.rst
lengau: Good catch - fixed!

docs/howto/charm-to-poetry.rst
```diff
@@ -0,0 +1,107 @@
...
+
+    parts:
+      my-charm:  # This can be named anything you want
+        plugin: poetry
+        source: .
+
+Select compatible versions of ``pip`` and ``poetry``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The ``poetry`` plugin requires at least pip 22.3, released in October 2022. If the
```
medubelko, line 31:
```suggestion
The Poetry plugin requires at least `pip 22.3
<https://pypi.org/project/pip/22.3>`_, released in October 2022. If the
```

docs/howto/charm-to-poetry.rst
```diff
@@ -0,0 +1,107 @@
...
+
+    find lib -name "*.py" -exec awk '/PYDEPS = \[/,/\]/' {} +
+
+If run from the base directory of a charm, this will show all the PYDEPS declarations
+from all loaded charm libs.
+
+Include extra files
+-------------------
+
+A ``poetry`` plugin only includes the contents of the ``src`` and ``lib`` directories
```
medubelko, line 86:
```suggestion
A Poetry plugin only includes the contents of the ``src`` and ``lib`` directories
```

docs/howto/charm-to-python.rst
```diff
@@ -0,0 +1,126 @@
+.. _howto-migrate-to-python:
+
+Migrate from the Charm plugin to the Python plugin
+==================================================
+
+The Python Charmcraft plugin offers a faster, stricter means of packing an operator
```
medubelko, line 6:
```suggestion
The Python plugin in Charmcraft offers a faster, stricter means of packing an operator
```

medubelko (changes requested):

medubelko (approved):

PR #1970, opened 2024-10-18
fix: fastapi-framework import typo

javierdelapuente:
Fix bug in fastapi-framework. The import was incorrect, as it is `paas-charm` instead.


jdkandersson (approved):

javierdelapuente:
@lengau Hi Alex, can you check this PR, as we have a bug in fastapi? Thanks!

tigarmo (approved):

Issue #1973, opened 2024-10-22
`charmcraft list-lib` does not work without a `metadata.yaml` file


DanielArndt:
### Bug Description

The command `charmcraft list-lib` seems to expect the charm name to be in a `metadata.yaml` file, but newer charms do not have this file, as everything is in `charmcraft.yaml`.

### To Reproduce

1. Go to a directory which contains a charm with only a `charmcraft.yaml` file (no `metadata.yaml`)
2. Run `charmcraft list-lib`

### Environment

Latest charmcraft (3a094837)

### charmcraft.yaml

```yaml
# This file configures Charmcraft.
# See https://juju.is/docs/sdk/charmcraft-config for guidance.

name: listlibs
type: charm
title: Charm Template
summary: A very short one-line summary of the charm.
description: |
  A single sentence that says what the charm is, concisely and memorably.

  A paragraph of one to three short sentences, that describe what the charm does.

  A third paragraph that explains what need the charm meets.

  Finally, a paragraph that describes whom the charm is useful for.
bases:
  - build-on:
    - name: ubuntu
      channel: "22.04"
    run-on:
    - name: ubuntu
      channel: "22.04"
config:
  options:
    log-level:
      description: |
        Configures the log level of gunicorn.

        Acceptable values are: "info", "debug", "warning", "error" and "critical"
      default: "info"
      type: string
containers:
  httpbin:
    resource: httpbin-image
resources:
  httpbin-image:
    type: oci-image
    description: OCI image for httpbin
    upstream-source: kennethreitz/httpbin
```

### Relevant log output

```shell
Can't access name in 'metadata.yaml' file. The 'list-lib' command must either be executed from a valid project directory, or specify a charm name using the --charm-name option.
```

DanielArndt:
Fixed by #1974

PR #1974, opened 2024-10-22
Check `charmcraft.yaml` for charm name

DanielArndt:
I don't know if this is the best way to solve the problem, but it's an attempt to solve #1973 

Tested locally and it seems to solve the problem.


lengau (approved):
Thanks! These commands need some bigger work, but this looks like a good fix for the bug so I'm not going to block on that.

mr-cal (approved):

DanielArndt:
Force pushed to remove the merge commits, as they defaulted to my personal e-mail which wasn't passing the CLA.

Hopefully this fixes it.

No code changes.

DanielArndt:
@lengau I don't think these status checks are related to my PR. Any way I can help move this forward or do those need to be fixed first?

PR #1975, opened 2024-10-23
Create scenario tests for init templates

javierdelapuente:
To avoid issues like https://github.com/canonical/charmcraft/pull/1970, a scenario test has been added to the spread test for the `{go,flask,fastapi,django}-framework` init profiles.

This test is a [scenario test (unit test)](https://juju.is/docs/sdk/write-a-functional-test-for-a-charm-with-scenario), that uses the `paas-charm` library and runs a pebble ready event. The real goal of the test is that there are no syntax errors or wrong imports in the init template.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -144,6 +144,7 @@ xfail_strict = true
 markers = [
     "slow: marks tests as slow (deselect with '-m \"not slow\"')",
 ]
+addopts = "--ignore=tests/spread"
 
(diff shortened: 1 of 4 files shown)


javierdelapuente:
@lengau Can you have a look at this PR? Did you have a similar idea in mind for these tests? Thanks!

lengau (approved):

bepri (approved):

PR #1978, opened 2024-10-31
fix: default support to include 24.04

jameinel:
The existing default templates generate a charm that doesn't have any tests available on Noble systems. I didn't touch the definition for github, as I'm not sure whether there are limitations of what can be run there.


lengau (approved):
Thanks!

bepri (approved):

lengau:
Rebased on main (which should fix the test failures), set to auto-merge.

PR #1984, opened 2024-11-08
fix(snap): check if LXD is actually installed

### LINKED ISSUE #1982: Snap configure hook timing out in CI
### Bug Description

The `configure` hook will often time out in CI, as reported by @jnsgruk 

### To Reproduce

See:
- https://github.com/jnsgruk/zinc-k8s-operator/actions/runs/11626972684/job/32379535656?pr=297
- https://github.com/jnsgruk/concierge/actions/runs/11627000431/job/32379648837#step:6:162

### Environment

GitHub CI (see above projects

### charmcraft.yaml

```yaml
n/a
```

### Relevant log output

```shell
snap install charmcraft --channel latest/stable --classic
2024-11-01T09:08:42Z INFO Waiting for automatic snapd restart...
error: cannot perform the following tasks:
- Run configure hook of "charmcraft" snap if present (run hook "configure": <exceeded maximum runtime of 5m0s>)
```

tigarmo:
On recent Ubuntu systems, "lxc" might be "/usr/sbin/lxc", which is provided by the "lxd-installer" package and will install the LXD snap if it's not installed. This installation can then take a long time if the store is having issues.

For the purposes of the configure and remove hooks we *don't* want to install LXD just to check that it has no stale images, so update the hooks to do some early detection and bail out if LXD is not installed.

Fixes #1982

### DIFF
--- a/snap/hooks/configure
+++ b/snap/hooks/configure
@@ -17,13 +17,11 @@ remote image version: core22 -----------------------------------------------
 """
 
 import json
-import platform
 import re
 import subprocess
 import sys
 
 from datetime import datetime
-from pathlib import Path
 
 from charmcraft import snap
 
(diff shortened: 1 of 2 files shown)


mr-cal (approved):
Nice and easy

mattculler:
> @mattculler, you made need to grab this for CRAFT-1483

Thanks, I definitely will!

tigarmo:
No idea what's going on with the macos/windows failures

lengau (approved):
LGTM; thanks!

Gonna bypass the failing Windows/mac tests in this case since it's clearly unrelated.

PR #1989, opened 2024-11-14
feat(docs): add information about CRAFT_SHARED_CACHE

lengau:


lengau:
@medubelko and @mr-cal - I'm not sure I like this (which is why it's in draft), but I'd appreciate fix suggestions.

mr-cal (commented):
For my context:
- Is there a use case for setting `CRAFT_SHARED_CACHE` locally?
- Will you need to document other CI performance practices in the future, such as base instance caching?

docs/howto/shared-cache.rst
```diff
@@ -0,0 +1,73 @@
+.. _howto-shared-cache:
+
+Cache intermediate build artefacts
+==================================
+
+Because Charmcraft builds Python packages from source rather than using pre-built
+wheels, the initial builds of charms can take a while. However, future builds of
+the same charm are sped up significantly by reusing the built wheels. When installed
```
medubelko, line 8: For better flow, I think we should split the paragraph at "When".

docs/howto/shared-cache.rst
```diff
@@ -0,0 +1,73 @@
...
+use instead.
+
+This can be especially useful in CI, where you may wish to specify a directory that
+gets cached between CI runs. This is done automatically when using the
+``charmcraft/pack`` action in Canonical's `craft-actions`_ repository for GitHub,
+but it can also be done manually in any CI.
+
+CI examples
+-----------
+
```
medubelko, line 22: We have a rule not to have two headings touching, so ideally we should add a sentence here.

docs/howto/shared-cache.rst
```diff
@@ -0,0 +1,73 @@
...
+          paths:
+            - .charmcraft_cache/
+      variables:
+        CRAFT_SHARED_CACHE: .charmcraft_cache/
+      script:
+        - mkdir -p .charmcraft_cache
+        - snap install charmcraft
+        - charmcraft pack
+
+On GitHub
```
medubelko, line 45: Since we mention GitHub in the preceding paragraph and it has a greater userbase, I think we should put it first.

docs/howto/shared-cache.rst
```diff
@@ -0,0 +1,73 @@
...
+beneficial to change this directory. This can be done by setting the
+``CRAFT_SHARED_CACHE`` environment variable to the path of an existing directory to
+use instead.
+
+This can be especially useful in CI, where you may wish to specify a directory that
+gets cached between CI runs. This is done automatically when using the
+``charmcraft/pack`` action in Canonical's `craft-actions`_ repository for GitHub,
+but it can also be done manually in any CI.
+
+CI examples
```
medubelko, line 20: Since this _can_ be useful for CI, we're suggesting it could be useful elsewhere, too. The heading suggests this. Are there other examples we want to impart at this time? If not, we should axe the heading.

docs/reference/changelog.rst
```diff
@@ -375,3 +375,4 @@ page.
 .. _3.1.2: https://github.com/canonical/charmcraft/releases/tag/3.1.2
 .. _3.2.0: https://github.com/canonical/charmcraft/releases/tag/3.2.0
 .. _3.2.1: https://github.com/canonical/charmcraft/releases/tag/3.2.1
+.. _3.2.2: https://github.com/canonical/charmcraft/releases/tag/3.2.1
```
medubelko, line 378: Adding a blocker to this link so we remember to fix it.

```suggestion
.. _3.2.2: https://github.com/canonical/charmcraft/releases/tag/3.2.2
```

docs/howto/shared-cache.rst
```diff
@@ -0,0 +1,73 @@
+.. _howto-shared-cache:
+
+Cache intermediate build artefacts
+==================================
+
+Because Charmcraft builds Python packages from source rather than using pre-built
+wheels, the initial builds of charms can take a while. However, future builds of
```
medubelko, line 7: > can take a while

To really sell the benefits of this task, we should probably share our opinion and not kneecap ourselves. Really, we're saying:

> initial builds are characteristically slow, _and_ it's best not to repeat them.

docs/howto/shared-cache.rst
```diff
@@ -0,0 +1,73 @@
...
+the same charm are sped up significantly by reusing the built wheels. When installed
+as a snap, Charmcraft automatically caches these wheels in the
+``~/snap/charmcraft/common/cache`` directory. However, in some cases, it may be
+beneficial to change this directory. This can be done by setting the
```
medubelko, line 11: With how-tos we want to prioritise tangible steps, so I think we should slice off "This can be done..." and place it in its own section.

lengau:
Thanks @mr-cal! To answer your questions:

1. Potentially, if you wanted to have different cache directories for different charms.
2. Probably, with https://github.com/canonical/craft-actions/pull/31 being my reference implementation.

medubelko (commented):
Some commentary mostly about the flow and structure.

docs/howto/shared-cache.rst
lengau: It can be useful outside CI, but writing that is out of scope for now.

docs/howto/shared-cache.rst
lengau: It's unclear right now whether we want to make this the preferred way to do it, which is why I'm hedging here.

medubelko: How about:

> However, future builds of the same charm are sped up significantly...

↓

> You can significantly speed up future builds of the same charm...

medubelko: That works, too! Noting here that there's no ban on the word "you" in how-tos. It's perfectly fine to use it.

medubelko (approved):

mr-cal (approved):
Excellent!

Issue #2004, opened 2024-11-25
charm plugin should include rust compiler in build dependencies


james-garner-canonical:
### What needs to get done

Add `"cargo"` to the charm plugin build dependencies for deb based charms (see PR #2005).

I'm not 100% sure what's needed for `yum` and `dnf` systems.

### Why it needs to get done

It's increasingly common that some python packages require a rust compiler at build time. We first noticed this a few months ago in the Juju SDK's ['Write your first Kubernetes charm' tutorial](https://juju.is/docs/sdk/observe-your-charm-with-cos-lite#heading--build-packages). Here we needed to [add](https://github.com/canonical/juju-sdk-tutorial-k8s/pull/42) `cargo` to `build-packages` for the `charm` `part` when adding [cos-lite](https://charmhub.io/cos-lite).

This came up again recently with the 2023 Operator Day tutorial ([video](https://www.youtube.com/watch?v=GkQ3OLJp2Gs), [repo](https://github.com/operatorinc/zinc-k8s-operator)) that most new starters [are asked to do](https://sites.google.com/canonical.com/engineeringonboarding/fundamentals/charms). Here the need for `cargo` comes from adding the [`traefik_k8s.v2.ingress`](https://github.com/canonical/traefik-k8s-operator/blob/main/lib/charms/traefik_k8s/v2/ingress.py) lib, because it includes the popular PyPI package `pydantic` (note: the lib is using `PYDEPS`). We considered another documentation fix here (which would have to include both the tutorial and the `ingress` lib, as well as any other libs using `pydantic`), but it seems like it might be a better idea to just include `cargo` in the charm plugin's build dependencies by default.

dimaqq:
When I played with this I had to include both `rustc` and `cargo`.
https://discourse.charmhub.io/t/charmcraft-3-0-in-the-beta-channel/13469/7?u=dimaqq

In any case, it would be great if rust toolchain was included by default just like the C toolchain.

artivis:
I've just run into this missing rust compiler issue building a charm with no part. Log can be found below.
In short it seems that a force-reinstall on pip tries to build maturin which requires a rust compiler installed and setup (i.e. `snap install rustup --classic && rustup default stable`).

<details>
<summary>Logs</summary>
<br>
<pre>
$ charmcraft pack --verbose --debug
Starting charmcraft, version 3.2.2                                                                                                                                                                                  
[...]

PR #2005, opened 2024-11-25
feat: add "cargo" to deb build dependencies for charm plugin

james-garner-canonical:
This PR adds `"cargo"` to the charm plugin build dependencies for deb based charms.

I don't know if a similar fix is needed for `yum` and `dnf` systems -- review appreciated here.

This will prevent various charm-libs from needing to document that including them requires adding `cargo` to `build-packages` in `charmcraft.yaml`, and prevent charmers from having to do this, since `cargo` will be included be default, and python packages requiring a rust compiler to build (e.g. `pydantic`) will work out of the box.

See issue:
- #2004

### DIFF
--- a/charmcraft/parts/plugins/_charm.py
+++ b/charmcraft/parts/plugins/_charm.py
@@ -194,6 +194,7 @@ class CharmPlugin(plugins.Plugin):
                 "python3-venv",
                 "python3-wheel",
                 "libyaml-dev",
+                "cargo",
             }
         elif platform.is_yum_based():
             try:


lengau:
This implementation may break a charm if the user installs their own `cargo` (and `rustc`) through another mechanism such as rustup, which might be required if the base has too old a version of rustc.

Please look at the way the [poetry plugin](https://github.com/canonical/craft-parts/blob/main/craft_parts/plugins/poetry_plugin.py#L87-L96) and its [environment validator](https://github.com/canonical/craft-parts/blob/afb0d652eb330b6aaad4f40fbd6e5357d358de47/craft_parts/plugins/poetry_plugin.py#L65-L70) implement the checks

lengau:
Closing due to inactivity. Feel free to reopen with the requested changes!

PR #2010, opened 2024-11-29
docs: add, refactor, and reformat discourse docs feat. charmcraft

tmihoc:
This PR adds all the juju.is/docs content featuring Charmcraft.

The content was originally formatted for Discourse; this PR reformats it to serve the RTD project the docs are destined for.

The content was also originally entangled with Ops content; this PR significantly refactors -- as well as clarifies and tries to complete -- everything to serve the stand-alone Charmcraft docs set it is destined for.

Issues that should be fixed in subsequent PRs:

- Everywhere: (1) (a) Links to juju.is will work, as even after moving docs out of there we'll have redirects; however, they should be updated. (b) Links to ops docs are empty -- they reference content I'm planning to move there post-refactoring; once we have them in place, which should be in the next few days and pre-official-Charmcraft-RTD-launch, we should add them in Charmcraft docs.
- Tutorial: The tutorial should also be offered in the "regular" flavor, that is, for apps that are not 12-factor. That flavor should also feature the publish step. (See, e.g., our old Kubernetes tutorial, which includes https://juju.is/docs/sdk/publish-your-charm-on-charmhub . PS If the Charmcraft tutorial starts featuring this, as it should, the old Kubernetes tutorial, which will be moved to the stand-alone Ops docs, can focus on just development, which is what Ops is really all about.)
- How-to guides: While this PR surfaces a lot more of Charmcraft and arguably a lot more clearly, there are still corners with cobwebs or gaps. These include: (1) Manage charms (12-factor app) -- I think this doc should be incorporated into the main Manage charms doc, which could have tabs for "regular', "django", etc. (This would align with plans for the juju.is/docs entrypoint into the "federated docs"). (2) The Manage parts doc, which is full of "TBA" -- i.e., has gaps. (3) The docs in "Misc", which are not well integrated with everything else here. (4) Maybe the Manage bundles doc, which doesn't by far tell the full story; given that we're phasing bundles out, it's also perhaps fine not to worry about this too much. (5) Commands `analyse` and `test` are not featured in any how-to guide.
- Reference: (1) The autogenerated `charmcraft` CLI command reference docs should have anchors on the template `command-charmcraft-x` and titles on the template "Command `charmcraft x`. This will not better align with existing practices in the Juju world but also results in more clarity when linking to the page using `{ref}`...`` (2) The various file reference docs should be as much as possible autogenerated from source (likely the source will have to be updated first) -- at the very least, file `charmcraft.yaml`. They may also need further clean-up. (3) The part docs are a bit of a mess. We should move any relevant content to the part docs, then link up to that and only keep in Charmcraft the material specific to Charmcraft docs, and then revisit that material as well to streamline it as much as possible (e.g., currently the dump plugin is documented in 3 places). (4) Charmcraft analyzers and linters -- this gives details on pack and analyse but it's not clear in what way it's useful to the reader, so we need to rethink.
- Explanation: We have nothing there. I didn't delete that section as I think there is potentially a topic we could have there -- e.g., Charmcraft in the bigger context of Starcraft (a clarification of the general design philosophy there and trends, maybe).

PS Some docs may have a Contributors list on the bottom, listing the Discourse handles of the contributors to date. We'll want to rethink how we do that.


lengau:
We can ignore the test failures, but the linter that's failing is codespell and is related to these changes: https://github.com/canonical/charmcraft/actions/runs/12084317065/job/33699249108?pr=2010#step:6:637

docs/common/craft-parts
lengau, line 1: This needs to be excluded from the changes please

tmihoc:
> We can ignore the test failures, but the linter that's failing is codespell and is related to these changes: https://github.com/canonical/charmcraft/actions/runs/12084317065/job/33699249108?pr=2010#step:6:637

I think I've addressed all of these now.

docs/tutorial/write-your-first-kubernetes-charm-for-a-django-app.md
```diff
@@ -0,0 +1,567 @@
...
+
+```bash
+sudo microk8s enable hostpath-storage
+# Required to host the OCI image of the Django application
+sudo microk8s enable registry
+# Required to expose the Django application
+sudo microk8s enable ingress
+```
+
+Juju is required to deploy the Django application. Install Juju and boostrap a development controller:
```
lengau, line 100:
```suggestion
Juju is required to deploy the Django application. Install Juju and bootstrap a development controller:
```

lengau (approved):
Looks good, thanks!

medubelko:
The following pages have extensive maintenance planned post-migraiton, so they don't need detailed review:

- reference/extensions/charmcraft-extension-django-framework
- reference/extensions/charmcraft-extension-fastapi-framework
- reference/extensions/charmcraft-extension-flask-framework
- reference/extensions/charmcraft-extension-go-framework
- howto/manage-12-factor-app-charms
- tutorial/write-your-first-kubernetes-charm-for-a-django-app
- tutorial/write-your-first-kubernetes-charm-for-a-fastapi-app
- tutorial/write-your-first-kubernetes-charm-for-a-go-app

medubelko (approved):
Approving without review. If we merge this, we can resolve any code issues during separate RST conversion tasks.

docs/tutorial/index.rst
lengau, line 1: @medubelko please look at this file

docs/reference/files/file-libname-py.md
lengau, line 1: @medubelko please look at this file

docs/reference/files/file-charmcraft-yaml.rst
```diff
@@ -0,0 +1,901 @@
+.. _file-charmcraft-yaml:
+.. highlight:: yaml
```
medubelko, line 3: Since `highlight` acts as a switch, I think the code would be easier to read if it were placed immediately before the first block that uses it.

docs/reference/files/file-charmcraft-yaml.rst
```diff
@@ -0,0 +1,901 @@
...
+
+``charmcraft.yaml``
+*******************
```
medubelko, line 6: A nitpick, but I've been trying to stick to the [heading conventions](https://canonical-documentation-with-sphinx-and-readthedocscom.readthedocs-hosted.com/style-guide/#headings) recommended by the style guide.

docs/reference/files/file-charmcraft-yaml.rst
```diff
@@ -0,0 +1,901 @@
...
+
+    Charmcraft will use the information you provide here to generate :ref:`file-actions-yaml`, :ref:`file-config-yaml`, and :ref:`file-metadata-yaml`, as well as all the other files it usually does.
+
+``charmcraft.yaml`` is a file in your charm project that contains keys that allow you to declare information about the project in a form that can be used by Charmcraft.
+
+.. note::
+    If you're starting from an empty file, the only required key is the ``type`` key. However, depending on what value you set it to (``charm`` or ``bundle``), other keys become required as well.
+
+
+.. dropdown:: Expand to view a full charm with sample content all at once
```
medubelko, line 37: In other Starcraft and Store projects we use [`collapse`](https://github.com/canonical/snapcraft/blob/a34296875e9b90d7c27632d49407993bca26e649/docs/conf.py#L55) because it's [less visually jarring](https://canonical-snapcraft.readthedocs-hosted.com/en/latest/howto/craft-a-snap/example-pre-built-app/#example-recipe-for-geekbench).

docs/reference/files/file-charmcraft-yaml.rst
```diff
@@ -0,0 +1,901 @@
...
+
+    .. literalinclude:: charmcraft-sample-charm.yaml
+
+.. _file-charmcraft-yaml-actions::
+``actions``
```
medubelko, line 43: For legibility I prefer blank lines before headings.

docs/reference/files/file-charmcraft-yaml.rst
```diff
@@ -0,0 +1,901 @@
...
+
+``charmcraft.yaml``
+*******************
+
+..
+    AUTHOR NOTE:
+    The full list of keys is defined in the Charmcraft project (but this implies upstream keys from craft-application):
+    https://github.com/canonical/charmcraft/blob/3.2.0/charmcraft/models/project.py#L381-L1070
+
+    Technically, the only key required upfront is type. But then, depending on what you choose, other keys become required as well. (The required keys are the ones that are set to a value, unless that value is a pydantic.Field that doesn't have either a default or a default_factory parameter)
```
medubelko, line 13: We've been very consistent about 2 spaces for indentation. You may recall we recently discussed switching my YAML examples in Snapcraft back to 2 spaces.

docs/reference/files/file-charmcraft-yaml.rst
```diff
@@ -0,0 +1,901 @@
...
+
+.. _file-charmcraft-yaml-assumes::
+``assumes``
+===========
+
+**Status:** Optional. Recommended for Kubernetes charms.
+
+**Purpose:** Allows charm authors to explicitly state in the metadata of a charm various features that a Juju model must be able to provide to ensure that the charm can be successfully deployed on it. When a charm comes preloaded with such requirements, this enables Juju to perform a pre-deployment check and to display user-friendly error messages if a feature requirement cannot be met by the model that the user is trying to deploy the charm to. If the assumes section of the charm metadata is omitted, Juju will make a best-effort attempt to deploy the charm, and users must rely on the output of `juju status` to figure out whether the deployment was successful. The `assumes` key is available since 2.9.23.
+
+**Structure:** The key consists of a list of features that can be given either directly or, depending on the complexity of the condition you want to enforce, nested under one or both of the boolean expressions ``any-of`` or ``all-of``, as shown below. In order for a charm to be deployed, all entries in the ``assumes` block must be satisfied.
```
medubelko: Pandoc's fault?

```suggestion
**Structure:** The key consists of a list of features that can be given either directly or, depending on the complexity of the condition you want to enforce, nested under one or both of the boolean expressions ``any-of`` or ``all-of``, as shown below. In order for a charm to be deployed, all entries in the ``assumes`` block must be satisfied.
```

docs/reference/files/file-charmcraft-yaml.rst
```diff
@@ -0,0 +1,901 @@
...
+.. literalinclude:: charmcraft-sample-charm.yaml
+    :start-at: charm-libs:
+    :end-before: config:
+
+.. _file-charmcraft-yaml-charmhub:
+``charmhub``
+=====
+
+
+.. warning::
```
medubelko, line 351: Per the [rST style guide](https://canonical-documentation-with-sphinx-and-readthedocscom.readthedocs-hosted.com/style-guide/#notes), we use `caution` instead of `warning`.

docs/reference/files/file-charmcraft-yaml.rst
```diff
@@ -0,0 +1,901 @@
...
+
+    name: <name>
+
+**Example:**
+
+.. literalinclude:: charmcraft-sample-charm.yaml
+        :start-at: name:
+        :end-before: parts:
+
+.. file-charmcraft-yaml-parts:
```
medubelko:
```suggestion
.. _file-charmcraft-yaml-parts:
```

docs/reference/files/file-charmcraft-yaml.rst
```diff
@@ -0,0 +1,901 @@
...
+
+**Purpose:** To define the interface schema that this relation conforms to.
+
+**Structure:**
+
+*Type:* String.
+
+*Value:* The name of the interface. Usually defined by the author of the charm providing the interface.  Cannot be ``juju``. Cannot begin with ``juju-``. Must only contain characters `a-z` and `-` and cannot start with `-`.
+
+.. warning:: The interface name is the only means of establishing whether two charms are compatible for integration; and carries with it nothing more than a mutual promise that the provider and requirer know the communication protocol implied by the name.
```
medubelko:
```suggestion
.. caution::
  The interface name is the only means of establishing whether two charms are compatible for integration; and carries with it nothing more than a mutual promise that the provider and requirer know the communication protocol implied by the name.
```

docs/reference/files/file-charmcraft-yaml.rst
```diff
@@ -0,0 +1,901 @@
...
+
+**Status:** Required.
+
+**Purpose:** Indicates whether charmcraft will pack a charm or a bundle.
+
+**Structure:**
+
+*Type:* String.
+
+*Value:* ``charm`` or ``bundle``.
```
medubelko, line 1002:
```suggestion
**Type:** String.

**Value:** ``charm`` or ``bundle``.
```

docs/reference/files/file-libname-py.md
```diff
@@ -0,0 +1,160 @@
...
+
+<!--
+SEE IF WE CAN INTEGRATE THIS CONTENT FROM THIS DOC: https://juju.is/docs/sdk/document-your-library 
+One of the most distinct features of a charm is the way it can readily relate to another charm. For this interface, there is one charm that will provide information in a specific way, and a charm that will consume information in a matching way to connect and work together. To ensure that the charm interface is a perfect pairing, the operator framework tooling and the store support a mechanism to publish and reuse this logic in a form of simple files, that we’ll call libraries. 
+-->
+
+
+## Location
+
+Charm libraries are located in a  directory inside the charm with the following structure:
```
medubelko: For safety, we might want to check these files for `\b {2,}\b`

docs/reference/files/file-libname-py.md
```diff
@@ -0,0 +1,160 @@
...
+
+where `$CHARMDIR` is the project's root (contains `src/`, `hooks/`, etc.), and the `<charm>` placeholder represents the charm responsible for the library named as `<libname>.py` with API version `<API>`. 
+
+For example, inside a charm `mysql`, the library `db` with major version 3 will be in a directory with the structure below:
+
+   $CHARMDIR/lib/charms/mysql/v3/db.py
+
+When you pack your charm, Charmcraft copies the top `lib` directory into the root directory of the charm. Thus, to import this library in Python use the full path minus the top `lib` directory, as below:
+
+```
```
medubelko: Missing language.

docs/reference/files/file-libname-py.md
```diff
@@ -0,0 +1,160 @@
...
+This list may be missing libraries from charms that are not publicly listed on Charmhub. If you would like to add a library to the list, please drop a comment using the feedback link below.
+
+```
+
+### Libraries that define relations
+
+The following libraries provide programmatic instructions for relating to a specific charm.
+
+| Library | Used in | Description |
+| --- | --- | --- |
```
medubelko: We can make tables even less noisy with:

```
Col | Col | Col
-|-|-
cell | cell | cell
```

The outer delimiters are optional.

docs/tutorial/index.rst
medubelko: This one is encroaching on CRAFT-3702, therefore I don't think we need to exercise a huge amount of discretion at this point. I've sent you the landing page exemplar we use.

medubelko (changes requested):
I left comments for the files you requested.

docs/reference/files/file-charmcraft-yaml.rst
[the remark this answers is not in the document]
medubelko: Done!

docs/reference/files/file-libname-py.md
medubelko: Done!

docs/reference/files/file-charmcraft-yaml.rst
lengau: That's pretty unintuitive to me - if I'm looking for why a block is highlighted in an unexpected manner, I'm going to check that block (which is likely not the first block in the page, especially for reference pages like this), then the directives at the very top of the page, then the global `highlight_language` configuration.

A `highlight` directive somewhere in the middle of a file would be highly unappreciated, much like an `import` coming between two class definitions in a source file. Worse still would be a switch of the `highlight` directive mid-file, which would be a form of code smell that this document is about two different things (and thus should be broken up into multiple pages), since in cases where one needs to change the highlight language a `code-block` can set its own language as a one-off.

docs/reference/files/file-charmcraft-yaml.rst
lengau: The current [configuration for markdown and rst](https://github.com/canonical/starbase/blob/c5c71673c8957fe8bd914f04b15e200403f58902/.editorconfig#L31-L32) in starbase is the global default of [`indent_size = 4`](https://github.com/canonical/starbase/blob/c5c71673c8957fe8bd914f04b15e200403f58902/.editorconfig#L8), so if we want to change that well need to change the standard and [the existing documentation files](https://github.com/canonical/charmcraft/blob/main/docs/reference/models/bases-charm.rst?plain=1), including [in Starbase itself](https://github.com/canonical/starbase/blob/c5c71673c8957fe8bd914f04b15e200403f58902/docs/release-notes/index.rst?plain=1#L55-L72).

docs/reference/files/file-charmcraft-yaml.rst
lengau: A combination of pandoc and alexdoc (pandoc left most of these as single-backticks and I missed that. Thanks for catching it!

docs/reference/files/file-charmcraft-yaml.rst
lengau: This one missed my lint-fixes because I'd also accidentally made it a directive rather than a link. Thanks for catching it!

docs/reference/files/file-charmcraft-yaml.rst
lengau: Thanks - yeah, @tmihoc pointed this out to me yesterday too, which was actually my first knowledge about us having an rST style guide specifically.

docs/reference/files/file-charmcraft-yaml.rst
lengau: I've created [an issue on Starbase](https://github.com/canonical/starbase/issues/307) so we can get clarification for everything that needs changing here.

docs/reference/files/file-libname-py.md
lengau: Sorry, I pointed you to the wrong file here (the deleted markdown one, not the new rst one). This is fixed with a [list-table](https://github.com/canonical/charmcraft/pull/2010/files#diff-28bae703b56662486de50e2c07ee4f3a89de783d46374a673f46081cdef9d8a4) in the markdown version (that also matches the style guide).

docs/reference/files/file-libname-py.md
lengau: Corrected in the rst version

docs/reference/files/file-charmcraft-yaml.rst
lengau: Oh actually - it looks like the updated file for the linter stuff did get it :-)

docs/reference/files/file-charmcraft-yaml.rst
lengau:
```suggestion
.. caution::
```

docs/reference/files/file-charmcraft-yaml.rst
lengau:
```suggestion
.. collapse:: Expand to view a full charm with sample content all at once
```

docs/reference/files/file-charmcraft-yaml.rst
lengau: I had done it this way initially to match what was on the Discourse docs, but I'm happy to change it.

docs/reference/files/file-charmcraft-yaml.rst
medubelko: I suppose my reaction comes down to general patterns. I haven't come across our using `highlight` elsewhere. It gives off a different kind of smell for me, similar to jumping back and forth from mid-page to the end to look at a URL for a reference. But I find your points fair in this instance.

docs/reference/files/file-libname-py.rst
```diff
@@ -0,0 +1,334 @@
...
+charm.
+
+.. list-table::
+    :header-rows: 1
+
+    * - Library
+      - Used in
+      - Description
+    * - `fluentbit <https://charmhub.io/fluentbit/libraries/fluentbit>`_
+      - - `fluentbit <https://charmhub.io/fluentbit>`_
```
medubelko, line 185:
```suggestion
      - `fluentbit <https://charmhub.io/fluentbit>`_
```

A handful of column 2 cells in this block have single-item lists, whereas in the original text they were regular paragraphs.

docs/reference/files/file-charmcraft-yaml.rst
medubelko: Here I was trying to keep consistent with the [plugin docs](https://github.com/canonical/charmcraft/blob/e820f399981ba691767e7c9c6498ff7c9456d08f/docs/reference/plugins/poetry_plugin.rst?plain=1#L15).

medubelko (approved):

medubelko:
@lengau I spent a bit of extra time on the clean up. Should be good to go. This includes part of the fix from #2070. I don't think a complete re-review is needed – a look at `conf.py` and the project files should be enough.

lengau (approved):

PR #2015, opened 2024-12-05
build: use starflow

lengau:
Requires #2014 

This is probably best to review commit by commit, as we'll be rebasing it anyway.


.github/workflows/security-scan.yaml
```diff
@@ -10,11 +10,14 @@ on:
 jobs:
   python-scans:
     name: Scan Python project
-    uses: canonical/starflow/.github/workflows/scan-python.yaml@main
+    # uses: canonical/starflow/.github/workflows/scan-python.yaml@main
+    uses: lengau/starflow/.github/workflows/scan-python.yaml@work/CRAFT-3707/uv
```
lengau, line 14: After merging https://github.com/canonical/starflow/pull/27:

```suggestion
    uses: canonical/starflow/.github/workflows/scan-python.yaml@main
```

snap/snapcraft.yaml
```diff
@@ -123,10 +123,7 @@ parts:
   charmcraft:
     after: [charmcraft-libs, libgit2]
     source: .
-    plugin: python
-    python-requirements:
-      - requirements-jammy.txt
-      - requirements.txt
+    plugin: nil
```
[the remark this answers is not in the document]
lengau: The requirements files are going away and this is roughly what the uv plugin is going to do, so long term this and the `uv sync` line will be replaced with just using the uv plugin.

tigarmo:
I'm keeping an eye on this and the changes seem fine - I'm just waiting for the changes that will be necessary to fix the spread tests breakage

snap/hooks/configure
```diff
@@ -1,4 +1,6 @@
-#!/usr/bin/env python3
+#!/bin/sh
+'''exec' "$(dirname -- "$(realpath -- "$0")")"/../../bin/'python3' "$0" "$@"
```
mr-cal, line 5: Will other snaps need to do this for their hooks once we have a uv plugin?

snap/snapcraft.yaml
```diff
@@ -136,27 +133,51 @@ parts:
...
+      - -bin/jsonschema
+      - -bin/deactivate*
+      - -bin/keyring
+      - -bin/normalizer
+      - -bin/pip*
+      - -bin/pydoc*
+      - -bin/snap-helpers
+      - -bin/tabulate
+    # organize:
+    #   bin/craftctl: libexec/charmcraft/craftctl
```
mr-cal, line 181: Did this stop working?

mr-cal (approved):
There are a lot of changes here, but I don't see anything of concern. I've seen a starbase PR from you, so I assume you have been upstreaming changes where appropriate.

tigarmo (approved):

snap/hooks/configure
lengau: Yes, though I'm not entirely sure why. But `/usr/bin/env` was finding `/usr/bin/python3` rather than `/snap/charmcraft/current/bin/python3`. (Or it was to do with something else I did and it would work now. Will check.)

lengau: UPDATE: It was something else I did. I'll revert these changes.

snap/snapcraft.yaml
lengau: The shebang changed because we're in a relocatable venv, and it was easier to just symlink in `override-build`. Should I not do that (does it need not not be available in the default PATH?)

lengau: I changed it so I'm using organize again and just rewriting the shebang.

PR #2021, opened 2024-12-10
feat(reactive): run charm build in verbose mode

lengau:
While a user could previously have added `--verbose` to `reactive-charm-build-arguments`, this automatically adds `--verbose` if the user hasn't specified a log level.

The verbose text gets handled by craft-cli anyway, but this changes the default so the text gets added to the log.

Spread test failures are unrelated: https://github.com/canonical/charmcraft/issues/2023


mr-cal (approved):
Very clean!

PR #2022, opened 2024-12-10
build(deps): update craft-parts

lengau:
Fixes the poetry plugin

NOTE: main removed `requirements.txt`, but `requirements-dev.txt` still exists for some workflows. `uv.lock` already has the correct craft-parts.


dariuszd21 (approved):
Thank you!

mr-cal (approved):

PR #2031, opened 2024-12-11
ci: use charmcraft from the store for weekly spread runs

lengau:
The only reason this would be different is if there happened to be a commit to main that hasn't pushed to edge yet, but that would be very rare and is not worth the extra time we spend building the snap.


mr-cal (approved):
This is a great idea

PR #2035, opened 2024-12-12
chore: add CODEOWNERS file

lengau:


mr-cal (approved):

upils (approved):
Should we write our strategy regarding CODEOWNERS file somewhere and reference it in the file to ease consistency and maintenance of these files?

PR #2036, opened 2024-12-12
Fix & always include bin/activate in python & poetry plugin

carlcsaposs-canonical:
Including `bin/activate` in the virtual environment will make debugging much easier (charm developers can easily run Python code using the charm's virtual environment to debug issues)

### DIFF
--- a/charmcraft/utils/parts.py
+++ b/charmcraft/utils/parts.py
@@ -66,11 +66,25 @@ def get_venv_cleanup_commands(venv_path: pathlib.Path, *, keep_bins: bool) -> li
     """
     venv_bin = venv_path / "bin"
     venv_lib64 = venv_path / "lib64"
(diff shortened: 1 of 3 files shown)


charmcraft/parts/plugins/_poetry.py
```diff
@@ -25,8 +25,8 @@
 
 
 class PoetryPluginProperties(poetry_plugin.PoetryPluginProperties, frozen=True):
-    poetry_keep_bins: bool = False
-    """Keep the virtual environment's 'bin' directory."""
+    poetry_keep_bin_activate: bool = True
```
lengau, line 28: Let's still leave an option to keep the whole `bins` directory, in case people want it for some reason.

lengau:
@carlcsaposs-canonical let's just always modify it.

lengau (approved):
Thanks!

bepri (approved):

PR #2048, opened 2024-12-17
docs: add "How to manage 12-factor app charms"

lengau:
I'm breaking down the content from PR #2010 into smaller chunks.

This is "How to manage 12-factor app charms" converted to RST: https://github.com/canonical/charmcraft/blob/737976c943a601adccd883ffa6a276bce625c9a1/docs/howto/manage-12-factor-app-charms.md

The main differences from the Markdown version:

- Uses tabs instead of dropdowns (matching the Rockcraft docs: https://documentation.ubuntu.com/rockcraft/en/latest/how-to/build-a-12-factor-app-rock/)
- "See also" blocks are added as cards


docs/howto/manage-12-factor-app-charms.rst
```diff
@@ -0,0 +1,342 @@
...
+   ## Update the OCI image for a 12-Factor app charm
+
+   See {ref}`Rockcraft Documentation > How to update the OCI image... <15018md>`
+
+   -->
+
+.. card:: See also
+    :link: https://juju.is/docs/juju/charmed-operator
+
+    ``juju`` | 12-factor app charms
```
medubelko, line 21: I'm not jazzed by using cards, for several reasons. I think we can make do by making this a note or a plain paragraph, which is closer to the original structure (blockquote). That way we won't need to introduce new patterns and directives.

Longer-term I'd like to come up with smoother wording for introducing these links. Dressing them up as notes will be a good way to encourage me to correct them, since you know I don't like notes. :)

docs/howto/manage-12-factor-app-charms.rst
```diff
@@ -0,0 +1,342 @@
...
+
+
+Manage configurations for a 12-factor app charm
+-----------------------------------------------
+
+A charm configuration can be added if your 12-factor app requires
+environment variables, for example, to pass a token for a service. Add
+the configuration in ``charmcraft.yaml``:
+
+.. code:: text
```
medubelko, line 82: Any reason not to use `yaml`?

docs/howto/manage-12-factor-app-charms.rst
```diff
@@ -0,0 +1,342 @@
...
+        converted to underscores and the ``FLASK_`` prefix will be added. In the
+        example above, the ``token`` configuration will be mapped as the
+        ``FLASK_TOKEN`` environment variable. In addition to the environment
+        variable, the configuration is also available in the Flask variable
+        ``app.config`` without the ``FLASK_`` prefix.
+
+        The configuration can be set on the deployed charm using
+        ``juju config <app name> token=<token>``.
+
+        .. card-carousel:: 2
```
medubelko, line 109: Oof, I'm very strongly not in favour of how this looks, either.

medubelko (commented):
Left some thoughts about the markup and structure.

docs/howto/manage-12-factor-app-charms.rst
```diff
@@ -0,0 +1,342 @@
...
+
+   <!--TODO: There are a few more low-hanging fruit topics that we can and should cover:
+
+   When we create such sections we should include thinks at both the rock and the charm level (because the 12-Factor app app charm wants you to think about them in this coupled fashion). However, for the Rockcraft things, we should have those sections point to Rockcraft docs. For example:
+
+   ## Update the OCI image for a 12-Factor app charm
+
+   See {ref}`Rockcraft Documentation > How to update the OCI image... <15018md>`
+
+   -->
```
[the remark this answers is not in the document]
lengau: It's not just left as Markdown, it's left as the raw pandoc output 🙂

docs/howto/manage-12-factor-app-charms.rst
lengau: Just missed this one in the pandoc output.

```suggestion
.. code:: yaml
```

docs/howto/manage-12-factor-app-charms.rst
medubelko: I revisited this question with Dora. Let's instead use block quotes (two prepended spaces) for these.

That's also a structure we don't use often, so I'll easily be able to pick it up in later reviews/rewrites.

lengau:
Superseded by #2083

PR #2049, opened 2024-12-17
docs: reorganise the Part reference

lengau:
This reorganises the part reference in line with #2010

### DIFF
--- a/docs/reference/plugins/index.rst
+++ b/docs/reference/plugins/index.rst
@@ -1,19 +1,23 @@
 .. _plugins:
 
-Parts plugins
-*************
+List of part plugins
+********************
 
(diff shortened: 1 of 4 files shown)


docs/reference/plugins/index.rst
```diff
@@ -1,19 +1,23 @@
...
-Parts plugins
-*************
+List of part plugins
+********************
 
 Most charms only need one, maybe two parts, typically consisting of one of Charmcraft's
 application-specific plugins such as the `charm plugin`_ or the `reactive plugin`_ and
 potentially the addition of further files using the :ref:`craft_parts_dump_plugin`.
 
+.. note::
```
lengau, line 10: I also added this note as a drive-by just so people know where to find that doc.

PR #2053, opened 2024-12-18
docs: fix poetry-with description

carlcsaposs-canonical:
--with is required for optional and non-optional dependencies

see https://github.com/python-poetry/poetry-plugin-export/pull/212 and https://github.com/python-poetry/poetry-plugin-export/issues/299

corresponding PR in https://github.com/canonical/craft-parts/pull/946

### DIFF
--- a/docs/howto/charm-to-poetry.rst
+++ b/docs/howto/charm-to-poetry.rst
@@ -56,9 +56,9 @@ for charms that build on Ubuntu 22.04 or earlier:
 Add optional dependency groups
 ------------------------------
 
-If the charm has optional `dependency groups`_ that should be included when creating
-the virtual environment, the ``poetry-with`` key can be used to include those groups
-when creating the virtual environment.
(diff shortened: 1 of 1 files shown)


lengau (approved):
Thanks!

docs/howto/charm-to-poetry.rst
```diff
@@ -56,9 +56,9 @@ for charms that build on Ubuntu 22.04 or earlier:
 Add optional dependency groups
 ------------------------------
 
-If the charm has optional `dependency groups`_ that should be included when creating
-the virtual environment, the ``poetry-with`` key can be used to include those groups
-when creating the virtual environment.
+If the charm has optional or non-optional `dependency groups`_ that should be included
```
medubelko, line 59: Technically, this means all dependency groups:

```suggestion
If the charm has `dependency groups`_ that should be included
```

medubelko (approved):
Leaving a non-blocking suggestion.

Issue #2009, opened 2024-11-27
Document multi-base support and build-plans


mr-cal:
### What needs to get done

Add explanation, reference and how-to documentation for build plans, bases, platforms, and multi-base builds.

The craft-platforms reference and explanation for charm build plans can be imported.

Snapcraft's documentation may be useful starting point ([ref](https://canonical-snapcraft.readthedocs-hosted.com/en/latest/reference/architectures/), [exp](https://canonical-snapcraft.readthedocs-hosted.com/en/latest/explanation/architectures/), [how-to](https://canonical-snapcraft.readthedocs-hosted.com/en/latest/howto/architectures/))

### Why it needs to get done

To enable charm developers to build charms using the `platforms` key.

PR #2054, opened 2024-12-18
feat: support multi-base recipes

### LINKED ISSUE #1789: Unable to build multi-base subordinate charms on 24.04
### Bug Description

Subordinate charms use the base of the principal charm. Therefore, subordinate charms such as MySQL Router need to support multiple bases, since we can't enforce the principal charm to use a particular base. Currently MySQL Router supports 20.04 and 22.04 and plans to support 24.04.

However, with the new `base` and `platforms` syntax in charmcraft 3, it is not possible to build a subordinate charm on 20.04, 22.04, and 24.04 with a single charmcraft.yaml and git branch

Additional context:
I believe @jnsgruk and @sergiusens were/are discussing this. Created an issue to track officially per @taurus-forever's request

### To Reproduce

1. Create charmcraft.yaml with
```yaml
bases:
  - name: ubuntu
    channel: "20.04"
    architectures: [amd64]
  - name: ubuntu
    channel: "22.04"
    architectures: [amd64]
  - name: ubuntu
    channel: "22.04"
    architectures: [arm64]
  - name: ubuntu
    channel: "24.04"
    architectures: [amd64]
  - name: ubuntu
    channel: "24.04"
    architectures: [arm64]
```
2. charmcraft pack

### Environment

Ubuntu 22.04

### charmcraft.yaml

```shell
# Copyright 2022 Canonical Ltd.

type: charm
bases:
  # Whenever "bases" is changed:
  # - Update tests/conftest.py::pytest_configure()
  # - Update .github/workflow/ci.yaml integration-test matrix
  - name: ubuntu
    channel: "20.04"
    architectures: [amd64]
  - name: ubuntu
    channel: "22.04"
    architectures: [amd64]
  - name: ubuntu
    channel: "22.04"
    architectures: [arm64]
  - name: ubuntu
    channel: "24.04"
    architectures: [amd64]
  - name: ubuntu
    channel: "24.04"
    architectures: [arm64]
parts:
  charm:
    override-pull: |
      craftctl default
      if [[ ! -f requirements.txt ]]
      then
          echo 'ERROR: Use "tox run -e build-dev" instead of calling "charmcraft pack" directly' >&2
          exit 1
      fi
    # TODO: enable after https://github.com/canonical/charmcraft/issues/1456 fixed
    charm-strict-dependencies: false
    charm-entrypoint: src/machine_charm.py
    prime:
      - charm_version
      - workload_version
    build-packages:
      - libffi-dev
      - libssl-dev
      - pkg-config
      - rustc
      - cargo
```

### Relevant log output

```shell
Bad charmcraft.yaml content:
- base requires 'platforms' definition: {'name': 'ubuntu', 'channel': '24.04', 'architectures': ['amd64']} (in field 'bases[3]')
- base requires 'platforms' definition: {'name': 'ubuntu', 'channel': '24.04', 'architectures': ['arm64']} (in field 'bases[4]')
```

### LINKED ISSUE #2008: Multi-base platform support
### What needs to get done

Polish the prototype in [work/CRAFT-3668/multibase-prototype](https://github.com/canonical/charmcraft/tree/work/CRAFT-3668/multibase-prototype) to support multi-base builds as defined in [ST124](https://docs.google.com/document/d/1QVHxZumruKVZ3yJ2C74qWhvs-ye5I9S6avMBDHs2YcQ/edit?tab=t.0). 

### Missing work

1. Validate that the newest base is >= `ubuntu@24.04`.
2. Unit tests
3. Spread tests

### Why it needs to get done

To resolve https://github.com/canonical/charmcraft/issues/1789

### Upstream work
- https://github.com/canonical/craft-platforms/issues/73
- https://github.com/canonical/craft-application/issues/569

mr-cal:
Adds support for multi-base recipes with the `platforms` keyword.

Internally, this moves to a new practice where the application stops modifying project data before passing the it to craft-platforms.  This will ensure consistent processing by charmcraft and by launchpad.

Documentation is coming soon via https://github.com/canonical/craft-platforms/issues/75 and https://github.com/canonical/charmcraft/issues/2009.

To summarize the usage, `base` and `build-base` are removed from the project and the base is encoded into the platform.

Platforms can be defined in a shorthand notation:
```yaml
platforms:
  ubuntu@22.04:amd64:
  ubuntu@24.04:amd64:
```

Or they can be defined in standard form:
```yaml
platforms:
  jammy:
    build-on: [ubuntu@22.04:amd64]
    build-for: [ubuntu@22.04:amd64]
  noble:
    build-on: [ubuntu@24.04:amd64]
    build-for: [ubuntu@24.04:amd64]
```

Fixes #1789 
Fixes #2008 
Creates https://github.com/canonical/craft-platforms/issues/85
Found during implementation https://github.com/canonical/craft-platforms/issues/80
(CRAFT-3718)

### DIFF
--- a/charmcraft/models/project.py
+++ b/charmcraft/models/project.py
@@ -38,7 +38,7 @@ from craft_cli import CraftError, emit
 from craft_platforms import charm
 from craft_providers import bases
 from pydantic import dataclasses
-from typing_extensions import Self
+from typing_extensions import Self, override
 
 from charmcraft import const, preprocess, utils
 from charmcraft.const import (
@@ -323,6 +323,15 @@ class CharmcraftBuildPlanner(models.BuildPlanner):
     build_base: str | None = None
     platforms: dict[str, models.Platform | None] | None = None  # type: ignore[assignment]
 
+    @override
+    @pydantic.field_validator("platforms", mode="before")
+    @classmethod
+    def _expand_shorthand_platforms(cls, platforms: dict[str, Any]) -> dict[str, Any]:
+        """Overrides the validator to prevent platforms from being modified.
+
+        Modifying the platforms field can break multi-base builds."""
+        return platforms
+
     def get_build_plan(self) -> list[models.BuildInfo]:
         """Get build bases for this charm.
 
@@ -346,12 +355,8 @@ class CharmcraftBuildPlanner(models.BuildPlanner):
                     ),
                 )
             ]
-        if not self.base:
+        if not self.base and not self.platforms:
             return list(CharmBuildInfo.gen_from_bases_configurations(*self.bases))
-
-        build_base = self.build_base or self.base
-        base_name, _, base_version = build_base.partition("@")
-        base = bases.BaseName(name=base_name, version=base_version)
 
         if self.platforms is None:
             raise CraftError("Must define at least one platform.")
@@ -373,7 +378,9 @@ class CharmcraftBuildPlanner(models.BuildPlanner):
                 platform=info.platform,
                 build_on=str(info.build_on),
                 build_for=str(info.build_for),
-                base=base,
+                base=bases.BaseName(
+                    name=info.build_base.distribution, version=info.build_base.series
+                ),
             )
             for info in build_infos
         ]
@@ -1064,7 +1071,7 @@ class PlatformCharm(CharmProject):
     """Model for defining a charm using Platforms."""
 
     # Silencing pyright because it complains about missing default value
-    base: BaseStr  # pyright: ignore[reportGeneralTypeIssues]
+    base: BaseStr | None = None
     build_base: BuildBaseStr | None = None
     platforms: dict[str, models.Platform | None]  # type: ignore[assignment]
 
@@ -1075,6 +1082,15 @@ class PlatformCharm(CharmProject):
                 f"Base {self.base} requires a build-base (recommended: 'build-base: ubuntu@devel')"
             )
         return self
+
+    @override
+    @pydantic.field_validator("platforms", mode="before")
+    @classmethod
+    def _expand_shorthand_platforms(cls, platforms: dict[str, Any]) -> dict[str, Any]:
+        """Overrides the validator to prevent platforms from being modified.
+
+        Modifying the platforms field can break multi-base builds."""
+        return platforms
 
 
 Charm = BasesCharm | PlatformCharm
--- a/charmcraft/models/project.py
+++ b/charmcraft/models/project.py
@@ -326,7 +326,7 @@ class CharmcraftBuildPlanner(models.BuildPlanner):
     @override
     @pydantic.field_validator("platforms", mode="before")
     @classmethod
-    def _expand_shorthand_platforms(cls, platforms: dict[str, Any]) -> dict[str, Any]:
+    def _populate_platforms(cls, platforms: dict[str, Any]) -> dict[str, Any]:
         """Overrides the validator to prevent platforms from being modified.
 
         Modifying the platforms field can break multi-base builds."""
@@ -1086,7 +1086,7 @@ class PlatformCharm(CharmProject):
     @override
     @pydantic.field_validator("platforms", mode="before")
     @classmethod
-    def _expand_shorthand_platforms(cls, platforms: dict[str, Any]) -> dict[str, Any]:
+    def _populate_platforms(cls, platforms: dict[str, Any]) -> dict[str, Any]:
         """Overrides the validator to prevent platforms from being modified.
 
         Modifying the platforms field can break multi-base builds."""
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ dynamic = ["version"]
 description = "The main tool to build, upload, and develop in general the Juju charms."
 readme = "README.md"
 dependencies = [
-    "craft-application @ git+https://github.com/canonical/craft-application@work/CRAFT-3721-multi-base-support",
+    "craft-application~=4.7",
     "craft-cli>=2.3.0",
     "craft-grammar>=2.0.0",
     "craft-parts>=2.2.0",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,12 +4,12 @@ dynamic = ["version"]
 description = "The main tool to build, upload, and develop in general the Juju charms."
 readme = "README.md"
 dependencies = [
-    "craft-application~=4.2",
+    "craft-application @ git+https://github.com/canonical/craft-application@work/CRAFT-3721-multi-base-support",
     "craft-cli>=2.3.0",
     "craft-grammar>=2.0.0",
     "craft-parts>=2.2.0",
     "craft-providers>=2.0.0",
-    "craft-platforms~=0.3",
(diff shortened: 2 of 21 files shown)


charmcraft/models/project.py
```diff
@@ -323,6 +323,15 @@ class CharmcraftBuildPlanner(models.BuildPlanner):
     build_base: str | None = None
     platforms: dict[str, models.Platform | None] | None = None  # type: ignore[assignment]
 
+    @override
+    @pydantic.field_validator("platforms", mode="before")
+    @classmethod
+    def _populate_platforms(cls, platforms: dict[str, Any]) -> dict[str, Any]:
+        """Overrides the validator to prevent platforms from being modified.
+
+        Modifying the platforms field can break multi-base builds."""
```
lengau, line 332: Is this related to https://github.com/canonical/craft-platforms/pull/84? If so, can we undo this?

pyproject.toml
```diff
@@ -4,12 +4,12 @@ dynamic = ["version"]
 description = "The main tool to build, upload, and develop in general the Juju charms."
 readme = "README.md"
 dependencies = [
-    "craft-application~=4.2",
+    "craft-application @ git+https://github.com/canonical/craft-application@work/CRAFT-3721-multi-base-support",
```
lengau, line 7: For merge-time:

```suggestion
    "craft-application~=4.7",
```

pyproject.toml
```diff
@@ -4,12 +4,12 @@ dynamic = ["version"]
...
     "craft-cli>=2.3.0",
     "craft-grammar>=2.0.0",
     "craft-parts>=2.2.0",
     "craft-providers>=2.0.0",
-    "craft-platforms~=0.3",
+    "craft-platforms @ git+https://github.com/canonical/craft-platforms@work/CRAFT-3271-multi-base-fix#egg=craft-platforms",
```
lengau, line 12:
```suggestion
    "craft-platforms~=0.5",
```

requirements-dev.txt
```diff
@@ -8,12 +8,12 @@ cffi==1.17.1
 chardet==5.2.0
 charset-normalizer==3.4.0
 coverage==7.6.9
-craft-application==4.4.0
+git+https://github.com/canonical/craft-application@work/CRAFT-3721-multi-base-support#egg=craft-application
```
lengau, line 11:
```suggestion
craft-application==4.7.0
```

requirements-dev.txt
```diff
@@ -8,12 +8,12 @@ cffi==1.17.1
...
 craft-archives==2.0.2
-craft-cli==2.10.1
+craft-cli==2.13.0
 craft-grammar==2.0.1
 craft-parts==2.2.0
-craft-platforms==0.4.0
+git+https://github.com/canonical/craft-platforms@work/CRAFT-3271-multi-base-fix#egg=craft-platforms
```
lengau, line 16:
```suggestion
craft-platforms==0.5.0
```

charmcraft/models/project.py
mr-cal: I don't think this can be avoided. Now that the shorthand and longhand notation are not equivalent, craft-application's expansion of shorthand into longhand will break in one way or another.

If you write:
```
platforms:
  ubuntu@24.04:riscv64:
```

This validator in craft-application could expand it to:
```
platforms:
  ubuntu@24.04:riscv64:
    build-on: [riscv64]
    build-for: [riscv64]
```
but this is a problem because craft-platforms does not allow for a base in platform name and `build-on` or `build-for` entries.

Craft-application could expand it to
```
platforms:
  riscv64:
    build-on: [ubuntu@24.04:riscv64]
    build-for: [ubuntu@24.04:riscv64]
```
but this is a problem for 2 reasons: the platform name has changed, which is confusing when using `--platform` and the build plan could contain duplicate platform names.

tests/spread/smoketests/multi-base/all/charmcraft.yaml
```diff
@@ -0,0 +1,20 @@
...
+  jammy:
+    build-on: [ubuntu@22.04:amd64]
+    build-for: [ubuntu@22.04:all]
+  noble:
+    build-on: [ubuntu@24.04:amd64]
+    build-for: [ubuntu@24.04:all]
+
+parts:
+  my-charm:
+    plugin: charm
```
lengau, line 19: Since the charm plugin can't actually build architecture-independent charms anymore due to `ops` dependencies and for the sake of speed, let's do:

```suggestion
    plugin: dump
```

charmcraft/models/project.py
mr-cal: The problem is that you can't have a base in the platform name and `build-on` or `build-for` entries. At least, that is how it is implemented in craft-platforms.  This is one of the discussion items in the spec that is not clearly defined.

mr-cal: Yeah, I wish I would have clarified this earlier as we were all interpreting the spec differently. 

[This is already in](https://github.com/canonical/craft-platforms/blob/7680af231de67febeb6e97d1f572339140c7b0a9/craft_platforms/charm/_build.py#L65-L72) craft-platforms, @cmatsuoka is OK with keeping as-is, and it's more restrictive, I think we should remain on this course.  In a subsequent iteration, we can relax the constraint in craft-platforms and clarify the spec.

mr-cal:
Looks like I have one store-related spread test to address but otherwise this is ready for review.

charmcraft/models/project.py
cmatsuoka: I see the problem there. The problem I see is that we're not expanding the shorthand notation according to the shorthand notation rules, which say that the platform name should be the same as `build-on` and `build-for` values. I think we'll need to check if there's a shorthand name and `build-on`/`build-for` specified only on entries that have not been expanded, otherwise this will always fail.

cmatsuoka: So
```
plaforms:
  ubuntu@24.04:amd64:
```
would expand to
```
platforms:
  ubuntu@24.04:amd64:
    build-on: [ubuntu@24.04:amd64]
    build-for: [ubuntu@24.04:amd64]
```
but this entry wouldn't be checked again for presence of shorthand name + attributes (because we already know that this will be exactly the case).

cmatsuoka: Maybe the easiest way to do that is to check for shorthand name + attributes *before* doing the expansion.

cmatsuoka: As discussed with @mr-cal, preventing Charmcraft from changing platform entries and sending the input as-is to craft-platforms seems to be the right thing to do: everything related to platforms would be handled by craft-platforms, and this would allow consistent platforms processing for anyone using the library (including Launchpad in the future).

cmatsuoka (approved):
Thanks! It would be nice to have a description of the overall platforms handling strategy in the PR description to help our confused future selves understand what's happening here.

charmcraft/models/project.py
lengau: I think we're on the same page now, though we should probably work upstream on craft-application not changing the platforms entries.

lengau (approved):
Nicely done! Lots of moving parts here.

PR #2074, opened 2025-01-13
style: update to ruff 0.9.0 formatting

lengau:
This handles the cases where lint starts failing with ruff 0.9.0.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,7 +13,7 @@ repos:
       - id: fix-byte-order-marker
       - id: mixed-line-ending
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: "v0.8.1"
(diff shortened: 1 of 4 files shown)


bepri (approved):
Approving ahead of time as this is nitpicky, but...

In the spirit of updating a linter, here's a few pyproject errors that could be added on to this:

- `git_describe_command` should be an array - just take care to not repeat the mistakes fixed in https://github.com/canonical/craft-parts/pull/957
- `lint.select` `TCH` rules should now be `TC`
- `lint.ignore` should drop the deprecated `ANN10`
- `tool.ruff.lint.per-file-ignores` should drop the deprecated `PT004` for test files

lengau:
@bepri agreed in all cases except the first - that craft-parts change should be upstreamed to starbase and we can get it from there.

PR #2080, opened 2025-01-14
docs: add platforms reference and how-to

### LINKED ISSUE #2009: Document multi-base support and build-plans

mr-cal:
Adds [reference](https://canonical-charmcraft--2080.com.readthedocs.build/en/2080/reference/platforms/) and [how-to](https://canonical-charmcraft--2080.com.readthedocs.build/en/2080/howto/build-guides/platforms/) documentation for platforms, bases, and architectures in Charmcraft, including information on the new [multi-base support](https://github.com/canonical/charmcraft/pull/2054).

Fixes #2009
(CRAFT-3720)


mr-cal:
@jnsgruk - do you mind reviewing this documentation?  I'm looking for feedback from a charm dev who hasn't already tried building multi-base charms.

For some reason, I can't add you as a PR reviewer.

docs/howto/build-guides/platforms.rst
```diff
@@ -0,0 +1,159 @@
...
+
+.. code-block:: yaml
+
+   base: ubuntu@24.04
+   platforms:
+     riscv64-cross:
+       build-on: [amd64]
+       build-for: [riscv64]
+
+Building on ``amd64`` will produce one charm that runs on ``riscv64``.
```
lengau, line 122: Can you add a note here that the user is responsible for ensuring that their charm works on the other architecture?

All of the current plugins (`charm`, `python`, `poetry`, `uv`) will install python packages for the host architecture rather than the target architecture.

lengau: I realised we don't have an issue for this: https://github.com/canonical/craft-parts/issues/974

docs/howto/build-guides/platforms.rst
```diff
@@ -0,0 +1,159 @@
...
+
+
+Create a set of charms for multiple bases
```
lengau, line 125: I don't have a better suggestion, but I'd like to shy away from using the term "bases" right now just because of the old "bases" key. Maybe multiple operating systems or OS releases?

docs/reference/platforms.rst
```diff
@@ -0,0 +1,215 @@
...
+page.
+
+Standard notation
+~~~~~~~~~~~~~~~~~
+
+.. code-block:: yaml
+
+   platforms:
+     <platform-name>:
+       build-on: [<arch>, <arch>]
```
lengau, line 21:
```suggestion
       build-on: [<arch1>, <arch2>]
```

docs/reference/platforms.rst
```diff
@@ -0,0 +1,215 @@
...
+* armhf
+* i386
+* ppc64el
+* riscv64
+* s390x
+
+.. _reference-bases:
+
+Bases
+-----
```
lengau, line 122: Can you add some form of admonishment (@medubelko will be able to say what's best) that this is not the same as the deprecated `bases` key?

docs/reference/platforms.rst
```diff
@@ -0,0 +1,215 @@
...
+
+.. code-block:: yaml
+
+    base: <base>
+
+``base`` determines the runtime environment for the charm. It's formatted as
+``<distribution>@<series>`` where ``distribution`` is the name of a Linux
+distribution and ``series`` is the release series name.
+
+Supported bases are:
```
lengau, line 141: This isn't actually true. Only Ubuntu 22.04+ are supported currently with the `base` keyword.

Also, while short term having this list here is fine, I think long-term I'd prefer to have a subcommand that tells us the supported bases.

docs/howto/build-guides/platforms.rst
```diff
@@ -0,0 +1,159 @@
...
+       build-for: [amd64]
+
+The ``build-on`` and ``build-for`` entries are identical for each platform, so
+the :ref:`multi-base shorthand notation
+<reference-platforms-multi-base-shorthand>` can be used:
+
+.. code-block:: yaml
+
+   platforms:
+     ubuntu@22.04:amd64:
```
jnsgruk, line 154: I’m kinda surprised by the difference between ‘ubuntu-22.04-amd64’ and ‘ubuntu@22.04:amd64’ - the two seem quite different and not in a very ‘obvious’ way?

mr-cal: Yeah, I think I was too strict with the validation and am circumventing my own rules.

I made the validation strict to prevent ambiguous definitions, like this one that is a hybrid of shorthand and longhand notation with mismatched bases:
```yaml
platforms:
  ubuntu@24.04:riscv64:
    build-on: [ubuntu@22.04:amd64, ubuntu@22.04:riscv64]
    build-for: [ubuntu@22.04:riscv64]
```

But I can circumvent my own validation by not using `<base>:<arch>`  in the platform name:
```yaml
platforms:
  ubuntu-24.04-riscv64:
    build-on: [ubuntu@22.04:amd64, ubuntu@22.04:riscv64]
    build-for: [ubuntu@22.04:riscv64]
```

I could use simple platform names like `noble-riscv64` or I could drop the validation, such that the platform name is completely arbitrary when `build-on` and `build-for` are defined.  @lengau, any thoughts?

docs/reference/platforms.rst
mr-cal: Agreed, and this is going to be outdated when 24.10 goes EOL in July 2025.  FWIW, dropping 24.10 will require manual intervention in craft-providers and charmcraft anyways.

docs/reference/platforms.rst
mr-cal: I like this suggestion but I realize both ways read like it needs to be exactly 2 items.

What I'm trying to convey in this reference is that `build-on` has no size constraints but `build-for` is a single element.  What do you think of?
```yaml
   platforms:
     <platform-name>:
       build-on: [<arch>, ...]
       build-for: [<arch>]
     ...
```

This also shows there can be more than 1 platform.

We could also do a style like `<arch1>, <arch2>, ..., <archX>`.

docs/reference/platforms.rst
mr-cal: Added.

docs/howto/build-guides/platforms.rst
mr-cal: (and my examples were wrong, to add to the confusion)

docs/howto/build-guides/platforms.rst
mr-cal: OSs releases would be the most accurate and bases is certainly overloaded, even if it is technically accurate.  I'll update it and add a sentence to tie the concepts of OSs, bases, and multi-base support together.

docs/howto/build-guides/platforms.rst
mr-cal: Added, do you have a preference for the name, "users" or "charm developers"?

docs/howto/build-guides/platforms.rst
```diff
@@ -0,0 +1,159 @@
+Build platform-specific charms
```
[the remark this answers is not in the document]
mr-cal: I like that but I have a question for @medubelko - is there a guideline for "How to" in the title?

It seems there is a mix in Charmcraft.  Further, some pages have "how to" in the title, but not in their entry in `howto/index.rst`.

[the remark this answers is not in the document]
medubelko: I'm avoiding repetition of "How to" in titles and filenames – it doesn't give much insight and it adds unnecessary length.

All the migrated files have the desired format, but there might be some stragglers that use the old one.

docs/howto/build-guides/platforms.rst
medubelko: I think that here, "users" and "developers" are sufficiently distinct. That's what I'm using to signal between snap users and snap crafters.

docs/reference/platforms.rst
lengau: How about a compromise of `<arch1>, ...`?

docs/howto/build-guides/platforms.rst
lengau: My preference would be that the only time we have limitations on the name of a platform is when it's used to infer the `build-on` and `build-for`. Otherwise `Järvfjället` is a perfectly good platform name.

mr-cal: I'm fine with that.  Once you, Claudio, and I agree on it, I can update the spec, relax the validation (which is trivial), and update the docs.

I don't think we need to block this PR.

docs/howto/build-guides/platforms.rst
```diff
@@ -0,0 +1,169 @@
...
+==============================
+
+A ``charmcraft.yaml`` defines where charms are built and where charms are run
```
medubelko, line 4:
```suggestion
A charm's ``charmcraft.yaml`` file determines which silicon architectures it builds on and which it can run on
```

docs/howto/build-guides/platforms.rst
```diff
@@ -0,0 +1,169 @@
...
+Building on ``amd64`` will produce one charm that runs on ``riscv64``.
+
+Note that the charm developer must ensure the charm is compatible with the
+target architectures. By default, the `charm`_,
+:ref:`python <craft_parts_python_plugin>`, :ref:`poetry <craft_parts_poetry_plugin>`,
+and :ref:`uv <craft_parts_uv_plugin>` plugins will install wheels for python packages
+for the ``build-on`` architecture rather than the ``build-for``. For more information,
+see `craft-parts#974`_.
+
+Create a set of charms for multiple OS releases
```
medubelko, line 131:
```suggestion
Select multiple platform bases
```

Alternatively, "multiple bases".

docs/howto/build-guides/platforms.rst
```diff
@@ -0,0 +1,169 @@
+Build platform-specific charms
+==============================
+
+A ``charmcraft.yaml`` defines where charms are built and where charms are run
+using the ``platforms``, ``base``, and ``build-base`` keywords.
+
+See the :doc:`Platforms reference </reference/platforms>` for more information.
+
```
medubelko, line 8: To be pedantic, the first section should probably be a simple task about setting the `base`, and ignore `platforms` entirely.

docs/howto/build-guides/platforms.rst
```diff
@@ -0,0 +1,169 @@
...
+Create a set of charms for multiple OS releases
+-----------------------------------------------
+
+A charm can only run on a single OS release or ``base``, using
+``charmcraft.yaml`` nomenclature. A ``charmcraft.yaml`` can use multi-base
+syntax to create a set of charms, each for a different base. To do this, the
+base is defined in each platform entry instead of being defined with the
+top-level ``base`` and ``build-base`` keywords.
+
+To build a charm for Ubuntu 22.04 and a charm for Ubuntu 24.04, use the
```
medubelko, line 140:
```suggestion
To build a set of two charms, one for Ubuntu 22.04 LTS and the other for Ubuntu 24.04 LTS, use the
```

docs/howto/build-guides/platforms.rst
```diff
@@ -0,0 +1,169 @@
...
+       build-for: [ubuntu@24.04:amd64]
+
+The ``build-on`` and ``build-for`` entries are identical for each platform, so
+the :ref:`multi-base shorthand notation
+<reference-platforms-multi-base-shorthand>` can be used:
+
+.. code-block:: yaml
+
+   platforms:
+     ubuntu@22.04:amd64:
```
medubelko, line 161: In the text we ought to point out the trailing colons for these entries, since they're easy to miss.

docs/reference/platforms.rst
medubelko: +1 to @lengau's suggestion. You could also use letters (or nonsense words) instead of numbers to less strongly imply there's a sequence.

docs/reference/platforms.rst
```diff
@@ -0,0 +1,218 @@
...
+``<base>:<arch>``.
+
+.. _reference-build-on:
+
+``build-on``
+~~~~~~~~~~~~
+
+``build-on`` is a list of architectures and optional bases that describes the
+environments where the platform can build. Each entry is formatted as
+``[<base>:]<arch>``.
```
medubelko, line 88: This notation twists me in light of `build-for: [<base>:<arch>]` and `<base>:<arch>:`. I don't see any examples where we use it. Are the square brackets literal?

docs/reference/platforms.rst
```diff
@@ -0,0 +1,218 @@
...
+
+The following architectures are supported:
+
+* amd64
+* arm64
+* armhf
+* i386
+* ppc64el
+* riscv64
+* s390x
```
medubelko, line 117: It might be worth mapping these to human-friendly names in a table. I compiled the list in the [Snapcraft system reqs](https://github.com/canonical/snapcraft/blob/685a8d61a6f378179334a12ab868ddc414e7bd41/docs/reference/system-requirements.rst?plain=1#L14).

docs/reference/platforms.rst
```diff
@@ -0,0 +1,218 @@
...
+
+.. _reference-bases:
+
+Bases
+-----
+
+.. note::
```
medubelko, line 124: I think this should be `important`. :)

docs/howto/build-guides/platforms.rst
```diff
@@ -0,0 +1,169 @@
+Build platform-specific charms
+==============================
+
+A ``charmcraft.yaml`` defines where charms are built and where charms are run
+using the ``platforms``, ``base``, and ``build-base`` keywords.
```
medubelko, line 5: Again being pedantic, but we don't show usage of `build-base` in this doc.

docs/reference/platforms.rst
```diff
@@ -0,0 +1,218 @@
...
+| 2      | amd64-debug   | amd64         | Ubuntu 24.04    | amd64          | Ubuntu 24.04 |
++--------+---------------+---------------+-----------------+----------------+--------------+
+| 3      | riscv64-cross | amd64         | Ubuntu 24.04    | riscv64        | Ubuntu 24.04 |
++--------+---------------+---------------+-----------------+----------------+--------------+
+| 4      | riscv64-cross | riscv64       | Ubuntu 24.04    | riscv64        | Ubuntu 24.04 |
++--------+---------------+---------------+-----------------+----------------+--------------+
+
+If Charmcraft executes on an ``riscv64`` system, it filters the build plan to
+only builds with a ``build-on`` of ``riscv64``. This means Charmcraft will only
+build charm #4.
```
medubelko, line 210:
```suggestion
build the fourth charm.
```

docs/reference/platforms.rst
```diff
@@ -0,0 +1,218 @@
...
+| 1      | amd64         | amd64         | Ubuntu 24.04    | amd64          | Ubuntu 24.04 |
++--------+---------------+---------------+-----------------+----------------+--------------+
+| 2      | amd64-debug   | amd64         | Ubuntu 24.04    | amd64          | Ubuntu 24.04 |
++--------+---------------+---------------+-----------------+----------------+--------------+
+| 3      | riscv64-cross | amd64         | Ubuntu 24.04    | riscv64        | Ubuntu 24.04 |
++--------+---------------+---------------+-----------------+----------------+--------------+
+| 4      | riscv64-cross | riscv64       | Ubuntu 24.04    | riscv64        | Ubuntu 24.04 |
++--------+---------------+---------------+-----------------+----------------+--------------+
+
+If Charmcraft executes on an ``riscv64`` system, it filters the build plan to
```
medubelko, line 208:
```suggestion
If Charmcraft executes on a ``riscv64`` system, it filters the build plan to
```

docs/reference/platforms.rst
```diff
@@ -0,0 +1,218 @@
...
+only builds with a ``build-on`` of ``riscv64``. This means Charmcraft will only
+build charm #4.
+
+If Charmcraft executes on an ``amd64`` system, it will build charms #1, #2, and
+#3. This can be further filtered with the ``--platform`` argument or the
```
medubelko, line 213:
```suggestion
If Charmcraft executes on an ``amd64`` system, it will build charms one, two, and
three. This can be further filtered with the ``--platform`` argument or the
```

medubelko (changes requested):
Since this is a critical topic, I've made some fairly in-depth suggestions.

I'm probably late to the party, but some of my points relate to the platform-architecture question, and affect the page and section titles. For clearer language that mimics natural speech, I'm leaning toward the notion of "build architecture" and "target architecture" (Docker uses this pair). I hope these proposals won't lead to blockages. If they do, feel free to ignore them.

I didn't give every line the same amount of attention, but I'm sure JJ and I will revisit these pages more than once later.

docs/howto/build-guides/platforms.rst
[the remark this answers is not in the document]
lengau: I've changed this to "Customise charm platforms" and added a separate issue for making our documentation more uniform. https://github.com/canonical/charmcraft/issues/2086

docs/reference/platforms.rst
lengau: I've adjusted this here to say "my-arch", "your-arch" and "our-arch", as well as the ellipsis. I chose the language because as far as I know it's fairly universal that those aren't ordered, but I'm leaving this open for @medubelko 's comments if that's not appropriate.

docs/howto/build-guides/platforms.rst
lengau: https://github.com/canonical/charmcraft/issues/2087

docs/howto/build-guides/platforms.rst
lengau: Removed for now, and added an explicit documentation notice to https://github.com/canonical/charmcraft/pull/2080/files#r1919338151

docs/howto/build-guides/platforms.rst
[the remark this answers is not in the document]
lengau: Actually based on the [other suggestion](https://github.com/canonical/charmcraft/pull/2080/files#r1917563465) I've changed this to "Select charm platforms".

docs/howto/build-guides/platforms.rst
```diff
@@ -0,0 +1,169 @@
+Build platform-specific charms
```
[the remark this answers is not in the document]
lengau: Closing this to keep discussion together. https://github.com/canonical/charmcraft/pull/2080/files#r1920598339

docs/howto/build-guides/platforms.rst
lengau: It defines more than that. New language coming up.

docs/howto/build-guides/platforms.rst
```diff
@@ -0,0 +1,169 @@
...
+==============================
+
+A ``charmcraft.yaml`` defines where charms are built and where charms are run
+using the ``platforms``, ``base``, and ``build-base`` keywords.
+
+See the :doc:`Platforms reference </reference/platforms>` for more information.
+
+Create a charm for a specific base and architecture
```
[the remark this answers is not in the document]
lengau: `architecture` means only the silicon. `platform` can mean only that, but can mean more, depending.

docs/howto/build-guides/platforms.rst
```diff
@@ -0,0 +1,169 @@
...
+  The OS of the build system doesn't affect what is built. Using LXD or
+  Multipass as a backend, Charmcraft will create a build environment with the
+  correct OS.
+
+Create a charm for all architectures
+------------------------------------
+
+For charms that are architecture-independent, use the following
+``charmcraft.yaml`` snippet to create a single charm that can run on any
+architecture:
```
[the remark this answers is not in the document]
lengau: This might imply that you can make a charm architecture-independent just by setting `build-for: [all]`. I've started with your language and adjusted it a bit.

docs/howto/build-guides/platforms.rst
```diff
@@ -0,0 +1,169 @@
...
+   platforms:
+     all:
+       build-on: [amd64, riscv64]
+       build-for: [all]
+
+Building on ``amd64`` or ``riscv64`` will produce one charm that runs on all
+architectures on Ubuntu 24.04 systems.
+
+
+Create a set of charms for multiple architectures
```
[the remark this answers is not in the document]
lengau: I've made this "Select multiple..." to distinguish from the different architectures below.

docs/reference/platforms.rst
medubelko: Let's give it a try!

docs/howto/build-guides/platforms.rst
lengau: I'd rather not use the term "bases" if possible since it could confuse people about the deprecated `bases` keyword that `base` and `platforms` replace. For now I've gone with "Select multiple OS releases", but we should discuss this further.

docs/howto/build-guides/platforms.rst
lengau: I don't want to imply that these are different charms. (I missed this language in my earlier review - sorry!) We already have charm authors complaining that this is going to require separate revisions depending on the base.

For right now I've changed the language to say "`.charm` file" where it refers to the outcome of the pack step, but I want to get people used to the fact that the same charm version can result in multiple "binaries" (<-- not the right word either, but philosophically equivalent if you think of `charmcraft pack` as compiling) depending on the base and architecture.

We should probably set aside time to come up with specific language for this and put it in a glossary, so I've made https://github.com/canonical/charmcraft/issues/2089 for that, but that doesn't answer the question in the short term :-)

docs/howto/build-guides/platforms.rst
lengau: I agree, but where in the text? We show the shorthand notation in several places. Do we want to highlight it after each sample? (Maybe best if people are coming to a specific section for reference, but repetitive if they're reading the entire document.) After only the first sample? (The opposite problem)

Your call, but for the moment I haven't made any changes to this.

docs/reference/platforms.rst
lengau: Yeah, this is hard to define as-is. I've added an English description of the format and some examples to help disambiguate.

docs/howto/build-guides/platforms.rst
lengau: I've also added a description of this in the reference file.

docs/reference/platforms.rst
lengau: I've also converted the examples here to use YAML's multi-line sequence syntax so we shouldn't have any load-bearing brackets

lengau:
@medubelko requesting you again as I've got some follow-up questions. I haven't yet handled all the changes.

docs/howto/build-guides/platforms.rst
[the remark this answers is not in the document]
sergiusens: FWIW, I said platform is the combination of base OS and hardware features, architecture being one part of a hardware feature.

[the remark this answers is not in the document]
medubelko: Thanks for the correction!

sergiusens (approved):

docs/howto/build-guides/select-platforms.rst
```diff
@@ -0,0 +1,196 @@
+.. _select-platforms:
+
+Select charm platforms
+======================
+
+The ``platforms`` and ``base`` keys in a charm's :ref:`charmcraft-yaml-file` selects
```
medubelko, line 6:
```suggestion
The ``platforms`` and ``base`` keys in a charm's :ref:`charmcraft-yaml-file` select
```

docs/howto/build-guides/select-platforms.rst
```diff
@@ -0,0 +1,196 @@
...
+
+.. code-block:: yaml
+
+    base: ubuntu@24.04
+
+Select one build and one target architecture
+--------------------------------------------
+
+The following ``charmcraft.yaml`` snippet will define a charm that both builds and
+runs on Ubuntu 24.04 LTS, only on an amd64 architecture:
```
medubelko, line 30:
```suggestion
runs on Ubuntu 24.04 LTS, only on an AMD64 architecture:
```

There are a few instances elsewhere where it's lowercase.

docs/howto/build-guides/select-platforms.rst
```diff
@@ -0,0 +1,196 @@
...
+Select multiple OS releases
+---------------------------
+
+The resulting ``.charm`` file packed by charmcraft can only run on a single OS release
+or ``base``, using ``charmcraft.yaml`` nomenclature. A ``charmcraft.yaml`` can use
+multi-base syntax to create a set of charms, each for a different base. To do this, the
+base is defined in each platform entry instead of being defined with the
+top-level ``base`` and ``build-base`` keys.
+
+To build a charm for Ubuntu 22.04 and a charm for Ubuntu 24.04, use the
```
medubelko, line 167:
```suggestion
To build a charm for Ubuntu 22.04 LTS and a charm for Ubuntu 24.04 LTS, use the
```

In the last paragraph, too.

docs/howto/build-guides/select-platforms.rst
```diff
@@ -0,0 +1,196 @@
...
+
+.. code-block:: yaml
+
+   platforms:
+     ubuntu@22.04:amd64:
+     ubuntu@24.04:amd64:
+
+With both snippets, building on amd64 will produce two charms, one for
+amd64 systems running Ubuntu 22.04 and one for amd64 systems running
+Ubuntu 24.04.
```
medubelko, line 193:
```suggestion
With both snippets, building on AMD64 will produce two charms, one for
AMD64 hosts running Ubuntu 22.04 LTS and one for AMD64 hosts running
Ubuntu 24.04 LTS.
```

Since you changed it to "hosts" earlier in the page.

docs/howto/build-guides/select-platforms.rst
```diff
@@ -0,0 +1,196 @@
...
+Select charm platforms
+======================
+
+The ``platforms`` and ``base`` keys in a charm's :ref:`charmcraft-yaml-file` selects
+both the operating system releases and silicon architectures on which the charm
+can build and run.
+
+See the :doc:`Platforms reference </reference/platforms>` for more information.
+
+Select an operating system
```
medubelko, line 12:
```suggestion
Select an OS
```

For consistency.

docs/howto/build-guides/select-platforms.rst
```diff
@@ -0,0 +1,196 @@
+.. _select-platforms:
+
+Select charm platforms
+======================
+
+The ``platforms`` and ``base`` keys in a charm's :ref:`charmcraft-yaml-file` selects
+both the operating system releases and silicon architectures on which the charm
```
medubelko, line 7:
```suggestion
both the OS releases and silicon architectures on which the charm
```

docs/howto/build-guides/select-platforms.rst
```diff
@@ -0,0 +1,196 @@
...
+can build and run.
+
+See the :doc:`Platforms reference </reference/platforms>` for more information.
+
+Select an operating system
+--------------------------
+
+The :ref:`charmcraft-yaml-base` key in ``charmcraft.yaml`` defines the operating
+system and version on which a charm can build and run. Supported operating systems
```
medubelko, line 16:
```suggestion
system and version on which a charm can build and run. Supported OSs
```

docs/howto/build-guides/platforms.rst
medubelko: Understood. I was a bit confused by the original text as well.

Long-term, I also prefer that users understand the correct mental model and so we can continue treating "charm" as a first-class term the same way we do "snap".

docs/howto/build-guides/platforms.rst
medubelko: Let's leave it for now. This could be a case where having a shiny note somewhere high-up in the page would be valuable.

docs/reference/platforms.rst
medubelko: Ah, so they're not literal. That's a tricky one to format, since YAML uses square brackets for inline collections. Gah. I've suggested a workaround in my upcoming review.

docs/reference/platforms.rst
```diff
@@ -0,0 +1,245 @@
...
+``<base>:<arch>``.
+
+.. _reference-build-on:
+
+``build-on``
+~~~~~~~~~~~~
+
+``build-on`` is a list of architectures and optional bases that describes the
+environments where the platform can build. Each entry is formatted with an optional
+base and a mandatory architecture: ``[<base>:]<arch>``
```
medubelko, line 98:
```suggestion
environments where the platform can build. Entries can be formatted either as only
the architecture (``<arch>``), or as base and architecture (``<base>:<arch>``).
```

medubelko (changes requested):
@lengau I like the changes you've made. I've closed the ones we can't handle for now. The rest of these are mechanical improvements which should be easy to do, with one suggestion to resolve a semantic question. If you're short on time I can commit them today.

docs/howto/build-guides/select-platforms.rst
lengau: IMO this language is cumbersome. Not your fault I know... I just want to complain

docs/howto/build-guides/select-platforms.rst
lengau: I'm distinguishing here the build host from the running system, which is why I kept these as systems.

docs/reference/platforms.rst
lengau: Turned it into a table, removed i386 (which Charmcraft doesn't support), and added a note about Juju support. Sadly there doesn't seem to be a Juju architecture support doc I can reference.

lengau (approved):
I feel cheeky approving my own changes, but I'm hitting approve here so we have ample approvals for a merge.

docs/howto/build-guides/select-platforms.rst
medubelko: Non-blocking, but what makes runtime different? Don't charms run on every kind of machine?

medubelko (approved):

docs/reference/platforms.rst
```diff
@@ -0,0 +1,259 @@
...
+
+   platforms:
+     <platform-name>:
+       build-on:
+        - <some-base>:<some-arch>
+        - <some-base>:<another-arch>
+        - <another-base>:<some-arch>
+        - <another-base>:<another-arch>
+       build-for:
+        - <base>:<arch>
```
mr-cal, line 64: I think using `some-base`, `another-base`, and `base` may be misleading because craft-platforms will fail to parse a platform entry with multiple bases as described a few lines below with:

> Within each platform entry, the base must be the same for all ``build-on`` and ``build-for`` entries.

There will probably be an exception to this rule once we get closer to 26.04 and need to support `devel` bases.

docs/reference/platforms.rst
```diff
@@ -0,0 +1,259 @@
...
+      - amd64
+      - riscv64
+
+Example ``build-on`` values containing both a base and an architecture include:
+
+.. code:: yaml
+
+    build-on:
+      - ubuntu@24.04:riscv64
+      - ubuntu@22.04:amd64
```
mr-cal, line 114: Similarly, this entire snippet would fail but we could split this into 2 discrete snippets:

```
.. code:: yaml
    build-on:
      - ubuntu@24.04:riscv64
```
```
.. code:: yaml
    build-on:
      - ubuntu@22.04:amd64

File: docs/reference/platforms.rst [hunk 34]
```diff
@@ -0,0 +1,259 @@
...
+
+.. _reference-bases:
+
+Bases
+-----
+
+.. important::
+
+   The bases described in this section are a different concept than the
+   deprecated ``bases`` keyword in a ``charmcraft.yaml``.
```
mr-cal, line 168: > Can we run a find-replace for "keyword" → "key" in this doc?

That comment was for the how-to, should we do it in the reference as well?

mr-cal (commented):
The edits look great. I added a few minor thoughts that I'm happy to submit as a follow-up PR.

PR #2082, opened 2025-01-16
feat(store): promote command

### LINKED ISSUE #1633: `charmcraft promote-charm` command
### What needs to get done

Promote a charm from one channel to another in a command.

This would be a charm version of `promote-bundle`.

### Why it needs to get done

To match snapcraft and developer convenience.

lengau:
This adds a `promote` command allowing promotion from one channel to another, with limitations (e.g. cannot promote to a less stable channel).

Fixes #1633


charmcraft/application/commands/store.py
```diff
@@ -830,6 +833,141 @@ def run(self, parsed_args):
...
+                from_channel = dataclasses.replace(from_channel, track=default_track)
+            if to_channel.track is None:
+                to_channel = dataclasses.replace(to_channel, track=default_track)
+
+        if to_channel == from_channel:
+            raise CraftError(
+                "Cannot promote from a channel to the same channel.",
+                retcode=64,  # Replace with os.EX_USAGE once we drop Windows.
+            )
+        if to_channel.risk > from_channel.risk:
```
dariuszd21, line 906: I'm not sure if it will work, as risk is just a `string`:
```
>>> 'stable' > 'edge'
True
>>> 'beta' > 'edge'
False
>>> 'candidate' > 'edge'
False
```

charmcraft/application/commands/store.py
```diff
@@ -830,6 +833,141 @@ def run(self, parsed_args):
...
+        ]
+        emit.progress(
+            f"The following revisions are on the {from_channel.name} channel:",
+            permanent=True,
+        )
+        with emit.pause():
+            print(
+                tabulate(presentable_candidates, tablefmt="plain", headers="keys"),
+                file=sys.stderr,
+            )
```
dariuszd21, line 954: Any particular reason why it can't be a progress as well (beside not logging it) ?

charmcraft/application/commands/store.py
```diff
@@ -830,6 +833,141 @@ def run(self, parsed_args):
...
+        if not parsed_args.yes and not utils.confirm_with_user(
+            f"Do you want to promote these revisions to the {to_channel.name} channel?"
+        ):
+            emit.message("Channel promotion cancelled.")
+            return 1
```
dariuszd21, line 959: Isn't `craft_cli` supporting `.confirm()` already or it wasn't released yet ?

uv.lock
```diff
@@ -463,7 +462,7 @@ requires-dist = [
     { name = "craft-platforms", specifier = "~=0.5" },
     { name = "craft-providers", specifier = ">=2.0.0" },
     { name = "craft-providers", specifier = ">=2.1.0" },
-    { name = "craft-store", specifier = ">=3.1.0" },
+    { name = "craft-store", git = "https://github.com/canonical/craft-store" },
```
dariuszd21, line 465: Reminder to update to the proper release

dariuszd21 (changes requested):
Just few minor comments but nothing major

charmcraft/application/commands/store.py
lengau: `emit.progress` tries to represent it as a single line, even with `permanent=True`.

charmcraft/application/commands/store.py
lengau: Not yet released. Once it is I'll move all cases over and delete the `confirm_with_user` function entirely.

charmcraft/application/commands/store.py
```diff
@@ -830,6 +833,141 @@ def run(self, parsed_args):
         emit.message(msg.format(*args))
 
 
+class PromoteCommand(CharmcraftCommand):
+    """Promote a charm in the Store."""
+
+    name = "promote"
+    help_msg = "Promote a charm from one channel to another on Charmhub."
+    overview = "TODO"
```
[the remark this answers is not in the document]
lengau: Will do - I've asked for your second review for other things ITMT

charmcraft/application/commands/store.py
dariuszd21: Got it, thanks for the explanation!

charmcraft/application/commands/store.py
dariuszd21: Would it work this way ? It seems to be a bit more consistent and does not leak the `sys.stderr` implementation detail
```suggestion
        with emit.open_stream() as stream:
            print(
                tabulate(presentable_candidates, tablefmt="plain", headers="keys"),
                file=stream,
            )
```

dariuszd21 (commented):
Thank you!

charmcraft/application/commands/store.py
```diff
@@ -830,6 +833,141 @@ def run(self, parsed_args):
...
+        to_channel = charmcraft.store.models.ChannelData.from_str(
+            parsed_args.to_channel
+        )
+        if None in (from_channel.track, to_channel.track):
+            package_metadata = store.get_package_metadata(name)
+            default_track = package_metadata.default_track
+            if from_channel.track is None:
+                from_channel = dataclasses.replace(from_channel, track=default_track)
+            if to_channel.track is None:
+                to_channel = dataclasses.replace(to_channel, track=default_track)
```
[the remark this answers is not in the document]
cmatsuoka: If both are missing, will it try to promote from default to default?

cmatsuoka (approved):
The implementation looks reasonable under my assumptions, but I missed the complete context because the PR has no description.

charmcraft/application/commands/store.py
```diff
@@ -830,6 +833,141 @@ def run(self, parsed_args):
...
+                f"than the source channel ({from_channel.name}).",
+                resolution=f"Did you mean: {command}",
+            )
+        if to_channel.track != from_channel.track:
+            if not parsed_args.yes and not utils.confirm_with_user(
+                "Did you mean to promote to a different track? (from "
+                f"{from_channel.track} to {to_channel.track})",
+            ):
+                emit.message("Cancelling.")
+                return 64  # Replace with os.EX_USAGE once we drop Windows.
```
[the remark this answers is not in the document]
lengau: I don't see why not. As an example, in CI you might want to promote `3/edge` to `latest/edge`

charmcraft/application/commands/store.py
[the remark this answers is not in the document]
lengau: I don't see why we would want to error out rather than using the default track. An example command that would trigger both of these conditions is:

```
charmcraft promote --from-channel=candidate --to-channel=stable
```

In this case, we'll look up the default track (typically `latest`) and insert it, so the command above is equivalent to:

```
charmcraft promote --from-channel=latest/candidate --to-channel=latest/stable
```

if the default track is `latest`.

charmcraft/application/commands/store.py
lengau: `open_stream()` doesn't result in a file-like object, just an integer with the file pointer, so you'd have to then open that, and it gets ugly.

dariuszd21: You can probably do `.os.write()` then instead of print, but it's up to you.

I did emit.progress() previously on multi line output and it worked, that's why I asked in the first place.

charmcraft/application/commands/store.py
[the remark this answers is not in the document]
dariuszd21: What if there are multiple tracks like this ?
```
24_04_13.2
22_04_13.2
20_04_11.1
``` 

Are you sure that you know what user has intended by that ?

charmcraft/application/commands/store.py
```diff
@@ -830,6 +833,152 @@ def run(self, parsed_args):
...
+    name = "promote"
+    help_msg = "Promote a charm from one channel to another on Charmhub."
+    overview = textwrap.dedent(
+        """Promote a charm from one channel to another on Charmhub.
+
+        Promotes the current revisions of a charm in a specific channel, as well as
+        their related resources, to another channel.
+
+        The most common use is to promote a charm to a more stable risk value on a
+        single track:
```
dariuszd21, line 848:
```suggestion
        single track::
```

charmcraft/application/commands/store.py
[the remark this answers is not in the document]
dariuszd21: Going further with your example, you can also promote `latest/edge` to `3/stable` without a confirmation

charmcraft/application/commands/store.py
```diff
@@ -830,6 +833,152 @@ def run(self, parsed_args):
...
+        )
+        parser.add_argument(
+            "--yes",
+            default=False,
+            action="store_true",
+            help="Answer yes to all questions.",
+        )
+
+    @override
+    def run(self, parsed_args: argparse.Namespace) -> int | None:
```
dariuszd21, line 887: I've also noticed that this command has plenty of logic, but there is not a single test that covers it.

dariuszd21 (changes requested):
Besides my different stand on allowing "implicit" track, I think this is

charmcraft/application/commands/store.py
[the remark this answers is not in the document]
lengau: You can, but again this is specifically an opt-in option to override the interactions for running in a non-interactive mode.

charmcraft/application/commands/store.py
lengau: I used `emit.progress` initially and it didn't work. I don't want to use `os.write` here because it requires handling a bunch of things that `print` already handles, so if we really wanted to use the stream I'd probably use `open` to create a file object from that descriptor and then `print` to that.

charmcraft/application/commands/store.py
lengau: Good point - with the latest craft-cli I can add the backticks.

charmcraft/application/commands/store.py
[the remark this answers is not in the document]
lengau: Per separate conversation, I'm changing this so `--yes` requires fully-defined tracks.

charmcraft/application/commands/store.py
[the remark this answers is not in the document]
lengau: @sergiusens mentioned that he doesn't like this, but I don't know whether this is going to be useful without the user being able to do cross-channel promotions in non-interactive mode.

Since we're no longer inferring the default branch in non-interactive mode though, I think this is now safer. If we still don't think that's safe enough, what about only allowing cross-track promotions to occur along the same risk? e.g. you can promote from `latest/stable` to `3/stable`, but you can't promote from `latest/candidate` to `3/stable`

charmcraft/application/commands/store.py
```diff
@@ -830,6 +833,167 @@ def run(self, parsed_args):
...
+        emit.progress(
+            f"{self._app.name} {self.name} does not have a stable CLI interface. "
+            "Use with caution in scripts.",
+            permanent=True,
+        )
+        store = cast(StoreService, self._services.get("store"))
+
+        name = parsed_args.name or self._services.project.name
+
+        # Check snapcraft for equiv logic
```
mr-cal, line 897: leftover todo?

charmcraft/application/commands/store.py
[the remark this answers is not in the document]
mr-cal: > Since we're no longer inferring the default branch in non-interactive mode though, I think this is now safer.

Agreed, I'm happy with this.  I frequently run `snapcraft promote snapcraft --from-channel 8.x/candidate --to-channel latest/stable`.

pyproject.toml
```diff
@@ -11,7 +11,8 @@ dependencies = [
     "craft-providers>=2.1.0",
     "craft-platforms~=0.5",
     "craft-providers>=2.0.0",
-    "craft-store>=3.1.0",
+    # "craft-store>=3.1.0",
+    "craft-store @ git+https://github.com/canonical/craft-store",
```
mr-cal, line 15: reminder to bump before landing

mr-cal (approved):
Looks great!

charmcraft/application/commands/store.py
```diff
@@ -830,6 +833,167 @@ def run(self, parsed_args):
...
+    name = "promote"
+    help_msg = "Promote a charm from one channel to another on Charmhub."
+    overview = textwrap.dedent(
+        """Promote a charm from one channel to another on Charmhub.
+
+        Promotes the current revisions of a charm in a specific channel, as well as
+        their related resources, to another channel.
+
+        The most common use is to promote a charm to a more stable risk value on a
+        single track:
```
dariuszd21, line 848:
```suggestion
        """
        Promote a charm from one channel to another on Charmhub.

        Promotes the current revisions of a charm in a specific channel, as well as
        their related resources, to another channel.

        The most common use is to promote a charm to a more stable risk value on a
        single track:
```

I noticed that during the demo and I can see it in documentation as well, I think to make `dedent` work correctly you have to have same indentation in all lines.
It's a bit weirdly rendered here right now
https://canonical-charmcraft--2082.com.readthedocs.build/en/2082/reference/commands/promote/

dariuszd21 (approved):
Looks good, Thank you!

charmcraft/application/commands/store.py
lengau: Good catch, thanks. Dedent doesn't require the same indentation everywhere, but it does use the first line's preceding whitespace to determine what to remove. That means the actual correct way is:

```suggestion
        """\
        Promote a charm from one channel to another on Charmhub.

        Promotes the current revisions of a charm in a specific channel, as well as
        their related resources, to another channel.

        The most common use is to promote a charm to a more stable risk value on a
        single track:
```

charmcraft/application/commands/store.py
lengau:
```suggestion
```

Indood!

PR #2083, opened 2025-01-17
docs: translate remaining .md files to .rst

jahn-junior:
Translated remaining Markdown files from #2010 to reStructuredText.

The contents of each page remain largely unchanged, with the main exceptions being:
- Dropdowns in [Manage a 12-factor app charm](https://canonical-charmcraft.readthedocs-hosted.com/en/stable/howto/manage-a-12-factor-app-charm/) were replaced with tabs.
- Output of `juju status` in [tutorials](https://canonical-charmcraft.readthedocs-hosted.com/en/stable/tutorial/) was changed from a bash code directive to a terminal directive.


docs/reference/files/manifest-yaml-file.rst
```diff
@@ -0,0 +1,19 @@
+.. _manifest-yaml-file:
+
+
+``manifest.yaml`` file
+======================
+
+    See also: `How to configure
+    Charmcraft <https://juju.is/docs/sdk/charmcraft-config>`_
```
lengau, line 8: This should probably instead go to our internal howto/manage-charms document now, which is the successor to that document.

docs/reference/files/metadata-yaml-file.rst
```diff
@@ -0,0 +1,896 @@
...
+    automatically from information you provide in the :ref:`charmcraft-yaml-file`. For
+    backwards compatibility, Charmcraft will continue to allow the use of the
+    ``metadata.yaml`` file, but you may not duplicate keys across the two files.
+
+.. important::
+
+    The ``charmcraft.yaml`` file follows keywords similar to snaps. On the other hand, the
+    ``metadata.yaml`` does not. As a result, when ``charmcraft pack`` generates the file, some
+    keywords may differ -- e.g., the ``title``  key from ``charmcraft.yaml`` is rendered as
+    ``display-name`` in ``metadata.yaml``.
```
lengau, line 24:
```suggestion
    The ``charmcraft.yaml`` file follows keywords similar to snaps, but 
    ``metadata.yaml`` does not. As a result, when ``charmcraft pack`` generates 
    the file, some keys may differ -- e.g., the ``title``  key from ``charmcraft.yaml``
    is rendered as ``display-name`` in ``metadata.yaml``.
```

docs/reference/files/metadata-yaml-file.rst
```diff
@@ -0,0 +1,896 @@
+.. _metadata-yaml-file:
+
+
+``metadata.yaml`` file
+======================
+
+.. The last version of this doc to show all the keys at once, in a long spec with
+.. comments, was version 22.
```
lengau, line 8:
```suggestion
```

This is no longer relevant since that discourse history is going to disappear.

docs/reference/files/metadata-yaml-file.rst
```diff
@@ -0,0 +1,896 @@
...
+
+    .. code-block:: yaml
+
+        name: mongodb
+        summary: A document database
+        description: A database that stores JSON-like data.
+
+.. dropdown:: ``metadata.yaml`` for a Kubernetes charm
+
+    .. code-block:: yaml
```
lengau, line 49: The minimal file above is probably not worth separating, but this example and the full spec may be better off as their own yaml files that we use `literalinclude`s for. Amongst other things, they'll get linting applied to them by yamllint.

docs/reference/files/metadata-yaml-file.rst
```diff
@@ -0,0 +1,896 @@
...
+.. code-block:: yaml
+
+    assumes:
+        - <feature_name>
+        - any-of:
+            - <feature_name>
+            - <feature_name>
+        - all-of:
+            - <feature_name>
+            - <feature_name>
```
lengau, line 385: This is rendering incorrectly:

image

docs/reference/files/metadata-yaml-file.rst
```diff
@@ -0,0 +1,896 @@
...
+
+**Structure:** This key consists of a list of containers along with their specification.
+Each container can be specified in terms of ``resource``, ``bases``, ``uid``, ``gid`` and
+``mounts``, where one of either the ``resource`` or the ``bases`` subkeys must be defined, and
+``mounts`` is optional. ``resource`` stands for the OCI image resource used to create the
+container; to use it, specify  an OCI image resource name (that you will then define
+further in the `resources`_ block). ``bases`` is a list of bases to be used
+for resolving a container image, in descending order of preference; to use it, specify a
+base name (for example, ``ubuntu``, ``centos``, ``windows``, ``osx``, ``opensuse``), a
+[channel](https://snapcraft.io/docs/channels), and an architecture. ``mounts`` is a list
```
lengau, line 477: [Hey! You got your Markdown in my rST!](https://www.youtube.com/watch?v=O7oD_oX-Gio)

docs/reference/files/metadata-yaml-file.rst
```diff
@@ -0,0 +1,896 @@
...
+
+
+``Other keys``
+--------------
+
+In addition to the keys mentioned above, a ``metadata.yaml`` file may also contain other
+arbitrary keys. These can serve to keep track of other choices a charmer might make. In
+some cases these become semi-official, being adopted by many charmers and even
+incorporated into CI processes. An example is
+:literalref:`upstream source <https://github.com/canonical/charmcraft/blob/b22fcdba3b894004468abfbf45caa54d93fbf7d0/charmcraft/templates/init-simple/metadata.yaml.j2#L40-L43>`_.
```
lengau, line 896:
```suggestion
:literalref:`upstream source <https://github.com/canonical/charmcraft/blob/b22fcdba3b894004468abfbf45caa54d93fbf7d0/charmcraft/templates/init-simple/metadata.yaml.j2#L40-L43>`.
```

docs/reference/files/metadata-yaml-file.rst
```diff
@@ -0,0 +1,896 @@
...
+    terms:
+        - <term>
+
+
+``website``
+-----------
+
+.. important::
+
+    In :ref:`recipes <charmcraft-yaml-file>`, this is now the ``links.website`` key.
```
lengau, line 879:
```suggestion
    In :literalref:`charmcraft.yaml <charmcraft-yaml-file>`, this is now the ``links.website`` key.
```

docs/reference/files/pyproject-toml-file.rst
```diff
@@ -0,0 +1,62 @@
...
+=======================
+
+The ``pyproject.toml`` file in your charm's root directory is a typical
+Python ``pyproject.toml`` file.
+
+    See more: `pip |
+    pyproject.toml <https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/>`_
+
+This file is generated automatically by ``charmcraft init`` with the
+contents below:
```
lengau, line 14: This could vary by which profile one uses. IDK if it's possible, but something that would be **amazing** would be if you were to include (probably under a dropdown or with tabs?) each actual Jinja template for the file based on the profile. E.g. for the [django-framework profile](https://github.com/canonical/charmcraft/blob/main/charmcraft/templates/init-django-framework/pyproject.toml.j2)

docs/reference/files/requirements-txt-file.rst
```diff
@@ -0,0 +1,13 @@
...
+
+
+``requirements.txt`` file
+=========================
+
+The ``requirements.txt`` file is a standard Python `requirements file
+<https://pip.pypa.io/en/stable/reference/pip_install/#requirements-file-format>`_
+used to declare and pin the version of any Python libraries required by
+a charm in production. This will be pre-populated with
+:literalref:`ops <https://ops.readthedocs.io/en/latest/>`_. Any dependencies
```
lengau, line 11:
```suggestion
:literalref:`ops <https://ops.readthedocs.io/en/latest/>`. Any dependencies
```

Ops should also probably be a reusable link in docs/reuse/links.txt tbh

docs/reference/files/src-charm-py-file.rst
```diff
@@ -0,0 +1,13 @@
...
+``src/charm.py`` file
+=====================
+
+The ``src/charm.py`` file is the default entry point for a charm. This
+file must be executable, and should include a
+`shebang <https://en.wikipedia.org/wiki/Shebang_(Unix)>`_ to indicate
+the desired interpreter. For many charms, this file will contain the
+majority of the charm code. It is possible to change the name of this
+file, but additional changes are then required to enable the charm to be
+built with Charmcraft.
```
lengau, line 13:
```suggestion
file, but additional changes are then required to build the charm with Charmcraft.
```

docs/reference/files/tests-integration-test-charm-py-file.rst
```diff
@@ -0,0 +1,52 @@
+.. _tests-integration-test-charm-py-file:
+
+
+``tests/integration/test_charm.py`` file
+========================================
+
+The ``tests/integration/test_charm.py`` file is the companion to
+``src/charm.py`` for integration testing.
+
+This file is created automatically by ``charmcraft init`` and it is
```
lengau, line 10: Not all profiles generate this file.

docs/reference/files/tests-unit-test-charm-py-file.rst
```diff
@@ -0,0 +1,41 @@
+.. _tests-unit-test-charm-py-file:
+
+
+``tests/unit/test_charm.py`` file
+=================================
+
+    See also: `How to write unit tests for a charm <https://juju.is/docs/sdk/write-a-unit-test-for-a-charm>`_
+
+The ``tests/unit/test_charm.py`` file is the companion to
```
lengau, line 9: Same here - not all profiles generate this file. I think specifically the 12-factor ones don't.

docs/reference/profile.rst
```diff
@@ -0,0 +1,19 @@
+.. _profile:
```
lengau, line 1: Out of scope for this PR, but this whole page is probably a prime target for deletion and inclusion into [a glossary file](https://github.com/canonical/charmcraft/issues/2089)

docs/tutorial/write-your-first-kubernetes-charm-for-a-django-app.rst
```diff
@@ -0,0 +1,698 @@
...
+    the actual application and disable debug mode. We will do this in the tutorial
+    for now and later demonstrate how we can set these to production-ready values.
+
+Test the deployment using ``curl`` to send a request to the root endpoint. The IP
+address is the ``Address`` listed in the ``Unit`` section of the ``juju status``
+output (e.g., ``10.1.157.80`` in the sample output above):
+
+.. code-block:: bash
+
+    curl 10.1.157:8000
```
lengau, line 431:
```suggestion
    curl 10.1.157.80:8000
```

lengau (changes requested):
Thanks for doing this!

My comments are a combination of specific requests and general observations, some of which may be better for a separate task.

docs/howto/manage-a-12-factor-app-charm.rst
```diff
@@ -0,0 +1,323 @@
...
+        will be mapped as ``FLASK_CONFIG_OPTION_NAME`` where the
+        option name will be converted to upper case, dashes will be
+        converted to underscores and the ``FLASK_`` prefix will be
+        added. In the example above, the ``token`` configuration will
+        be mapped as the ``FLASK_TOKEN`` environment variable. In
+        addition to the environment variable, the configuration is
+        also available in the Flask variable ``app.config`` without
+        the ``FLASK_`` prefix.
+
+            See also: `How to add configuration to a charm <https://juju.is/docs/sdk/config>`_,
```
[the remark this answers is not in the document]
medubelko: This will be converted to an intersphinx link with #2090.

docs/reference/extensions/go-framework-extension.rst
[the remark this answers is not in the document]
medubelko: Agreed, we should restructure this so the repetition isn't needed.

docs/reference/files/manifest-yaml-file.rst
medubelko: Agreed, I wasn't certain about this link in my initial review.

docs/reference/files/metadata-yaml-file.rst
medubelko: For the sake of time, we'll just leave as-is and come back to this later.

docs/reference/files/pyproject-toml-file.rst
medubelko: We would do this (probably in a different PR) but seven elements is too much for a tab container. Is there any way we could logically group them into multiple containers?

docs/reference/files/requirements-txt-file.rst
medubelko: This link is incorrect. We should be pointing to the actual reference page in ops, or this shouldn't be a literalref.

docs/reference/files/tests-integration-test-charm-py-file.rst
medubelko: Let's change it to "profiles other than the one for 12-factor apps automatically create..."

medubelko (changes requested):
We'll need to fix all the indentation for the YAML code blocks; it needs to be based on two spaces or it won't render correctly.

docs/reference/files/metadata-yaml-file.rst
medubelko: This is caused by accidental 4-space indentation.

docs/reference/files/metadata-yaml-file.rst
```diff
@@ -0,0 +1,896 @@
...
+.. code-block:: yaml
+
+    # (Optional) A string (or a list of strings) containing a link (or links) to project websites.
+    # In general this is likely to be the upstream project website, or the formal website for the
+    # charmed offering.
+    website: <url> | [<urls>]
+
+
+``Other keys``
+--------------
```
medubelko, line 890:
```suggestion
Other keys
----------
```

docs/reference/files/metadata-yaml-file.rst
medubelko, line 896:
```suggestion
incorporated into CI processes.
```

Let's cut this, because it isn't appropriate to provide such a specific example for an open-ended syntax.

docs/howto/manage-a-12-factor-app-charm.rst
```diff
@@ -4,13 +4,14 @@
 Manage a 12-factor app charm
 ============================
 
-    See also: `Juju | 12-factor app charms <https://juju.is/docs/juju/charmed-operator>`_
+    See also:
+    `Juju | 12-factor app charms <https://juju.is/docs/juju/charmed-operator>`_
```
lengau, line 8: If you add Juju to intersphinx like I did in https://github.com/canonical/charmcraft/pull/2095/files#diff-85933aa74a2d66c3e4dcdf7a9ad8397f5a7971080d34ef1108296a7c6b69e7e3:

```suggestion
    :external+juju:ref:`Juju | Charm <charm>`
```

docs/howto/manage-a-12-factor-app-charm.rst
```diff
@@ -98,8 +99,9 @@ token for a service. Add the configuration in ``charmcraft.yaml``:
         also available in the Flask variable ``app.config`` without
         the ``FLASK_`` prefix.
 
-            See also: `How to add configuration to a charm <https://juju.is/docs/sdk/config>`_,
-            `Configuration Handling -- Flask Documentation <https://flask.palletsprojects.com/en/3.0.x/config/>`_
+            See also: `How to add configuration to a
+            charm <https://juju.is/docs/sdk/config>`_, `Configuration Handling --
```
lengau, line 103: Likewise we can use intersphinx for juju here. I'm not sure it's worth doing for the Flask docs.

docs/reference/extensions/fastapi-framework-extension.rst
```diff
@@ -136,7 +136,7 @@ These endpoint definitions are as below:
 
 .. note::
 
-    The key optional with value ``False`` means that the charm will
+    The ``key optional`` with value ``False`` means that the charm will
```
lengau, line 139:
```suggestion
    The key ``optional`` with value ``False`` means that the charm will
```

docs/reference/files/pyproject-toml-file.rst
lengau: I'm going to close this for now since in practice it looks like these are the same, although this raises those questions for, say, `charmcraft.yaml` which does vary for every profile.

docs/howto/manage-a-12-factor-app-charm.rst
```diff
@@ -0,0 +1,323 @@
...
+Manage a 12-factor app charm
+============================
+
+    See also: `Juju | 12-factor app charms <https://juju.is/docs/juju/charmed-operator>`_
+
+
+Prepare an OCI image for a 12-factor app charm
+----------------------------------------------
+
+Follow this `how-to guide <https://documentation.ubuntu.com/rockcraft/en/latest/how-to/build-a-12-factor-app-rock>`_
```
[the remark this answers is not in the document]
tmihoc: For external pages it's a good idea to prefix the link. E.g., Rockcraft | ... (This is crucial for links between Juju, Charmcraft, and Ops, and a good idea in general.)

Note also that in the Juju world docs so far we have also in general followed a cross referencing pattern that is not only very transparent about the link target but also minimizes metalanguage. So, instead of "Follow this ...", just, on a separate line and set off from the main text: See more: ...

docs/reference/files/metadata-yaml-file.rst
tmihoc: This was added on purpose as an example of how thisv unofficial key is used by convention to point to this upstream source. 

Links to specific pieces of existing charm yaml and code from the charm ecosystem are a pattern we've deliberately tried to use in both Charmcraft and Ops docs as they are a lightweight yet very powerful way to make a point.

docs/howto/manage-a-12-factor-app-charm.rst
tmihoc: The intersphinx links are nice but it's crucial that we find a way to prefix them with the target documentation set on the pattern you see before the change ("Juju | ..."). Otherwise, given that we have same title stories in multiple documentation sets (a feature, not a bu, as I've argued in the past), the links become very confusing.

docs/howto/manage-a-12-factor-app-charm.rst
jahn-junior: I can't seem to find that specific doc on Juju's RTD. Was it moved or replaced by something more up-to-date?

docs/reference/extensions/fastapi-framework-extension.rst
jahn-junior: Yikes, that's the second time I've butchered that line.

docs/howto/manage-a-12-factor-app-charm.rst
tmihoc: All links to Juju or Ops docs should point to the new docs on RTD.

tmihoc: It's under manage applications > configure

docs/howto/manage-a-12-factor-app-charm.rst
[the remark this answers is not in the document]
jahn-junior: Rewrote to follow Juju doc conventions.

lengau (approved):
LGTM - the rest of the discussion is out of scope for me.

docs/howto/manage-a-12-factor-app-charm.rst
jahn-junior: Changed to an intersphinx link pointing to Ops documentation

docs/reference/files/metadata-yaml-file.rst
medubelko: We were wondering about this one. The linked code was removed from the repo a while back, so it looked to me like it was a historical artifact that's no longer current.

The main reason for removal was the link breaks the linter. There's no way around that at the moment, and GH doesn't have a built-in shortener.

Since the code isn't live in the repo and therefore the reader can't be reassured of its currency, I suggest we simply copy it verbatim here.

tmihoc: As I was saying, linking to real examples of charm code has been a pattern we've followed in other places in the docs too, on purpose and for a good reason -- it makes it possible to keep the docs quite generic and clean while still giving people a sense of what the content covered in the docs looks like in the wild. The fact that the linter doesn't like it doesn't seem to me a good reason to undo this pattern.

medubelko: So that we don't lose this important snippet, can we copy the code in verbatim? We do that quite a bit in Starcraft docs.

lengau: Since we're now using sphinx I wonder if we can use a `literalinclude`  to the path? I don't know if that works with a path outside the docs root, but maybe a symlink could work around that?

I would prefer to include other text on the page rather than linking to it, but if we can avoid having another copy to maintain that would also be preferable.

docs/reference/extensions/django-framework-extension.rst
```diff
@@ -0,0 +1,292 @@
...
+
+Juju secrets can be passed as environment variables to your Django
+application. The secret ID has to be passed to the application as a
+config option in the file ``charmcraft.yaml`` file of type ``secret``.
+This config option has to be populated with the secret ID, in the format
+``secret:<secret ID>``.
+
+The environment variable name passed to the application will be:
+
+.. code-block:: bash
```
medubelko, line 284: `text` provides no highlighting, so it's probably more appropriate than `bash`. This isn't quite bash-compatible text, just a pseudocode name.

Strangely, an empty language directive (which we would use in Markdown) defaults to _some_ amount of highlighting, but I can't tell what rules it's using. Sphinx just says it's "default". Not sure what that means.

docs/reference/files/metadata-yaml-file.rst
```diff
@@ -0,0 +1,897 @@
...
+
+.. important::
+
+    In :ref:`recipes <charmcraft-yaml-file>`, this is now the ``links.contact`` key.
+
+.. code-block:: yaml
+
+    # (Optional) A list of maintainers in the format "First Last <email>"
+    maintainers:
+        - <maintainer>
```
medubelko, line 673:
```suggestion
      - <maintainer>
```

docs/reference/files/requirements-dev-txt-file.rst
```diff
@@ -0,0 +1,11 @@
+.. _requirements-dev-txt-file:
+
+
+``requirements-dev.txt`` file
+=============================
+
+The ``requirements-dev.txt`` is a standard Python `requirements
+file <https://pip.pypa.io/en/stable/reference/pip_install/#requirements-file-format>`__
```
medubelko, line 8:
```suggestion
file <https://pip.pypa.io/en/stable/reference/requirements-file-format>`_
```

The page has moved.

docs/reference/files/requirements-txt-file.rst
```diff
@@ -0,0 +1,14 @@
+.. _requirements-txt-file:
+
+
+``requirements.txt`` file
+=========================
+
+The ``requirements.txt`` file is a standard Python `requirements file
+<https://pip.pypa.io/en/stable/reference/pip_install/#requirements-file-format>`_
```
medubelko, line 8:
```suggestion
<https://pip.pypa.io/en/stable/reference/requirements-file-format>`_
```

docs/reference/files/tox-ini-file.rst
```diff
@@ -0,0 +1,103 @@
...
+
+    See more: `Tox |
+    Configuration <https://tox.wiki/en/latest/user_guide.html#configuration>`_
+
+This file is generated automatically by ``charmcraft init`` with the
+contents below:
+
+.. code-block:: yaml
+
+      Copyright 2023 Ubuntu
```
medubelko, line 18:
```suggestion
    # Copyright 2023 Ubuntu
```

docs/reference/extensions/fastapi-framework-extension.rst
```diff
@@ -0,0 +1,260 @@
...
+
+Juju secrets can be passed as environment variables to your Flask
+application. The secret ID has to be passed to the application as a
+config option in the file ``charmcraft.yaml`` file of type ``secret``.
+This config option has to be populated with the secret ID, in the format
+``secret:<secret ID>``.
+
+The environment variable name passed to the application will be:
+
+.. code-block:: bash
```
medubelko, line 252:
```suggestion
.. code-block:: text
```

docs/reference/extensions/flask-framework-extension.rst
```diff
@@ -0,0 +1,273 @@
...
+-------
+
+Juju secrets can be passed as environment variables to your Flask application. The
+secret ID has to be passed to the application as a config option in the file
+``charmcraft.yaml`` file of type ``secret``. This config option has to be populated with
+the secret ID, in the format ``secret:<secret ID>``.
+
+The environment variable name passed to the application will be:
+
+.. code-block:: bash
```
medubelko, line 265:
```suggestion
.. code-block:: text
```

docs/reference/extensions/go-framework-extension.rst
```diff
@@ -0,0 +1,270 @@
...
+-------
+
+Juju secrets can be passed as environment variables to your Go application. The secret
+ID has to be passed to the application as a config option in the file
+``charmcraft.yaml`` file of type ``secret``. This config option has to be populated with
+the secret ID, in the format ``secret:<secret ID>``.
+
+The environment variable name passed to the application will be:
+
+.. code-block:: bash
```
medubelko, line 262:
```suggestion
.. code-block:: text
```

docs/reference/extensions/go-framework-extension.rst
```diff
@@ -0,0 +1,270 @@
...
+HTTP Proxy
+----------
+
+Proxy settings should be set as model configurations. Charms generated using the
+``go-framework`` extension will make the Juju proxy settings available as the
+``HTTP_PROXY``, ``HTTPS_PROXY`` and ``NO_PROXY`` environment variables. For example, the
+``juju-http-proxy`` environment variable will be exposed as ``HTTP_PROXY`` to the Go
+service.
+
+    See more: `List of model configuration
```
medubelko, line 237:
```suggestion
    See more: `Juju | List of model configuration
```

docs/reference/files/lxd-profile-yaml-file.rst
```diff
@@ -0,0 +1,43 @@
...
+``lxd-profile.yaml`` file
+=========================
+
+This document describes the ``lxd-profile.yaml`` file in the root direct
+of your machine charm project.
+
+This file is optional, though it may be useful for charms intended for a
+machine cloud of the LXD type.
+
+    See more: `Juju \| The LXD cloud and
```
medubelko, line 13:
```suggestion
    See more: `Juju | The LXD cloud and
```

medubelko (changes requested):
Thanks for the very quick and excellent work, @jahn-junior!

I picked up a few small code-related blips that only fresh eyes would notice. Once those are tweaked we should be able to merge.

docs/reference/files/metadata-yaml-file.rst
medubelko: > if we can use a literalinclude to the path

I think we're conflating intersphinx features and literalinclude. The latter expects a local file, so we'd have to first get the text from a specific Git commit – not simple.

JJ's latest commit is a simple way to address the question for now: [f45b51b](https://github.com/canonical/charmcraft/pull/2083/commits/f45b51b672c0b7ee9da1e5795cf72a5193a1d91b)

medubelko (approved):
Thanks for accommodating everyone's suggestions, @jahn-junior!

I believe this is in a good state and is reasonably in step with prevailing patterns.

Unless there are major reservations about how this completes the original scope (MD to rST) or code functionality, I'd like to see this merged and tackle further improvements as new tasks.

PR #2111, opened 2025-01-24
feat: add improved bash completion script

bepri:
Adds an improved bash completion script and removes the old one.

CRAFT-3874.


snap/snapcraft.yaml
```diff
@@ -164,6 +162,12 @@ parts:
       [ -n "$(echo $version | grep "post")" ] && grade=devel || grade=stable
       /snap/snapcraft/current/libexec/snapcraft/craftctl set grade="${grade}"
 
+      # uv pip uninstall charmcraft
+      # uv pip install ooooooooh\!
```
lengau, line 166:
```suggestion
```

🤣

lengau (approved):

snap/snapcraft.yaml
bepri: 🥹

tigarmo (approved):

PR #2126, opened 2025-01-28
ci(docs): spread test documentation

lengau:
This adds a CI workflow for spread testing flows in the documentation. In so doing I needed to adjust the documentation spread tasks to pass on Google's runners too.

CRAFT-3798

### DIFF
--- a/.github/workflows/spread-docs.yaml
+++ b/.github/workflows/spread-docs.yaml
@@ -1,0 +1,45 @@
+name: Spread tests for docs
+on:
+  pull_request:
+    paths:
+      - spread.yaml
+      - docs/**/code/**
+  push:
+    branches:
+      - main
+    paths:
+      - spread.yaml
+      - docs/**/code/**
+  schedule:
+    - cron: "0 0 * * 0" # Midnight UTC on Sundays
+
+jobs:
+  docs-tests:
+    runs-on: spread-installed
+    steps:
+      - name: Cleanup job workspace
+        run: |
+          rm -rf "${{ github.workspace }}"
+          mkdir "${{ github.workspace }}"
+
+      - name: Checkout charmcraft
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          submodules: true
+
+      - name: Download snap from store
+        run: |
+          snap download --edge charmcraft
+
+      - name: Run spread
+        run: |
+          spread google:docs/howto/code/ google:docs/tutorial/code/
+
+      - name: Discard spread workers
+        if: always()
+        run: |
+          shopt -s nullglob
+          for r in .spread-reuse.*.yaml; do
+            spread -discard -reuse-pid="$(echo "$r" | grep -o -E '[0-9]+')"
+          done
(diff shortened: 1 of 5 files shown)


.github/workflows/spread-docs.yaml
```diff
@@ -0,0 +1,45 @@
...
+
+      - name: Checkout charmcraft
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          submodules: true
+
+      - name: Download snap from store
+        run: |
+          snap download --edge charmcraft
```
tigarmo, line 33: I'm afraid we'll eventually reach a point where the doc tests need the PR's version of charmcraft, which will make this quite hard to debug
I don't really know how to fix this though, short of packing the snap again here (for the 4th or so time probably)

docs/howto/code/flask-async/task.yaml
```diff
@@ -89,8 +93,6 @@ execute: |
   charmcraft init --profile flask-framework --name flask-async-app
   # [docs:charm-init-end]
 
-  sed -i "s/paas-charm.*/https:\/\/github.com\/canonical\/paas-charm\/archive\/async-workers.tar.gz/g" requirements.txt
```
tigarmo, line 92: how come?

tigarmo (approved):

docs/howto/code/flask-async/task.yaml
lengau: The file no longer exists so that line was causing the spread test to fail.

.github/workflows/spread-docs.yaml
lengau: I agree, but I'm kicking that can down the road for now because it's something I'd like to eventually solve with a workflow that starts by building the snap and then conditionally kicks off every job that consumes the snap, ideally using Starflow to do this for all our apps.

spread.yaml
```diff
@@ -134,12 +135,37 @@ suites:
...
     summary: tests howto from the docs
     systems:
-      - ubuntu-24.04-64
+      - ubuntu-22.04-64
+    manual: true
+    prepare: |
+      juju_channel=3.5/stable
+      microk8s_channel=1.31-strict/stable
+
+      mkdir -p ~/.local/share  # Workaround for Juju not being able to create the
```
mr-cal, line 144:
```suggestion
      mkdir -p ~/.local/share  # Workaround for Juju not being able to create the directory
```

spread.yaml
```diff
@@ -134,12 +135,37 @@ suites:
...
   docs/tutorial/code/:
     summary: tests tutorial from the docs
     systems:
       - ubuntu-22.04-64
     manual: true
+    prepare: |
+      juju_channel=3.5/stable
+      microk8s_channel=1.31-strict/stable
+
+      mkdir -p ~/.local/share  # Workaround for Juju not being able to create the
```
mr-cal, line 161:
```suggestion
      mkdir -p ~/.local/share  # Workaround for Juju not being able to create the directory
```

mr-cal (approved):

PR #2156, opened 2025-02-10
docs(how-to): add how-to migrate to uv

lengau:
Adds a how-to for migrating to the uv plugin. It's based on the poetry plugin page.

Drive-by: typo fix

https://canonical-charmcraft--2156.com.readthedocs.build/en/2156/howto/migrate-plugins/charm-to-uv/

### DIFF
--- a/docs/howto/migrate-plugins/charm-to-uv.rst
+++ b/docs/howto/migrate-plugins/charm-to-uv.rst
@@ -41,6 +41,7 @@ To find these dependencies, check each loaded library file for its ``PYDEPS`` by
 the following command at the root of the charm project:
 
 .. code-block:: bash
+
     find lib -name "*.py" -exec awk '/PYDEPS = \[/,/\]/' {} +
 
 Next, in ``pyproject.toml``, list them in a ``charmlibs`` dependency group.
@@ -61,11 +62,8 @@ Add dependency groups
 ---------------------
 
 If the charm has dependency groups that should be included when creating the virtual
-environment, such as one for charm libraries, the ``uv-groups`` key can be set to
-include them.
-
-Including this dependency group in your charm is as easy as adding it to
-``charmcraft.yaml``:
+environment, such as one for charm libraries, the
+:ref:`uv plugin's <craft_parts_uv_plugin>` ``uv-groups`` key can be set to include them:
 
 .. code-block:: yaml
     :caption: charmcraft.yaml
@@ -77,6 +75,9 @@ Including this dependency group in your charm is as easy as adding it to
         source: .
         uv-groups:
           - charmlibs
+
+Likewise, optional dependencies under the ``pyproject.toml`` key
+``project.optional-dependencies`` can be added with the ``uv-extras`` key.
 
 Include extra files
 -------------------
--- a/docs/howto/migrate-plugins/charm-to-uv.rst
+++ b/docs/howto/migrate-plugins/charm-to-uv.rst
@@ -1,0 +1,107 @@
+.. _howto-migrate-to-uv:
+
+Migrate from the Charm plugin to the uv plugin
+==============================================
+
+For charms that use `uv`_, Charmcraft has a :ref:`craft_parts_uv_plugin`. Migrating
+from the Charm plugin provides some benefits, such as using uv during the build
+process not having to maintain a separate ``requirements.txt`` file. If the
+charm to be migrated does not currently use uv, refer to the
+`uv documentation <https://docs.astral.sh/uv/guides/projects/>`_ for instructions
+on how to use uv for a Python project.
+
+Update the project file
+-----------------------
+
+The first step is to update the project file to include the correct parts definition.
+Depending on the history of a specific charm, it may not have an explicitly-included
+``parts`` section determining how to build the charm. In this case, a ``parts`` section
+can be created as follows:
+
+.. code-block:: yaml
+    :caption: charmcraft.yaml
+
+    parts:
+      my-charm:  # This can be named anything you want
+        plugin: uv
+        source: .
+
+Add optional dependency groups
+------------------------------
+
+If the charm has `dependency groups`_ that should be included when creating the virtual
+environment, the ``uv-groups`` key can be used to include those groups when creating
+the virtual environment.
+
+.. note::
+    This is useful and encouraged, though not mandatory, for keeping track of
+    library dependencies, as covered in the next section.
+
+Include charm library dependencies
+----------------------------------
+
+Unlike the Charm plugin, the uv plugin does not install the dependencies for
+included charmlibs. If any of the charm libraries used have ``PYDEPS``, these will
+need to be added to the charm's dependencies, potentially as their own
+`dependency group <dependency groups_>`_.
+
+To find these dependencies, check each library file for its ``PYDEPS``. A command
+that can find these is::
+
+    find lib -name "*.py" -exec awk '/PYDEPS = \[/,/\]/' {} +
+
+If run from the base directory of a charm, this will show all the PYDEPS declarations
+from all loaded charm libs. These can then be included in ``pyproject.toml``.
+
+.. code-block:: toml
+    :caption: pyproject.toml
+
+    [dependency-groups]
+    # Dependencies brought from libraries the charm uses.
+    charmlibs = [
+        "cosl",
+        "pydantic",
+        "cryptography",
+        "ops>=2.0.0",
+    ]
+
+Including this dependency group is as easy as adding it to ``charmcraft.yaml``:
+
+.. code-block:: yaml
+    :caption: charmcraft.yaml
+    :emphasize-lines: 5-6
+
+    parts:
+      my-charm:
+        plugin: uv
+        source: .
+        uv-groups:
+          - charmlibs
+
+Include extra files
+-------------------
+
+The uv plugin only includes the contents of the ``src`` and ``lib`` directories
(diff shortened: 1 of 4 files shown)


docs/howto/migrate-plugins/charm-to-uv.rst
```diff
@@ -0,0 +1,107 @@
...
+Add optional dependency groups
+------------------------------
+
+If the charm has `dependency groups`_ that should be included when creating the virtual
+environment, the ``uv-groups`` key can be used to include those groups when creating
+the virtual environment.
+
+.. note::
+    This is useful and encouraged, though not mandatory, for keeping track of
+    library dependencies, as covered in the next section.
```
jahn-junior, line 38: Not sure about the necessity of this. If we do need to mention that the key is encouraged but not mandatory, I feel that it could be worked into the previous paragraph rather than being in a note.

docs/howto/migrate-plugins/charm-to-uv.rst
```diff
@@ -0,0 +1,107 @@
...
+    [dependency-groups]
+    # Dependencies brought from libraries the charm uses.
+    charmlibs = [
+        "cosl",
+        "pydantic",
+        "cryptography",
+        "ops>=2.0.0",
+    ]
+
+Including this dependency group is as easy as adding it to ``charmcraft.yaml``:
```
jahn-junior, line 68:
```suggestion
Including this dependency group is as easy as adding it to the project file:
```

jahn-junior (commented):
Submitting as comments because I may not be able to do a timely re-review this week. No blocking changes, just a few things I think could be improved. Thanks for writing this up!

docs/howto/migrate-plugins/charm-to-uv.rst
medubelko: If it's useful and encouraged, it shouldn't be a note!

docs/howto/migrate-plugins/charm-to-uv.rst
```diff
@@ -0,0 +1,107 @@
+.. _howto-migrate-to-uv:
+
+Migrate from the Charm plugin to the uv plugin
+==============================================
+
+For charms that use `uv`_, Charmcraft has a :ref:`craft_parts_uv_plugin`. Migrating
+from the Charm plugin provides some benefits, such as using uv during the build
+process not having to maintain a separate ``requirements.txt`` file. If the
```
medubelko, line 8: > not having to maintain a separate ``requirements.txt`` file

:partying_face:

docs/howto/migrate-plugins/charm-to-uv.rst
```diff
@@ -0,0 +1,107 @@
...
+    This is useful and encouraged, though not mandatory, for keeping track of
+    library dependencies, as covered in the next section.
+
+Include charm library dependencies
+----------------------------------
+
+Unlike the Charm plugin, the uv plugin does not install the dependencies for
+included charmlibs. If any of the charm libraries used have ``PYDEPS``, these will
+need to be added to the charm's dependencies, potentially as their own
+`dependency group <dependency groups_>`_.
```
medubelko, line 46: We linked to this quite recently – generally not necessary to do it again so soon.

docs/howto/migrate-plugins/charm-to-uv.rst
```diff
@@ -0,0 +1,107 @@
...
+
+To find these dependencies, check each library file for its ``PYDEPS``. A command
+that can find these is::
+
+    find lib -name "*.py" -exec awk '/PYDEPS = \[/,/\]/' {} +
+
+If run from the base directory of a charm, this will show all the PYDEPS declarations
+from all loaded charm libs. These can then be included in ``pyproject.toml``.
```
medubelko, line 54:
```suggestion
To find these dependencies, check each loaded library file for its ``PYDEPS`` by running
the following command at the root of the charm project:

.. code-block:: bash

    find lib -name "*.py" -exec awk '/PYDEPS = \[/,/\]/' {} +

Next, in ``pyproject.toml``, list them in a ``charmlibs`` dependency group.
```

docs/howto/migrate-plugins/charm-to-uv.rst
```diff
@@ -0,0 +1,107 @@
...
+.. code-block:: yaml
+    :caption: charmcraft.yaml
+
+    parts:
+      my-charm:  # This can be named anything you want
+        plugin: uv
+        source: .
+
+Add optional dependency groups
+------------------------------
```
medubelko, line 30: I recommend consolidating or reflowing this section and the next. It reads askew to start with an optional step about dependency groups and then discuss a second, required step about dependency groups.

Is the order critical?

docs/howto/migrate-plugins/charm-to-uv.rst
```diff
@@ -0,0 +1,107 @@
...
+        "pydantic",
+        "cryptography",
+        "ops>=2.0.0",
+    ]
+
+Including this dependency group is as easy as adding it to ``charmcraft.yaml``:
+
+.. code-block:: yaml
+    :caption: charmcraft.yaml
+    :emphasize-lines: 5-6
```
medubelko, line 72: :partying_face:

medubelko (changes requested):
Thanks for this quick one, @lengau!

The small stuff isn't blocking, but if you could look into the intro and the question about the optional dependency group section, I think the guide could be made even more robust.

docs/howto/migrate-plugins/charm-to-uv.rst
lengau: I very intentionally used `charmcraft.yaml` here to distinguish between the two project file's we're talking about (`charmcraft.yaml` and `pyproject.toml`) in this section.

docs/howto/migrate-plugins/charm-to-uv.rst
lengau: Kept this link and removed the other one because I rearranged the sections a bit.

docs/howto/migrate-plugins/charm-to-uv.rst
```diff
@@ -0,0 +1,108 @@
...
+
+Migrate from the Charm plugin to the uv plugin
+==============================================
+
+For charms that use `uv`_, Charmcraft has a :ref:`craft_parts_uv_plugin`. This guide
+shows how to migrate from the default Charm plugin to the uv plugin.
+
+Migrating from the Charm plugin provides some benefits, not having to maintain a
+separate ``requirements.txt`` file and using the much faster ``uv`` for package
+management rather than ``pip``.
```
medubelko, line 11:
```suggestion
Migrating from the Charm plugin provides some benefits, notably not having to maintain a
separate ``requirements.txt`` file. For package management, uv is much faster than pip.
```

Reviewing this again to make the paragraph self-consistent.

docs/howto/migrate-plugins/charm-to-uv.rst
```diff
@@ -0,0 +1,108 @@
...
+        "cryptography",
+        "ops>=2.0.0",
+    ]
+
+Add dependency groups
+---------------------
+
+If the charm has dependency groups, such as one for charm libraries, that should be
+included when creating the virtual environment, the ``uv-groups`` key can be used to
+include those groups when creating the virtual environment.
```
medubelko, line 66:
```suggestion
If the charm has dependency groups that should be included when creating the virtual
environment, such as one for charm libraries, the ``uv-groups`` key can be set to
include them.
```

Yay, the diff worked.

docs/howto/migrate-plugins/charm-to-uv.rst
```diff
@@ -0,0 +1,108 @@
...
+
+Including this dependency group in your charm is as easy as adding it to
+``charmcraft.yaml``:
```
medubelko, line 69: With the new paragraph, this feels redundant. Cut?

medubelko (changes requested):

docs/howto/migrate-plugins/charm-to-uv.rst
lengau: Don't want to fully cut (I kept it because without it the jump to the example felt a bit abrupt), but I'll work on a way to enbriefen and attach it to the previous paragraph.

lengau: OK, I've adjusted it, and once again our line length limit is producing some uhh... "interesting" formatting.

I also added a brief bit after the example about ``uv-extras``. I chose not to add an example for that because it's the same format as groups, but I did link again to the reference page down here since I think they're likely to want to look at the reference from here again.

lengau:
Another question for @medubelko: I feel like a link to this new page from the [uv plugin reference page](https://canonical-charmcraft--2156.com.readthedocs.build/en/2156/reference/plugins/uv_plugin/#craft-parts-uv-plugin) might be useful, but I'm not sure what the best way to do it is. My lazy thought is a quick "see also" after the first paragraph.

docs/howto/migrate-plugins/charm-to-uv.rst
medubelko: LGTM.

medubelko:
> My lazy thought is a quick "see also" after the first paragraph.

Let's do it. Sometimes, that's the simplest way.

medubelko (approved):

cmatsuoka (approved):

PR #2168, opened 2025-02-12
docs: update readme and logo

tmihoc:
The Charmcraft RTD docs logo still shows the Ubuntu circle friends whereas it should be showing the Juju logo and the README still points to the old docs. This PR fixes both these issues.

PS This work parallels similar updates in the Juju and Ops repo.


lengau (approved):
Thanks!

medubelko (approved):

PR #2172, opened 2025-02-13
docs: update release step with resource details

### LINKED ISSUE #2171: "Publish a charm to Charmhub" should include an example with your first resource.
### What needs to get done

Over here: https://canonical-charmcraft.readthedocs-hosted.com/en/latest/howto/manage-charms/#publish-a-charm-on-charmhub
and probably over here too: https://canonical-charmcraft.readthedocs-hosted.com/en/latest/howto/manage-resources/#publish-a-resource-on-charmhub

We should show how to attach a specific resource revision when releasing a charm.

### Why it needs to get done

This is not clear and causes even charming experts headaches since they need to bootstrap it so infrequently. https://discourse.charmhub.io/t/unable-to-release-charm-to-charmhub-with-optional-resource/16786/7

tmihoc:
The HTG Manage charms / Publish > Release example did not include resources, although (1) resources had been mentioned previously and (2) the way to attach them at release time is easy to forget. This has resulted in an issue even for an otherwise expert user: https://github.com/canonical/charmcraft/issues/2171 This PR fixes that. (It also makes the name of the charm the same across all the examples in the Publish section.)

@erik78se Thanks for the example output for releasing with a resource. 

Fixes #2171

### DIFF
--- a/docs/howto/manage-charms.rst
+++ b/docs/howto/manage-charms.rst
@@ -633,12 +633,12 @@ Publish a charm on Charmhub
 
    .. code-block:: bash
 
-       charmcraft upload-resource my-super-charm someresource
+       charmcraft upload-resource my-awesome-charm someresource
        --filepath=/tmp/superdb.bin
 
    .. terminal::
 
-       Revision 1 created of resource 'someresource' for charm 'my-super-charm'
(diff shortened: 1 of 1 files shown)


lengau (approved):
Thanks!

medubelko (approved):

PR #2187, opened 2025-02-21
feat(extension): Block on missing required config

alithethird:
This PR is related to 12 Factor [Block on missing required config PR](https://github.com/canonical/paas-charm/pull/42)

## Rationale

Some applications require certain config options to run and would crash if these options are not provided. To prevent this, we need to add an optional field to the config options and block the charm if a config option marked as non-optional has not been provided with a message that explains which config options are missing.

To stay backward compatible the default value for the optional field will be true.

Since it doesn’t make sense for config options that are non-optional to have a default value, charmcraft will look for this condition at pack time and error out.

## Extension Changes

Adds a check for non-optional options with default values and throws `ExtensionError` when found.
Adds documentation about the non-optional options.

Reviewers:
@jdkandersson 
@javierdelapuente 
@erinecon

### DIFF
--- a/docs/reference/extensions/django-framework-extension.rst
+++ b/docs/reference/extensions/django-framework-extension.rst
@@ -61,7 +61,7 @@ will set the ``DJANGO_ALLOWED_HOSTS`` environment variable, the ingress
 URL or the Kubernetes service URL if there is no ingress integration,
 will be set automatically.
 
-In addition to this, you can set the configuration options to be
+In addition to this, you can set the configuration option to be
 mandatory by setting the ``optional`` key to ``false``. This will
 block the charm and stop services until the configuration is supplied. For example,
 if your application needs an ``api-token`` to function correctly you can set
@@ -69,6 +69,7 @@ if your application needs an ``api-token`` to function correctly you can set
 services until the ``api-token`` configuration is supplied.
 
 .. code-block:: yaml
+    :caption: charmcraft.yaml
 
     config:
       options:
@@ -79,9 +80,8 @@ services until the ``api-token`` configuration is supplied.
 
 .. note::
 
-    A configuration with the ``optional: false`` option cannot have a
-     ``default`` value. If you set a ``default`` value to a configuration
-     with ``optional: false`` ``charmcraft`` will fail when packing the charm.
+    A configuration with ``optional: false`` can't also have a ``default`` key.
+    If it has both, the charm will fail to pack.
 
 
 ``peers``, ``provides``, and ``requires`` keys
--- a/docs/reference/extensions/django-framework-extension.rst
+++ b/docs/reference/extensions/django-framework-extension.rst
@@ -62,11 +62,11 @@ URL or the Kubernetes service URL if there is no ingress integration,
 will be set automatically.
 
 In addition to this, you can set the configuration options to be
-``non-optional`` by setting the ``optional`` key to ``False``. This will
-block the charm and stop services until the config is supplied. For example,
+mandatory by setting the ``optional`` key to ``false``. This will
+block the charm and stop services until the configuration is supplied. For example,
 if your application needs an ``api-token`` to function correctly you can set
-it ``non-optional``, as below. This will block the charm and stop the
-services until the ``api-token`` is supplied.
+``optional``, as shown below. This will block the charm and stop the
+services until the ``api-token`` configuration is supplied.
 
 .. code-block:: yaml
 
@@ -79,9 +79,9 @@ services until the ``api-token`` is supplied.
 
 .. note::
 
-    The ``non-optional`` options can not have ``default`` values. If you
-    set a ``default`` value to a ``non-optional`` configuration option
-    ``charmcraft`` will fail when packing the charm.
+    A configuration with the ``optional: false`` option cannot have a
+     ``default`` value. If you set a ``default`` value to a configuration
+     with ``optional: false`` ``charmcraft`` will fail when packing the charm.
 
 
 ``peers``, ``provides``, and ``requires`` keys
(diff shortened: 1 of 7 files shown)


docs/reference/extensions/django-framework-extension.rst
```diff
@@ -61,6 +61,28 @@ will set the ``DJANGO_ALLOWED_HOSTS`` environment variable, the ingress
 URL or the Kubernetes service URL if there is no ingress integration,
 will be set automatically.
 
+In addition to this, you can set the configuration options to be
+``non-optional`` by setting the ``optional`` key to ``False``. This will
```
erinecon, line 65:
```suggestion
mandatory by setting the ``optional`` key to ``false``. This will
```

docs/reference/extensions/django-framework-extension.rst
```diff
@@ -61,6 +61,28 @@ will set the ``DJANGO_ALLOWED_HOSTS`` environment variable, the ingress
...
+block the charm and stop services until the config is supplied. For example,
```
erinecon, line 66:
```suggestion
block the charm and stop services until the configuration is supplied. For example,
```

docs/reference/extensions/django-framework-extension.rst
```diff
@@ -61,6 +61,28 @@ will set the ``DJANGO_ALLOWED_HOSTS`` environment variable, the ingress
...
+if your application needs an ``api-token`` to function correctly you can set
+it ``non-optional``, as below. This will block the charm and stop the
```
erinecon, line 68:
```suggestion
``optional`` to ``false``, as shown below. This will block the charm and stop the
```

docs/reference/extensions/django-framework-extension.rst
```diff
@@ -61,6 +61,28 @@ will set the ``DJANGO_ALLOWED_HOSTS`` environment variable, the ingress
...
+services until the ``api-token`` is supplied.
```
erinecon, line 69:
```suggestion
services until the ``api-token`` configuration is supplied.
```

docs/reference/extensions/django-framework-extension.rst
```diff
@@ -61,6 +61,28 @@ will set the ``DJANGO_ALLOWED_HOSTS`` environment variable, the ingress
...
+        api-token:
+          description: The token necessary for the service to run.
+          type: string
+          optional: false
+
+.. note::
+
+    The ``non-optional`` options can not have ``default`` values. If you
+    set a ``default`` value to a ``non-optional`` configuration option
+    ``charmcraft`` will fail when packing the charm.
```
erinecon, line 84:
```suggestion
    A configuration with the ``optional: false`` option cannot have a ``default`` value. If you
    set a ``default`` value to a configuration with ``optional: false``
    ``charmcraft`` will fail when packing the charm.
```
Probably the first line length is too long -- sorry about that

erinecon (commented):
Thanks for adding this section to the reference pages! I have some feedback on the wording that I only added to the Django page but applies to all of the pages. 

The biggest change I'm suggesting: I removed the ``non-optional`` syntax because it's not used from a coding perspective, meaning that a user wouldn't add ``non-optional`` anywhere in the YAML file. Please correct me if I'm mistaken.

erinecon (approved):
Looks great! The note at the end is rendering strangely in the RTD build, otherwise I left two nitpicks on the Django doc that apply to all the pages :)

lengau (approved):
Approved on the code side. Deferring to @medubelko and @erinecon on docs

docs/reference/extensions/fastapi-framework-extension.rst
```diff
@@ -39,7 +39,29 @@ by running ``juju config <application> token=<token>``.
...
+        api-token:
+          description: The token necessary for the service to run.
+          type: string
+          optional: false
+
+.. note::
+
+    A configuration with the ``optional: false`` option cannot have a
+     ``default`` value. If you set a ``default`` value to a configuration
+     with ``optional: false`` ``charmcraft`` will fail when packing the charm.
```
medubelko, line 64:
```suggestion
.. note::

    A configuration with ``optional: false`` can't also have a ``default`` key.
    If it has both, the charm will fail to pack.
```

docs/reference/extensions/fastapi-framework-extension.rst
```diff
@@ -39,7 +39,29 @@ by running ``juju config <application> token=<token>``.
...
+
+In addition to this, you can set the configuration options to be
+mandatory by setting the ``optional`` key to ``false``. This will
+block the charm and stop services until the configuration is supplied. For example,
+if your application needs an ``api-token`` to function correctly you can set
+``optional``, as shown below. This will block the charm and stop the
+services until the ``api-token`` configuration is supplied.
+
+.. code-block:: yaml
+
```
medubelko, line 52:
```suggestion
    :caption: charmcraft.yaml
```

medubelko (changes requested):
My suggestions apply to all affected docs. Thanks!

docs/reference/extensions/go-framework-extension.rst
```diff
@@ -48,6 +48,28 @@ charm can set it by running ``juju config <application> token=<token>``.
...
+      options:
+        api-token:
+          description: The token necessary for the service to run.
+          type: string
+          optional: false
+
+.. note::
+
+    A configuration with ``optional: false`` can't also have a ``default`` key.
+    If it has both, the charm will fail to pack.
```
jdkandersson, line 71: Could we make this a reusable block? Like: https://github.com/canonical/charmcraft/blob/b4aee1e97868997e878fcad42109740f536f044a/docs/tutorial/write-your-first-kubernetes-charm-for-a-django-app.rst?plain=1#L59

tests/extensions/test_app.py
```diff
@@ -404,6 +404,20 @@ def test_flask_merge_charm_libs(flask_input_yaml, tmp_path):
...
+                "options": {
+                    "non-optional": {
+                        "type": "string",
+                        "optional": False,
+                        "default": "default value",
+                    }
+                }
+            }
+        },
+        id="non-optional-config-with-defautl",
```
jdkandersson, line 419: Let's fix the typo

Should we add a test case with a optional: False without a default?

medubelko (approved):

PR #2195, opened 2025-02-26
docs: add `charm-user` in the charmcraft.yaml file reference

tonyandrewmeyer:
The `charm-user` field is documented in the metadata.yaml file reference, but is missing from the charmcraft.yaml file reference, even though it does work there. This PR adds it to that reference.

As the docs currently are, it gives the incorrect impression that you cannot use the combined charmcraft.yaml file to specify `charm-user`, and must have separate metadata, actions, and config files.


lengau (approved):
Thanks! Could you do me a favour and add this key to one of the charmcraft YAML files under the [basic smoke tests](https://github.com/canonical/charmcraft/tree/main/tests/spread/smoketests/basic)? I just looked and it doesn't seem like we're actually e2e testing this anywhere.

docs/reference/files/charmcraft-yaml-file.rst
```diff
@@ -346,6 +346,42 @@ and the lib version (in ``"<api version>[.<patch version>]"`` string format).
...
+
+.. _charmcraft-yaml-key-charm-user:
+
+``charm-user``
+--------------
+
+.. important::
+
+    ``charm-user`` was added in Juju 3.6.0. Currently only supported on Kubernetes
+    charms and has no effect on machine charms.
```
medubelko, line 360:
```suggestion
    ``charm-user`` was added in Juju 3.6.0. It's currently only supported in Kubernetes
    charms and has no effect on machine charms.
```

docs/reference/files/charmcraft-yaml-file.rst
```diff
@@ -346,6 +346,42 @@ and the lib version (in ``"<api version>[.<patch version>]"`` string format).
...
+
+**Status:** Optional. Recommended for Kubernetes charms.
+
+**Purpose:** The ``charm-user`` key  allows charm authors to specify that their charm
+hook code does not need to be run as root. This key, in addition to ``uid`` + ``uid``
+fields in ``containers``, allows charms to be run rootless. The value of ``root``
+ensures the charm runs as root. Both ``sudoer`` and ``non-root`` will run as a user
+other than root. In the case of the value ``sudoer``, the charm will be run as a user
+with access to sudo to elevate its privileges.
```
medubelko, line 369: Here's an attempt at tidying up some of the language. It contains a mix of style guide conventions and semantic changes.

I assumed this...

> ``uid`` + ``uid``

... is a typo, hence the change.

```suggestion
**Purpose:** The ``charm-user`` key allows charm authors to specify that their charm
hook code doesn't need to run as root. This key, in addition to the ``containers.uid``
and ``containers.gid`` keys, allows the charm to be run rootless. If set to ``root``,
the charm runs as root. If set to ``sudoer`` or ``non-root``, the charm runs as a user
other than root. If the value is ``sudoer``, the charm runs as a superuser, giving it
elevated privileges.
```

docs/reference/files/charmcraft-yaml-file.rst
```diff
@@ -346,6 +346,42 @@ and the lib version (in ``"<api version>[.<patch version>]"`` string format).
...
+
+.. important::
+
+    ``charm-user`` was added in Juju 3.6.0. Currently only supported on Kubernetes
+    charms and has no effect on machine charms.
+
+**Status:** Optional. Recommended for Kubernetes charms.
+
+**Purpose:** The ``charm-user`` key  allows charm authors to specify that their charm
+hook code does not need to be run as root. This key, in addition to ``uid`` + ``uid``
```
medubelko, line 365: > in addition to

Should this be "in combination with"?

docs/reference/files/charmcraft-yaml-file.rst
tonyandrewmeyer: This is all copied from the `metadata.yaml` definition. Do you want it changed there as well?

"does not need to be run as root" does not have the same meaning as "doesn't need root access", so that would be a bad change. "runs as a superuser" also does not have the same meaning as "run as a user with access to sudo", so that would also be a bad change.

tonyandrewmeyer:
> Could you do me a favour and add this key to one of the charmcraft YAML files under the [basic smoke tests](https://github.com/canonical/charmcraft/tree/main/tests/spread/smoketests/basic)? I just looked and it doesn't seem like we're actually e2e testing this anywhere.

Done, assuming that adding the key is all that's required.

tonyandrewmeyer:
I'm not sure of the conventions here - do I need to squash all the commits, and/or rebase? Or will you do that as part of the merge? Let me know if I need to do more.

medubelko:
@tonyandrewmeyer We'll squash on merge.

medubelko (approved):

PR #2196, opened 2025-02-27
docs: maintain Django tutorial

erinecon:
Some specific fixes to the Django tutorial:
* Don't need to explicitly have `python3 manage.py`. Django developers are more used to `./manage.py` instead. I have updated the tutorial and `task.yaml` file. 
* Any reference to `/django-hello-world` in text has been updated to `~/django-hello-world`.
* Updated indentation in a code snippet that was rendering incorrectly.

Additional reviewers: @jdkandersson

### DIFF
--- a/docs/tutorial/kubernetes-charm-django.rst
+++ b/docs/tutorial/kubernetes-charm-django.rst
@@ -138,7 +138,7 @@ Now, run the Django app to verify that it works:
 
 .. code-block:: bash
 
-    python3 manage.py runserver 0.0.0.0:8000
+    ./manage.py runserver 0.0.0.0:8000
 
 .. note::
 
--- a/docs/tutorial/kubernetes-charm-django.rst
+++ b/docs/tutorial/kubernetes-charm-django.rst
@@ -118,7 +118,7 @@ Run the Django app locally
 We will test the Django app by visiting the app in a web
 browser.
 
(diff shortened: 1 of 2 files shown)


medubelko (approved):
Changes LGTM.

docs/tutorial/kubernetes-charm-django.rst
```diff
@@ -138,7 +138,7 @@ Now, run the Django app to verify that it works:
 
 .. code-block:: bash
 
-    python3 manage.py runserver 0.0.0.0:8000
+    ./manage.py runserver 0.0.0.0:8000
```
lengau, line 141: I know it seems weird given that it's one line, but using a `literalinclude` from the task here might save us some headaches if this changes again.

lengau (approved):
LGTM! Let me know what you think about my comment and I can merge at the appropriate time.

docs/tutorial/kubernetes-charm-django.rst
erinecon: Thank you for pointing this out. The issue with using the line from the task file is that the task file has a `&` at the end. So it's a slightly different command AFAIK. 

In any event I didn't update the command in the task file (for now), and I don't expect the line to change. I'm good with no change (for now) and having this PR merged :)

PR #2197, opened 2025-02-28
docs: Add structure to 12-factor how-to guides

erinecon:
Currently the 12-factor how-to guides were split into two documents: one for the async configuration, and another containing all other how-to sections. This pull request consolidates the two documents into a directory, `manage-12-factor-charms`, with added structure by splitting up the sections into separate pages. 

### Summary of changes
* Created `manage-12-factor-charms` directory with an index page and four different categories: Initialization, Configuration, Integration, and Usage
* Consolidated the flask-async document into the Configuration category 
* Added redirects

Additional reviewers: @jdkandersson


jdkandersson (commented):
I think the structure looks good, perhaps we change the name from `manage-12-factor-charms` to `manage-web-app-charms` and similarly for the other files

docs/howto/index.rst
```diff
@@ -8,7 +8,7 @@ How-to guides
 
    manage-charmcraft
    manage-charms
-   manage-12-factor-app-charms
+   Manage 12-factor app charms <manage-12-factor-charms/index>
```
medubelko, line 11: None of the TOC changes are visible through due to this.

```suggestion
   Manage 12-factor app charms <manage-web-app-charms/index>
```

docs/howto/manage-web-app-charms/configure-web-app-charms.rst
```diff
@@ -296,33 +164,66 @@ Add the Juju secret ID to the application:
...
+   ok
+   ok
 
-If the migration script fails, it will retry upon ``update-status``. The migration
-script will run on every unit. The script is assumed to be idempotent (in other words,
-can be run multiple times) and that it can be run on multiple units simultaneously
-without issue. Handling multiple migration scripts that run concurrently
-can be achieved by, for example, locking any tables during the migration.
+It can take up to a minute for the configuration to take effect. When the
+configuration changes, the charm will re-enter the active state.
```
lengau, line 229: Might be nice to have a sample of what this looks like in a terminal from `juju status`

docs/howto/manage-web-app-charms/integrate-web-app-charms.rst
```diff
@@ -0,0 +1,68 @@
...
+- ``POSTGRESQL_DB_PARAMS``
+- ``POSTGRESQL_DB_QUERY``
+- ``POSTGRESQL_DB_FRAGMENT``
+- ``POSTGRESQL_DB_USERNAME``
+- ``POSTGRESQL_DB_PASSWORD``
+- ``POSTGRESQL_DB_HOSTNAME``
+- ``POSTGRESQL_DB_PORT``
+
+    See also: `How to add an integration to a charm
+    <https://juju.is/docs/sdk/implement-integrations-in-a-charm>`_
```
lengau, line 68:
```suggestion
    See also: :external+ops:doc:`Ops | How to manage relations <howto/manage-relations>`
```

docs/howto/manage-web-app-charms/index.rst
```diff
@@ -0,0 +1,52 @@
...
+.. toctree::
+   :maxdepth: 2
+
+   Initialize your 12-factor app charm <init-web-app-charms>
+
+Configuration
+-------------
+
+Once your 12-factor app charm has been set up, you can customize many aspects
+of the charm using Juju.
```
medubelko, line 23:
```suggestion
of it using Juju.
```

docs/howto/manage-web-app-charms/index.rst
```diff
@@ -0,0 +1,52 @@
...
+
+.. toctree::
+   :maxdepth: 2
+
+   Configure your 12-factor app charm <configure-web-app-charms>
+
+Integration
+-----------
+
+Once deployed, your 12-factor app can be integrated to nearly anything in the
```
medubelko, line 33:
```suggestion
Once deployed, your 12-factor app can be integrated into nearly anything in the
```

docs/howto/manage-web-app-charms/index.rst
```diff
@@ -0,0 +1,52 @@
...
+charm ecosystem.
```
medubelko, line 34: > charm

"Juju"?

docs/howto/manage-web-app-charms/index.rst
```diff
@@ -0,0 +1,52 @@
...
+.. toctree::
+   :maxdepth: 2
+
+   Integrate your 12-factor app charm <integrate-web-app-charms>
+
+Usage
+-----
+
+Now that you've initialized and configured your 12-factor app charm, you're
+ready to use it!
```
medubelko, line 45:
```suggestion
ready to use it.
```

I'm a party pooper.

docs/howto/manage-web-app-charms/index.rst
```diff
@@ -0,0 +1,52 @@
+.. _manage-12-factor-app-charms:
+
+How to manage 12-factor app charms
```
medubelko, line 3: I generally prefer singular for all these titles. What do you think?

docs/howto/manage-web-app-charms/init-web-app-charms.rst
medubelko, line 1: Any reason for the discrepancy between the singular in the title, the filename, and the ref name?

This is one of the reasons I push for making all three as similar as possible, to limit ambiguity between the three different reference points.

docs/howto/manage-web-app-charms/init-web-app-charms.rst
```diff
@@ -0,0 +1,59 @@
...
+will need to check the project file and ``README.md`` to verify that the charm's name
+and description are correct.
+
+    See also: :ref:`ref_commands_init`
+
+.. tabs::
+
+    .. group-tab:: Flask
+
+        Specify the ``flask-framework`` profile:
```
medubelko, line 23: We could simplify further by instantiating a generic version of this line before the tab block, and cutting it from the tabs themselves.

docs/howto/manage-web-app-charms/init-web-app-charms.rst
```diff
@@ -0,0 +1,59 @@
...
+
+            charmcraft init --profile go-framework
+
+
+Prepare an OCI image for a 12-factor app charm
+----------------------------------------------
+
+    See more:
+    :external+rockcraft:doc:`Rockcraft | How to build a 12-factor app rock
+    <how-to/build-a-12-factor-app-rock>`
```
medubelko, line 59:
```suggestion
 Next, :external+rockcraft:doc:`prepare a rock for your 12-factor app charm
 <how-to/build-a-12-factor-app-rock>` with Rockcraft.
```

docs/howto/manage-web-app-charms/integrate-web-app-charms.rst
```diff
@@ -0,0 +1,67 @@
+.. _integrate-12-factor-charms:
+
+Manage relations for a 12-factor app charm
+==========================================
+
+A charm integration can be added to your charmed 12-factor app by providing
+the integration and endpoint definition in your project file:
+
+.. code-block:: yaml
+
```
medubelko, line 10:
```suggestion
    :caption: charmcraft.yaml

```

Here and for the rest. Sorry for this drive-by. We quite recently added these to the tutorials.

docs/howto/manage-web-app-charms/configure-web-app-charms.rst
```diff
@@ -1,73 +1,10 @@
...
-
-    .. group-tab:: Go
-
-        Specify the ``go-framework`` profile:
-
-        .. code-block:: bash
-
-            charmcraft init --profile go-framework
-
+.. _configure-12-factor-charms:
```
medubelko, line 1: As mentioned elsewhere, the ref and filename are significant departures from the human title.

medubelko (approved):
@erinecon Most of these are my impressions on the nitty gritty of the structural split, but I also have some language suggestions for the new index.

medubelko:
Let's also not forget the necessary additions to `redirects.txt`.

docs/howto/manage-web-app-charms/index.rst
erinecon: I have no strong preference. I'm fine with adopting your preference here :)

docs/howto/manage-web-app-charms/configure-web-app-charms.rst
erinecon: While I agree with you, I'd prefer to do that in a follow-up PR if that's all right

docs/howto/manage-web-app-charms/index.rst
```diff
@@ -0,0 +1,52 @@
...
+==================================
+
+These guides walk you through all ways you can manage 12-factor app charms,
+from initialization to usage.
+
+Initialization
+--------------
+
+You need both an rock and a charm to have a deployable 12-factor app on Juju.
```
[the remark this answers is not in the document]
erinecon: Ah yeah, probably. Good catch! I'll make the change

docs/howto/manage-web-app-charms/index.rst
erinecon: I was trying to emphasize the value of Charmhub here -- with the correct infrastructure in place, a 12-factor app charm can be integrated with anything in Charmhub. That's what I was going for with "charm ecosystem".

Do you think "Juju ecosystem" would provide the same meaning?

docs/howto/manage-web-app-charms/init-web-app-charms.rst
erinecon: The reason was I was not thinking too hard about consistency :) I can update the filenames and titles to be singular, but I lean toward keeping the ref name the same for link preservation. If that's not a concern for you, then I can go ahead and update the ref names as well.

docs/howto/manage-web-app-charms/init-web-app-charms.rst
erinecon: I mean, we already have a generic version above with `Use ``charmcraft init`` and specify the relevant profile:`. So it's probably safe to cut out of the tab block, and I have no objections to it.

docs/howto/manage-web-app-charms/index.rst
medubelko: It's an opinionated point, and I don't know the answer. I've heard "Juju ecosystem" before, though, so I thought I'd check.

docs/howto/manage-web-app-charms/init-web-app-charms.rst
medubelko: Agreed – let's leave the ref names as-is.

docs/howto/manage-web-app-charms/index.rst
erinecon: I'm fine with proceeding with Juju ecosystem and we can always iterate if UX research shows this is a confusing term.

docs/howto/manage-web-app-charms/index.rst
```diff
@@ -0,0 +1,52 @@
...
+
+.. toctree::
+   :maxdepth: 2
+
+   Initialize your 12-factor app charm <init-web-app-charms>
```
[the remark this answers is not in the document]
erinecon: I think a potential source of confusion was that I didn't use a subheading for the initialization of the charm and instead had the title serving two purposes (as both the page title and a section header). I added a new header to try and remove the hierarchy implication.

I also tested the numbered lists locally, and my preference is not to include it. I personally don't like how the sidebar navigation includes the numbering, and I personally didn't see much added value. Let me know what you think.

Also, I think technically the order doesn't matter -- and probably in most use cases the OCI image would be prepared first, which would imply that the ordering in the init file needs to be updated. I'm fine changing the order in this doc, but I was trying to put the Charmcraft-specific stuff at the forefront :)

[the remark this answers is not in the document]
medubelko: > didn't use a subheading for the initialization of the charm and instead had the title serving two purposes (as both the page title and a section header)

From a purely web-driven POV, that's fine. That's completely standard in larger doc sets. There's no reader expectation that a link always reaches a heading or a document. Readers might expect that, on average, _most_ links inside docs go to the tops of pages, so using a few here and there should be fine.

> I think technically the order doesn't matter

I believe we should strive to provide a sequenced workflow, even if from our measly perspective as TAs. From my understanding, these steps are constituents of the same task. In many of our other how-tos, we have steps that are organised purely logically, that could have been ordered differently or not ordered at all. But we prescribe a sequence to them anyway, as the user's guide.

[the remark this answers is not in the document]
erinecon: > I believe we should strive to provide a sequenced workflow 

Sure, that's totally reasonable to me. I'll swap the order of the two initialization sections.

Unless there are any other concerns, I think we should be good to go!

PR #2215, opened 2025-03-06
chore: add a documentation issue template

### LINKED ISSUE #2214: Change the issue templates so documentation issues get added here rather than linking to Discourse
### What needs to get done

Documentation issues should be a specific template type in GitHub, not just a link to Discourse

### Why it needs to get done

We no longer have our docs in Discourse.

lengau:
Replaces the documentation discourse link with an issue template to help describe documentation requests.

Fixes #2214 
CRAFT-4306

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -3,6 +3,3 @@ contact_links:
   - name: Charmcraft Discourse
     url: https://discourse.charmhub.io/c/charmcraft/3
     about: Issues creating charms
-  - name: Documentation Issues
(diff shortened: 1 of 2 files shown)


medubelko (commented):
Can we copy all the other nifty bits from the [template](https://github.com/canonical/rockcraft/blob/main/.github/ISSUE_TEMPLATE/docs.yaml?plain=1) in Rockcraft?

lengau:
@medubelko only if by "copy" you mean "steal wholesale because it's much better" :-)

medubelko (approved):
image

jahn-junior (approved):
After testing Rockcraft's issue template, I can confirm that this one was indeed worth "borrowing".

PR #2224, opened 2025-03-13
chore: remove tox and freeze-requirements files

lengau:
These are no longer required


upils (approved):
LGTM. Do we still need the `requirements-<series>.txt` files or can they also be removed?

lengau:
@upils those are still needed for now (for the snap), but I'll be able to drop them soon.

PR #2226, opened 2025-03-17
feat: mark the "promote-bundle" command as deprecated

lengau:
This command is deprecated and will be removed in Charmcraft 4.0. Bundle commands will remain available in Charmcraft 3.

CRAFT-4005

### DIFF
--- a/charmcraft/application/commands/store.py
+++ b/charmcraft/application/commands/store.py
@@ -1037,6 +1037,11 @@ class PromoteBundleCommand(CharmcraftCommand):
 
     def run(self, parsed_args: "Namespace") -> None:
         """Run the command."""
+        emit.progress(
+            "Bundle commands are deprecated and will be removed in Charmcraft 4.",
+            permanent=True,
+        )
+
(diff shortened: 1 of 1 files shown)


lengau:
Closed/reopened because workflows weren't being run

charmcraft/application/commands/store.py
```diff
@@ -1037,6 +1037,11 @@ def fill_parser(self, parser: "ArgumentParser") -> None:
 
     def run(self, parsed_args: "Namespace") -> None:
         """Run the command."""
+        emit.progress(
```
upils, line 1040: question: A test exists for the deprecated `register-bundle` command. Do you want to add one for this?

upils (approved):

charmcraft/application/commands/store.py
lengau: Nah, because it's going away in a couple of months anyway.

lengau:
Failures are unrelated (see: https://github.com/canonical/charmcraft/pull/2229), overriding

PR #2228, opened 2025-03-17
docs: Tutorial updates

erinecon:
Three updates that expand over all four tutorials. These changes were suggested from one or more UX sessions. 

Summary of changes:
1. If the user needs to open rockcraft.yaml or charmcraft.yaml to update the architecture to `arm64`, have them also comment out `amd64`. While there's no harm or issue with having both architectures available, for novice users, this change will eliminate any possible concern or worry over including both architectures in the YAML file(s).
2. Provide more explanation and reassurance about the web app entering a blocked state when adding the PostgreSQL database.
3. Rephrase the "tear things down" section to tell the users that they can just delete the whole instance if they're not going to keep the VM. 

Additional reviewer: @jdkandersson

### DIFF
--- a/docs/tutorial/kubernetes-charm-django.rst
+++ b/docs/tutorial/kubernetes-charm-django.rst
@@ -219,8 +219,8 @@ the architecture of your system:
 
     dpkg --print-architecture
 
-If your host uses ARM64 architecture, open ``rockcraft.yaml`` in a
-text editor and include ``arm64`` under ``platforms``.
+If your host uses ARM architecture, open ``rockcraft.yaml`` in a
+text editor, comment out ``amd64``, and include ``arm64`` under ``platforms``.
 
 Django apps require a database. Django will use a sqlite
 database by default. This won't work on Kubernetes because the database
--- a/docs/tutorial/kubernetes-charm-django.rst
+++ b/docs/tutorial/kubernetes-charm-django.rst
@@ -765,16 +765,7 @@ development process, including:
 - Exposing the app using an ingress
 - Adding an initial app and configuring the app
 
-If you'd like to reset your working environment, you can run the
-following in the rock directory ``~/django-hello-world`` for the tutorial:
-
-.. literalinclude:: code/django/task.yaml
-    :language: bash
-    :start-after: [docs:clean-environment]
-    :end-before: [docs:clean-environment-end]
-    :dedent: 2
-
-You can also clean up your Multipass instance. Start by exiting it:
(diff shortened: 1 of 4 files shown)


docs/tutorial/kubernetes-charm-django.rst
```diff
@@ -219,8 +219,8 @@ the architecture of your system:
 
     dpkg --print-architecture
 
-If your host uses ARM64 architecture, open ``rockcraft.yaml`` in a
-text editor and include ``arm64`` under ``platforms``.
+If your host uses ARM architecture, open ``rockcraft.yaml`` in a
+text editor, comment out ``amd64``, and include ``arm64`` under ``platforms``.
```
medubelko, line 223: @lengau Can Charmcraft run on ARMhf systems?

docs/tutorial/kubernetes-charm-django.rst
```diff
@@ -785,6 +778,17 @@ And then you can proceed with its deletion:
     multipass delete charm-dev
     multipass purge
 
+If you'd like to manually reset your working environment, you can run the
+following in the rock directory ``~/django-hello-world`` for the tutorial:
```
medubelko, line 782: Note for work in the next cycle: In other tutorials I made this a required step, even though it was optional. We're not to fork in tutorials.

docs/tutorial/kubernetes-charm-fastapi.rst
```diff
@@ -654,28 +660,30 @@ development process, including:
...
-You can then proceed with its deletion:
+And then you can proceed with its deletion:
 
 .. code-block:: bash
 
     multipass delete charm-dev
     multipass purge
 
+If you'd like to manually reset your working environment, you can run the
+following in the rock directory ``~/fastapi-hello-world`` for the tutorial:
```
medubelko, line 677:
```suggestion
If you'd like to manually reset your working environment, you can run the
following in the ``~/fastapi-hello-world`` directory:
```

This flows more cleanly to me, but I might be missing the meaning.

medubelko (approved):

jdkandersson (approved):

docs/tutorial/kubernetes-charm-django.rst
lengau: Technically yes, but the last time snapcraft.io reports a nonzero number of users on armhf (2) was April of last year. One would have to deploy their own juju runners to run charms on armhf AFAIK.

lengau (approved):

PR #2236, opened 2025-03-26
docs: add venv activation to the contribution instructions

dwilding:
I was working through the instructions in CONTRIBUTING.md but wasn't able to run `python -m charmcraft`. Then I realised that I needed to first activate the virtual environment created by `make setup`. That's probably obvious to many, but it tripped me up :slightly_smiling_face: 

This PR adds an explicit command for activating the virtual environment.


medubelko (approved):
LGTM, thanks!

PR #2237, opened 2025-03-26
docs: add security policy

tigarmo:
The policy distinguishes between Charmcraft itself and the charms it generates, describes the technologies used in build isolation and the how Charmcraft base support is related to Ubuntu's.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,61 @@
+# Security policy
+
+## Charmcraft and charms
+
+Charmcraft is a tool to create and publish charms. A charm is a software artifact that
+contains instructions to manage a separate application in the context of a [Juju] model,
+and without regular maintenance and updates it can become vulnerable.
+
+A charm's author or maintainer is the sole party responsible for its security. Charm
+authors should be diligent and keep the software inside their charms up-to-date with the
+latest releases, security patches, and security measures.
+
+Any vulnerabilities found in a charm should be reported to the charm's author or
+maintainer.
+
+## Build isolation
+
+In typical operation, Charmcraft makes use of tools like [LXD] and [Multipass] to create
+isolated build environments. Charmcraft itself provides no extra security, relying on
+these tools to provide secure sandboxing. The security of these build environments
+are the responsibility of these tools and should be reported to their respective
+project maintainers.
+
+Additionally, [destructive] builds are designed to give full access to the running host
+and are not isolated in any way.
+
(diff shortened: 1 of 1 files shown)


SECURITY.md
```diff
@@ -0,0 +1,61 @@
+# Security policy
+
+## Charmcraft and charms
+
+Charmcraft is a tool to create and publish charms. A charm is a software artifact that
+contains instructions to manage a separate application in the context of a [Juju] model,
```
tigarmo, line 6: possibly we should also mention things like resources but I'm not sure I know enough to write something decent

medubelko, line 6:
```suggestion
Charmcraft is a tool to create and publish charms. A component of the [Juju]
ecosystem, a charm is a piece of software that orchestrates and manages software
services,
```

Trying to simplify/clarify this part. @lengau Thoughts on this wording?

medubelko (approved):
LGTM, with a suggestion for the definition of a charm.

SECURITY.md
lengau: Not sure we should bother - the only thing Charmcraft really does with resources is upload them and do basic store operations.

lengau: I'm fine either way.

SECURITY.md
```diff
@@ -0,0 +1,61 @@
...
+
+### Supported bases
+
+Bases are tied to Ubuntu LTS releases. For example, the `ubuntu@24.04` base uses Ubuntu
+24.04 LTS as its build and runtime environment. This means that Charmcraft's support
+of bases aligns with the [Ubuntu LTS release cycle].
+
+The most recent major release of Charmcraft will always support bases still in their
+regular maintenance lifecycle. When a major release of Charmcraft drops support for a
+base, the previous major release remains supported until the dropped base reaches the
```
lengau, line 41: I'm not sure how to specify that we're only supporting it for those bases that aren't in the newer version.

E.g. Charmcraft 4.0 is going to drop centos 7, but that means that Charmcraft 3 will be supported *only* for centos 7 (but not for Ubuntu 20.04).

medubelko: > those bases that aren't in the newer version

That's the intended meaning behind "a major release of Charmcraft drops support for a base...".

It took some time to workshop this paragraph due to the difficulty in expressing this idea. I'd be happy to consider any improvements.

lengau (approved):

Issue #2235, opened 2025-03-25
Feedback on the Django tutorial


evildmp:
### Request type

Fix

### What needs to get done

* [ ] _If you’re new to the charming world_ note - the eyes skip over this somewhat. Should probably just be a normal paragraph, after the very first one int he document.
* [ ] _Warning: This tutorial requires version 3.2.0 or later of Charmcraft_ - seems quite out of order here, we don't get near charm craft till much later. If we are serious about using Multipass as described, we'll _know_ what the user has installed.
* [ ] _ First, install Multipass_ > _Install Multipass if you don't already have it installed. See How to install Multipass._
* [ ] _[Multipass | How to install Multipass](https://multipass.run/docs/install-multipass)_ link is broken.
* [ ] After _juju bootstrap microk8s dev-controller_ seems like the right point for a rest, before `mkdir django-hello-world` - perhaps the `Create the Django app` heading should be here.
* [ ] _Create the Django app_  - _app_ is highly overloaded, perhaps _Start a new Django project_ 
* [ ] `requirements.txt` - `touch` seems superfluous; why not just edit with nano?
* [ ] The `django-hello-world/django_hello_world/django_hello_world` structure is very confusing. We might not be able to escape `django_hello_world/django_hello_world`, but at least the first part could be different, e.g. `django-charm-project`.
* [ ] `MULTIPASS_PRIVATE_IP` - all-caps suggests a constant
* [ ] At that point, http://192.168.64.14:8000 worked perfectly in Safari but not in Chrome ("the site could not be reached"). Unclear why and not sure what to suggest.
* [ ] _Pack the Django app into a rock_ - this is the moment at which I'd appreciate a few words about what a rock is, as reminder if nothing else. Also helps signal that we are entering rock world.
* [ ] _We’ll use Juju later to easily deploy a database._ - no _easily_ please.
* [ ] _Ensure that platforms includes the architecture of your host_. It's ambiguous, should it include anything else? Suggest: _Platforms must match the architecture of your host_. 
* [ ] _If your host uses ARM architecture, open rockcraft.yaml in a text editor, comment out amd64, and include arm64 under platforms._ > _Edit the platforms entry in rockcraft.yaml if required. 
* [ ] Paragraph _ We’ll need to update the settings.py file to prepare_ is a headful. Why not _Back in settings.py..._ and let the user work it out, they need that muscle memory of where to find it.
* [ ] database settings code has indentation of 3 and 6 instead of 4 and 8.
[...]

PR #2251, opened 2025-04-11
Docs: Updated 12-factor tutorials

erinecon:
Updates to the 12-factor tutorials that affect the reusable setup files and all four tutorials. Some of these changes were suggested by @evildmp in https://github.com/canonical/charmcraft/issues/2235

### Summary of changes
* Updated the `setup_edge.rst` file to install the `latest/edge` channel of Rockcraft.
* Removed an admonition block in the Django tutorial that warned users about using an older version of Rockcraft. 
* Removed any mentions of "easily" in the tutorials.
* Removed `touch` from the instructions and ask users to instead open the new file directly with `nano`.
* Removed the admonition block warning users about the Charmcraft version.

### DIFF
--- a/docs/tutorial/kubernetes-charm-django.rst
+++ b/docs/tutorial/kubernetes-charm-django.rst
@@ -83,9 +83,8 @@ Create the Django app
 Let's start by creating the "Hello, world" Django app that
 will be used for this tutorial.
 
-Create and open a new file ``requirements.txt`` in a text editor using
-``nano requirements.txt``, copy the following text into it and then
-save the file:
+Create a new requirements file with ``nano requirements.txt``.
+Then, copy the following text into it, and save:
 
 .. literalinclude:: code/django/requirements.txt
     :caption: ~/django-hello-world/requirements.txt
--- a/docs/tutorial/kubernetes-charm-django.rst
+++ b/docs/tutorial/kubernetes-charm-django.rst
@@ -25,7 +25,7 @@ This tutorial should take 90 minutes for you to complete.
     A rock is a special kind of OCI-compliant container image, while a
     charm is a software operator for cloud operations that use the Juju
     orchestration engine. The result is Django apps that
-    can be easily deployed, configured, scaled, integrated, etc.,
+    can be deployed, configured, scaled, integrated, etc.,
     on any Kubernetes cluster.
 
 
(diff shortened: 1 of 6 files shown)


docs/tutorial/kubernetes-charm-django.rst
```diff
@@ -25,7 +25,7 @@ This tutorial should take 90 minutes for you to complete.
     A rock is a special kind of OCI-compliant container image, while a
     charm is a software operator for cloud operations that use the Juju
     orchestration engine. The result is Django apps that
-    can be easily deployed, configured, scaled, integrated, etc.,
+    can be deployed, configured, scaled, integrated, etc.,
```
medubelko, line 28: While we're editing this line, can we do this too? :)

```suggestion
    can be deployed, configured, scaled, integrated, and so on,
```

docs/tutorial/kubernetes-charm-django.rst
```diff
@@ -83,9 +83,9 @@ Create the Django app
 Let's start by creating the "Hello, world" Django app that
 will be used for this tutorial.
 
-Create a ``requirements.txt`` file using ``touch requirements.txt``.
-Then, open the file in a text editor using ``nano requirements.txt``,
-copy the following text into it and then save the file:
+Create and open a new file ``requirements.txt`` in a text editor using
+``nano requirements.txt``, copy the following text into it and then
+save the file:
```
medubelko, line 88: Following the advice about functional and mechanical actions, I think this could be even clearer:

```suggestion
Create a new requirements file with ``nano requirements.txt``.
Then, copy the following text into it, and save:
```

Also consider that for this audience, we probably don't need to specify a particular text editor. I think for many readers the mention of nano would be distracting, since it's usually an editor of last resort.

We could just say:

> Create a `requirements.txt` file and copy the following text into it:

medubelko (approved):

docs/tutorial/kubernetes-charm-django.rst
erinecon: Specifying a specific text editor was a request from several UX sessions. For a part of our user base (mostly students, the "Novice User" as we've named this particular user persona), they don't have background knowledge on using terminal text editors. Without being specific about a terminal text editor, the Novice User gets immediately stuck and cannot continue without assistance. I have some recorded UX sessions where this happens, and I can share them with you if you'd like.

I'm aware that it's probably a minority of the users, and I know that `nano` isn't everyone's preferred choice of terminal text editor (it isn't mine, for instance) but I do feel strongly like we need to provide *something* specific here.

medubelko: Do you think the original suggestion would work?

erinecon: The first suggestion will work. I'll update now, thanks :)

lengau (approved):

PR #2253, opened 2025-04-16
docs: add charm docs exemplars links

izmalk:
Improve charm docs guidance to include smaller/bigger charm docs. 
Add links to the charm docs exemplars.

This is a part of the [Epic](https://warthogs.atlassian.net/browse/DOCPR-948) to Standardise charm documentation by creating exemplars of good.

### DIFF
--- a/docs/howto/manage-charms.rst
+++ b/docs/howto/manage-charms.rst
@@ -234,16 +234,12 @@ charm's project file, specify an item under the :ref:`links.website
 Add docs
 ~~~~~~~~
 
-When you publish your charm on Charmhub, reference documentation about the charm's
+If you publish your charm on Charmhub, reference documentation about the charm's
 resources, actions, configurations, relations, and libraries is generated and 
 published automatically in respective tabs.
-Use the Description tab to add additional documentation, such as tutorials or 
-how-to guides, focusing specifically on the charm itself. 
-For workload-specific or Juju-related content, 
-link to the appropriate upstream documentation.
 
-To add content to the Description tab, create a Discourse topic and include its URL 
-in your charm's project file under the 
+To add content to the Description tab, 
+create a Discourse topic and include its URL in your charm's project file under the 
 :ref:`links.documentation <charmcraft-yaml-key-documentation>` key:
 
 .. code-block:: yaml
@@ -255,15 +251,23 @@ in your charm's project file under the
 
     See more: :ref:`charmcraft-yaml-key-links`
 
-Smaller charms can have a single-page documentation in the Description tab. 
-Bigger charms with multi-page documentation, should have a full
-[Diátaxis](https://diataxis.fr/) navigation tree in the Description tab, 
-or just provide a brief description with a link to an external documentation set. 
+The Description tab can contain additional documentation, such as tutorials or 
+how-to guides.
+
+.. note::
+   A charm's documentation should focus on the charm itself.
+   For workload-specific or Juju-related content, link to the appropriate upstream 
+   documentation.
+
+A smaller charm can have a single-page documentation for its description. 
+A bigger charm, that needs a multi-page documentation, can have either
+a brief description with a link to an external documentation set, or
+a full [Diátaxis](https://diataxis.fr/) navigation tree in the Description tab.
 
 .. admonition:: Best practice
     :class: hint
 
-    Smaller charm documentation exemplars:
+    Smaller charm documentation examples:
 
     * [Azure storage integrator](https://charmhub.io/azure-storage-integrator) charm
     * [Repo policy compliance](https://charmhub.io/repo-policy-compliance) charm
@@ -271,7 +275,7 @@ or just provide a brief description with a link to an external documentation set
 .. admonition:: Best practice
     :class: hint
 
-    Bigger charm documentation exemplars:
+    Bigger charm documentation examples:
 
     * [OpenSearch](https://charmhub.io/opensearch) charm 
     * [Wordpress-k8s](https://charmhub.io/wordpress-k8s) charm
--- a/docs/howto/manage-charms.rst
+++ b/docs/howto/manage-charms.rst
@@ -234,12 +234,17 @@ charm's project file, specify an item under the :ref:`links.website
 Add docs
 ~~~~~~~~
 
-If you publish your charm on Charmhub, reference documentation about the charm's
-resources, actions, configurations, relations, and libraries is extracted automatically.
-However, you should also aim to add further docs, e.g., a tutorial, how-to guides, etc.
-To provide a link to these docs, in your charm's project file, specify a
-value for the :ref:`links.documentation <charmcraft-yaml-key-documentation>` key. Note
-that at present this must be a Discourse page. E.g.,
+When you publish your charm on Charmhub, reference documentation about the charm's
+resources, actions, configurations, relations, and libraries is generated and 
+published automatically in respective tabs.
+Use the Description tab to add additional documentation, such as tutorials or 
+how-to guides, focusing specifically on the charm itself. 
+For workload-specific or Juju-related content, 
+link to the appropriate upstream documentation.
+
+To add content to the Description tab, create a Discourse topic and include its URL 
+in your charm's project file under the 
+:ref:`links.documentation <charmcraft-yaml-key-documentation>` key:
 
 .. code-block:: yaml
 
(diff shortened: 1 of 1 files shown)


docs/howto/manage-charms.rst
```diff
@@ -234,12 +234,17 @@ charm's project file, specify an item under the :ref:`links.website
 Add docs
 ~~~~~~~~
 
-If you publish your charm on Charmhub, reference documentation about the charm's
-resources, actions, configurations, relations, and libraries is extracted automatically.
-However, you should also aim to add further docs, e.g., a tutorial, how-to guides, etc.
-To provide a link to these docs, in your charm's project file, specify a
-value for the :ref:`links.documentation <charmcraft-yaml-key-documentation>` key. Note
-that at present this must be a Discourse page. E.g.,
+When you publish your charm on Charmhub, reference documentation about the charm's
```
medubelko, line 237: > When

I have a feeling that the "if" was there to suggest that publication to Charmhub wasn't mandatory. We usually keep private stores in mind.

docs/howto/manage-charms.rst
```diff
@@ -234,12 +234,17 @@ charm's project file, specify an item under the :ref:`links.website
...
+resources, actions, configurations, relations, and libraries is generated and 
+published automatically in respective tabs.
+Use the Description tab to add additional documentation, such as tutorials or
```
medubelko, line 240:
```suggestion

On the **Description** tab, add additional documentation, such as tutorials or 
```

I think this would read better with a paragraph break here.

docs/howto/manage-charms.rst
```diff
@@ -234,12 +234,17 @@ charm's project file, specify an item under the :ref:`links.website
...
+resources, actions, configurations, relations, and libraries is generated and 
+published automatically in respective tabs.
+Use the Description tab to add additional documentation, such as tutorials or 
+how-to guides, focusing specifically on the charm itself. 
+For workload-specific or Juju-related content, 
+link to the appropriate upstream documentation.
+
+To add content to the Description tab, create a Discourse topic and include its URL 
+in your charm's project file under the 
+:ref:`links.documentation <charmcraft-yaml-key-documentation>` key:
```
medubelko, line 247: Keep in mind I've never managed a charm, but I feel like these two paragraphs step on each other. Both describe adding content on the tab but in different ways. Should they be combined?

docs/howto/manage-charms.rst
```diff
@@ -250,6 +255,26 @@ that at present this must be a Discourse page. E.g.,
 
     See more: :ref:`charmcraft-yaml-key-links`
 
+Smaller charms can have a single-page documentation in the Description tab. 
+Bigger charms with multi-page documentation, should have a full
+[Diátaxis](https://diataxis.fr/) navigation tree in the Description tab, 
+or just provide a brief description with a link to an external documentation set.
```
medubelko, line 261:
```suggestion
A smaller charm can have a single-page documentation for its description. 
A bigger charm, with multi-page documentation, should have a full
[Diátaxis](https://diataxis.fr/) navigation tree in its description, 
or just provide a brief description with a link to an external documentation set. 
```

I think we can reduce how much we speak about the tab itself.

docs/howto/manage-charms.rst
```diff
@@ -250,6 +255,26 @@ that at present this must be a Discourse page. E.g.,
...
 
+Smaller charms can have a single-page documentation in the Description tab. 
+Bigger charms with multi-page documentation, should have a full
+[Diátaxis](https://diataxis.fr/) navigation tree in the Description tab, 
+or just provide a brief description with a link to an external documentation set. 
+
+.. admonition:: Best practice
+    :class: hint
+
+    Smaller charm documentation exemplars:
```
medubelko, line 266: > exemplar

This term came up before in a discussion about user-facing docs. Do you think for ease of understanding we should use "example" instead?

medubelko (approved):
Thanks, @izmalk! I left some copy suggestions. The only one that's significant is the question about the tab.

docs/howto/manage-charms.rst
lengau: Or indeed charms that never go into a store and just get deployed through CI/CD

docs/howto/manage-charms.rst
lengau: Agreed - combined witch your previous suggestion of putting this in a new paragraph that flows better IMO.

docs/howto/manage-charms.rst
izmalk: Reverted the changes back to "if".

docs/howto/manage-charms.rst
izmalk: I've experimented with it a bit. Please check the [new version](https://github.com/canonical/charmcraft/pull/2253/commits/8ef59e0da29fd0535a9d45fe03fa8ec90af3ef13)

docs/howto/manage-charms.rst
izmalk: I don't think we can replace all mentions of the tab, since we need to refer to the exact UI element sometimes, but I've adjusted the text to minimise the usage. Up to one per paragraph.

docs/howto/manage-charms.rst
izmalk: Yes, I see how the word example can be easier to understand. Although the meaning is not precisely the same, I guess in this context we can replace it. Fixed in https://github.com/canonical/charmcraft/pull/2253/commits/8ef59e0da29fd0535a9d45fe03fa8ec90af3ef13

docs/howto/manage-charms.rst
izmalk: Fixed in https://github.com/canonical/charmcraft/pull/2253/commits/8ef59e0da29fd0535a9d45fe03fa8ec90af3ef13

docs/howto/manage-charms.rst
```diff
@@ -250,6 +251,34 @@ that at present this must be a Discourse page. E.g.,
...
+The Description tab can contain additional documentation, such as tutorials or 
+how-to guides.
+
+.. note::
+   A charm's documentation should focus on the charm itself.
+   For workload-specific or Juju-related content, link to the appropriate upstream 
+   documentation.
+
+A smaller charm can have a single-page documentation for its description. 
+A bigger charm, that needs a multi-page documentation, can have either
```
medubelko, line 263:
```suggestion
A smaller charm can have a single-page document for its description. 
A bigger charm, that needs multi-page documentation, can have either
```

Grammar-related.

docs/howto/manage-charms.rst
```diff
@@ -250,6 +251,34 @@ that at present this must be a Discourse page. E.g.,
 
     See more: :ref:`charmcraft-yaml-key-links`
 
+The Description tab can contain additional documentation, such as tutorials or 
+how-to guides.
+
+.. note::
```
medubelko, line 257: I like the update! I think it flows well.

Do you think we need the note block? I think it could work as a regular paragraph.

docs/howto/manage-charms.rst
```diff
@@ -235,11 +235,12 @@ Add docs
...
 If you publish your charm on Charmhub, reference documentation about the charm's
-resources, actions, configurations, relations, and libraries is extracted automatically.
-However, you should also aim to add further docs, e.g., a tutorial, how-to guides, etc.
-To provide a link to these docs, in your charm's project file, specify a
-value for the :ref:`links.documentation <charmcraft-yaml-key-documentation>` key. Note
-that at present this must be a Discourse page. E.g.,
+resources, actions, configurations, relations, and libraries is generated and 
+published automatically in respective tabs.
+
+To add content to the Description tab,
```
medubelko, line 241: Any chance we can use bold to fit with the recommended UI formatting? https://docs.ubuntu.com/styleguide/en/#bold-vs.-italics

docs/howto/manage-charms.rst
medubelko: Looks great!

docs/howto/manage-charms.rst
izmalk: Yes, we can delete the articles, but I don't think the other word has the same meaning. We really want to talk about charm's documentation that fits into a single web page.

```suggestion
A smaller charm can have single-page documentation for its description. 
A bigger charm, that needs multi-page documentation, can have either
```

docs/howto/manage-charms.rst
izmalk: Yes, this fragment is important enough to use the note block.
Given that we are in a How-to guide, it feels too generic to put it in the middle of more practical instructions as just usual text.

medubelko:
Looks good from my end. Awaiting review from @lengau.

lengau (approved):

PR #2256, opened 2025-04-17
12-factor docs: How to add a custom action

erinecon:
Add a new section in `configure-web-app-charms.rst` called "Add a custom action". The section provides a brief generic description covered in more detail in the Ops documentation followed by two example custom actions. 

Each example has a Spread test and accompanying materials. They have been tested and pass with the `multipass` backend on my local machine. I've turned on the how-to Spread tests in `spread.yaml` to see whether they pass, but I will revert the change before the changes are merged.


docs/howto/manage-web-app-charms/configure-web-app-charm.rst
```diff
@@ -85,8 +85,132 @@ The configuration can be set on the deployed charm using:
 
     juju config <app name> token=<token>
 
-Manage secrets for a 12-factor app charm
-----------------------------------------
+Add a custom action
+-------------------
+
+For certain commands or processes that you will run many times, you can set up
+custom Juju actions to expedite your workflow. Actions are handled in the Ops
```
medubelko, line 92: > you will run many times

The "you will" throws me a bit here because of the tone of a how-to. Can we be more descriptive with:

```suggestion
For commands and processes that that see shared or frequent use,
custom Juju actions can expedite your workflow. Actions are handled in the Ops
```

docs/howto/manage-web-app-charms/configure-web-app-charm.rst
```diff
@@ -85,8 +85,132 @@ The configuration can be set on the deployed charm using:
...
+library, and Juju manages the action's workflow via tasks and operations. To
+implement a custom action, you need to declare the action in
+``charmcraft.yaml`` and define an event handler into the Ops framework of your
```
medubelko, line 95: > define... into

"into" wouldn't be my choice for a preposition. Should it be "for"? Or maybe "add... to"?

docs/howto/manage-web-app-charms/configure-web-app-charm.rst
```diff
@@ -85,8 +85,132 @@ The configuration can be set on the deployed charm using:
...
+charm.
+
+.. seealso::
+
+    :external+ops:ref:`Ops | Manage actions <manage-actions>`
+
+The custom action must be defined in ``charmcraft.yaml`` under the
```
medubelko, line 102:
```suggestion
The custom action must be defined in the project file under the
```

I suppose it's time to bring this phrase up. It's the generic phrase for all `*craft.yaml` files in our products, and reduces the amount of code in the text.

Kindly consider it for all instances.

docs/howto/manage-web-app-charms/configure-web-app-charm.rst
```diff
@@ -85,8 +85,132 @@ The configuration can be set on the deployed charm using:
...
+    Waiting for task 2...
+    result: session cleared!
+
+
+Example: check Flask app status and write output to container file
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+As another example, let's say you want to add a custom action to your charmed
+Flask app that checks the status of the running app and writes the status to a
+logfile in the app container. The action performs the following steps:
```
medubelko, line 158:
```suggestion
log file in the app container. The action performs the following steps:
```

docs/howto/manage-web-app-charms/configure-web-app-charm.rst
```diff
@@ -85,8 +85,132 @@ The configuration can be set on the deployed charm using:
...
+Example: clear cookies in a Django app
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+For example, let's say you want to add a custom action to your charmed Django
+app that removes expired session cookies.
+
+Add the custom action in ``charmcraft.yaml``:
+
+.. literalinclude:: ../code/custom-action-clearsession/clearsession_action_charmcraft.yaml
+    :language: yaml
```
medubelko, line 124:
```suggestion
    :language: yaml
    :caption: charmcraft.yaml
```

It would be nice if all the code blocks had these.

docs/howto/manage-web-app-charms/configure-web-app-charm.rst
```diff
@@ -85,8 +85,132 @@ The configuration can be set on the deployed charm using:
...
+
+As another example, let's say you want to add a custom action to your charmed
+Flask app that checks the status of the running app and writes the status to a
+logfile in the app container. The action performs the following steps:
+
+- Sends a request to the Flask app at its available port.
+- If the request is successful, updates the event container with the status
+  message in a logfile.
+
+Add the custom action in ``charmcraft.yaml``:
```
medubelko, line 164:
```suggestion
Add the custom action to ``charmcraft.yaml``:
```

docs/howto/manage-web-app-charms/configure-web-app-charm.rst
```diff
@@ -85,8 +85,132 @@ The configuration can be set on the deployed charm using:
...
+
+.. literalinclude:: ../code/custom-action-updatelogfile/updatelogfile_action_charmcraft.yaml
+    :language: yaml
+
+Modify the ``src/charm.py`` to include ``import requests`` at the top, define
```
medubelko, line 169:
```suggestion
Add ``import requests`` to the start of ``src/charm.py``, then define
```

docs/howto/manage-web-app-charms/configure-web-app-charm.rst
```diff
@@ -85,8 +85,132 @@ The configuration can be set on the deployed charm using:
...
+Modify the ``src/charm.py`` to define your custom action as part of the class
+and provide the function definition:
+
+.. literalinclude:: ../code/custom-action-clearsession/clearsession_action_charm.py
+    :language: python
+    :lines: 20-46
+    :emphasize-lines: 8-27
+
+Build the charm using ``charmcraft pack`` and deploy the app with Juju.
+Finally, call the action using:
```
medubelko, line 135:
```suggestion

Finally, call the action:
```

docs/howto/manage-web-app-charms/configure-web-app-charm.rst
```diff
@@ -85,8 +85,132 @@ The configuration can be set on the deployed charm using:
...
+
+    Running operation 1 with 1 task
+      - task 2 on unit-flask-app-0
+
+    Waiting for task 2...
+    result: |-
+      Hello, world!
+       written to file /tmp/example.log in app container
+
+Check that the file was updated using Pebble:
```
medubelko, line 197: This feels like the step that ought to start with "finally".

medubelko (approved):

docs/howto/code/custom-action-clearsession/clearsession_action_charm.py
```diff
@@ -0,0 +1,49 @@
...
+        try:
+            self._container.exec(
+                ["python3", "manage.py", "clearsessions"],
+                service_context="django",
+                combine_stderr=True,
+                working_dir=str(self._workload_config.app_dir),
+            ).wait_output()
+            event.set_results({"result": "session cleared!"})
+        except ops.pebble.ExecError as e:
+            event.fail(str(e.stdout))
```
dwilding, line 46: This should be `stderr` shouldn't it? Same for the other charm

```suggestion
            event.fail(str(e.stderr))
```

dwilding (commented):
These doc updates are exactly what I would have needed when adding an action to a Django app charm recently. Nice :smiley: 

One concern/suggestion about the Flask example: Is that adapted from something that app developers want to do? On the face of it, it seems a bit indirect for the charm to push the output to the app container and then have the admin manually SSH into the container to read the output.

My feeling is, the example doesn't need to worry about pushing files around - showing how to talk to the workload (with `requests.get` and `self._workload_config.port` etc) is already really helpful. How about hitting a placeholder endpoint such as `http://127.0.0.1:{self._workload_config.port}/status` and directly returning `response.text` as the action result?

spread.yaml
```diff
@@ -136,7 +136,7 @@ suites:
     summary: tests howto from the docs
     systems:
       - ubuntu-22.04-64
-    manual: true
+    manual: false
```
lengau, line 139: If you're not planing on turning this back to manual any time soon, this might as will just go back to the default :-)

```suggestion
```

erinecon:
> These doc updates are exactly what I would have needed when adding an action to a Django app charm recently. Nice 😃
> 
> One concern/suggestion about the Flask example: Is that adapted from something that app developers want to do? On the face of it, it seems a bit indirect for the charm to push the output to the app container and then have the admin manually SSH into the container to read the output.
> 
> My feeling is, the example doesn't need to worry about pushing files around - showing how to talk to the workload (with `requests.get` and `self._workload_config.port` etc) is already really helpful. How about hitting a placeholder endpoint such as `http://127.0.0.1:{self._workload_config.port}/status` and directly returning `response.text` as the action result?

Hi @dwilding , thanks for taking a look at this PR! 

Originally, the Flask example was written in the way you described: Hitting the placeholder endpoint and directly returning the result. The feedback I received from one of my team's engineering managers was that there are circumstances where a developer will want to modify or write directly to the app container, and it's a more useful example to show how to do that. I added the part about SSHing into the container as a way to verify that the action worked.

I'm not a web app developer, so I don't have a great feeling about the specific circumstances in which a developer would want to directly change the app container. But this is the reason why the Flask example action pushes `response.text` to the app container itself.

spread.yaml
erinecon: Thanks -- I'll turn back off in a separate commit so I can run the Spread tests one last time!

erinecon: Spread tests are passing -- turned them back off :)

erinecon: Well I thought I had turned off the tests, but they ran again...do I need to explicitly remove `manual: true`?

docs/howto/code/custom-action-clearsession/clearsession_action_charm.py
dwilding: Oh, I just see that for `_on_clearsession_action` you have `combine_stderr=True` in the `exec` call, which means that `e.stdout` would be the correct thing to use (stdout and stderr will be combined into stdout) - sorry for the misleading comment

For `_on_updatelogfile_action` you're not doing an `exec` call, so I think the errors need to be handled differently from what you currently have. I'll double check now

docs/howto/code/custom-action-updatelogfile/updatelogfile_action_charm.py
```diff
@@ -0,0 +1,57 @@
...
+            response.raise_for_status()
+            self._container.push(event.params["logfile"], response.text)
+            output = response.text
+            output += " written to file " + event.params["logfile"]
+            event.set_results({"result": output})
+        except ops.pebble.ExecError as e:
+            event.fail(str(e.stderr))
+        except requests.exceptions.RequestException as e:
+            # if it failed with http bad status code or the connection failed
+            event.fail(str(e.stderr))
```
dwilding, line 54: `RequestException` won't have a stdout/stderr attribute. But this should work (I just tested it locally):

```suggestion
        except requests.exceptions.RequestException as e:
            if e.response is None:
                event.fail(f"unable to connect on port {self._workload_config.port}")
            else:
                event.fail(f"workload responded with code {e.response.status_code}")

```

dwilding:
> The feedback I received from one of my team's engineering managers was that there are circumstances where a developer will want to modify or write directly to the app container, and it's a more useful example to show how to do that. I added the part about SSHing into the container as a way to verify that the action worked.

It still feels a bit strange to me because the app doesn't use the data that's being written to the container. I guess the original thinking might have been to push config data (or similar) into the container, for the app to use?

How about splitting this into two separate examples:

* One example of hitting an HTTP endpoint and returning the output in the action results (a reduced version of what you have now)
* Inspired by the [dashboard charm](https://github.com/canonical/dashboard), an example of pushing a custom [initial_data.yaml](https://github.com/canonical/dashboard/blob/main/dashboard/initial_data.yaml) into the app container

What do you think? I know it's easy to make suggestions, so if you take this approach, let me know if I should pitch in!

erinecon:
> It still feels a bit strange to me because the app doesn't use the data that's being written to the container. I guess the original thinking might have been to push config data (or similar) into the container, for the app to use?
> 
> How about splitting this into two separate examples:
> 
> * One example of hitting an HTTP endpoint and returning the output in the action results (a reduced version of what you have now)
> * Inspired by the [dashboard charm](https://github.com/canonical/dashboard), an example of pushing a custom [initial_data.yaml](https://github.com/canonical/dashboard/blob/main/dashboard/initial_data.yaml) into the app container
> 
> What do you think? I know it's easy to make suggestions, so if you take this approach, let me know if I should pitch in!

Hi @dwilding ,

The original request for the action was "...create an action that updates the [content] of a file in the container based on input and then that content gets returned on one of the API endpoints". The exact wording of the feedback (after the original action was written) was "I was thinking the action would update a file in the container and to have an endpoint that returns the contents of that file. You can use this for a kind of app status message for users or something". 

I don't see the value in splitting this example into two separate ones. The first example you suggested is very similar to the original version of this action, which got pushback for not being demonstrative enough. Furthermore, I think the two demonstrations you're proposing -- hitting the HTTP endpoint, and having the app use data pushed into its container -- could be combined into a single example.

The current example could be expanded to e.g. return the contents of the file to an API endpoint, as the original request intended -- then the action would demonstrate how data pushed to the app container can be used, and the instruction to SSH into the container could be dropped. I think I will go in that direction. Thanks again for the feedback!

dwilding:
> The current example could be expanded to e.g. return the contents of the file to an API endpoint, as the original request intended -- then the action would demonstrate how data pushed to the app container can be used, and the instruction to SSH into the container could be dropped

Sounds good!

docs/howto/code/custom-action-updatelogfile/updatelogfile_action_charm.py
```diff
@@ -0,0 +1,57 @@
...
+            response = requests.get(
+                    f"http://127.0.0.1:{self._workload_config.port}", timeout=5
+                    )
+            response.raise_for_status()
+            self._container.push(event.params["logfile"], response.text)
+            output = response.text
+            output += " written to file " + event.params["logfile"]
+            event.set_results({"result": output})
+        except ops.pebble.ExecError as e:
+            event.fail(str(e.stderr))
```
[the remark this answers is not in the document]
erinecon: Thanks! Will update the line

docs/howto/code/custom-action-updatelogfile/updatelogfile_action_charm.py
```diff
@@ -0,0 +1,67 @@
...
+            event.fail("flask-app container is not ready")
+        try:
+            response = requests.get(
+                    f"http://127.0.0.1:{self._workload_config.port}", timeout=5
+                    )
+            response.raise_for_status()
+            # push response to file in app container
+            self._container.push(event.params["logfile"], response.text)
+            output = "App response: " + response.text
+            # access file in container and read its contents
```
dwilding, line 49: How about adding an extra comment here:

```suggestion
            # access file in container and read its contents
            # (you could put this part in a separate action called readlogfile)
```

docs/howto/manage-web-app-charms/configure-web-app-charm.rst
```diff
@@ -85,8 +85,131 @@ The configuration can be set on the deployed charm using:
...
+
+Example: check Flask app status and write output to container file
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+As another example, let's say you want to add a custom action to your charmed
+Flask app that checks the status of the running app and writes the status to a
+log file in the app container. The action performs the following steps:
+
+- Sends a request to the Flask app at its available port.
+- If the request is successful, updates the event container with the status
```
dwilding, line 164:
```suggestion
- If the request is successful, updates the app container with the status
```

docs/howto/manage-web-app-charms/configure-web-app-charm.rst
```diff
@@ -85,8 +85,131 @@ The configuration can be set on the deployed charm using:
...
+  message in a logfile.
+- Reads the contents of the logfile in the event container and outputs the
+  status message as part of the action result.
```
dwilding, line 167: I think that someone reading this (and the code) might wonder why we push the file to the container then immediately pull it back again. How about adding a couple of extra parts:

```suggestion
- As a consistency check, reads the contents of the log file in the app
  container and outputs the status message as part of the action result.
  In practice, you could create a separate action for reading the logfile.
```

dwilding (approved):

lengau (approved):
Thanks! I'm happy with this once you've resolved @dwilding's comments

docs/howto/code/custom-action-updatelogfile/updatelogfile_action_charm.py
```diff
@@ -0,0 +1,67 @@
...
+            output_comp = self._container.pull(event.params["logfile"]).read()
+            output += "Output written to file: " + output_comp
+            event.set_results({"result": output})
+        except ops.pebble.ExecError as e:
```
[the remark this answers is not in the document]
erinecon: Good catch -- thanks!

erinecon:
Thanks so much for the thorough review, @dwilding ! 
If every looks good on your ends @medubelko and @lengau , please go ahead and merge the PR :)

medubelko:
@lengau Shall we run the tests locally to double-check?

PR #2257, opened 2025-04-17
Docs: 12-factor tutorial updates

erinecon:
Six updates proposed by @evildmp in https://github.com/canonical/charmcraft/issues/2235 that affect all four of the 12-factor tutorials.

### Summary of changes
* Converted first note admonition into a normal paragraph.
* Updated the ambiguity of "Ensure that platforms includes the architecture of your host." by using the suggestion "Platforms must match the architecture of your host."
* Updated the instruction to edit `rockcraft.yaml` using the suggestion. 
* Removed an explicit reference to AMD64 architecture at the Juju model constraint step to let the user build up muscle memory. Architecture is still mentioned because that's what the command is explicitly doing.
* (Django tutorial only) Updated "The generated Django app doesn’t come with an app" to reference the Django *project* instead.
* (Django tutorial only) Updated `MULTIPASS_PRIVATE_IP` to use regular text formatting, as the original spelling implied a variable.

### DIFF
--- a/docs/tutorial/kubernetes-charm-django.rst
+++ b/docs/tutorial/kubernetes-charm-django.rst
@@ -217,7 +217,7 @@ the architecture of your system:
 
     dpkg --print-architecture
 
-Edit the ``platforms`` entry in ``rockcraft.yaml`` if required.
+Edit the ``platforms`` key in ``rockcraft.yaml`` if required.
 
 Django apps require a database. Django will use a sqlite
 database by default. This won't work on Kubernetes because the database
--- a/docs/tutorial/kubernetes-charm-django.rst
+++ b/docs/tutorial/kubernetes-charm-django.rst
@@ -544,7 +544,7 @@ Add a "Hello, world" app
 
 In this iteration, we'll add a greeting app that returns a ``Hello, world!`` greeting.
 
-The generated Django app doesn't come with an app, which is why
+The generated Django project doesn't come with an app, which is why
 we had to initially enable debug mode for testing.  We will need to go back
 out to the ``~/django-hello-world`` directory where the rock is and enter
 into the ``./django_hello_world`` directory where the Django app
--- a/docs/tutorial/kubernetes-charm-django.rst
+++ b/docs/tutorial/kubernetes-charm-django.rst
@@ -406,8 +406,7 @@ the Django app. Let's create a new model:
     :end-before: [docs:add-juju-model-end]
     :dedent: 2
 
-Include a constraint to the Juju model to specify your architecture if you
-aren't on a host with AMD64:
+Include a constraint to the Juju model to specify your architecture:
 
 .. literalinclude:: code/django/task.yaml
     :language: bash
--- a/docs/tutorial/kubernetes-charm-django.rst
+++ b/docs/tutorial/kubernetes-charm-django.rst
@@ -23,7 +23,7 @@ specifically supported with a template to quickly generate a
 **rock** and a matching template to generate a **charm**.
 A rock is a special kind of OCI-compliant container image, while a
 charm is a software operator for cloud operations that use the Juju
-orchestration engine. The result is Django apps that
+orchestration engine. The combined result is a Django app that
 can be deployed, configured, scaled, integrated, and so on,
 on any Kubernetes cluster.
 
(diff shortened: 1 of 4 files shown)


docs/tutorial/kubernetes-charm-django.rst
```diff
@@ -18,15 +18,14 @@ running with Juju. Let's get started!
...
-    specifically supported with a template to quickly generate a
-    **rock** and a matching template to generate a **charm**.
-    A rock is a special kind of OCI-compliant container image, while a
-    charm is a software operator for cloud operations that use the Juju
-    orchestration engine. The result is Django apps that
-    can be deployed, configured, scaled, integrated, and so on,
-    on any Kubernetes cluster.
+If you're new to the charming world, Django apps are
+specifically supported with a template to quickly generate a
+**rock** and a matching template to generate a **charm**.
```
lengau, line 23: Since we're adjusting this paragraph, should we make `rock` a link to the equivalent rockcraft tutorial?

lengau (approved):

docs/tutorial/kubernetes-charm-django.rst
medubelko: In the past I've advised including links so early in the tutorial. Since the user will be splitting their attention between this page and the terminal, I think we should strive not to distract them with alternate documents unless essential. I think the next paragraph is sufficient context for the introduction.

docs/tutorial/kubernetes-charm-django.rst
```diff
@@ -18,15 +18,14 @@ running with Juju. Let's get started!
...
+A rock is a special kind of OCI-compliant container image, while a
+charm is a software operator for cloud operations that use the Juju
+orchestration engine. The result is Django apps that
```
medubelko, line 26: Maybe the singular would lead the reader's thoughts toward the activity they're about to do?

```suggestion
orchestration engine. The combined result is a Django app that
```

docs/tutorial/kubernetes-charm-django.rst
```diff
@@ -211,15 +210,14 @@ The top of the file should look similar to the following snippet:
 
 Verify that the ``name`` is ``django-hello-world``.
 
-Ensure that ``platforms`` includes the architecture of your host. Check
+The ``platforms`` must match the architecture of your host. Check
```
medubelko, line 213:
```suggestion
The ``platforms`` key must match the architecture of your host. Check
```

docs/tutorial/kubernetes-charm-django.rst
```diff
@@ -211,15 +210,14 @@ The top of the file should look similar to the following snippet:
...
 the architecture of your system:
 
 .. code-block:: bash
 
     dpkg --print-architecture
 
-If your host uses ARM architecture, open ``rockcraft.yaml`` in a
-text editor, comment out ``amd64``, and include ``arm64`` under ``platforms``.
+Edit the ``platforms`` entry in ``rockcraft.yaml`` if required.
```
medubelko, line 220: > entry

"key" would be more consistent and accurate, but I'm not too fussed.

docs/tutorial/kubernetes-charm-django.rst
```diff
@@ -406,8 +404,7 @@ the Django app. Let's create a new model:
     :end-before: [docs:add-juju-model-end]
     :dedent: 2
 
-Include a constraint to the Juju model to specify your architecture if you
-aren't on a host with AMD64:
+Include a constraint to the Juju model to specify your architecture:
```
medubelko, line 407: An alternative phrasing could be:

```suggestion
Constrain the Juju model to your architecture:
```

medubelko (approved):
Looks great. Just leaving a few thoughts about the wording.

docs/tutorial/kubernetes-charm-django.rst
erinecon: I agree with @medubelko about not including a link here. Thanks all for your feedback!

docs/tutorial/kubernetes-charm-django.rst
erinecon: Sounds good to me -- will make the change everywhere

docs/tutorial/kubernetes-charm-django.rst
erinecon: I fine with being consistent and using "key" here -- will make the change

PR #2263, opened 2025-04-23
docs: 12-Factor tutorial updates

erinecon:
Five updates proposed by @evildmp in https://github.com/canonical/charmcraft/issues/2235 that affect all four of the 12-factor tutorials and the reusable setup scripts.

### Summary of changes
* Updated the Multipass install link to point to the correct how-to guide.
* Removed explicit mentions of "If you aren't on a host with AMD64 architecture" and the note admonitions around the rock and charm names. This is to help the user build up the muscle memory of the naming and take responsibility for the file names.
* Added a brief description of the ``--profile`` flag in the rock and charm creation sections.
* (Django tutorial only) Reduced the text in the "We'll need to update the settings.py file..." paragraph, again to help the user build muscle memory of where the file is located.
* (Django tutorial only) Rephrased the text about connecting the Django app to PostgreSQL, as that's not what's actually going on at that time. Instead we're adding the PostgreSQL relation endpoint to the charm project file, so the text now reflects that change.

In addition to these changes, a warning admonition block about older versions of Charmcraft was removed in the Django tutorial.


docs/tutorial/kubernetes-charm-django.rst
```diff
@@ -300,14 +301,10 @@ finish.
 
 Once Rockcraft has finished packing the Django rock, the
 terminal will respond with something similar to
-``Packed django-hello-world_0.1_<architecture>.rock``. After the initial
+``Packed django-hello-world_0.1_<architecture>.rock``. The name of the
+``.rock`` file will reflect your system's architecture. After the initial
```
medubelko, line 305:
```suggestion
``Packed django-hello-world_0.1_<architecture>.rock``. The file name
reflects your system's architecture. After the initial
```

docs/tutorial/kubernetes-charm-django.rst
```diff
@@ -352,7 +350,8 @@ Initialize a charm named ``django-hello-world``:
 
 The files will automatically be created in your working directory.
 
-We will need to connect the Django app to the PostgreSQL database.
+We will need to integrate our Django app to the PostgreSQL database,
+which means we need to declare a requirement in the charm project file.
```
medubelko, line 354:
```suggestion
which means we must declare a requirement in the charm project file.
```

One less "need to".

docs/tutorial/kubernetes-charm-django.rst
```diff
@@ -165,9 +165,11 @@ original terminal of the Multipass VM using :kbd:`Ctrl` + :kbd:`C`.
 Pack the Django app into a rock
 -------------------------------
 
-First, we'll need a ``rockcraft.yaml`` file. Using the
-``django-framework`` profile, Rockcraft will automate the creation of
-``rockcraft.yaml`` and tailor the file for a Django app. Change
+First, we'll need a ``rockcraft.yaml`` file. We'll take advantage of a
```
medubelko, line 168:
```suggestion
First, we'll need a ``rockcraft.yaml`` project file. We'll take advantage of a
```

This should prime the reader that `rockcraft.yaml` = project file.

medubelko (approved):
Left some hopefully helpful tidbits that apply to all four tutorials.

lengau (approved):

PR #2266, opened 2025-04-24
build(build-deps): install cargo if not installed

lengau:
This installs cargo if it's not installed already. This is needed for CI runners that aren't AMD64 or ARM64 for cryptography.


sergiusens:
You also need pkg-config, should these not require be conditionally installed per architecture? Would we be spending to many resources rebuilding these for our unit tests? Can we not go the python apt route instead?

PR #2278, opened 2025-05-01
docs: add cryptographic docs

lengau:
CHARMCRAFT-616

### DIFF
--- a/docs/explanation/cryptography.rst
+++ b/docs/explanation/cryptography.rst
@@ -7,7 +7,7 @@ Charmcraft uses cryptographic technologies to fetch arbitrary files over the int
 communicate with local processes, and store credentials. It does not directly implement
 its own cryptography, but it does depend on external libraries to do so.
 
-Charmcraft is built upon `Craft Application`_ and derives much of its functionality from
+Charmcraft is built on `Craft Application`_ and derives much of its functionality from
 it, so much of Charmcraft's cryptographic functionality is described in `Cryptographic
 technology in Craft Application`_. The additional cryptography functionality in
 Charmcraft is documented below.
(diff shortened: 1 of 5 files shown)


bepri (approved):

docs/explanation/cryptography.rst
```diff
@@ -0,0 +1,34 @@
+.. _explanation_cryptographic-technology:
+
+Cryptographic technology in Charmcraft
+======================================
+
+Charmcraft uses cryptographic technologies to fetch arbitrary files over the internet,
+communicate with local processes, and store credentials. It does not directly implement
+its own cryptography, but it does depend on external libraries to do so.
+
+Charmcraft is built upon `Craft Application`_ and derives much of its functionality from
```
medubelko, line 10:
```suggestion
Charmcraft is built on `Craft Application`_ and derives much of its functionality from
```

I wish I hadn't let this word propagate...

docs/explanation/cryptography.rst
```diff
@@ -0,0 +1,34 @@
...
+it, so much of Charmcraft's cryptographic functionality is described in `Cryptographic
+technology in Craft Application`_. The additional cryptography functionality in
+Charmcraft is documented below.
+
+Local Docker images
+~~~~~~~~~~~~~~~~~~~
+
+Charmcraft uses the `Docker SDK for Python`_ to communicate with a local Docker instance
+(if one is installed or running). This communicates with Docker over a Unix-domain
+socket if the user running Charmcraft has the necessary permission.
```
medubelko, line 20:
```suggestion
If a Docker instance is installed or running on the host, Charmcraft uses the `Docker
SDK for Python`_ to communicate with it. The SDK communicates with Docker over a
Unix-domain socket if the user running Charmcraft has the necessary permission.
```

medubelko, line 20: > installed or running on the host

I would rather we explain what the container could be doing and when Charmcraft would need to communicate with it.

docs/explanation/cryptography.rst
lengau: "We built this city upon rock and roll." -- Starcraft

docs/explanation/cryptography.rst
medubelko: Sublime!

medubelko (approved):

Issue #2259, opened 2025-04-23
Builds fail for 20.04 (in some common circumstances)


dimaqq:
### Bug Description

Due to https://github.com/pypa/setuptools-scm/issues/1131
If requirements.txt is used (and, I think, not the uv plugin, but YMMV)
If any non-trivial package is included (e.g. opentelemetry),
Because charmcraft builds everything from source,
The build-time dependency "install" thing ends up in a loop and fails.

Both Python 3.8 (EOL) and 3.9 (supported) are affected, let's wait for upstream.

Dep chain:
- opentelemetry-api
- hatchling (build dep)
- pluggy
- setuptools_scm

### To Reproduce

build on 20.04
add e.g. `opentelemetry-api` into `requirements.txt`

### Environment

default settings

### charmcraft.yaml

```yaml
type: charm
bases:
  - build-on:
      - name: "ubuntu"
        channel: "20.04"
    run-on:
      - name: "ubuntu"
        channel: "20.04"
parts:
  charm:
    build-packages:
      - git
```

### Relevant log output

```shell
# Manual MRE

                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                              14:01:49
(code) ⋊> dima@bb ⋊> /code pip install --no-binary=:all: setuptools_scm                                                                                                                                                                                                                                                                                                                                                                                                                                                                                    14:01:49
Collecting setuptools_scm
  Using cached setuptools_scm-8.3.0.tar.gz (78 kB)
  Installing build dependencies ... error
  error: subprocess-exited-with-error

  × pip subprocess to install build dependencies did not run successfully.
  │ exit code: 1
  ╰─> [92 lines of output]
      Collecting setuptools>=61
        Using cached setuptools-79.0.0.tar.gz (1.4 MB)
        Getting requirements to build wheel: started
        Getting requirements to build wheel: finished with status 'done'
        Preparing metadata (pyproject.toml): started
        Preparing metadata (pyproject.toml): finished with status 'done'
      Collecting tomli<=2.0.2
        Using cached tomli-2.0.2.tar.gz (16 kB)
        Installing build dependencies: started
        Installing build dependencies: finished with status 'done'
        Getting requirements to build wheel: started
        Getting requirements to build wheel: finished with status 'done'
        Preparing metadata (pyproject.toml): started
        Preparing metadata (pyproject.toml): finished with status 'done'
      Collecting importlib-metadata>=4.6
[...]

PR #2279, opened 2025-05-01
fix: bugfix updates and release for 3.4.6

lengau:
Prerequisite: https://github.com/canonical/charmcraft/issues/2259


mr-cal (approved):

jahn-junior:
This might be out of scope, but I still get the warnings for the pre-8.9 confdb schema command references not being included in the ToC. Do we want to include a fix for that as well?

docs/release-notes/charmcraft-3.4.rst
```diff
@@ -165,13 +165,20 @@ Charmcraft 3.4.3
...
 ~~~~~~~~~~~~~~~~
 
 - Snap dependencies were updated to resolve `CVE-2025-43859
-  <https://www.cve.org/CVERecord?id=CVE-2025-43859>`_
+  <https://www.cve.org/CVERecord?id=CVE-2025-43859>`_.
+
+Charmcraft 3.4.6
+~~~~~~~~~~~~~~~~
+
+- The uv plugin breaks with uv 0.7.
```
medubelko, line 179: Sorry to retread this, but as-is it reads like a known issue. Is there no link to the ticket?

lengau: There's no issue filed in charmcraft and no related craft-parts issue because @bepri fixed it before anyone could file one :-)

medubelko: Can we change it to "was breaking" so it doesn't sound like an active issue? We didn't account for a format for undocumented issues, so it's best if the wording makes sense at the time of reading.

lengau:
```suggestion
- The uv plugin was breaking with uv 0.7.
```

medubelko (approved):

PR #2282, opened 2025-05-06
docs: Updates to the 12-factor tutorials

erinecon:
Three updates proposed by @evildmp (two updates in https://github.com/canonical/charmcraft/issues/2235 plus one other) that affect all four tutorials and their `task.yaml` files.

### Summary of changes
* Added an admonition for `rockcraft pack` to provide more context to the user. 
* Provided more information for the `rockcraft.skopeo copy` command to give a better sense of what each of the options and tags are doing. Updated the `task.yaml` files so that the command appears over more lines, and changed the `seealso` admonition to a `note` block to provide a brief description of each part while also including the external link.
* Updated the description of the MicroK8s registry to describe the registry as a temporary Dockerhub.


docs/tutorial/kubernetes-charm-django.rst
```diff
@@ -304,25 +304,39 @@ Now let's pack the rock:
 Depending on your system and network, this step can take several minutes to
 finish.
 
+.. admonition:: For more options when packing rocks
+
+    See the :external+rockcraft:ref:`ref_commands_pack` command reference.
```
medubelko, line 309: :tada:

docs/tutorial/kubernetes-charm-django.rst
```diff
@@ -304,25 +304,39 @@ Now let's pack the rock:
...
+deployed in the Kubernetes cluster. Copy the rock:
 
 .. literalinclude:: code/django/task.yaml
     :language: bash
     :start-after: [docs:skopeo-copy]
     :end-before: [docs:skopeo-copy-end]
     :dedent: 2
 
-.. seealso::
+.. note::
```
medubelko, line 327: You probably know what I'm going to ask. :)

I do think, keeping in mind that a tutorial is a lesson, that the breakdown here is a worthwhile thing to have the reader think about, and making it look optional might do disservice to that chance at learning.

erinecon: Thanks for the classic comment :)

I am not opposed to moving the breakdown into the text. I liked including it in the `note` block because I could get away with placing each explanation on an individual line, and I like that because I think the mixture of code formatted text and regular text is easier to parse.

I suppose that I could still do that in text if I use a bulleted list? I'll push something and see whether I like it. Please let me know if you have any thoughts.

medubelko: I think it's better!

I'd like to float this possibility by you:

image

No need to make any further changes, but I think this is the kind of flow we're being asked to strive for.

<details>
<summary><strong>Code</strong></summary>
<pre>
That was a long skopeo command, so here's a short breakdown of what it did. The
``--insecure-policy`` argument lets us proceed without a dedicated policy file. The
``--dest-tls-verify=false`` argument bypasses the need for HTTPS and certificate
verification while interacting with the MicroK8s registry. Most importantly,
``oci-archive:`` points to the rock we created for our Django app. And lastly,
``docker:`` specifies the name of the image in the MicroK8s registry.
</pre>
</details>

lengau (approved):

PR #2289, opened 2025-05-15
chore: add charm-tech as template owners

lengau:


mr-cal (approved):
Approving with the assumption that it's OK to give charm-tech write access to charmcraft.

PR #2242, opened 2025-03-31
feat(templates): add workload module to Kubernetes and machine profiles

dwilding:
This PR update the `kubernetes` and `machine` profiles to include a standalone module for workload-specific logic. Other profiles are not changed in any way.

In addition to the new workload module, the updated profiles have a bit more structure in `charm.py` to illustrate the basic flow of logic through the charm & workload module, and improved unit tests in `test_charm.py`.

**This PR includes**
- New/updated template files for the `kubernetes` and `machine` profiles.
- Changes to `charmcraft/application/commands/init.py`, to determine the name of the workload module from the charm name (using `workload.py` if the charm has a generic name such as `charm` or `k8s-operator`).
- Changes to `tests/integration/commands/test_init.py`, to test the updated functionality (with tests passing).
- Updated help and output of `charmcraft init`, to make users aware of the workload module.

Hopefully it will be easy to follow my commits in sequence. I tried to make the changes in a logical step-by-step order.

**TODO**
- Check whether any documentation needs updating, and include relevant changes in this PR.

PR #2295, opened 2025-05-21
feat(templates)!: replace simple profile by kubernetes

dwilding:
This PR makes **breaking changes** to the `charmcraft init` profiles:

- Removes the `simple` profile (a deployable Kubernetes charm for httpbin)
- Makes `kubernetes` the default profile

These changes also affect [Manage charms](https://canonical-charmcraft--2295.com.readthedocs.build/en/2295/howto/manage-charms/) in the docs: `simple` is no longer mentioned, and we now say that `kubernetes` is the default profile.

The httpbin charm still has value as a demo, rather than as a minimal starting point, so I'm adding it to our collection of example charms in the Ops repo. See https://github.com/canonical/operator/pull/1743.

---

Follow-up actions after the profile changes are released: https://github.com/canonical/operator/issues/1742


dwilding:
@lengau this PR is for the profile changes in Charmcraft 4.0 that we discussed in Frankfurt. @medubelko this also includes a doc change. Please let me know if I've missed anything - thanks!

FYI @tmihoc on this upcoming change. I had a look in the Juju docs, but I couldn't find anything that would be affected.

In Frankfurt, we also discussed changing the `kubernetes` and `machine` profiles to use the uv plugin. I'll do that in a separate PR.

tmihoc:
Thanks for the ping, @dwilding ! Out of curiosity: Why the change? Is it just that `simple` was a bit opaque and we now do the job it was trying to do in the `kubernetes` profile already, or...?

dwilding:
> Thanks for the ping, @dwilding ! Out of curiosity: Why the change? Is it just that `simple` was a bit opaque and we now do the job it was trying to do in the `kubernetes` profile already, or...?

That's it :slightly_smiling_face:

In #2242 we added a better `kubernetes` profile that has the basic scaffolding you'd need (which is also releasing in Charmcraft 4.0). `simple` is a more complete demo, so less suitable as a general starting point I feel.

medubelko (approved):
Doc changes LGTM.

lengau (approved):

PR #2298, opened 2025-05-23
chore:  minor dev environment improvements

lengau:
- Remove no-longer-used requirements files
- Replace Oracular with Questing for python-apt
- Update uv.lock
- Swap the order of brew installs on macos to prevent system conflicts


mr-cal (approved):

pyproject.toml
```diff
@@ -91,13 +91,12 @@ dev-noble = [
     # 2.7 for Noble
     "python-apt~=2.7.0;sys_platform=='linux'",
 ]
-dev-oracular = [
-    # 2.9 for Oracular+
-    "python-apt>=2.9.0,<2.9.9;sys_platform=='linux'",
-]
 dev-plucky = [
     # 2.9 for Oracular+
```
bepri, line 95: Should we just remove these version number comments? I see them drift from accuracy somewhat often and they don't describe much.

bepri (approved):

PR #2305, opened 2025-06-03
feat!(bundle): remove bundle-specific store commands

lengau:
This removes bundle-specific store commands as bundles are deprecated.

Anyone who needs bundle-specific commands will be able to get them with Charmcraft 3.


lengau:
Superseded by #2320

Issue #1858, opened 2024-08-27
Remove bundle commands


lengau:
### What needs to get done

Remove the commands that are specific to bundles.

### Why it needs to get done

Bundle registration is discontinued as of 2024-11-01

https://discourse.charmhub.io/t/discontinuing-new-charmhub-bundle-registrations/15344

PR #2307, opened 2025-06-03
feat!: drop Windows support

### LINKED ISSUE #1810: Remove Windows support

lengau:
This removes everything that's been added for specific Windows support.

Fixes #1810 

CRAFT-3237

### DIFF
--- a/tests/unit/commands/test_lifecycle.py
+++ b/tests/unit/commands/test_lifecycle.py
@@ -70,26 +70,6 @@ def pack(service_factory: services.ServiceFactory) -> lifecycle.PackCommand:
 
 
 @pytest.mark.parametrize(
-    ("platform", "expected"),
-    [
-        ("linux", False),
-        ("macos", True),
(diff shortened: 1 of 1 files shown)


charmcraft/application/commands/lifecycle.py
```diff
@@ -161,26 +160,6 @@ def _validate_bases_indices(self, bases_indices):
             if bases_index >= len_configured_bases:
                 raise CraftError(msg.format(bases_index))
 
-    def run_managed(self, parsed_args: argparse.Namespace) -> bool:
```
lengau, line 164: This technically changes things on Linux, but it's irrelevant because we're going to be eliminating bundles in #1858

PR #2309, opened 2025-06-04
fix(templates): fix layer in K8s template unit test

dwilding:
This PR fixes an issues with the unit tests of the `kubernetes` profile charm. The Pebble layer didn't explicitly specify some properties, which caused inconsistency errors from `ops.testing`.

### DIFF
--- a/charmcraft/templates/init-kubernetes/tests/unit/test_charm.py.j2
+++ b/charmcraft/templates/init-kubernetes/tests/unit/test_charm.py.j2
@@ -10,7 +10,15 @@ from charm import SERVICE_NAME, {{ class_name }}
 
 CHECK_NAME = "service-ready"  # Name of Pebble check in the mock workload container.
 
-base_layer: pebble.LayerDict = {"services": {SERVICE_NAME: {}}, "checks": {CHECK_NAME: {}}}
+base_layer: pebble.LayerDict = {
+    "services": {SERVICE_NAME: {}},
+    "checks": {
+        CHECK_NAME: {
+            "startup": "enabled",
+            "level": "ready",
+        }
+    },
+}
 layers = {"base": pebble.Layer(base_layer)}
 
 


dwilding:
@tonyandrewmeyer would you mind reviewing this? Do you think we ought to explicitly specify additional parts of the Pebble layer, to avoid issues in the future?

dwilding:
Using the latest pre-release version of ops-scenario (from [main](https://github.com/canonical/operator/tree/main/testing), I also get this error:

```
scenario.errors.InconsistentScenarioError: Inconsistent scenario. The following errors were found: container 'some-container' has a check 'service-ready' with a different 'threshold' (3) than the plan (None).
```

I've just pushed another commit to include this.

In a way, it seems strange to have to specify `threshold` and `startup`, as these properties have default values in Pebble.

lengau (approved):
Thanks!

charmcraft/templates/init-kubernetes/tests/unit/test_charm.py.j2
```diff
@@ -10,7 +10,16 @@ from charm import SERVICE_NAME, {{ class_name }}
 
 CHECK_NAME = "service-ready"  # Name of Pebble check in the mock workload container.
 
-base_layer: pebble.LayerDict = {"services": {SERVICE_NAME: {}}, "checks": {CHECK_NAME: {}}}
+base_layer: pebble.LayerDict = {
+    "services": {SERVICE_NAME: {}},
+    "checks": {
+        CHECK_NAME: {
+            "startup": "enabled",
```
tonyandrewmeyer, line 17: Similarly here, `override` is required so should be included I think. There ought to be one of the checks (http/tcp/exec) as well, although that gets to the same issue as `command` with services, that you have to define something and there's probably not an obvious choice for a template.

tonyandrewmeyer (commented):
I think it's a good improvement to silence the warnings by having a compliant layer. Since we're doing part of that, it seems like it would be better to go the whole way and add at least `override` (since `replace` is an easy choice for a value), and maybe `command` and `http`/`exec`/`tcp` as well.

Overall, it feels a bit off that the test defines a layer and uses it, rather than using the layer from the charm. In a real charm, I'd much rather re-use the layer from src/charm.py (or from rockcraft.yaml) than re-define it in the tests.

Related to that, it's a little odd that the tests have a layer with a check but the layer in the charm doesn't. I get that it's so that there's nice generic code when collecting the status that will 'just work' when a check does get added, and that this is the test so that makes sense.

I think what I'd rather see is either src/charm.py gaining a placeholder check in its layer and the tests pull the layer from src/charm.py, *or* the layer in src/charm.py stays the same, but the tests pull the layer from there, and then add in a check (either as a separate layer or editing the imported one) for the test that's about checks.

I think that change could be in a follow-up PR, though, and this one can just be adding in the required fields to make the layers match the spec.

tonyandrewmeyer:
> In a way, it seems strange to have to specify `threshold` and `startup`, as these properties have default values in Pebble.

They have default values in Pebble *layers*, but they're always included in the Pebble *response* when you ask for check info, and they could be anything at that point.

What we currently recommend (this is [buried in the README](https://github.com/canonical/operator/tree/main/testing#pebble-checks) at the moment, but will soon be in the how-to) is something like this:

```python
ctx = testing.Context(MyCharm, meta={"name": "foo", "containers": {"my_container": {}}})
layer = pebble.Layer({
    "checks": {"http-check": {"override": "replace", "startup": "enabled", "failures": 7}},
})
check_info = testing.CheckInfo(
    "http-check",
    status=pebble.CheckStatus.UP,
    level=layer.checks["http-check"].level,
    startup=layer.checks["http-check"].startup,
    threshold=layer.checks["http-check"].threshold,
)
container = testing.Container("my_container", check_infos={check_info}, layers={"layer1": layer})
state = testing.State(containers={container})
ctx.run(ctx.on.pebble_check_failed(info=check_info, container=container), state=state)
```

The layer doesn't have to have `startup` and `failures` explicitly set. The `CheckInfo` loads the `level`, `startup`, and `threshold` values from the layer, because that's essentially what Pebble does: it copies the values into the check info and they'll always be the same as in the current plan, whereas the other fields (`status`, `failures`) will change based on the state of the check.

> scenario.errors.InconsistentScenarioError: Inconsistent scenario. The following errors were found: container 'some-container' has a check 'service-ready' with a different 'threshold' (3) than the plan (None).

Yeah, this isn't ideal. `LayerDict` and `CheckDict` are just dictionaries, so can't have default values, even though there are ones in Pebble (fields are marked as `Optional` in most of those cases). `Layer` and `Check` are actual classes so can do some defaults, but don't provide them in most cases. I'm not sure whether that's fixable in a backwards-compatible way, although I've wondered if we should have an `ops.Layer` that is a dataclass and fixes issues like this and would be accepted (and converted to a `ops.pebble.Layer`) in `Container` methods that want a `ops.pebble.Layer`).

In any case, these are ops problems, so we should continue discussion (if needed) there rather than in charmcraft :)

dwilding:
Thank you for your detailed replies! I'm working through changes now - adding extra properties to the layer like you suggested. I agree that we should do a follow-up change to get the layer from the charm, with a check defined in the charm too. (FWIW, when I was revising the template after a couple of months, it did seem too magical that no check was actually defined in the charm. I think it would be more helpful to include a placeholder check.)

Re this approach:

> ```python
> ctx = testing.Context(MyCharm, meta={"name": "foo", "containers": {"my_container": {}}})
> layer = pebble.Layer({
>     "checks": {"http-check": {"override": "replace", "startup": "enabled", "failures": 7}},
> })
> check_info = testing.CheckInfo(
>     "http-check",
>     status=pebble.CheckStatus.UP,
>     level=layer.checks["http-check"].level,
>     startup=layer.checks["http-check"].startup,
>     threshold=layer.checks["http-check"].threshold,
> )
> container = testing.Container("my_container", check_infos={check_info}, layers={"layer1": layer})
> state = testing.State(containers={container})
> ctx.run(ctx.on.pebble_check_failed(info=check_info, container=container), state=state)
> ```

This is giving me a type checking error:

```
error: Argument of type "CheckLevel | str" cannot be assigned to parameter "level" of type "CheckLevel | None" in function "__init__"
```

Which I can resolve with:

```python
level=pebble.CheckLevel(layer.checks["http-check"].level)
```

Is there a better way to resolve the error?

tonyandrewmeyer (approved):

PR #2311, opened 2025-06-05
docs(templates): list all 12-factor profiles in init help

dwilding:
The output of `charmcraft help init` only shows one 12-factor profile (`flask-framework`). This PR adds the other ones.

Note that I'm separately proposing to remove `simple`, in #2295


medubelko (approved):

lengau (approved):
Thanks David! Would you mind re-targeting this at the `hotfix/3.5` branch?

tonyandrewmeyer (approved):
Do you think that the descriptions for 12-factor should say that they're for K8s, or is that something people would already know by the time they're going to use one of these profiles? "A basic Kubernetes charm for a 12-factor {...} app." for example.

Not related to this PR, but related to the above, shouldn't the template charmcraft.yaml have an `assumes` for K8s?

erinecon:
Hi @tonyandrewmeyer and @medubelko ,

I think it's worth adding "Kubernetes" to the descriptions for the 12-factor extensions. I think this because we've received questions in past UX sessions about whether the extensions could be used for machine charms (they can't AFAIK), and because adding "Kubernetes" will make these profile descriptions consistent with the other descriptions.

dwilding:
Sounds good to me too, thanks! I've just pushed a commit that adds "Kubernetes" to the descriptions.

medubelko:
Thanks, @dwilding and @erinecon!

@lengau This should be good to merge.

Issue #2238, opened 2025-03-27
uv how-to should include how to install uv in the build environment


lengau:
### Request type

Fix

### What needs to get done

The how-to doesn't mention how to install uv. We should either how to add the uv snap or how to manually install uv in another part.

### Documentation location

https://canonical-charmcraft.readthedocs-hosted.com/en/latest/howto/migrate-plugins/charm-to-uv/

dwilding:
I just ran into this and was about to raise an issue!

@lengau, would your recommendation be to do it this way?

```yaml
parts:
  charm-project:
    plugin: uv
    source: .
    build-snaps:
      - astral-uv
```

It looks like both of these docs would need updating:

- https://canonical-charmcraft.readthedocs-hosted.com/en/latest/howto/migrate-plugins/charm-to-uv/
- https://canonical-charmcraft.readthedocs-hosted.com/en/latest/reference/plugins/uv_plugin/#example

Also, I found that `uv.lock` has to be present. That's implied by the second doc, but I suppose it would be good to say it explicitly.

FYI @medubelko. Would it be helpful if I contribute some doc updates?

lengau:
@dwilding that's the way I've been doing it and I think it's a resounding "probably." The interesting catch is that we (Canonical) don't have an "official" way to provide uv. The uv snap is probably closer than the official builds since it runs on all of our supported architectures and I publish it (albeit in my personal time).

medubelko:
@dwilding 

> Would it be helpful if I contribute some doc updates?

Always!

@lengau My vote is for the snap.

lengau:
Makes sense to me! We can always update it if/when we have a more official method.

PR #2312, opened 2025-06-05
feat(templates): use uv for testing and packing kubernetes and machine charms

dwilding:
This PR changes the `kubernetes` and `machine` profiles of `charmcraft init` to use the tox-uv plugin (for testing) and the uv Charmcraft plugin (for packing).

[Here's a minimal Kubernetes charm generated from the updated profiles](https://github.com/dwilding/vm/tree/main/demo-api-charm)

### Profile changes

- Dependencies are now listed in *pyproject.toml*. I've removed *requirements.txt*.

- The environments in *tox.ini* now require uv. The user should install tox with `uv tool install tox --with tox-uv`.

- The scope of `tox -e static` now includes *tests/unit/* and *tests/integration/* for the charms. Strictly, this isn't part of switching to uv.

- *uv.lock* is provided as part of the profiles. (We went back and forth on this in discussions.) But the file is **not** generated using `uv lock`. The Charm Tech team will periodically refresh the *uv.lock.j2* templates to keep up with major Ops features or dependency issues.

- *charmcraft.yaml* now specifies the uv plugin. After discussion (in the comments of this PR), I felt it's best not to have a dedicated dependency group for PYDEPS from charmlibs.

- Misc cleanup: I updated some old links to Juju docs and adjusted the formatting of *charmcraft.yaml* to avoid double blank lines.

### Help & doc changes

- Modified the help output of `charmcraft init` to:
    - Explain that *pyproject.toml* specifies dependencies if the profile is `kubernetes` or `machine`
    - Explain that *requirements.txt* is created for 12-factor profiles only
    - List *uv.lock* and say that this file is created for `kubernetes` and `machine`

- Updated the following docs (links to previews):

    - [pyproject.toml file](https://canonical-charmcraft--2312.com.readthedocs.build/2312/reference/files/pyproject-toml-file/)
    - [requirements.txt file](https://canonical-charmcraft--2312.com.readthedocs.build/2312/reference/files/requirements-txt-file/)
    - [tox.ini file](https://canonical-charmcraft--2312.com.readthedocs.build/2312/reference/files/tox-ini-file/)
    - [uv.lock file](https://canonical-charmcraft--2312.com.readthedocs.build/2312/reference/files/uv-lock-file/) (new)

- Removed [requirements-dev.txt file](https://canonical-charmcraft.readthedocs-hosted.com/stable/reference/files/requirements-dev-txt-file/) because it doesn't seem to be referenced anywhere and I can't find any Charmcraft code that creates/uses this file. I've redirected the page to [requirements.txt file](https://canonical-charmcraft--2312.com.readthedocs.build/2312/reference/files/requirements-txt-file/).

---

Follow-up actions:

- Expand docs about the uv plugin - see https://github.com/canonical/charmcraft/issues/2238#issuecomment-2900521717
- After the profile changes are released, https://github.com/canonical/operator/issues/1742


charmcraft/application/commands/init.py
```diff
@@ -34,15 +34,14 @@
...
-    "simple": "init-simple",
     "kubernetes": "init-kubernetes",
     "machine": "init-machine",
     "flask-framework": "init-flask-framework",
     "django-framework": "init-django-framework",
     "go-framework": "init-go-framework",
     "fastapi-framework": "init-fastapi-framework",
 }
-DEFAULT_PROFILE = "simple"
+DEFAULT_PROFILE = "kubernetes"
```
dwilding, line 44: These changes are unrelated to switching to uv. I've included them from https://github.com/canonical/charmcraft/pull/2295 to help with testing `charmcraft init`.

tests/integration/commands/test_init.py
```diff
@@ -232,12 +230,16 @@ def test_gecos_user_has_no_name(monkeypatch, new_path, init_command):
...
+def test_create_directory(new_path, init_command, subdir, base_expected_files):
     init_dir = new_path / subdir
 
     try:
-        init_command.run(create_namespace(project_dir=pathlib.Path(subdir)))
+        params = create_namespace(name="foo-bar-k8s", project_dir=pathlib.Path(subdir))
+        init_command.run(params)
+
+        workload_module_path = pathlib.Path("src/foo_bar.py")
+        expected_files = base_expected_files.union({workload_module_path})
```
dwilding, line 242: These changes are unrelated to switching to uv. I've included them from https://github.com/canonical/charmcraft/pull/2295 to help with testing `charmcraft init`.

charmcraft/application/commands/init.py
```diff
@@ -52,10 +51,6 @@
 charm operator package. By default it will work in the current directory.
 
 Available profiles are:
-    simple:
-        A basic kubernetes charm with lot of texts helping the developer
-        to navigate their first charm by following the instructions.
-
```
dwilding, line 58: This change is unrelated to switching to uv. I've included it from https://github.com/canonical/charmcraft/pull/2295 to help with testing `charmcraft init`.

charmcraft/templates/init-kubernetes/tests/integration/test_charm.py.j2
```diff
@@ -29,6 +29,7 @@ async def test_build_and_deploy(ops_test: OpsTest):
     }
 
     # Deploy the charm and wait for active/idle status
+    assert ops_test.model is not None
```
dwilding, line 32: This change is unrelated to switching to uv. It makes sure that the integration test code passes type checking. See https://github.com/charmed-kubernetes/pytest-operator/issues/143

charmcraft/templates/init-kubernetes/tests/unit/test_charm.py.j2
```diff
@@ -10,8 +10,28 @@ from charm import SERVICE_NAME, {{ class_name }}
 
 CHECK_NAME = "service-ready"  # Name of Pebble check in the mock workload container.
 
-base_layer: pebble.LayerDict = {"services": {SERVICE_NAME: {}}, "checks": {CHECK_NAME: {}}}
```
dwilding, line 13: The changes in this file are unrelated to switching to uv. I've included them from https://github.com/canonical/charmcraft/pull/2309 to help with testing the charms.

charmcraft/templates/init-kubernetes/pyproject.toml.j2
```diff
@@ -1,3 +1,37 @@
+[project]
+name = "{{ name }}"
+version = "0.0.1"
+requires-python = ">=3.10"
+
+# Dependencies of the charm code in src/
+dependencies = [
+    "ops >=2.22,<3",
```
benhoyt, line 8: Given 3 is coming out soon, shall we do `<4`?

charmcraft/templates/init-kubernetes/pyproject.toml.j2
```diff
@@ -1,3 +1,37 @@
...
+# Dependencies of code in tests/integration/
+integration = [
+    "pytest",
+    "pytest-operator",
+]
+# Dependencies needed to actually run the tests
+run-static = [
+    "pyright",
+]
+run-unit = [
```
benhoyt, line 28: I don't understand why we need these to be separate -- what about just a single `unit` and `integration` (and `static`) group? Is it important or useful that they're separate?

charmcraft/templates/init-kubernetes/tox.ini.j2
```diff
@@ -44,13 +44,23 @@ commands =
     ruff check {[vars]all_path}
     ruff format --check --diff {[vars]all_path}
 
+[testenv:static]
+description = Run static type checks
+runner = uv-venv-lock-runner
+dependency_groups =
+    unit
+    integration
+    run-static
```
benhoyt, line 53: Wouldn't we need `charmlib-pydeps` here too (if there are any in that list)?

charmcraft/templates/init-kubernetes/pyproject.toml.j2
```diff
@@ -1,3 +1,37 @@
...
+requires-python = ">=3.10"
+
+# Dependencies of the charm code in src/
+dependencies = [
+    "ops >=2.22,<3",
+]
+
+[dependency-groups]
+# Dependencies of Charmcraft libraries (from their PYDEPS)
+charmlib-pydeps = []
```
benhoyt, line 13: I wonder -- it seems that this could easily be missed (as I think you already do in this PR in `static`). I wonder if it's better to just include them in `dependencies`, maybe with a separator and comment, like:

```
# Dependencies of the charm code in src/
dependencies = [
    "ops >=2.22,<3",

    # Dependencies of Charmcraft libraries below (from their PYDEPS)
]
```

benhoyt (commented):
Nice, I quite like this. I've added a couple of comments and suggestions for simplifications. @dimaqq should take a look too.

charmcraft/templates/init-kubernetes/pyproject.toml.j2
dwilding: Do you mean `>=2.22,<4` or minimum 3.0? In any case, the profile changes would be released in Charmcraft 4, and at that point we may have already released Ops 3, in which case we can set `>=3.x,<4`.

charmcraft/templates/init-kubernetes/tox.ini.j2
dwilding: No, I don't think so. At least, not from my testing. We don't run static checks on libs themselves (and we probably shouldn't, I guess, as they're outside the charmer's control. The dependencies specified in `charmlibs-deps` aren't directly referenced from the charm code or the test code.

charmcraft/templates/init-kubernetes/pyproject.toml.j2
benhoyt: Yeah, I'd prefer the simplicity. Interested to see what Alex says, but if he agrees we can update that doc instead.

Does `uv` tooling keep comments like I've shown above in place correctly?

tonyandrewmeyer: Yes, but `pyproject.toml` gets updated when you `uv add` (or `uv remove`).

It looks like comments are preserved where they are, but an `add` appends so would end up in the pydeps section:

```
$ cat pyproject.toml 
[project]
name = "pp"
version = "0.1.0"
description = "Add your description here"
readme = "README.md"
requires-python = ">=3.13"
dependencies = [
    'ops',
    # These are from pydeps.
    'cosl',
]
$ uv add requests
[...]
$ cat pyproject.toml 
[project]
name = "pp"
version = "0.1.0"
description = "Add your description here"
readme = "README.md"
requires-python = ">=3.13"
dependencies = [
    'ops',
    # These are from pydeps.
    'cosl',
    "requests>=2.32.3",
]
```

Of course, maybe you are `uv add`ing a dependency from a new charmhub charm lib, so that would be correct. But it seems like it would be more likely that you're adding one for the charm itself, particularly as we ween people off charmhub charm libs that have pydeps.

One solution would be to put them the other way around, like:

```toml
dependencies = [
    # These are from pydeps.
    'cosl',
    # These are for the code in src/
    'ops',
]
```

But it feels a bit backwards, and will be odd with that section empty in the template/profile, and you're still going to have to manually shuffle them around when it is for a lib.

This is one of the reasons that I like the separate pydeps group.

benhoyt: I think @tonyandrewmeyer's push-back is fair here: how `uv add` works with the comment approach (or doesn't work very well) shows that a separate group might be a good idea.

HOWEVER, David's right that uv considers dependency groups to be *development* deps. And it's not just uv, the dependency group PEP ([PEP 735](https://peps.python.org/pep-0735/)) explicitly says right at the top:

> This PEP specifies a mechanism for storing package requirements in pyproject.toml files such that they are not included in any built distribution of the project.

And it mentions "development dependencies" too. The [PyPA docs on dependency groups](https://packaging.python.org/en/latest/specifications/dependency-groups/#dependency-groups) that come from that PEP also says "Dependency Groups are suitable for internal development use-cases like linting and testing, as well as for projects which are not built for distribution, like collections of related scripts."

So maybe it's not required by the tooling, but that pretty strongly states the intent of these groups is not for runtime dependencies. Given that, I'd lean towards what David suggests in the final paragraph of his last comment: putting the charm + charmlib deps together in `dependencies` with an explanation comment above the list (and document the approach).

charmcraft/templates/init-kubernetes/pyproject.toml.j2
dwilding: Tony and I just discussed this. We feel the best approach is `"ops <4"`. When the user first runs `uv lock` they'll automatically get the latest-released version of Ops. Then if they really want to use an older version for some reason (e.g., to ensure Python 3.8 support), they can manually override the version specifier.

dwilding:
Thanks for the feedback so far!

I've simplified the dependency groups, to have a single group for each of `static`, `unit`, and `integration`. We'll still need to make a final decision about how to handle charmlib dependencies (see [discussion](https://github.com/canonical/charmcraft/pull/2312#discussion_r2130879565)).

@tonyandrewmeyer and I discussed whether `init` should generate the uv lockfile. We felt that it's best to explicitly make the user aware (through command output and the docs) that we expect them to run `uv lock`. So I've included a message at the end of the command output:

```
Now edit the following package files to provide fundamental charm metadata
and other information:

charmcraft.yaml
src/charm.py
src/workload.py
README.md

Before packing, run `uv lock` to create a lockfile.
```

charmcraft/application/commands/init.py
```diff
@@ -109,6 +104,10 @@ def _make_success_message(src_files: list[str]) -> str:
 {src_files_str}
 README.md
 """
+    if profile in ("kubernetes", "machine"):
+        message = f"""{message}
+Before packing, run `uv lock` to create a lockfile."""
```
dimaqq, line 109: **Idea:** it would be grand if the tool ran the initial `uv.lock` instead.

I've no clue how hard that may be / if we want to rely on `uv` being available at run time, etc.

charmcraft/templates/init-kubernetes/pyproject.toml.j2
dimaqq: The upstream states:

> **potentially** as their own [dependency group](https://docs.astral.sh/uv/concepts/projects/dependencies/#dependency-groups).

So it's not like mandatory... in fact it doesn't seem to imply strongly that dependency groups specifically should be used.

I wonder what the intention was behind that.
Perhaps being able to sync libs' DEPS with some other tooling?

charmcraft/templates/init-kubernetes/pyproject.toml.j2
```diff
@@ -1,3 +1,33 @@
...
+unit = [
+    "coverage[toml]",
+    "ops[testing]",
+    "pytest",
+]
+# Dependencies of integration tests
+integration = [
+    "juju",
+    "pytest",
+    "pytest-operator",
```
dimaqq, line 28: Let's add `jubilant` here as well.

charmcraft/templates/init-kubernetes/pyproject.toml.j2
```diff
@@ -38,4 +68,7 @@ max-complexity = 10
 skip = "build,lib,venv,icon.svg,.tox,.git,.mypy_cache,.ruff_cache,.coverage"
```
dimaqq, line 68: Personally I'd add `htmlcov` to this line too, on par with coverage.
However, that may be a subject of another PR?

charmcraft/templates/init-kubernetes/pyproject.toml.j2
```diff
@@ -1,3 +1,33 @@
...
+[dependency-groups]
+# Dependencies of charmlibs (from their PYDEPS)
+charmlib-pydeps = []
+# Dependencies of static type checks
+static = [
+    "pyright",
+]
+# Dependencies of unit tests
+unit = [
+    "coverage[toml]",
```
dimaqq, line 20: **Discussion:** my personal preference is for `pytest-cov` over plain `coverage`.

That should probably be a separate PR though, shouldn't it?
Uaing `pytest-cov` also changes the commands and settings elsewhere in the template.

charmcraft/templates/init-kubernetes/tests/unit/test_charm.py.j2
```diff
@@ -10,8 +10,28 @@ from charm import SERVICE_NAME, {{ class_name }}
...
+            }
+        },
+        "checks": {
+            CHECK_NAME: {
+                "override": "replace",
+                "level": "ready",
+                "threshold": 3,
+                "startup": "enabled",
+                "http": {
+                    "url": "http://localhost:8000/version",  # The specific URL isn't important.
```
dimaqq, line 29: **minor:** the comment seems odd.

It took me three reads to understand that this is boilerplate config that's ignored by Scenario.

dimaqq (commented):
Looking good!

charmcraft/templates/init-kubernetes/pyproject.toml.j2
dwilding: Captured this in https://github.com/canonical/operator/issues/1818

charmcraft/templates/init-kubernetes/pyproject.toml.j2
dwilding: Captured this in https://github.com/canonical/operator/issues/1818

charmcraft/templates/init-kubernetes/pyproject.toml.j2
dwilding: 100% agree! I don't have a public issue, but planning to do that in a separate PR :slightly_smiling_face:

charmcraft/templates/init-kubernetes/tests/unit/test_charm.py.j2
dwilding: That's helpful to raise, thanks. The comment has already been merged, so I'll address this in a follow-up PR.

dwilding:
This is ready for review - thanks all for the input!

@lengau there are a couple of points that I especially want to see what you think:

- I decided not to have a dependency group for PYDEPS from charmlibs, because these dependencies are runtime dependencies not development dependencies. There's more discussion in the comments.

- I decided to require the user to create *uv.lock* instead of Charmcraft storing it. As Dima mentioned in a comment, a third alternative would be for Charmcraft to generate it at `init` time. But that would introduce a hard requirement on uv being installed.

charmcraft/application/commands/init.py
```diff
@@ -113,6 +115,10 @@ def _make_success_message(src_files: list[str]) -> str:
 {src_files_str}
 README.md
 """
+    if profile in ("kubernetes", "machine"):
+        message = f"""{message}
```
james-garner-canonical, line 119: Should this be `+=`?
```suggestion
        message += f"""{message}
```

charmcraft/application/commands/init.py
james-garner-canonical: I guess requiring the user to have a `uv.lock` and using it is a decent choice from the uv charmcraft plugin, since it pushes charmers towards keeping a lock file in version control. An option for the plugin to generate one when packing would be nice for testing/prototyping/etc., though that'd be out of scope for this PR.

I wonder if `charmcraft init` should include a known good `uv.lock` file when using these templates? (i.e. add such a file to these templates now). Thoughts @canonical/charm-tech?

charmcraft/templates/init-kubernetes/pyproject.toml.j2
```diff
@@ -1,3 +1,33 @@
+[project]
+name = "{{ name }}"
+version = "0.0.1"
+requires-python = ">=3.10"
+
+# Dependencies of the charm code
+# You should include the dependencies of the code in src/. You should also include the
+# dependencies of any charmlibs that the charm uses (copy the dependencies from PYDEPS).
+dependencies = [
+    "ops<4",
```
james-garner-canonical, line 10: It might be worth specifying a minimum version of `ops` here as well as a maximum. One ideal would be to use the minimum version that has the features the template charm uses (or that we suggest using in comments etc). But how about we just use the latest release and bump it when we make changes?
```suggestion
    "ops>=2.20,<4",
```
Wdyt, @canonical/charm-tech?

james-garner-canonical (approved):
LGTM, a couple of thoughts on deps in the templates, but happy to go with what you already have if that's preferred.

I haven't tried the new templates locally, or looked into what the charmcraft tests cover with regard to the templates.

charmcraft/application/commands/init.py
tonyandrewmeyer: I argued (privately with @dwilding) to not include it in the template, because then you're getting the set of dependencies at template-creation time, not at init time, which is what I think people want and would expect. It's busywork to continually update the template, but if we don't, we're going to end up shipping templates with bugs and so forth.

I think init time is interesting, conditional on `uv` being installed (with a warning otherwise) but should be a separate feature/PR.

I think pack time might be interesting but is definitely a seperate feature/PR and if I understand the StarCraft philosophy wouldn't fit - the idea would be a pack override that ran the lock, in the charmcraft file (although that could be in the template as a separate PR).

charmcraft/templates/init-kubernetes/pyproject.toml.j2
james-garner-canonical: I'm happy to go along with no minimum, if that's the general preference. I guess in practice not having a minimum is unlikely to cause trouble, and we can always add one if we extend the templates to use more modern `ops` features.

charmcraft/application/commands/init.py
james-garner-canonical: My mistake, I missed that `message` was included in the new message via f-string. Either way is fine, imo -- the current code is fine.

charmcraft/templates/init-kubernetes/tox.ini.j2
```diff
@@ -44,43 +44,40 @@ commands =
...
+runner = uv-venv-lock-runner
+dependency_groups =
+    unit
+commands =
+    coverage run --source={[vars]src_path} -m pytest \
+        -v \
+        -s \
+        --tb native \
+        {posargs} \
+        {[vars]tests_path}/unit
```
dimaqq, line 68: **minor:** I have a slight preference for `{posargs}` to be the last item on the command line, because that matches the invocation `tox -e some-env -- more args here` better.

```suggestion
        --tb native \
        {[vars]tests_path}/unit \
        {posargs}
```

dimaqq: Stand-up consensus: let's change it so that `{posargs}` are last.

charmcraft/templates/init-kubernetes/pyproject.toml.j2
lengau: I'd like a minimum set, but `>=2.0` or similar is perfectly fine for me. I'd like to eventually (<-- weasel word) have us actually testing these with the minimum requirements so we do know whether the template still works with older versions.

docs/reference/files/requirements-txt-file.rst
```diff
@@ -6,9 +6,13 @@
...
 <https://pip.pypa.io/en/stable/reference/pip_install/#requirements-file-format>`_
-used to declare and pin the version of any Python libraries required by a charm in
-production. This will be pre-populated with :literalref:`ops`. Any dependencies
-specified here will be bundled with the charm when it is built with
-:ref:`charmcraft pack <ref_commands_pack>`.
+used to declare and pin the version of any Python packages required by your charm.
+
+For :ref:`ref_commands_init` with the ``kubernetes`` or ``machine`` profile, Charmcraft
+doesn't create this file. Instead, Charmcraft creates a dependencies key in the
+:ref:`pyproject-toml-file`.
```
lengau, line 13: I think this would be better to have something like "not all init profiles create this file and it's not needed if [...]".

docs/reference/files/requirements-txt-file.rst
```diff
@@ -6,9 +6,13 @@
...
 
-.. _ops: https://ops.readthedocs.io/en/latest/
+For ``init`` with a 12-factor profile (``django-framework`` and so on), Charmcraft
```
lengau, line 15: @javierdelapuente @alithethird I'd love it if we moved the 12-factor templates over to `uv` too, though we don't have to do it right now. WDYT?

docs/reference/files/requirements-txt-file.rst
dwilding: I've updated to this wording:

> Charmcraft doesn’t create this file for all init profiles. It’s not needed if the profile is `kubernetes` or `machine`. These profiles use a pyproject.toml file instead.

docs/reference/files/tox-ini-file.rst
```diff
@@ -4,100 +4,35 @@
...
-This file is generated automatically by ``charmcraft init`` with the
-contents below:
+Charmcraft creates this file and defines the following environments:
 
-.. code-block:: yaml
+- ``format`` - Apply coding style standards
+- ``lint`` - Check code against coding style standards
+- ``static`` - Run static type checks
+- ``unit`` - Run the charm's unit tests
+- ``integration`` - Run the charm's integration tests
```
medubelko, line 19: I'd prefer a table, personally.

docs/reference/files/tox-ini-file.rst
```diff
@@ -4,100 +4,35 @@
...
+The ``tox.ini`` file in your charm's root directory is a typical tox
 configuration file.
 
-    See more: `Tox |
+    See more: `tox |
     Configuration <https://tox.wiki/en/latest/user_guide.html#configuration>`_
 
-This file is generated automatically by ``charmcraft init`` with the
-contents below:
+Charmcraft creates this file and defines the following environments:
```
medubelko, line 13:
```suggestion
Charmcraft creates this file and defines the following tox environments:
```

docs/reference/files/tox-ini-file.rst
```diff
@@ -4,100 +4,35 @@
...
-.. code-block:: yaml
+- ``format`` - Apply coding style standards
+- ``lint`` - Check code against coding style standards
+- ``static`` - Run static type checks
+- ``unit`` - Run the charm's unit tests
+- ``integration`` - Run the charm's integration tests
 
-      Copyright 2023 Ubuntu
-    # See LICENSE file for licensing details.
+To run an environment, use ``tox -e <env-name>``.
```
medubelko, line 21: Maybe?

```suggestion
To call an environment, run ``tox -e <env-name>``.
```

docs/reference/files/tox-ini-file.rst
```diff
@@ -4,100 +4,35 @@
...
 
-    [tox]
-    no_package = True
-    skip_missing_interpreters = True
-    env_list = format, lint, static, unit
-    min_version = 4.0.0
+For :ref:`ref_commands_init` with the ``kubernetes`` or ``machine`` profile,
```
medubelko, line 23:
```suggestion
For the Kubernetes and machine profiles,
```

docs/reference/files/tox-ini-file.rst
```diff
@@ -4,100 +4,35 @@
...
+``tox.ini`` requires the `tox-uv <https://github.com/tox-dev/tox-uv>`_ plugin, so you
+should use `uv <https://docs.astral.sh/uv/>`_ to install tox:
```
medubelko, line 25: Should we make it clear that this happens on the host?

docs/reference/files/tox-ini-file.rst
```diff
@@ -4,100 +4,35 @@
...
-        pytest-operator
-        -r {tox_root}/requirements.txt
-    commands =
-        pytest -v \
-                -s \
-                --tb native \
-                --log-cli-level=INFO \
-                {posargs} \
-                {[vars]tests_path}/integration
+For ``init`` with a 12-factor profile (``django-framework`` and so on), ``tox.ini``
```
medubelko, line 36:
```suggestion
For 12-factor app profiles, ``tox.ini``
```

docs/reference/files/uv-lock-file.rst
```diff
@@ -0,0 +1,27 @@
+.. _uv-lock-file:
+
+
+``uv.lock`` file
+================
+
+The ``uv.lock`` file in your charm's root directory is a lockfile that specifies the
```
medubelko, line 7:
```suggestion
The ``uv.lock`` file in your charm's root directory specifies the
```

Circular definition.

docs/reference/files/uv-lock-file.rst
```diff
@@ -0,0 +1,27 @@
...
+exact versions of your charm's dependencies and your development dependencies.
+
+    See more: `uv |
+    The lockfile
+    <https://docs.astral.sh/uv/concepts/projects/layout/#the-lockfile>`_
```
medubelko, line 12:
```suggestion
.. seealso::

    `uv | The lockfile
    <https://docs.astral.sh/uv/concepts/projects/layout/#the-lockfile>`_
```

docs/reference/files/uv-lock-file.rst
```diff
@@ -0,0 +1,27 @@
...
+
+This file is required if your charm uses the :ref:`craft_parts_uv_plugin`. It's also
+required if you use tox with tox-uv. For more information, see :ref:`tox-ini-file`.
+
+Charmcraft doesn't create this file when you run the :ref:`ref_commands_init` command.
+
+For ``init`` with the ``kubernetes`` or ``machine`` profile, you should create this
```
medubelko, line 19:
```suggestion
After initializing a charm with the Kubernetes or machine profile, create this
```

docs/reference/files/uv-lock-file.rst
```diff
@@ -0,0 +1,27 @@
...
+file by running ``uv lock``. Alternatively, you can create/update this file by using
+``uv add`` and ``uv remove`` to manage dependencies in the :ref:`pyproject-toml-file`.
+
+You shouldn't manually edit this file.
+
+For ``init`` with a 12-factor profile (``django-framework`` and so on), this file isn't
```
medubelko, line 25:
```suggestion
For the 12-factor app profiles, this file isn't
```

medubelko (approved):
I left a fair number of comments about flow and meaning, and less about technical correctness. Let me know what you think!

docs/reference/files/tox-ini-file.rst
dwilding: The tox terminology is awkward... I've gone with "To run the commands in an environment, [...]"

dwilding:
@lengau I just bumped the minimum Ops versions to 3 in the *pyproject.toml* and *uv.lock* templates. So everything is ready now, with the exception of a part of `init` output that I'm not 100% happy with ([see comment](https://github.com/canonical/charmcraft/pull/2312#discussion_r2177078297)).

charmcraft/application/commands/init.py
```diff
@@ -91,7 +93,9 @@
     │   │   └── test_charm.py      - Integration tests
     │   └── unit
     │       └── test_charm.py      - Unit tests
-    └── tox.ini                    - Configuration for tox, the tool to run all tests
+    ├── tox.ini                    - Configuration for tox, the tool to run all tests
+    ├── uv.lock                    - Specifies exact versions of Python dependencies,
+    │                                created if profile is 'kubernetes' or 'machine'
```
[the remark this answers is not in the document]
medubelko: I think it's sufficient for our purposes. The CLI interpreter is strict, so trickery with invisible chars and the like could be more trouble than they're worth.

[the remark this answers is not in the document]
medubelko: If you want to get fancy, we have a dotted pipe:

```
│
┆
```

And a simple point for a terminal vertex:

```
│
.
```

I don't think either are much better.

PR #2318, opened 2025-06-09
docs: fix broken intersphinx link

jahn-junior:
Looks like this intersphinx link broke when the Rockcraft docs got reorganized.


erinecon (approved):
Thanks for catching!

PR #2320, opened 2025-06-10
feat!(bundle): remove bundle features

lengau:
This removes the ability to pack bundles. When Charmcraft tries to load a charmcraft.yaml with `type: bundle`, it will error and tell the user to use Chramcraft 3.


lengau:
Weekly test failures are known and unrelated - they were triggered because I changed the weekly test workflow file to remove bundle tests.

bepri (changes requested):
Looks good, but I found a couple stragglers.

Also, two questions: there appear to be a lot of references to "bundles" remaining in `charmcraft/store/store.py` and `charmcraft/utils/store.py`. Similarly, there's still references to "bundles" in Craft Store's charm module. Are these there to support basic interfacing with charms from the store still? I would expect to have to downgrade to Charmcraft 3 for that functionality.

Second, do you want to track these changes in the release notes as the major release is made, or do you plan to write them all at once?

lengau:
@bepri related to the store-related bundle stuff: what remains is required for supporting the store API. The bundle removal is [already in the release notes](https://github.com/canonical/charmcraft/pull/2306/files#diff-89a3dbc2713e7f67cc1f64515e9891285ff9bff84d1b95fcb891eb9c5ca30f12R58).

docs/howto/manage-bundles.rst
medubelko, line 1: Please redirect all removed pages to the home page.

docs/reference/files/charmcraft-yaml-file.rst
```diff
@@ -29,7 +29,7 @@ to declare information about the project in a form that can be used by Charmcraf
 .. note::
 
     If you're starting from an empty file, the only required key is the ``type`` key.
-    However, depending on what value you set it to (``charm`` or ``bundle``), other
+    However, depending on what value you set it to (currently only ``charm``), other
```
medubelko, line 32: Shouldn't we remove this whole note?

docs/reference/files/charmcraft-yaml-file.rst
```diff
@@ -29,7 +29,7 @@ to declare information about the project in a form that can be used by Charmcraf
 .. note::
 
     If you're starting from an empty file, the only required key is the ``type`` key.
-    However, depending on what value you set it to (``charm`` or ``bundle``), other
```
medubelko, line 32: If `type: bundle`, the charm will simply fail to pack, right? No hint to switch to `charm`?

medubelko (changes requested):

docs/reference/files/charmcraft-yaml-file.rst
lengau: Correct. For Charmcraft 4.0 the only valid `type` is `charm`. (And switching to `charm` from a bundle would fail anyway since they're quite different things.)

docs/howto/manage-bundles.rst
lengau: Done! I also caught an intersphinx issue with Rockcraft that I've updated.

docs/reference/files/charmcraft-yaml-file.rst
lengau: Seems reasonable to me. Done!

docs/reference/files/charmcraft-yaml-file.rst
lengau: Actually I amend that: if you set `type: bundle` in `charmcraft.yaml` and run `charmcraft pack`, it will tell you that bundle support has been removed and to use charmcraft 3.

charmcraft/models/project.py
```diff
@@ -457,7 +443,13 @@ def unmarshal(cls, data: dict[str, Any]):
                 return BasesCharm.unmarshal(data)
             return PlatformCharm.unmarshal(data)
         if project_type == "bundle":
-            return Bundle.unmarshal(data)
+            raise errors.CraftValidationError(
+                "Invalid 'charmcraft.yaml' file"
+                details="Support for 'type: bundle' was removed in Charmcraft 4"
+                resolution: "To create charm bundles, use Charmcraft 3",
```
lengau, line 449:
```suggestion
                details="Support for 'type: bundle' was removed in Charmcraft 4",
                resolution="To create charm bundles, use Charmcraft 3",
```

medubelko (approved):

PR #2321, opened 2025-06-10
feat: add plucky and questing as devel bases

lengau:
This adds Ubuntu 25.04 and 25.10 as development bases for packing charms.


lengau:
Bundle-related test failures are known and unrelated to this. They are fixed in #2320

mr-cal (approved):

PR #2325, opened 2025-06-11
ci: run heavy workflows less frequently

lengau:


.github/workflows/qa.yaml
```diff
@@ -8,6 +8,13 @@ on:
       - "release/*"
       - "renovate/*"
   pull_request:
+    paths-ignore:
```
lengau, line 11: Right now this means we don't lint docs unless there are code changes, so that needs to change.

PR #2329, opened 2025-06-13
docs: correct env vars for flask secrets

javierdelapuente:
Fix environment name for Flask secrets.


docs/howto/manage-web-app-charms/configure-web-app-charm.rst
medubelko, line 1: Should we also update the env var names for FastAPI and Go on 275-276 and 285-286?

I took a quick look at the Charmcraft and Juju code but couldn't find where the names come from.

javierdelapuente: Hi @medubelko the other names are correct (for the rest of frameworks the env vars start with APP)

javierdelapuente:
> Thanks! Could you re-target this to the `hotfix/3.5` branch?

I changed the PR to be merged in `hotfix/3.5`. Do you apply it to main too from there?

PR #2334, opened 2025-06-13
ci: be more selective about when we run expensive tests

lengau:
We have a number of tests that take a long time (and a lot of energy) to run, and we don't need to run those on every single PR. This sets them not to run when the *only* changes are on certain paths that cannot affect them.

### DIFF
--- a/.github/workflows/spread.yaml
+++ b/.github/workflows/spread.yaml
@@ -1,10 +1,43 @@
 name: Spread
 on:
   pull_request:
+    paths-ignore:
+      - .github/**
+      - "!.github/workflows/spread.yaml" # Run on this file getting changed
+      - docs/**
+      - schema/**
+      - tests/** # Don't run if only tests changed, unless a spread test changed.
+      - "!tests/spread/**"
+      - .editorconfig
+      - .pre-commit-config.yaml
+      - .readthedocs.yaml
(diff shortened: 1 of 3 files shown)


.github/workflows/spread.yaml
```diff
@@ -1,10 +1,43 @@
...
+      - .github/**
+      - "!.github/workflows/spread.yaml" # Run on this file getting changed
+      - docs/**
+      - schema/**
+      - tests/** # Don't run if only tests changed, unless a spread test changed.
+      - "!tests/spread/**"
+      - .editorconfig
+      - .pre-commit-config.yaml
+      - .readthedocs.yaml
+      - "*.md" # Only top-level Markdown files
```
upils, line 14: I suspect these lists could contain more files/directories (ex. `tools`, `osv-scanner.toml`, etc.) but these are probably good first sets that can be tuned later.

.github/workflows/spread.yaml
```diff
@@ -1,10 +1,43 @@
...
   merge_group:
+    paths-ignore:
```
upils, line 16: It is a shame that we cannot reuse the list and have to duplicate it 3 times.

upils (approved):

PR #2338, opened 2025-06-17
docs: replace google for ci and fix tests

javierdelapuente:
This PR tries to improve the spread-docs, as they were not working in the CI. For that:

- Stop using google and instead run the tests in the self hosted github runners. The spread tests now use the new backend ci using the [.extension](https://github.com/canonical/snapcraft/blob/hotfix/8.9/snapcraft/templates/test/spread/.extension) from snapcraft. Thanks @lengau for this information.
- Several changes in all spread doc tests, mostly related to prepare/restore and retries.

### DIFF
--- a/docs/reuse/reference/extensions/integrations.rst
+++ b/docs/reuse/reference/extensions/integrations.rst
@@ -1,9 +1,9 @@
 
-``peers``, ``provides``, and ``requires`` keys
-----------------------------------------------
+Relations
+---------
 
-Your charm already has some ``peers``, ``provides``, and ``requires``
-integrations, for internal purposes.
+Your charm already has the following ``peers``, ``provides``, and ``requires``
+relations, as they were automatically supplied by the |framework| extension:
 
 .. code-block:: yaml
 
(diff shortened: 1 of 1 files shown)


.github/workflows/spread-docs.yaml
```diff
@@ -14,25 +14,70 @@ on:
...
+        id: charmcraft
+      - name: Upload snap artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: snap
+          path: ${{ steps.charmcraft.outputs.snap }}
+
   spread-select:
-    runs-on: spread-installed
+    runs-on: [self-hosted, self-hosted-linux-amd64-noble-medium]
```
lengau, line 34: Can we keep running the select job on a `spread-installed` container? We'll need to keep the addition of the checkout, but then we should be able to skip the setting up of go and spread.

.github/workflows/spread-docs.yaml
```diff
@@ -14,25 +14,70 @@ on:
...
+        with:
+          fetch-depth: 0
+          submodules: true
+      - name: Set up Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.24
+      - name: Install spread
+        run: |
+          go install github.com/snapcore/spread/cmd/spread@latest
```
lengau, line 80: Can we use the `spread` binary installed in the charmcraft snap instead? (Available as `charmcraft.spread`)

docs/tutorial/code/django/task.yaml
```diff
@@ -7,7 +7,13 @@
 ###########################################
 summary: Getting started with Django tutorial
 
-kill-timeout: 180m
+kill-timeout: 50m
+
+restore: |
+  rm -f $HOME/*.yaml $HOME/*.py $HOME/*.txt
```
lengau, line 13: Could put everything in a subdirectory of `$HOME` for each of these tasks so we can just delete that subdirectory?

.github/workflows/spread-docs.yaml
```diff
@@ -14,25 +14,58 @@ on:
     - cron: "0 0 * * 0" # Midnight UTC on Sundays
 
 jobs:
+  snap-build:
+    runs-on: [self-hosted, self-hosted-linux-amd64-noble-medium]
```
lengau, line 18:
```suggestion
    runs-on: [self-hosted, linux, amd64, noble, medium]
```

lengau (approved):
LGTM, thanks!

medubelko (approved):
Tests are running OK.

PR #2343, opened 2025-06-23
Chore: Fix minor documentation issues

alithethird:
ISD 3547

Fixed minor issues in the Go and FastAPI extension reference documentation:

- FastAPI reference remove "optional: false" in first config block
- FastAPI reference: remove Flask mentions
- Go reference: Fix the first list so that it renders correctly


erinecon (approved):
Thanks!

medubelko (approved):

PR #2347, opened 2025-06-23
docs: fix tree in init command docstring

medubelko:
For CHARMCRAFT-629:

- Fix an issue where Sphinx treated the code block of the file tree as a block quote.
- This doesn't fix an underlying issue where the lines are too long on 80-char terminals.

Other:

- Remove a link to a Juju page that no longer exists.

### DIFF
--- a/charmcraft/application/commands/init.py
+++ b/charmcraft/application/commands/init.py
@@ -76,7 +76,7 @@ Available profiles are:
         A basic Kubernetes charm for a 12-factor Go app.
 
 Depending on the profile choice, Charmcraft will setup the following tree of
-files and directories:
+files and directories::
 
     .
     ├── charmcraft.yaml            - Charm build configuration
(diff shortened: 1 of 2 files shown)


charmcraft/application/commands/init.py
```diff
@@ -76,7 +76,7 @@
         A basic Kubernetes charm for a 12-factor Go app.
 
 Depending on the profile choice, Charmcraft will setup the following tree of
-files and directories:
+files and directories::
```
lengau, line 79: This is going to make the colon appear twice on the CLI.

jahn-junior (approved):
While I agree that the RTD doc is a higher priority in this case, we should be careful making changes that affect the `help` output. This might look like a typo in plaintext.

charmcraft/application/commands/init.py
lengau: Per external discussion:

- Short term: We're going to accept this for now since it's less bad than how the sphinx docs look.
- Medium term: https://github.com/canonical/craft-cli/issues/351
- Long term: craft-cli is going to rethink how this is done.

lengau (approved):

PR #2350, opened 2025-06-26
docs: fix broken link to Juju docs in template CONTRIBUTING.md

tonyandrewmeyer:
juju.is/docs/sdk/dev-setup 404s now. This PR changes the links to that page in the template CONTRIBUTING.md to point to [the current equivalent](https://github.com/tonyandrewmeyer/charmcraft/pull/new/fix-link-in-contributing.md).

(There are many other juju.is links, but fixing these ones in particular seems straightforward, and many of the other redirect).

PR #2359, opened 2025-07-01
chore: merge 3.5.2 to main

lengau:


mr-cal (approved):

dwilding (approved):
Looks like the only change affecting templates is #2350, which is already in main. All good

PR #2370, opened 2025-07-10
ci: add experimental workflow using ty for type checking

lengau:
Since craft-platforms is passing with ty, let's give it something a bit harder to bang against.


mr-cal (approved):
The list of failures is too big to be informative to me, other than signaling there are a lot of issues.

I'm curious to see how many end up being actual issues vs `ty` issues and how quickly this list shrinks over time.

PR #2337, opened 2025-06-16
feat(extension): add Spring Boot extension

javierdelapuente:
Applicable spec: ISD216

### Overview
Adding Spring Boot framework extension.

### Rationale
Spring Boot is a very common Java framework and it would add a lot of potential users if we support it with the Paas-charm library and Charmcraft Extension.

PR #2373, opened 2025-07-14
feat: Add OpenID Connect integration to 12 Factor

javierdelapuente:
Adds the `oauth` relation to support the OpenID Connect authentication protocol.

Differently from other integrations, the configuration options names for `oauth` depend on the endpoint name. That way, many enpoints for then interface `oauth` can be added. That way, a charm will be able to integrate with two different OIDC providers (for example Google and GitHub at the same time).

Pending to add support for the Spring Boot framework once https://github.com/canonical/charmcraft/pull/2337  is merged.


javierdelapuente:
@erinecon

docs/reuse/reference/extensions/environment_variables.rst
```diff
@@ -124,4 +124,13 @@ Ingress relation.
...
+  * - OpenID Connect
+    -
+        - |framework_prefix|\ ``{endpoint_name}_CLIENT_ID``
+        - |framework_prefix|\ ``{endpoint_name}_CLIENT_SECRET``
+        - |framework_prefix|\ ``{endpoint_name}_ACCESS_TOKEN_URL``
+        - |framework_prefix|\ ``{endpoint_name}_AUTHORIZE_URL``
+        - |framework_prefix|\ ``{endpoint_name}_USERINFO_URL``
+        - |framework_prefix|\ ``{endpoint_name}_JWKS_URL``
+        - |framework_prefix|\ ``{endpoint_name}_API_BASE_URL``
+        - |framework_prefix|\ ``{endpoint_name}_CLIENT_KWARGS``
```
medubelko, line 136: There's a space in the final result:

Screenshot from 2025-08-20 14-31-30

If the markup doesn't work without the space, let me know, and I'll find a different method.

docs/reuse/reference/extensions/integrations.rst
```diff
@@ -130,6 +130,25 @@ the following charms:
               optional: True
               limit: 1
 
+  * - `Hydra <https://charmhub.io/hydra>`__ charm
+    - The ``oauth`` interface is used to provide the OpenID Connect authentication
+      protocol. The endpoint name will be used as the prefix to the configuration
+      options related to this endpoint.
```
medubelko, line 136:
```suggestion
    - The ``oauth`` interface is a conduit for the OpenID Connect authentication
      protocol. Each configuration option for this endpoint is prefixed with the
      endpoint name.
```

docs/reuse/reference/extensions/integrations.rst
```diff
@@ -130,6 +130,25 @@ the following charms:
...
+
+      .. code-block:: yaml
+
+          requires:
+            oauth-endpoint-name:
+              interface: oauth
+              optional: True
+              limit: 1
+
+      The integration will create two configuration options,
```
medubelko, line 146: > integration

Should this be "relation"?

medubelko (changes requested):
Hi @javierdelapuente! I'm sorry for the delay. This is quite close to ready. I'm making this review blocking because of the render error for the environment variable names.

docs/reuse/reference/extensions/environment_variables.rst
javierdelapuente: hi @medubelko. I tried to have the substitution inside the inline mark, but I could not make it work. Not sure if you know how to do it.

When that text is copied, the space is not there (the space is really padding in html). 

Another alternative is to remove the inline literals, but that will make that section different to the others (and inline sounds the correct way).

PR #2379, opened 2025-07-18
test(spread): per-base destructive mode tests

lengau:
Also removes the uv test since that is now redundant.

Broken spread tests are not related.


bepri (approved):
Tests seem reasonable and the code's all template stuff, LGTM

tests/spread/smoketests/uv/uv.lock
lengau, line 1: We're getting rid of the uv project in smoke tests because we're testing it per-platform instead.

tigarmo (approved):

PR #2384, opened 2025-07-22
docs: copy note from Juju docs, tidy up container structure doc

tonyandrewmeyer:
The ["Defining storage"](https://documentation.ubuntu.com/juju/3.6/reference/storage/#defining-storage) Juju doc is being removed. Most of this information is already in the `charmcraft.yaml` reference doc, but there is a note about the ability to specify the location of the mounted storage for K8s charms in either `storage` or `containers` that isn't there, so this copies that across.

While doing this, I've also updated the `containers` section of the doc to be less of a wall of text and more similar to the other sections of the doc, like `config`, where there's a shorter "Structure" section and then comments in example YAML that detail each of the fields.

This PR replaces #2380, targeting hotfix/3.5 rather than main as suggested there. The only difference between main and 3.5 in the existing text is dropping "windows" as an example base. I've dropped it here, since it's only in a list of examples not a list of all possible values, I assume we're not wanting to encourage its use, and it means that this will cleanly merge against main.


medubelko (approved):

lengau (approved):
Thanks!

PR #2395, opened 2025-07-28
feat!(create-track): make name and track named-args

lengau:
This is a breaking change in Charmcraft 4.0 for the create-track command.

CRAFT-3837


bepri (approved):
LGTM! Spread test failures appear unrelated, the only instance of `charmcraft create-track` in the tests was corrected here

lengau:
@bepri thanks for pointing that out - yes, the spread test failures are known and unrelated.

PR #2415, opened 2025-08-14
chore: merge 3.5.3 to main

lengau:
This merges the 3.5 hotfix branch to main so the history on main contains everything from the 3.5.3 release.

Spread test failures are known, unrelated, and reduced thanks to this :-)


charmcraft/templates/init-simple/charmcraft.yaml.j2
```diff
@@ -0,0 +1,66 @@
+# This file configures Charmcraft.
```
benhoyt: Is `main` the 4.0 branch? @dwilding In that case, isn't `init-simple` going away in 4.0?

lengau: It is - the merge grabbed this as a non-conflict, but I've amended it away :-)

tigarmo (approved):

mr-cal (approved):

PR #2416, opened 2025-08-14
fix(test): fix lifecycle spread tests for tox-uv

lengau:
This was left out when updating init templates. This PR should make the Ubuntu 24.04 spread tests succeed.

CHARMCRAFT-636


tests/spread/smoketests/full-lifecycle/charmcraft-simple.yaml
lengau, line 1: We no longer have a simple profile so this file was vestigial.

mr-cal (approved):

PR #2419, opened 2025-08-15
refactor: simplify resource option class

### LINKED ISSUE #1627: Simplify `ResourceOption`
### What needs to get done

[ResourceOption](https://github.com/canonical/charmcraft/blob/7640ef793a415c54701fe1472448b97a6dfde565/charmcraft/utils/cli.py#L56-L86) should just be a class that gets used in the CLI. We don't need to instantiate it and then call it

For an example of the simpler model, refer to: https://github.com/canonical/snapcraft/pull/4701#discussion_r1546725241

### Why it needs to get done

The current code is a bit weird to read.

aj4ayushjain:
Fixes #1627. The Resource option class can be directly instantiated with `__init__` making it easier to understand and read.


lengau (approved):
Thanks!

bepri (approved):

PR #2425, opened 2025-08-18

erinecon:
Relevant tickets: ISD-4006, ISD-3973

Five changes to the 12-factor tutorials based on feedback from UX sessions.

### Summary of changes
* Reorder instructions for `rockcraft.yaml` setup -- have the user check the architecture of their host first, then take a look at the project file to determine if `platforms` matches their host.
* Rephrase the uploading of the rock.
* Put more commands into code blocks, specifically all `curl` commands and checking that the newly added Juju configuration appears in the list.
* Update the links in the "Next steps" table to align more closely with the expected user journey. This included adding a link to the 12-factor RTD site and updating links to be more 12-factor specific.


docs/tutorial/kubernetes-charm-django.rst
```diff
@@ -854,16 +868,15 @@ in a number of typical ways. But there is a lot more to explore:
     * - If you are wondering...
       - Visit...
     * - "How do I...?"
-      - :ref:`How-to guides <how-to-guides>`,
-        :external+ops:ref:`Ops | How-to guides <how-to-guides>`
+      - :ref:`How to manage a 12-factor app charm <manage-12-factor-app-charms>`
```
medubelko, line 871:
```suggestion
      - :ref:`manage-12-factor-app-charms`
```

I don't think we need to rename this one.

docs/tutorial/kubernetes-charm-django.rst
```diff
@@ -854,16 +868,15 @@ in a number of typical ways. But there is a lot more to explore:
...
-      - `Charm debugging tools <https://juju.is/docs/sdk/debug-a-charm>`_
+      - :ref:`Troubleshoot a 12-factor app charm <troubleshoot-12-factor-charms>`,
+        :external+juju:ref:`Juju | Debug a charm <debug-a-charm>`
     * - "How do I get in touch?"
       - `Matrix channel <https://matrix.to/#/#12-factor-charms:ubuntu.com>`_
     * - "What is...?"
-      - :ref:`reference`,
-        :external+ops:ref:`Ops | Reference <reference>`,
+      - :external+rockcraft:ref:`Rockcraft profile <django-framework-reference>`,
+        :ref:`Charmcraft profile <django-framework-extension>`,
```
medubelko, line 879: The line break is a bit awkward here. What do you think about making each link its own line?

medubelko, line 879:
```suggestion
      - :external+rockcraft:ref:`django-framework extension in Rockcraft
        <django-framework-reference>`,
        :ref:`django-framework extension in Charmcraft <django-framework-extension>`,
```

I'm uncomfortable calling this simply the "*craft profile". That could mean a lot of different things depending on context.

docs/tutorial/kubernetes-charm-express.rst
```diff
@@ -533,13 +536,17 @@ We can now pack and deploy the new version of the Express app:
...
+    juju config expressjs-hello-world | grep -A 6 greeting:
+
+Check that the response is still ``Hello, world!`` using:
+
+.. code-block:: bash
 
-Using ``curl http://expressjs-hello-world  --resolve
-expressjs-hello-world:80:127.0.0.1``
-shows that the response is ``Hello, world!`` as expected.
+    curl http://expressjs-hello-world  --resolve expressjs-hello-world:80:127.0.0.1
```
medubelko, line 549: I didn't notice before, but on a few of these commands there are two spaces before the flags. They don't do any harm, but it might be a distraction to keen-eyed readers.

docs/tutorial/kubernetes-charm-django.rst
```diff
@@ -854,16 +868,15 @@ in a number of typical ways. But there is a lot more to explore:
...
         :external+juju:ref:`Juju | Reference <reference>`
     * - "Why...?", "So what?"
-      - :external+ops:ref:`Ops | Explanation <explanation>`,
-        :external+juju:ref:`Juju | Explanation <explanation>`
+      - :external+12-factor:ref:`12-Factor app support | Explanation <explanation>`
```
erinecon, line 882: Think about whether "support" is the best term here

docs/tutorial/kubernetes-charm-django.rst
erinecon: Sounds good, thanks!

docs/howto/manage-web-app-charms/use-web-app-charm.rst
```diff
@@ -38,6 +38,7 @@ popular libraries for programmatic database migration are:
 - `prisma <https://www.npmjs.com/package/prisma/>`__ and `knex
   <https://www.npmjs.com/package/knex/>`__ for Express.
 
+.. _troubleshoot-12-factor-charms:
```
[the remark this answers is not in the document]
erinecon: A very fair point, thanks. I'll update to your suggestion

docs/tutorial/kubernetes-charm-django.rst
erinecon: Sure, I'm fine with having each link be on a separate line

erinecon: Sounds good, I'll make this change :)

docs/tutorial/kubernetes-charm-django.rst
erinecon: Going to go with
```suggestion
      - :external+12-factor:ref:`12-Factor app principles and support in Charmcraft
         and Rockcraft <explanation>`
```

docs/tutorial/kubernetes-charm-express.rst
erinecon: Good catch, I'll remove the extra spaces since I'm already editing these lines :)

PR #2427, opened 2025-08-18
fix(lifecycle): fix --project-dir and --output

lengau:
This moves the final artifact to the correct location based on the `--project-dir` and `--output` parameters to `pack`.

### DIFF
--- a/tests/spread/smoketests/help-slug/task.yaml
+++ b/tests/spread/smoketests/help-slug/task.yaml
@@ -6,8 +6,9 @@ systems:
 prepare: |
   charmcraft init --project-dir=charm
 
-  # change requirements.txt to be invalid
-  sed -e 's/ops /oppss-invalid /' -i charm/requirements.txt
+  # change requirements to be invalid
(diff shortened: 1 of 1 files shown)


tools/gen_cli_docs.py
lengau, line 1: This came from the necessity of bringing in craft-application 5.7.

charmcraft/application/commands/lifecycle.py
```diff
@@ -163,4 +164,25 @@ def _run(
...
+                craft_cli.emit.debug(
+                    "Could not find artifacts in the state service. Not moving."
+                )
+            else:
+                project_dir = parsed_args.project_dir or pathlib.Path.cwd()
+                output_dir = parsed_args.output or pathlib.Path.cwd()
+
+                for artifact in artifacts.values():
+                    old_path = project_dir / artifact
+                    new_path = output_dir / artifact
```
bepri, line 184: Is it known that `artifact` will always be a relative path? I don't know a lot about the state service, so my immediate thought when seeing this was the possibility of `artifact` being an absolute path.

bepri (approved):

PR #2430, opened 2025-08-27
feat(templates): use Jubilant for integration tests of K8s and machine profiles

dwilding:
This PR updates the `kubernetes` and `machine` profiles to use the [Jubilant](https://documentation.ubuntu.com/jubilant/) library for integration tests. Main changes:

- Replace pytest-operator by Jubilant in `tests/integration/test_charm.py` and `pyproject.toml`
- Add `tests/integration/conftest.py`, which defines fixtures for creating a temporary Juju model and packing the charm under test
- Docs: update [tests/integration/test_charm.py file](https://canonical-ubuntu-documentation-library--2430.com.readthedocs.build/charmcraft/2430/reference/files/tests-integration-test-charm-py-file/) and add [tests/integration/conftest.py file](https://canonical-ubuntu-documentation-library--2430.com.readthedocs.build/charmcraft/2430/reference/files/tests-integration-conftest-py-file/)

For easier review, here are the charms that `charmcraft init` will produce:

- [Kubernetes charm](https://github.com/dwilding/initial-charms/tree/main/kubernetes-dev-jubilant)
- [Machine charm](https://github.com/dwilding/initial-charms/tree/main/machine-dev-jubilant)

### DIFF
--- a/charmcraft/templates/init-kubernetes/tests/integration/conftest.py.j2
+++ b/charmcraft/templates/init-kubernetes/tests/integration/conftest.py.j2
@@ -26,7 +26,7 @@ def juju(request: pytest.FixtureRequest):
             logger.info("Collecting Juju logs...")
             time.sleep(0.5)  # Wait for Juju to process logs.
             log = juju.debug_log(limit=1000)
-            print(log, end='', file=sys.stderr)
+            print(log, end="", file=sys.stderr)
 
 
 @pytest.fixture(scope="session")
--- a/charmcraft/templates/init-machine/tests/integration/conftest.py.j2
+++ b/charmcraft/templates/init-machine/tests/integration/conftest.py.j2
@@ -26,7 +26,7 @@ def juju(request: pytest.FixtureRequest):
             logger.info("Collecting Juju logs...")
             time.sleep(0.5)  # Wait for Juju to process logs.
             log = juju.debug_log(limit=1000)
-            print(log, end='', file=sys.stderr)
+            print(log, end="", file=sys.stderr)
 
 
 @pytest.fixture(scope="session")


charmcraft/templates/init-kubernetes/pyproject.toml.j2
```diff
@@ -70,7 +73,4 @@ max-complexity = 10
 skip = "build,lib,venv,icon.svg,.tox,.git,.mypy_cache,.ruff_cache,.coverage"
 
 [tool.pyright]
-include = [
-    "src/**/*.py",
-    "tests/**/*.py",
-]
+include = ["src", "tests"]
```
dwilding, line 76: Drive-by

dwilding:
macos-smoke-test failed, but I don't think that's related to my changes

benhoyt (approved):
Looks like a good start to me, thanks!

charmcraft/templates/init-machine/uv.lock.j2
lengau, line 1: This is just me being grumpy that a mechanised change is 1200 lines for something so minimal so ignore me if it's not doable, but what do you think about a couple of `pyproject.yaml` changes to make the lock file smaller?

Specifically, I don't think we care about limiting the charm to Linux:

```toml
[tool.uv]
environments = ["sys_platform == 'linux'"]
```

More problematic would be setting `requires-python = "~=3.10.0"` in the project section, but then people would have to change this to change their base.

charmcraft/templates/init-kubernetes/pyproject.toml.j2
lengau: A worthwhile change though, thanks!

charmcraft/templates/init-kubernetes/tests/integration/conftest.py.j2
lengau, line 1: These seem like they'd be worthwhile to put into a pytest plugin in Jubilant

charmcraft/templates/init-machine/uv.lock.j2
lengau: I suppose maybe we could use `python_version` and/or `implementation_name` in tool.uv.environments too...

charmcraft/templates/init-kubernetes/tests/integration/conftest.py.j2
tonyandrewmeyer: By design, Jubilant isn't going to have anything pytest specific (arg handling, fixtures, etc). It does provide some helpers to make it easy to write fixtures, like the `jubilant.temp_model()` used here. But it's not going to expand out to include the sorts of things that pytest-operator has.

There is a [pytest-jubilant](https://github.com/canonical/pytest-jubilant) that has those things. The current Charm Tech position is that we're happy to link to this as something charmers should be aware of and consider using, but not bake it into recommended practice, at least not yet.

For the `charm()` fixture, once we're all in on spread, this would go out of here and into a spread preparation step, right? In the short term, it seems ok to me to have it here, and it's definitely not going into Jubilant because we see it as something that happens outside of the Python code.

@dwilding how necessary do you think it is to have a `conftest.py`. I don't really like the magic of `conftest` anyway, so this is perhaps biased, but why not just have these two fixtures in the `test_charm.py` file itself? If/when a charm expands to have multiple integration test modules, the charmer could move them out into a `conftest.py` or might actually want to have different fixtures anyway, specifically suited to whatever that module is testing (like deploying a bunch of other apps, in a module focused on testing a specific integration, for example).

charmcraft/templates/init-machine/uv.lock.j2
dwilding: Almost all the changes are removals after we got rid of `juju` and `pytest-operator`, so (hopefully) this is the last time we'll have such a big one-off change. The line count is much lower now, at 431 lines. I tried `"sys_platform == 'linux'` and the line count went down to 368

But I think on balance I'm in favour of less extra config, that's easier for us and charmers to understand in future. So I'd lean towards leaving it as-is, if you'd be OK with that?

charmcraft/templates/init-kubernetes/tests/integration/conftest.py.j2
```diff
@@ -0,0 +1,33 @@
...
+            log = juju.debug_log(limit=1000)
+            print(log, end="")
+
+
+@pytest.fixture(scope="session")
+def charm():
+    """Pack the charm under test."""
+    subprocess.check_call(["charmcraft", "pack"])
+    # Modify below if you're building for multiple bases or architectures.
+    return next(pathlib.Path(".").glob("*.charm"))
```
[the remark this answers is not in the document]
dwilding: This is a great point. We had a discussion in Charm Tech daily today and here's what I'm proposing: We modify the `charm` fixture to check for an environment variable called `CHARM_PATH`. If that variable is set, we skip packing and pass the charm path through for Jubilant to deploy. Otherwise we pack the charm first. Would this approach work inside a spread instance?

See https://github.com/canonical/operator/pull/2018 for my idea of how this would be implemented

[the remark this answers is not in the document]
dwilding: Related: The `tox.ini` template passes these environment variables in every testing environment (see [here](https://github.com/canonical/charmcraft/blob/main/charmcraft/templates/init-kubernetes/tox.ini.j2#L21-L24)):

```
CHARM_BUILD_DIR
MODEL_SETTINGS
```

But these variables aren't used in any of the tests. Is there any value in retaining them?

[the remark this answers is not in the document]
dwilding: After further discussion within Charm Tech, I'm revising the proposal. We think it makes more sense for the integration tests to get out of the business of packing. New proposal:

- If `CHARM_PATH` is not set, we look for a `.charm` file, deploy that charm, then run the tests
- If `CHARM_PATH` is set, we deploy the specified charm file then run the tests

[Draft implementation for the httpbin charm in Ops](https://github.com/canonical/operator/pull/2018)

The charm author would be expected to run `charmcraft pack` before running integration tests. Technically, I suppose this would be a breaking change in the behaviour of the profiles. But I think it's cleaner and more flexible overall.

What do you think?

dwilding:
@lengau I added a couple of commits to this PR today. The main change is adding support for a `CHARM_PATH` variable and not having the integration testing code run `charmcraft pack`. Per [this comment](https://github.com/canonical/charmcraft/pull/2430#discussion_r2309246116)

I'd like to do some final testing tomorrow, if there's time before you need to merge to catch Charmcraft 4.

charmcraft/templates/init-machine/uv.lock.j2
lengau: Thanks!

lengau (approved):
LGTM, thanks!

dwilding:
Thanks @lengau! I just pushed a small adjustment to the integration tests after doing further testing. (Have all tests depend on the charm path, as in [this example](https://github.com/canonical/operator/blob/main/examples/httpbin-demo/tests/integration/test_charm.py), so that all tests fail immediately if the charm hasn't been packed.)

Good to go now :slightly_smiling_face:

dwilding:
Spread tests are failing, but I think the failed task is unrelated to this PR: `dependencies/setuptools-scm-python:python`

lengau:
Agreed - that's just flakiness in the archive server on an unrelated test.

PR #2445, opened 2025-09-15

erinecon:
Relevant ticket: ISD-4179

Five changes to the 12-factor tutorials based on feedback from UX sessions.

### Summary of changes

* Add a note in the "What you'll need" section to emphasize the importance of the required resources, specifically the RAM and disk space.
* Provide a bit more context about why we expose the app to ingress.
* Reword some awkward phrasing around changing directories and waiting for the app to be ready.
* Update the admonition block about configurations to clarify that the capitalization/prefix changes are made for the resulting environment variable. This is to help tie the concepts of app environment variables with charm configurations. 
* Flask and FastAPI tutorials: Update the admonition block about adding `psycopg2-binary` in the `requirements.txt` file to clarify that adding the PostgreSQL database happens at the end of the tutorial.

### DIFF
--- a/charmcraft/extensions/app.py
+++ b/charmcraft/extensions/app.py
@@ -57,12 +57,12 @@ SECRET_OPTIONS = {
     },
 }
 OAUTH_DYNAMIC_OPTIONS = {
-    "redirect-path": {
+    "{endpoint_name}-redirect-path": {
         "type": "string",
         "description": "The path that the user will be redirected upon completing login.",
         "default": "/callback",
     },
-    "scopes": {
+    "{endpoint_name}-scopes": {
         "type": "string",
         "description": "A list of scopes with spaces in between.",
         "default": "openid profile email",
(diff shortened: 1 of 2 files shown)


docs/tutorial/kubernetes-charm-django.rst
```diff
@@ -542,7 +545,7 @@ Set the configuration:
     Turning on debug mode shouldn't be done in production. We will do this in
     the tutorial for now and later disable debug mode.
 
-Let's expose the app using ingress. Deploy the
+Let's expose the app using ingress so that we can access the app. Deploy the
```
lengau, line 548:
```suggestion
Let's expose the app using ingress so that we can access it. Deploy the
```

docs/tutorial/kubernetes-charm-fastapi.rst
```diff
@@ -431,7 +434,7 @@ output:
     Unit                    Workload  Agent  Address      Ports  Message
     fastapi-hello-world/0*  active    idle   10.1.157.75
 
-Let's expose the app using ingress. Deploy the
+Let's expose the app using ingress so that we can access the app. Deploy the
```
lengau, line 437:
```suggestion
Let's expose the app using ingress so that we can access it. Deploy the
```

lengau (approved):
Thanks!

PR #2447, opened 2025-09-17
fix: align oidc redirect-uri to Spring Boot defaults

javierdelapuente:
Currently the default callback (redirect-uri) for oidc is `/callback` for all frameworks. 

This is not very Spring Boot specific, and requires code changes in the applications, see:
https://docs.spring.io/spring-security/reference/servlet/oauth2/login/advanced.html#oauth2login-advanced-redirection-endpoint 

Set the default value of redirect-uri to the Spring Boot default as `/login/oauth2/code/{endpoint_name}`. This will make working with Spring Boot seamlessly, and the user still can change it using the config option.

As the Spring Boot extension is still experimental, this is a not breaking change.

### DIFF
--- a/charmcraft/extensions/app.py
+++ b/charmcraft/extensions/app.py
@@ -57,12 +57,12 @@ SECRET_OPTIONS = {
     },
 }
 OAUTH_DYNAMIC_OPTIONS = {
-    "redirect-path": {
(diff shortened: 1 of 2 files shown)


lengau (approved):
LGTM, thanks!

alithethird (approved):

PR #2456, opened 2025-10-02
chore: add Matrix room to 12-factor YAML templates

erinecon:
Relevant ticket: ISD-3415

### Overview

Add Matrix room to the YAML templates for the 12-factor framework extensions.

### Rationale

As part of the 12-factor documentation roadmap for this cycle, we want to clarify the process for providing feedback and guidance in the user journey. As part of these discussions, one point of improvement was to advertise the 12-factor Matrix room as a way to answer questions and help in the setup and onboarding process.


medubelko (approved):
Looks fine to me.

PR #2458, opened 2025-10-06
feat: Add app profiles to Spring Boot

javierdelapuente:
Add the Spring Boot configuration option app-profiles that will be exposed as the environment variable spring.profiles.active. See
https://docs.spring.io/spring-boot/reference/features/profiles.html.

In Spring Boot, a profile is a way to group and activate configuration based on the environment in which the application is running. This is a usual configuration in Spring Boot that is not easy to configure in 12-factor. This PR adds that configuration so it can be easily set depending on the environment the application is deployed in.

This is the related PR in paas-charm: https://github.com/canonical/paas-charm/pull/172

### DIFF
--- a/docs/reference/extensions/spring-boot-framework-extension.rst
+++ b/docs/reference/extensions/spring-boot-framework-extension.rst
@@ -151,3 +151,38 @@ variable name will have the hyphens replaced by underscores and all the letters
 capitalised.
 
    See more: :external+juju:ref:`Juju | Secret <secret>`
+
+.. _spring-boot-grafana-graphs:
+
+Grafana dashboard graphs
+------------------------
+
+If the Spring Boot app is connected to the `Canonical Observability Stack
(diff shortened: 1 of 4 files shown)


lengau (approved):

docs/reference/extensions/spring-boot-framework-extension.rst
```diff
@@ -26,6 +26,9 @@ The predefined configuration options for the ``spring-boot-framework`` are:
   configured using this port. The environment variable passed to the app is
   ``APP_PORT``. Default value is 8080.
 
+* **app-profiles**: Comma-separated list of Spring Boot active profiles. The environment
+  variable passed to the app is ``spring.profiles.active``.
```
medubelko, line 30:
```suggestion
* **app-profiles**: Comma-separated list of active Spring Boot profiles. The environment
  variable passed to the app is ``spring.profiles.active``.
```

This is a grammar nitpick.

PR #2463, opened 2025-10-14

erinecon:
Relevant ticket: ISD-3531

Three changes to the setup instructions for the 12-factor tutorials based on feedback from UX sessions.

### Summary of changes

* Add an admonition box about experimental extensions, warning users that they need to install `latest/edge` of Charmcraft and Rockcraft.
* Rearrange the LXD setup instructions to check whether LXD is installed, install LXD if not installed, and then initialize LXD.
* Add a sentence to make it clear that, once we shell into the VM, all further instructions happen within the VM unless stated otherwise.

### DIFF
--- a/docs/reuse/tutorial/setup_edge.rst
+++ b/docs/reuse/tutorial/setup_edge.rst
@@ -25,7 +25,7 @@ Install Rockcraft and Charmcraft
 
 .. admonition:: This extension is under active development
 
-    This tutorial requires the edge releases of Rockcraft and Charmcraft, which are 
+    This tutorial requires the edge releases of Rockcraft and Charmcraft, which are
     riskier than the stable releases.
 
 In order to create the rock, you need to install Rockcraft with


docs/reuse/tutorial/setup_edge.rst
```diff
@@ -18,31 +18,46 @@ Once the VM is up, open a shell into it:
...
 
+Unless stated otherwise, we will work entirely within the VM from now on.
+
+Install Rockcraft and Charmcraft
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. admonition:: This extension is under active development
+
+    To follow this tutorial, you must install ``latest/edge``
+    of Rockcraft and Charmcraft.
```
medubelko: Since we describe installation specifics later, I think we can cover the why a bit more here:

```suggestion
    This tutorial requires the edge releases of Rockcraft and Charmcraft, which are 
    riskier than the stable releases.
```

PR #2464, opened 2025-10-15
docs: add related links section to docs homepage

james-garner-canonical:
This PR adds `relatedlinks` metadata to the Charmcraft documentation site homepage.

[**Preview build**](https://canonical-ubuntu-documentation-library--2464.com.readthedocs.build/charmcraft/2464/)

This uses the related links extension included in `canonical`'s Sphinx extensions.

I've used the same list that we use for [ops](https://documentation.ubuntu.com/ops/latest/) and [charmlibs](https://documentation.ubuntu.com/charmlibs/), with the addition of Craft Parts. Perhaps there are more craft specific links that could be included, or maybe some of the currently included links aren't related enough (Jubilant? Pebble?). The list is currently (manually) sorted alphabetically, but this can be changed too.

### DIFF
--- a/docs/reuse/tutorial/setup_edge.rst
+++ b/docs/reuse/tutorial/setup_edge.rst
@@ -18,6 +18,16 @@ Once the VM is up, open a shell into it:
 
     multipass shell charm-dev
 
+Unless stated otherwise, we will work entirely within the VM from now on.
+
+Install Rockcraft and Charmcraft
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. admonition:: This extension is under active development
+
+    This tutorial requires the edge releases of Rockcraft and Charmcraft, which are
+    riskier than the stable releases.
+
 In order to create the rock, you need to install Rockcraft with
 classic confinement, which grants it access to the whole file system:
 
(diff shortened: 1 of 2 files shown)


james-garner-canonical:
Thanks, @medubelko!

medubelko:
@james-garner-canonical #2466 is merged. :)

docs/index.rst
```diff
@@ -1,3 +1,5 @@
+:relatedlinks: [Charmlibs](https://canonical-charmlibs.readthedocs-hosted.com/), [Concierge](https://github.com/canonical/concierge), [Craft&#32;Parts](https://canonical-craft-parts.readthedocs-hosted.com/latest/), [Jubilant](https://documentation.ubuntu.com/jubilant/), [Juju](https://documentation.ubuntu.com/juju/), [Ops](https://documentation.ubuntu.com/ops/), [Pebble](https://documentation.ubuntu.com/pebble/)
```
medubelko: In my mind, it should be ordered based on immediate relevance to building charms, then managing or deploying them. This makes more sense to me:

- Juju
- Ops
- Charmlibs
- Jubilant
- Concierge
- Pebble

I think we should leave out Craft Parts, because it's not product/tool and not something would use with Charmcraft.

I'd like to hear your thoughts!

james-garner-canonical: I'm happy to make that change. We used alphabetical ordering because that's what Juju does, but I don't see any reason not to use a different ordering.

I added Craft Parts because I've found it to be a very useful reference when writing `charmcraft.yaml` files, specifically when writing custom `parts`. But happy to remove that if it's considered too low-level.

lengau: If we're not going alphabetical I'd even suggest we might want to promote Jubilant over charmlibs, but I have no strong opinions.

PR #2466, opened 2025-10-15
build(docs): remove line length check from linter

medubelko:


jahn-junior (approved):
Thanks! I'm sure external contributors will be overjoyed by this change.

If this is the new standard for our doc sets, we should update Starbase as well.

lengau (approved):
Thanks!

PR #2475, opened 2025-10-27
docs: fix incorrect indentation in YAML snippets

dwilding:
Some YAML snippets in the docs are rendered incorrectly, with some properties indented less than they should be. For example, in the [charmcraft.yaml doc](https://documentation.ubuntu.com/charmcraft/stable/reference/files/charmcraft-yaml-file/#containers):

Screenshot from 2025-10-24 19-37-50

Under `bases`, the `channel` and `architectures` properties should be indented to the same level as `name`. Similarly, under `mounts`, the `location` property should be indented to the same level as `storage`.

I've pinned the issue down to the sphinx_toolbox extension, which the Charmcraft docs use. Bug report: https://github.com/sphinx-toolbox/sphinx-toolbox/issues/198

The extension introduces a custom `code-block` directive that supports a `tab-width` option. It doesn't look like the Charmcraft docs require this custom directive. The docs presumably use the extension for a different purpose.

Workaround approaches (that I can think of):

1. Replace the offending code blocks by `literalinclude`.
2. <mark>Use `code` instead of `code-block`. The extension doesn't touch the `code` directive, it seems.</mark>
3. Reduce the level of indentation by extracting the deeply-nested YAML into standalone code blocks.
4. Don't import the whole extension; only import the modules that are needed.
5. Redefine `code-block` to the built-in directive.

After discussion (see comments) we went with approach 2. Only replacing `code-block` by `code` where the output was broken. [Preview of one of the fixed snippets](https://canonical-ubuntu-documentation-library--2475.com.readthedocs.build/charmcraft/2475/reference/files/charmcraft-yaml-file/#containers)

### DIFF
--- a/docs/reuse/reference/extensions/http_proxy.rst
+++ b/docs/reuse/reference/extensions/http_proxy.rst
@@ -5,13 +5,13 @@ HTTP Proxy
 Proxy settings can be sourced either from the model configurations or from the
 ``http-proxy`` relation. When the ``http-proxy`` relation is present, its
 values take precedence over those defined in the model configuration. Charms
-built using the TARGET extension automatically expose the Juju proxy
+built using the |framework| extension automatically expose the Juju proxy
 settings to the workload as the environment variables ``HTTP_PROXY``,
 ``HTTPS_PROXY`` and ``NO_PROXY``. The ``HTTP_PROXY`` and ``HTTPS_PROXY`` values
 are obtained from the relation, whereas the ``NO_PROXY`` value is obtained from
 the model configuration. For example, if the relation is absent, the
 ``juju-http-proxy`` environment variable will be exposed as ``HTTP_PROXY`` to
-the TARGET service.
+the |framework| service.
 
     See more:
     :external+juju:ref:`Juju | List of model configuration


medubelko:
Thanks for making this a priority!

> Use `code` instead of `code-block`. The extension doesn't touch the code directive, it seems.

I didn't know this was the case. They're supposed to be [aliases for each other](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-code-block).

I like that the solution the best. It's easy to fix and doesn't require customisation, just a different pattern that we put into practice.

medubelko:
> Would you want me to switch all code blocks within the doc source?

Until we have Vale running in the repos where we can validate its use, I don't think correcting it for all the normal instances will make much of a difference. For the specific cases where it's broken, it's fine to use it.

docs/reference/files/metadata-yaml-file.rst
```diff
@@ -549,14 +549,14 @@ and the default is 0 (root).
...
-          resource: super-app-image
-          mounts:
-            - storage: logs
-              location: /logs
+        containers:
+          super-app:
+            resource: super-app-image
+            mounts:
+              - storage: logs
+                location: /logs
```
dwilding: This one looked like it was meant to be within the dropdown, so I fixed that too

dwilding:
> For the specific cases where it's broken, it's fine to use it.

:+1: I've gone ahead and switched to `code` for the broken snippets. Let me know what you think

james-garner-canonical:
@lengau Are you happy to merge this PR too?

lengau (approved):

Issue #2494, opened 2025-11-30
charmcraft pack wont produce a proper charm


erik78se:
### Check existing issues

- [x] I've verified that this bug isn't described by any existing issues.

### Bug description

**When:** Creating a default machine charm with charmcraft init.
```
mkdir foobar
charmcraft init --profile machine
vi pyproject.toml # modify the dependencies
charmcraft pack
unzip -l foobar_amd64.charm # The dependencies has not been included as expected
```

**Expected 1:** A charm that builds a machine charm and respects the dependencies in pyproject.toml for added python libs.

**Expected 2:** The charm builds for 24.04 since this would be something reasonable.

**What I did:** I added a library to the pyproject.toml dependencies.

**What happened:** The built charm was built for a 22.04 (very old)
**What happened:** The built charm ignored the pyproject.toml

What I would like to see:

The charmcraft init --profile machine should behave properly and respect the pyproject.toml (and be 24.04)

### Steps to reproduce

```
mkdir foobar
charmcraft init --profile machine
vi pyproject.toml # modify the dependencies
charmcraft pack
unzip -l foobar_amd64.charm # The dependencies has not been included as expected
```

### Environment

charmcraft 4.0.1

### charmcraft.yaml

```yaml
type: charm
name: foobar
title: Charm Template
summary: A very short one-line summary of the charm.
description: |
  test

base: ubuntu@22.04
platforms:
  amd64:

parts:
  charm:
    plugin: uv
    source: .
    build-snaps:
      - astral-uv
```

### Log output

```shell

```

erik78se:
Also remove references to OPS if pyproject.toml isnt used or perhaps
indicate the exact way to add or remove packages from the project.

The generated skeleton is super important to provide a help to new
Chalmers, so getting this template right will affect anyone trying to use
juju.

Den tors 4 dec. 2025 04:20Dave Wilding ***@***.***> skrev:

PR #2506, opened 2025-12-04
feat: add adoption help at end of init message

dwilding:
The machine and Kubernetes profiles in Charmcraft 4 use the uv plugin, which requires a different way of managing dependencies than in earlier versions of Charmcraft. The difference is not apparent when running `charmcraft init`. See https://github.com/canonical/charmcraft/issues/2494 for an example of how this can trip people up.

This PR adds extra lines at the end of the success message from `charmcraft init`:

```
To manage your charm's dependencies, use uv.

To migrate from the 'charm' plugin, see:
https://documentation.ubuntu.com/charmcraft/stable/howto/migrate-plugins/charm-to-uv/
```

My thinking is that the second line (and doc link) could be removed in a future version of Charmcraft.

---

### DIFF
--- a/charmcraft/application/commands/init.py
+++ b/charmcraft/application/commands/init.py
@@ -126,7 +126,7 @@ README.md
 
 To manage your charm's dependencies, use uv.
 
-To migrate from the 'charm' plugin, see:
+To migrate from the Charm plugin to the uv plugin, see:
 https://documentation.ubuntu.com/charmcraft/stable/howto/migrate-plugins/charm-to-uv/
 """
 


charmcraft/application/commands/init.py
```diff
@@ -123,6 +123,11 @@ def _make_success_message(src_files: list[str]) -> str:
 charmcraft.yaml
 {src_files_str}
 README.md
+
+To manage your charm's dependencies, use uv.
+
+To migrate from the 'charm' plugin, see:
```
medubelko:
```suggestion
To migrate from the Charm plugin to the uv plugin, see:
```

PR #2664, opened 2026-04-22
tests: add reproducer for reactive plugin missing .build.manifest (#872)

### LINKED ISSUE #872: charm artifacts from reactive plugin are missing valuable build metadata
when a reactive charm is built with `charm build .` the resulting artifact has a file in the top level with the title `.build.manifest`

This file is generated by the `charm` snap and lists the commit hash for each charm layer, installed pip from the wheelhouse

1) It seems when i build a reactive charm using the charmcraft reactive plugin, this file is missing from the `.charm` artifact.

2) it would also be REALLY nice if the `manifest.yaml` generated would also list the `charmtool-version` used to build the charm. 
 the plugin can determine this version by running `/snap/bin/charm version --format json` and extracting the output

lengau:
## Summary

Adds a failing unit test that reproduces #872: when building a reactive charm, the `.build.manifest` file generated by `charm build` is not copied into the install directory.

## Bug

When `charm build` runs, it places `.build.manifest` in the `build_dir/` root alongside the charm subdirectory. The reactive plugin only symlinks `build_dir/<charm_name>/` into `install_dir/`, so `.build.manifest` is never included in the `.charm` artifact.

## Reproducer

```
uv run python -m pytest tests/unit/parts/plugins/test_reactive.py::test_build_includes_build_manifest -v
```

The test fails with:
```
AssertionError: .build.manifest generated by charm build must be present in the install directory
```

Closes #872


lengau:
@copilot Please use the instructions and skills from https://github.com/canonical/charmcraft/pull/2663 and create a fix for the bug reproduced by the failing test in this PR.

PR #2668, opened 2026-04-22
tests: add reproducer for dev multipass version causing InvalidVersion (#1917)

### LINKED ISSUE #1917: Charmcraft cannot parse a dev version build for multipass
### Bug Description

charmcraft throws invalid version for a dev build of multipass which follows sematic versioning spec.

### To Reproduce

* Install a dev build of multipass 
* Run `charmcraft pack`

### Environment

OSX 15.0 (Sequioa)

### charmcraft.yaml

```yaml
N/A
```

### Relevant log output

```shell
charmcraft internal error: InvalidVersion("Invalid version: '1.15.0-dev.2929.pr661'")
Full execution log: '/Users/X/Library/Logs/charmcraft/charmcraft-y-z.log'
```

lengau:
## Summary

Adds a failing unit test that reproduces #1917: when multipass has a dev version string like `1.15.0-dev.2929.pr661+gc67ef6641.mac`, charmcraft crashes with an unhandled `packaging.version.InvalidVersion` internal error.

## Bug

The root cause is in `craft_providers`: `multipass.is_supported_version()` calls `packaging.version.parse()` which raises `InvalidVersion` for non-PEP-440 strings. Charmcraft's `ProviderService.instance()` does not catch this exception, so it propagates as an internal error.

Upstream bug: https://github.com/canonical/craft-providers/issues/661

The fix in charmcraft would be to catch `InvalidVersion` in `ProviderService.instance()` and convert it to a user-friendly `CraftError`.

## Reproducer

```
uv run python -m pytest 'tests/unit/services/test_provider.py::test_instance_handles_invalid_multipass_version' -v
```

The test fails with:
```
packaging.version.InvalidVersion: Invalid version: '1.15.0-dev.2929.pr661'
```

Closes #1917


lengau:
@copilot Please use the instructions and skills from https://github.com/canonical/charmcraft/pull/2663 and create a fix for the bug reproduced by the failing test in this PR.

PR #2677, opened 2026-04-29
docs: specify tested framework versions in python tutorials

f-atwi:
Describe your changes.

---

- [x] In documents I changed, I [added a meta description](https://canonical-starflow.readthedocs-hosted.com/how-to/add-a-page-meta-description/) if one was missing.


docs/tutorial/kubernetes-charm-django.rst
```diff
@@ -18,6 +18,8 @@ running with Juju. Let's get started!
 
 This tutorial should take 90 minutes for you to complete.
 
+Tested with Django 5.1.4.
```
erinecon: WDYT about moving this note into the "What you'll do" section, updating the first step to say something like:
```
#. Create a Django app (tested with Django 5.1.4).
```
If you agree, I think we should do this for all 3 tutorials :)

medubelko: I like this. I would tweak the wording:

```
Create a Django app with Django 5.1.4.
```

But that's a nit.

docs/tutorial/code/flask/requirements.txt
lengau, line 1: Please add comments in these requirements files reminding people to update the docs if they update the requirements :-)

f-atwi: Even as a comment, it will show in the rendered docs. I wouldn't want instructions for doc maintainers to be visible to the people following a tutorial as that would cause confusion
the alternative would be to have the comment in the tutorial where the requirements file is included but I would argue that is not that visible when modifying the file

erinecon: Does it make sense to include this reminder in Charmcraft's contributing documentation? Or include a README in `docs/tutorial/code` with this reminder? 

I share the concern about forgetting to update both the docs and the requirements files, but I'm not entirely sure the best approach for guaranteeing these files are consistently maintained.

erinecon (approved):
Documentation LGTM -- thanks!

docs/tutorial/code/flask/requirements.txt
medubelko: @erinecon @f-atwi It should be sufficient to add a comment to the top of the files, and then start from line two (`:lines: 2-`) in the `literalinclude` blocks.

medubelko (approved):

lengau (approved):

PR #2717, opened 2026-06-05
docs: configure latest docs for canonical.com proxy

jahn-junior:
Configure the documentation to be migrated to https://canonical.com/juju/docs/charmcraft

I'll be opening similar PRs against the hotfix/4.2 and hotfix/3.5 branches when we're ready to complete the migration.

---


docs/conf.py
```diff
@@ -28,23 +28,20 @@
 project = "Charmcraft"
 author = "Canonical"
 
-# Sidebar documentation title; best kept reasonably short
 # The full version, including alpha/beta/rc tags
-release = charmcraft.__version__
+release = os.environ.get("READTHEDOCS_VERSION", charmcraft.__version__)
 # The commit hash in the dev release version confuses the spellchecker
 if ".post" in release:
     release = "dev"
```
medubelko: As an aside, I'd rather there be consistency between local and remote builds, so the upcoming logic will be something like this:

https://github.com/canonical/starbase/blob/work/add-version-tests/tests/unit/starbase/test_version.py#L25-L32

It's an active bug that the sites read "<product> dev", so at least your code fixes that for now.

docs/conf.py
```diff
@@ -90,27 +87,27 @@
...
+slug = "juju/docs/charmcraft"
 
 
 #########################
 # Sitemap configuration #
 #########################
 
 # Use RTD canonical URL to ensure duplicate pages have a specific canonical URL
-html_baseurl = os.environ.get("READTHEDOCS_CANONICAL_URL", "/")
+html_baseurl = f"{ogp_site_url}{release}/"
```
medubelko: Is it better to keep the separators out of the component strings? Would require editing `ogp_site_url`.

```suggestion
html_baseurl = f"{ogp_site_url}/{release}/"
```

medubelko (approved):

lengau (approved):

docs/conf.py
jahn-junior: https://github.com/canonical/charmcraft/pull/2717/commits/c8bc94a929b1a4c0614ab0f38ac705b4d747e169

docs/conf.py
jahn-junior: Probably makes more sense. Whether they're in the links or not, it should at least be consistent.

https://github.com/canonical/charmcraft/pull/2717/commits/005fd8e532cd1d517994475a8940d477d988aad0

PR #2721, opened 2026-06-08
fix(docs): auto-sync framework versions to 12-factor tutorials

erinecon:
Following https://github.com/canonical/charmcraft/pull/2703 , this PR syncs the Django, Flask, and FastAPI versions automatically to the tutorial. That way, we can surface the software versions without having to manually maintain them. 

---

- [ ] I've followed the [contribution guidelines](https://github.com/canonical/charmcraft/blob/main/CONTRIBUTING.md).
- [ ] I've signed the [CLA](http://www.ubuntu.com/legal/contributors/).
- [X] I've added or updated any relevant documentation.


docs/tutorial/kubernetes-charm-django.rst
```diff
@@ -46,7 +46,12 @@ than the sufficient resources, the tutorial will take longer to complete.
 What you'll do
 ~~~~~~~~~~~~~~
 
-#. Create a Django app with Django 5.1.4.
+#. Create a Django app using the following version of Django:
+
+    .. include:: code/django/requirements.txt
+        :start-after: Django==
```
medubelko: This is a clever way to transclude that I never considered. :eyes:

I don't like the way it appears on the page, though:

image

It's a lot of screen space for a single version number.

There are other substitution solutions we could pursue, namely Sphinx aliases (`|django-tutorial-version|`), which work inline. We could, in principle, hook a very small script at the end of `conf.py` that reads from that `requirements.txt` file and sets the value in the global Sphinx context at build time.

@jahn-junior Are there other simple methods we could look into?

erinecon: Thanks @medubelko , my original intention was to avoid touching `conf.py` since I didn't want to add more complexity to a project that I don't strictly own, but since you've invited this type of change, I gave it a try! 

Right now the script in `conf.py` is hardcoded for Django, but if it's acceptable to you and @jahn-junior , then I can convert it to a function that we can use to extract + inject the Flask and FastAPI versions, too.

For full disclosure, I used Gemini to iterate on this solution.

jahn-junior: > @jahn-junior Are there other simple methods we could look into?

I can't think of anything simpler than @erinecon's current approach. It's a shame the tutorial deps aren't project deps because then we could access them directly.

I'm more interested in whether this brings enough value over the list item simply stating "Create a Django app". This is a structured lesson, so the version seems arbitrary to me as long as the tutorial is reproducible. If this was a how-to guide where a user brings their own software and needs to apply this workflow, then a callout like this would be far more important in my eyes. Here, it's a nice detail, but I don't think it's worth the added complexity.

docs/conf.py
```diff
@@ -343,3 +345,29 @@ def link_common_docs(library_name: str) -> None:
...
+    with open(req_path, "r", encoding="utf-8") as f:
+        content = f.read()
+
+        match = re.search(r"Django==([\d.]+)", content)
+        if match:
+            extracted_number = match.group(1)
+
+# Dynamically append or create the global substitution variable
+rst_epilog = rst_epilog if 'rst_epilog' in locals() else ""
+rst_epilog += f"\n.. |django_version| replace:: {extracted_number}"
```
medubelko: I think we should pick a substitution name that's extremely unlikely for a user to happen upon. Maybe `conf_django_version`.

docs/conf.py
```diff
@@ -343,3 +345,29 @@ def link_common_docs(library_name: str) -> None:
...
+if os.path.exists(req_path):
+    with open(req_path, "r", encoding="utf-8") as f:
+        content = f.read()
+
+        match = re.search(r"Django==([\d.]+)", content)
+        if match:
+            extracted_number = match.group(1)
+
+# Dynamically append or create the global substitution variable
+rst_epilog = rst_epilog if 'rst_epilog' in locals() else ""
```
medubelko: I think it makes more sense for the data if the substitution is in `prolog`. If a user tries to override the substitution in the page, it will error regardless, so preceding over a user-set substitution name of the same name won't matter here.

docs/conf.py
medubelko: @copilot is there a way to rework these lines without assigning `rst_epilog` twice?

docs/conf.py
```diff
@@ -9,6 +9,8 @@
 
 import charmcraft
 
+import re
```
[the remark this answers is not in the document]
medubelko: This should be addressed.

.github/workflows/spread-docs.yaml
```diff
@@ -2,7 +2,7 @@ name: Spread tests for docs
 on:
   pull_request:
     paths:
-      - .github/workflow/spread-docs.yaml
+      - .github/workflows/spread-docs.yaml
```
medubelko: Good catch!

docs/conf.py
medubelko: Guess I can't ask questions on others' PRs. :thought_balloon:

docs/conf.py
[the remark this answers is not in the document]
erinecon: Ah okay, good to know. I'll re-order the imports, thanks!

docs/conf.py
erinecon: `conf_django_version` is reasonable to me, I'll update. Thanks!

erinecon: I was going to ask your question to copilot, but I get a :warning: that says "Copilot isn't available for cross-repository pull requests". I guess because I'm using a fork -- good to know. But I can get Copilot to work on this locally :)

erinecon: Local Copilot pointed out that `rst_epilog` is already defined earlier in the file (the `.. include:: /reuse/links.txt` block), so the conditional check is unnecessary (the check only matters if `rst_epilog` might not exist, but we know it does).

So I'll go ahead and remove the first line from `conf.py`, although it might not matter in any event since you also suggested to put the data into `prolog` instead.

docs/conf.py
erinecon: Sounds good, will update to use `rst_prolog` instead

medubelko:
@erinecon When you mark this as ready for review, I'll give this a full check on my machine.

docs/conf.py
```diff
@@ -343,3 +345,29 @@ def link_common_docs(library_name: str) -> None:
...
+extracted_number = "unknown"
+
+# extract the number
+if os.path.exists(req_path):
+    with open(req_path, "r", encoding="utf-8") as f:
+        content = f.read()
+
+        match = re.search(r"Django==([\d.]+)", content)
+        if match:
+            extracted_number = match.group(1)
```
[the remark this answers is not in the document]
erinecon: Not really sure if this is something we should incorporate.

medubelko (approved):
Works fine for me locally. Thanks @erinecon!

PR #2723, opened 2026-06-12
fix: match Juju's action name validation rules

tonyandrewmeyer:
The documentation implies that a Juju action name may contain underscores, but [that is not the case](https://github.com/juju/charm/blob/v12/actions.go#L20), and Juju will refuse to deploy a charm that does have one. This PR updates the documentation to be clearer what the real rules are.

It's also better for charmcraft to fail rather than Juju, so the Juju regular expression is also copied over to the charmcraft validation. Also, the validation wasn't hooked up, even though other fields (like name) are, which seems like a bug, so adjusted that too.

---

- [x] I've 'successfully' run `make lint && make test`.
- [x] In documents I changed, I [added a meta description](https://canonical-starflow.readthedocs-hosted.com/how-to/add-a-page-meta-description/) if one was missing.
- [x] I've updated the relevant release notes.

### DIFF
--- a/docs/reference/files/actions-yaml-file.rst
+++ b/docs/reference/files/actions-yaml-file.rst
@@ -1,5 +1,8 @@
 .. _actions-yaml-file:
 
+.. _meta::
+    :description: Reference for the ``actions.yaml`` file, including the structure, keys, and naming rules for declaring Juju actions that a charm exposes to operators.
+
 ``actions.yaml`` file
 =====================
 
--- a/docs/reference/files/actions-yaml-file.rst
+++ b/docs/reference/files/actions-yaml-file.rst
@@ -109,10 +109,12 @@ charm.
 **Structure:**
 
 *Name:* The name of the key (``<action name>``) is defined by the charm
-author. It must be a valid Python :external+python:ref:`identifier <identifiers>`
-that does not collide with Python :external+python:ref:`keywords <keywords>`
-except that it may contain hyphens (which will be mapped to underscores in the Python
-event handler).
+author. It must match the regular expression
+``^[a-z0-9](?:[a-z0-9-]*[a-z0-9])?$`` (lowercase ASCII letters and digits,
+with internal hyphens permitted) and must not collide with a Python
+:external+python:ref:`keyword <keywords>`. Underscores are not permitted;
+hyphens in the action name are mapped to underscores in the Python event
+handler.
 
 *Type:* Map.
 
--- a/docs/reference/files/actions-yaml-file.rst
+++ b/docs/reference/files/actions-yaml-file.rst
@@ -1,7 +1,7 @@
 .. _actions-yaml-file:
 
-.. _meta::
-    :description: Reference for the ``actions.yaml`` file, including the structure, keys, and naming rules for declaring Juju actions that a charm exposes to operators.
+.. meta::
+    :description: Reference for the ``actions.yaml`` file, with descriptions for each of the keys used to declare a charm's supported actions.
 
 ``actions.yaml`` file
 =====================
--- a/docs/reference/files/actions-yaml-file.rst
+++ b/docs/reference/files/actions-yaml-file.rst
@@ -112,12 +112,9 @@ charm.
 **Structure:**
 
 *Name:* The name of the key (``<action name>``) is defined by the charm
-author. It must match the regular expression
-``^[a-z0-9](?:[a-z0-9-]*[a-z0-9])?$`` (lowercase ASCII letters and digits,
-with internal hyphens permitted) and must not collide with a Python
-:external+python:ref:`keyword <keywords>`. Underscores are not permitted;
-hyphens in the action name are mapped to underscores in the Python event
-handler.
+author. It must consist only of lowercase ASCII letters, numbers, and hyphens (-).
+A hyphen can't be the first or last character. It must not consist entirely of a
+:external+python:ref:`Python keyword <keywords`.
 
 *Type:* Map.
 
--- a/docs/reference/files/charmcraft-yaml-file.rst
+++ b/docs/reference/files/charmcraft-yaml-file.rst
@@ -1,7 +1,7 @@
 .. _charmcraft-yaml-file:
 
-.. _meta::
-    :description: Reference for the ``charmcraft.yaml`` project file, covering every supported top-level key, nested structures, and the inline charm metadata, actions, and configuration that Charmcraft consumes when packing a charm.
+.. meta::
+    :description: Reference for the Charmcraft project file, with descriptions and examples of each of its keys.
 
 ``charmcraft.yaml`` file
 ========================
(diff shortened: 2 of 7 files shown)


docs/release-notes/charmcraft-4.2.rst
```diff
@@ -99,6 +102,8 @@ The following issues have been resolved in Charmcraft 4.2.2:
 
 - `#2598 <https://github.com/canonical/charmcraft/issues/2598>`__ Remote build fails
 when getting project name
+- `#TODO <https://github.com/canonical/charmcraft/pull/TODO>`__ Match Juju's
```
tonyandrewmeyer:
```suggestion
- `#2723 <https://github.com/canonical/charmcraft/pull/2723>`__ Match Juju's
```

lengau (approved):
Thanks!

tests/unit/models/test_actions.py
```diff
@@ -0,0 +1,66 @@
...
+        "-leading-hyphen",
+        "trailing-hyphen-",
+        "has space",
+        "with.dot",
+        "",
+    ],
+)
+def test_invalid_action_names_rejected(name):
+    with pytest.raises(ValueError, match="is not a valid action name"):
+        JujuActions(actions=_actions(name))
```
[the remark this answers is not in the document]
tonyandrewmeyer: I'm not sure this is correct. In Pydantic v2, ValidationError is a subclass of ValueError, so pytest.raises(ValueError) catches it correctly and the tests do pass as written. I've run the tests and they pass.

tests/unit/models/test_actions.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+@pytest.mark.parametrize("name", ["if", "for", "class", "return"])
+def test_python_keyword_action_names_rejected(name):
+    with pytest.raises(ValueError, match="reserved keyword"):
+        JujuActions(actions=_actions(name))
```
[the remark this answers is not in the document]
tonyandrewmeyer: Added a couple more tests.

docs/reference/files/actions-yaml-file.rst
```diff
@@ -109,10 +112,12 @@ charm.
...
-author. It must be a valid Python :external+python:ref:`identifier <identifiers>`
-that does not collide with Python :external+python:ref:`keywords <keywords>`
-except that it may contain hyphens (which will be mapped to underscores in the Python
-event handler).
+author. It must match the regular expression
+``^[a-z0-9](?:[a-z0-9-]*[a-z0-9])?$`` (lowercase ASCII letters and digits,
+with internal hyphens permitted) and must not collide with a Python
+:external+python:ref:`keyword <keywords>`. Underscores are not permitted;
+hyphens in the action name are mapped to underscores in the Python event
+handler.
```
medubelko: I think it would be best to leave out the regex ([compare to Snapcraft](https://documentation.ubuntu.com/snapcraft/stable/reference/snapcraft-yaml/#name)) and write this in plain English:

```suggestion
author. It must consist of lowercase ASCII letters, numbers, and hyphens (-). The first
character can't be a hyphen. It must not contain a 
:external+python:ref:`Python keyword <keywords>`. It can't contain underscores (_).
```

medubelko (commented):
Left a suggestion for the key text. @jahn-junior Can you take a look at the meta descriptions? You're closer to the subject matter than me.

docs/reference/files/actions-yaml-file.rst
tonyandrewmeyer: Maybe this?

```suggestion
author. It must consist only of lowercase ASCII letters, numbers, and hyphens (-).
A hyphen cannot be the first or last character. It must not be a
:external+python:ref:`Python keyword <keywords`.
```

The underscores bit shouldn't really be needed, since it's already said which characters are allowed, and an underscore was not there. I think it is important that a hyphen can't be the last character, and it's more accurate to say that it can't *be* a keyword, rather than *contain* one, since you certainly can have a "fork" action, even though that contains "for".

medubelko: Your text addresses the nuances much better than mine! I have some followup suggestions ("can't" and "consist entirely"):

```suggestion
author. It must consist only of lowercase ASCII letters, numbers, and hyphens (-).
A hyphen can't be the first or last character. It must not consist entirely of a
:external+python:ref:`Python keyword <keywords`.
```

docs/reference/files/charmcraft-yaml-file.rst
```diff
@@ -1,5 +1,8 @@
 .. _charmcraft-yaml-file:
 
+.. _meta::
+    :description: Reference for the ``charmcraft.yaml`` project file, covering every supported top-level key, nested structures, and the inline charm metadata, actions, and configuration that Charmcraft consumes when packing a charm.
```
jahn-junior: Here's what we've been using for some of the other craft apps' project file references.
```suggestion
.. meta::
    :description: Reference for the Charmcraft project file, with descriptions and examples of each of its keys.
```
I'd like to leave out the file name here since it's already the most prominent text in the search result. Aside from that, I had an attempt at making this a little more concise.

docs/release-notes/charmcraft-4.2.rst
```diff
@@ -1,5 +1,8 @@
 .. _release-4.2:
 
+.. _meta::
+    :description: Release notes for Charmcraft 4.2, listing the new features, behavioural changes, deprecations, and bug fixes shipped in the 4.2.x patch series.
```
jahn-junior, line 4:
```suggestion
.. meta::
    :description: Learn about the new features, changes, and fixes introduced in Charmcraft 4.2.
```
We've just been reusing the page's first sentence for release notes meta descriptions.

docs/reference/files/actions-yaml-file.rst
```diff
@@ -1,5 +1,8 @@
 .. _actions-yaml-file:
 
+.. _meta::
+    :description: Reference for the ``actions.yaml`` file, including the structure, keys, and naming rules for declaring Juju actions that a charm exposes to operators.
```
jahn-junior:
```suggestion
.. meta::
    :description: Reference for the ``actions.yaml`` file, with descriptions for each of the keys used to declare a charm's supported actions.
```
Tried to align this with the project file description, but I'm not as picky about this one. Let me know what you think.

jahn-junior (changes requested):
Thanks! I left a few suggestions on the meta descriptions, but this is only blocking because of the directive syntax issues.

docs/reference/files/actions-yaml-file.rst
tonyandrewmeyer: Seems fine to me, thanks. For the most part people shouldn't need to look at this file anyway, unless they are Juju devs. Although charmcraft.yaml does defer to it a bit.

docs/release-notes/charmcraft-4.2.rst
tonyandrewmeyer: This one I will leave alone, because I assume there will be / is already one of these and so it should get merged out of the PR. But I had to include it to tick all the boxes.

jahn-junior (approved):
Doc changes and meta descriptions LGTM. Thanks again.

PR #2732, opened 2026-06-17
docs: Update 12-factor Spread tests

erinecon:
This PR fixes typos and outdated material in the 12-factor Spread tests.

---


charmcraft/extensions/app.py
erinecon, line 1: Hi @lengau , can you take a look at these code changes please? 

For some context: When I was debugging the Spread tests, I found that the Flask and Django tests were failing at the `charmcraft pack` stage with errors like `Extension ... does not support base(s): ubuntu@24.04`. See for example [this failed run for the Django test](https://github.com/canonical/charmcraft/actions/runs/27421751368/job/81053185205). 

I was able to reproduce the error locally. The tutorials still work because users install `latest/stable` of Charmcraft, which doesn't report these errors. But if you try the tutorial with `latest/edge`, then you indeed run into the same error.

Since Copilot suggested this code change, I wanted to run it by a pair of technical eyes first. This type of code change has a precedent; other frameworks have similar functions to define the supported bases. Does it make sense that we also add these functions to the Flask and Django frameworks, too?

medubelko:
The Spread tests are failing because of the store. If they're still failing next week, I'll check locally.

.github/workflows/spread-docs.yaml
```diff
@@ -2,7 +2,7 @@ name: Spread tests for docs
 on:
   pull_request:
     paths:
-      - .github/workflow/spread-docs.yaml
+      - .github/workflows/spread-docs.yaml
```
lengau, line 5: 😆

charmcraft/extensions/app.py
lengau: Ope, yes that's an oversight. Thank you!

Once this is ready I'd like to cherry-pick it for 4.2

lengau (approved):
Thanks! I think we have a couple of still-broken spread tests, but they're *less* broken now so I'm still happy to merge regardless.

charmcraft/extensions/app.py
lengau: Oh actually I think this got separately fixed by @alithethird as well, so we *should* be able to revert the changes in this file to fix the tests :-)

erinecon: Ah okay, thanks! I can revert these changes and see what happens :) 
(Waiting for the current Spread tests to finish running because I'm curious)

PR #2733, opened 2026-06-17
chore: remove ops dependency in framework templates

Thanhphan1147:
## Summary
- Remove `ops` dependency from all framework `requirements.txt.j2` templates since `paas-charm` already lists `ops` as a dependency


Thanhphan1147:
Hi @lengau, This PR is ready for review. I found this issue during testing which preventing charms to install paas_charm v1.11.3 which contains the support for valkey

lengau (approved):

PR #2742, opened 2026-06-24
docs: 12 factor v2

alithethird:
Describe your changes.

---

- [ ] I've followed the [contribution guidelines](https://github.com/canonical/charmcraft/blob/main/CONTRIBUTING.md).
- [ ] I've signed the [CLA](http://www.ubuntu.com/legal/contributors/).
- [ ] I've successfully run `make lint && make test`.
- [ ] I've added or updated any relevant documentation.
- [ ] In documents I changed, I [added a meta description](https://canonical-starflow.readthedocs-hosted.com/how-to/add-a-page-meta-description/) if one was missing.
- [ ] I've updated the relevant release notes.

PR #2754, opened 2026-07-03
feat(profiles): bound dependencies, remove uv.lock files, improve init output

### LINKED ISSUE #2749: Remove uv.lock from `kubernetes` and `machine` profiles
At the Madrid sprints in May, we decided to stop shipping `uv.lock` when initing charms with `--profile kubernetes` and `--profile machine`. It's awkward to keep the locked versions of packages up-to-date. Instead, we expect the charm developer to run `uv lock` after initing their charm.

I also propose a clean-up of the `init` output. Something like:

```
Created project files for your charm:

charmcraft.yaml
pyproject.toml
README.md
src/charm.py
src/demo_app.py
...

To manage your charm's dependencies, use uv.

To migrate from the Charm plugin to the uv plugin, see:
https://documentation.ubuntu.com/charmcraft/stable/howto/migrate-plugins/charm-to-uv/

Next steps:
 1. Run 'uv lock'
 2. Edit charmcraft.yaml and pyproject.toml to provide metadata, then commit (including uv.lock)
 3. Write your charm code and tests
```

dwilding:
This PR makes several changes related to profiles. Fixes #2749.

---

1. For the `kubernetes` and `machine` profiles, bound all dependencies so that (on user update) they stay within the same major version. The minimum version of each dependency is the latest tested major.minor version. See:

    - `charmcraft/templates/init-kubernetes/pyproject.toml.j2`
    - `charmcraft/templates/init-machine/pyproject.toml.j2`

---

2. Remove `uv.lock` from the `kubernetes` and `machine` profiles. After internal discussion, we feel it makes more sense for the user to lock and ensure they have up-to-date dependencies, rather than us continually bumping the template lockfiles.

    After making this change, I updated the Spread tests as thoroughly as I could.

---

3. Make the output of `charmcraft init` better match the requested profile:

    - The summary of created files is now correct for each type of profile. E.g., for `flask-framework` the summary lists `requirements.txt`; for `test-kubernetes` the summary only lists Spread-related files.
    - For profiles that use uv to manage dependencies, include a message that explains about running `uv lock`. This applies to `kubernetes`, `machine`, and the upcoming 26.04 12-factor profiles. It doesn't apply to the existing 12-factor profiles.

**Example for `kubernetes`**

```text
$ charmcraft init
Created project files for your charm:

charmcraft.yaml
pyproject.toml
README.md
src/charm.py
src/myapp.py
...

To manage your charm's dependencies, use uv.

To migrate from the Charm plugin to the uv plugin, see:
https://canonical.com/juju/docs/charmcraft/stable/howto/migrate-plugins/charm-to-uv/

Next steps:

1. Run 'uv lock'
2. Edit charmcraft.yaml and pyproject.toml to provide metadata, then commit (including uv.lock)
3. Write your charm code and tests
```

**Example for `flask-framework`**

```text
$ charmcraft init --profile flask-framework
Created project files for your charm:

charmcraft.yaml
pyproject.toml
requirements.txt
src/charm.py
...
```

---

4. Updated `init` examples in the documentation:

    - [Manage charms > Initialise a charm](https://canonical-charmcraft--2754.com.readthedocs.build/2754/howto/manage-charms/#initialise-a-charm) (expand "Example session")
    - [Manage extensions > View details about the extension in use](https://canonical.com/juju/docs/charmcraft/4.3/howto/manage-extensions/#view-details-about-the-extension-in-use) (expand "Example")

---

5. Updated file reference docs to present uv-managed charms as the default, which will automatically cover the 26.04 12-factor profiles when they are released. The existing 12-factor profiles are called out as an exception.

    - [`pyproject.toml` file](https://canonical-charmcraft--2754.com.readthedocs.build/2754/reference/files/pyproject-toml-file/)
    - [`requirements.txt` file](https://canonical-charmcraft--2754.com.readthedocs.build/2754/reference/files/requirements-txt-file/)
    - [`tox.ini` file](https://canonical-charmcraft--2754.com.readthedocs.build/2754/reference/files/tox-ini-file/)
    - [`uv.lock` file](https://canonical-charmcraft--2754.com.readthedocs.build/2754/reference/files/uv-lock-file/)

I also updated the `init` command help. See `charmcraft/application/commands/init.py`.

For info about the upcoming 26.04 12-factor profiles, see #2742.

---


tests/spread/smoketests/lxd/task.yaml
```diff
@@ -4,13 +4,11 @@ summary: pack a init-created charm with lxd
...
 prepare: |
+  snap install --classic --no-wait astral-uv
+  snap watch --last=install
   charmcraft init --project-dir=charm
   cd charm
-
-  # The charm itself doesn't matter, so replace the dependencies with something
-  # simple and small so we don't have to build pyyaml.
-  # NOTE: We still do need at least one dependency here so we can trigger the cache.
-  echo "distro" > requirements.txt
```
dwilding, line 13: This part of the test has been unused since we migrated the `kubernetes` profile to use `pyproject.toml` for dependencies. (Likely I missed updating the test at the time.) Unless anyone objects, I think we can drop it.

docs/howto/manage-extensions.rst
```diff
@@ -31,71 +31,67 @@ extension.
 
 .. dropdown:: Example
```
dwilding, line 42: The example seems out of place. It's in a section called "View details about the extension in use", and I think the example is there to illustrate the "Suppose you've initialised [...]" part. But this seems distracting. We already have exactly the same example on the "Manage charms" page, so maybe the best thing would be to remove the example from this page?

Maybe we only keep the part that shows `charmcraft.yaml`, since that is the main point of the text:

> [...] your rockcraft.yaml / charmcraft.yaml > extensions lists this extension.

lengau (approved):
LGTM, thanks!

charmcraft/application/commands/init.py
```diff
@@ -86,10 +86,12 @@
     ├── LICENSE                    - Your charm license, we recommend Apache 2
     ├── pyproject.toml             - Configuration for testing, formatting and
     │                                linting tools. Specifies Python dependencies for
+    │                                all profiles except 12-factor app profiles
+    │                                targeting Ubuntu 24.04 or lower
```
tonyandrewmeyer: lower->earlier? Same below.

charmcraft/application/commands/init.py
```diff
@@ -86,10 +86,12 @@
...
     │                                your charm if profile is 'kubernetes' or 'machine'
```
tonyandrewmeyer: Is this line meant to be removed?

charmcraft/application/commands/init.py
```diff
@@ -111,23 +111,30 @@
...
+Next steps:
+
+1. Run 'uv lock'
+2. Edit charmcraft.yaml and pyproject.toml to provide metadata, then commit (including uv.lock)
+3. Write your charm code and tests
 """
+    if "pyproject.toml" in project_files and "requirements.txt" not in project_files:
+        return f"{default_message}\n{uv_message}"
+    else:
+        return default_message
```
tonyandrewmeyer: Very much just style, so feel free to ignore.

```suggestion
    return default_message
```

charmcraft/templates/init-kubernetes/pyproject.toml.j2
```diff
@@ -13,28 +13,28 @@ requires-python = ">=3.10"
 # 'charm-libs' block in charmcraft.yaml, run `charmcraft fetch-libs` to download the libraries,
 # then inspect the libraries for dependencies specified in PYDEPS. List those dependencies here.
 dependencies = [
-    "ops~=3.7",
+    "ops>=3.8,<4",
```
tonyandrewmeyer: I know not everyone agrees, and it's annoying that there isn't "one obvious way to do it", but I think `~` is a nicer way to express this.

charmcraft/templates/init-kubernetes/pyproject.toml.j2
```diff
@@ -13,28 +13,28 @@ requires-python = ">=3.10"
...
 ]
 
 [dependency-groups]
 # Dependencies of linting and static type checks
 lint = [
-    "ruff",
-    "codespell",
-    "pyright",
+    "ruff>=0.15,<1",
```
tonyandrewmeyer: I'm not sure `<1` makes sense. Generally, a 0.x release has no API guarantee at all, so any version bump contains the same risk. Going from 0.x to 1.0 actually seems likely to *increase* stability in terms of the API, because it should be saying "ok, we're committed to this API now, and will maintain it until there's a breaking 2.x release".

No upper bound is probably not ideal either. `~0.15` would give the intent better, but assuming we're not using `~`, what about making it `>=0.15<2`?

charmcraft/templates/init-kubernetes/pyproject.toml.j2
```diff
@@ -13,28 +13,28 @@ requires-python = ">=3.10"
...
+    "codespell>=2.4,<3",
+    "pyright>=1.1,<2",
 ]
 # Dependencies of unit tests
 unit = [
-    "coverage[toml]",
-    "ops[testing]",
-    "pytest",
+    "coverage[toml]>=7.15,<8",
+    "ops[testing]>=3.8,<4",
```
tonyandrewmeyer: We don't really need the versions here, because `ops` will force them. I'm not sure what `uv add` and the like will do here, but I think generally it would be nicer to have no constraint here, and perhaps a comment explaining why. That way this doesn't need to be kept in lockstep to avoid impossible resolution.

docs/howto/manage-charms.rst
```diff
@@ -44,26 +44,27 @@ profile, you get the ``kubernetes`` profile.
...
         ls -R
 
         .:
-        charmcraft.yaml  pyproject.toml  src  tox.ini
+        charmcraft.yaml  pyproject.toml  requirements.txt  src  tox.ini
 
         ./src:
         charm.py
 
+..
```
tonyandrewmeyer, line 66: Is this meant to be an ellipsis? Or is it a ReST syntax thing?

tests/integration/commands/test_init.py
```diff
@@ -148,6 +147,27 @@ def test_files_created_correct(
...
+def test_success_message(new_path, init_command, emitter, profile: str, uses_uv: bool):
+    init_command.run(create_namespace(profile=profile))
+    output = "\n".join(
+        c.args[1] for c in emitter.interactions if c.args[0] == "message"
+    )
+    if uses_uv:
+        assert "\nuv.lock\n" not in output  # uv.lock isn't listed as a created file.
+        assert "Run 'uv lock'" in output
+    else:
+        assert "uv.lock" not in output
```
tonyandrewmeyer, line 169: I get that these are doing different things - not being a listed file and not being present at all, but it seems like a very subtle difference for the test, and maybe a bit fragile given that they differ only by whitespace?

I guess this line can't do the work for both because there's an instruction to commit `uv.lock` that would be hit in that case. What about having the first one be a regex that matches being in the list of files?

tests/spread/hooks/configure/task.yaml
```diff
@@ -8,7 +8,11 @@ include:
 
 prepare: |
   tests.pkgs install unzip
+  snap install --classic --no-wait astral-uv
+  snap watch --last=install
   charmcraft init --project-dir=charm
+  cd charm
```
tonyandrewmeyer, line 14: Is it safe here to change into the charm directory and not change out? I'm not sure exactly what `prepare` does at the end, if it restores you to where you were for example. It seems like the pushd/popd style used elsewhere might be cleaner.

docs/reference/files/uv-lock-file.rst
```diff
@@ -13,16 +13,16 @@ your charm's dependencies and your development dependencies.
     <https://docs.astral.sh/uv/concepts/projects/layout/#the-lockfile>`_
 
 This file is required if your charm uses the :ref:`craft_parts_uv_plugin`. It's also
-required if you use tox with tox-uv. For more information, see :ref:`tox-ini-file`.
+required by the tox-uv plugin. For more information, see :ref:`tox-ini-file`.
 
-When a charm is initialized with the Kubernetes or machine profile, Charmcraft creates
-this file.
+Charmcraft doesn't create this file.
```
tonyandrewmeyer: Does this need to be a paragraph of its own? It seems like it could be the intro for the next one.

tonyandrewmeyer (approved):
Thanks! I only tried a couple of profiles myself, but they worked well, and the rest make sense to me reading through it. A few small questions/suggestions.

charmcraft/application/commands/init.py
jahn-junior: 'higher' and 'lower' are preferred in the craft apps https://documentation.ubuntu.com/starflow/latest/how-to/starcraft-style-guide/#version-names

However, we should change 'Ubuntu 24.04' to 'Ubuntu 24.04 LTS' for branding and consistency with the subsequent change to this file.

charmcraft/application/commands/init.py
```diff
@@ -111,23 +110,29 @@
...
 To manage your charm's dependencies, use uv.
 
 To migrate from the Charm plugin to the uv plugin, see:
-https://documentation.ubuntu.com/charmcraft/stable/howto/migrate-plugins/charm-to-uv/
+https://canonical.com/juju/docs/charmcraft/stable/howto/migrate-plugins/charm-to-uv/
+
+Next steps:
+
+1. Run 'uv lock'
+2. Edit charmcraft.yaml and pyproject.toml to provide metadata, then commit (including uv.lock)
```
jahn-junior, line 130: Optional:
```suggestion
2. Edit the charmcraft.yaml and pyproject.toml files to provide metadata, then commit (including uv.lock)
```
Since this is in CLI copy, I could go either way it adds characters. Up to you!

docs/howto/manage-charms.rst
jahn-junior: I believe this was added to denote the end of the preceding terminal directive. If it wasn't there, the subsequent "See more" links would get rendered as terminal output because of their indentation level.

docs/howto/manage-extensions.rst
jahn-junior: That seems reasonable to me. Since the section is about extensions that are _already_ in use, the example output you point out should've come and gone well before the user gets here.

This page needs more attention, but I think cutting this is a fine way to start :)

docs/howto/manage-extensions.rst
jahn-junior:
```suggestion
  .. terminal::
      :output-only:
```
Completely unrelated to your changes, but this is all getting rendered as input. If you don't end up making any more changes to this PR, feel free to resolve and I'll fix this separately.

docs/reference/files/pyproject-toml-file.rst
```diff
@@ -12,15 +12,15 @@ Python ``pyproject.toml`` file.
...
-this file with the following contents:
+Charmcraft creates this file with the following contents:
 
 - Dependencies of the charm code, pre-populated with :external+ops:doc:`Ops <index>`
 - Dependencies of tests and linters
 - Configuration of tests and linters
 
-If you manually modify the dependencies, you'll need to update the :ref:`uv-lock-file`.
+You'll need to create :ref:`uv-lock-file` and update it if you change your charm's
+dependencies.
```
jahn-junior:
```suggestion
You'll need to create and update the :ref:`uv-lock-file` if you change your charm's
dependencies.
```
Missing article before the link text. I also tweaked the wording slightly, but I'm less opinionated there.

docs/reference/files/pyproject-toml-file.rst
```diff
@@ -12,15 +12,15 @@ Python ``pyproject.toml`` file.
...
 
-When a charm is initialized with a 12-factor app profile, ``pyproject.toml`` contains
-the configuration for tests and linters. Dependencies are specified in the
+For 12-factor app profiles targeting Ubuntu 24.04 LTS or lower: ``pyproject.toml``
```
jahn-junior: (Is that an unsplit hair I see?) Optional:
```suggestion
For 12-factor app charms targeting Ubuntu 24.04 LTS or lower, the ``pyproject.toml`` file
```
Let me know how you feel about the distinction between profiles and the charms created with them.

I've always considered the 'profile' to be the means by which you initialize 12-factor charms. Profiles may support a range of bases, but they don't target one in particular, right? That's something the user decides for the charm they're crafting _with_ a profile.

docs/reference/files/requirements-txt-file.rst
```diff
@@ -4,14 +4,13 @@
 ``requirements.txt`` file
 =========================
 
-The ``requirements.txt`` file is a standard Python `requirements file
+Charmcraft only creates this file for 12-factor app profiles targeting Ubuntu 24.04 LTS
+or lower. Other profiles use a :ref:`pyproject-toml-file` instead.
```
jahn-junior: Optional:
```suggestion
Charmcraft creates this file for 12-factor charms targeting Ubuntu 24.04 LTS
or lower. Other profiles and bases use a :ref:`pyproject-toml-file` instead.
```

charmcraft/application/commands/init.py
```diff
@@ -86,10 +86,11 @@
     ├── LICENSE                    - Your charm license, we recommend Apache 2
     ├── pyproject.toml             - Configuration for testing, formatting and
     │                                linting tools. Specifies Python dependencies for
-    │                                your charm if profile is 'kubernetes' or 'machine'
+    │                                all profiles except 12-factor app profiles
```
jahn-junior: Optional:
```suggestion
    │                                all profiles except 12-factor app charms
```
See https://github.com/canonical/charmcraft/pull/2754/changes#r3634177095

charmcraft/application/commands/init.py
```diff
@@ -86,10 +86,11 @@
...
+    │                                targeting Ubuntu 24.04 or lower
     ├── README.md                  - Frontpage for your charmhub.io/charm/
-    ├── requirements.txt           - Python dependencies for your charm, with Ops,
-    │                                created for 12-factor app profiles only
+    ├── requirements.txt           - Python dependencies for 12-factor app profiles
```
jahn-junior: Optional:
```suggestion
    ├── requirements.txt           - Python dependencies for 12-factor app charms
```

docs/reference/files/requirements-txt-file.rst
```diff
@@ -4,14 +4,13 @@
...
 
-Charmcraft doesn't create this file for the Kubernetes and machine profiles. They use a
-:ref:`pyproject-toml-file` instead.
-
-When a charm is initialized with a 12-factor app profile, Charmcraft creates this file
-and pre-populates it with :external+ops:doc:`Ops <index>`. Any dependencies specified
-here are bundled with the charm when it's built with
+This file is pre-populated with :external+ops:doc:`Ops <index>`. Any dependencies
+specified here are bundled with the charm when it's built with
 :ref:`charmcraft pack <ref_commands_pack>`.
```
jahn-junior: Optional:
```suggestion
specified here are bundled with the charm when it's packed.
```
Unrelated to your changes, but this link feels unnecessary to me.

docs/reference/files/tox-ini-file.rst
```diff
@@ -27,19 +27,18 @@ Charmcraft creates this file and defines the following tox environments:
 
 To run the commands in an environment, use ``tox -e <env-name>``.
 
-For the Kubernetes and machine profiles, ``tox.ini`` requires the
-`tox-uv <https://github.com/tox-dev/tox-uv>`_ plugin. First make sure that
-`uv <https://docs.astral.sh/uv/>`_ is installed on the current host. Then use uv to
-install tox and tox-uv:
+``tox.ini`` requires the `tox-uv <https://github.com/tox-dev/tox-uv>`_ plugin. First
+make sure that `uv <https://docs.astral.sh/uv/>`_ is installed on the current host. Then
+use uv to install tox and tox-uv:
```
jahn-junior: Various optional nits:
```suggestion
The ``tox.ini`` file requires the `tox-uv <https://github.com/tox-dev/tox-uv>`__ plugin. First,
make sure that `uv <https://docs.astral.sh/uv/>`__ is installed on the current host. Then,
install tox and tox-uv with uv:
```

docs/reference/files/tox-ini-file.rst
```diff
@@ -27,19 +27,18 @@ Charmcraft creates this file and defines the following tox environments:
...
 Configuration of testing and linting tools is specified in the
-:ref:`pyproject-toml-file`. Dependencies are also specified in ``pyproject.toml``. If
-you manually modify the dependencies in ``pyproject.toml``, you'll need to update the
-:ref:`uv-lock-file` before using tox.
+:ref:`pyproject-toml-file`. Dependencies are also specified in ``pyproject.toml``.
+You'll need to create the :ref:`uv-lock-file` before using tox.
 
-For 12-factor app profiles, ``tox.ini`` doesn't require any plugins. Dependencies are
-specified in the :ref:`requirements-txt-file` instead of ``pyproject.toml``.
+For 12-factor app profiles targeting Ubuntu 24.04 LTS or lower: ``tox.ini`` doesn't
```
jahn-junior: Optional:
```suggestion
For 12-factor app charms targeting Ubuntu 24.04 LTS or lower, ``tox.ini`` doesn't
```

docs/reference/files/uv-lock-file.rst
```diff
@@ -12,17 +12,16 @@ your charm's dependencies and your development dependencies.
     `uv | The lockfile
     <https://docs.astral.sh/uv/concepts/projects/layout/#the-lockfile>`_
 
-This file is required if your charm uses the :ref:`craft_parts_uv_plugin`. It's also
-required if you use tox with tox-uv. For more information, see :ref:`tox-ini-file`.
+Charmcraft doesn't create this file. This file is required if your charm uses the
+:ref:`craft_parts_uv_plugin`. It's also required by the tox-uv plugin. For more
```
jahn-junior: Optional:
```suggestion
Charmcraft doesn't create this file. This file is required by the tox-uv plugin and
charms with environments managed by the :ref:`craft_parts_uv_plugin`. For more
```

docs/reference/files/uv-lock-file.rst
```diff
@@ -12,17 +12,16 @@ your charm's dependencies and your development dependencies.
...
 
-When a charm is initialized with the Kubernetes or machine profile, Charmcraft creates
-this file.
-
-If you change your charm's dependencies, use ``uv lock`` to update this file.
-Alternatively, you can update this file by using ``uv add`` and ``uv remove`` to manage
-dependencies in the :ref:`pyproject-toml-file`.
+Use ``uv lock`` to create this file and update it if you change your charm's
+dependencies. Alternatively, you can update this file by using ``uv add`` and
+``uv remove`` to manage dependencies in the :ref:`pyproject-toml-file`.
```
jahn-junior: Optional:
```suggestion
To create or update this file, run the ``uv lock`` command. Alternatively,
you can update this file by running ``uv add`` or ``uv remove`` to
manage dependencies in the :ref:`pyproject-toml-file`.
```

docs/reference/files/uv-lock-file.rst
```diff
@@ -12,17 +12,16 @@ your charm's dependencies and your development dependencies.
...
 
 You shouldn't manually edit this file.
 
-For the 12-factor app profiles, this file isn't necessary. Dependencies are specified
-in the :ref:`requirements-txt-file` instead of ``pyproject.toml``.
+For 12-factor app profiles targeting Ubuntu 24.04 LTS or lower: This file isn't
```
jahn-junior: Optional:
```suggestion
For 12-factor app charms targeting Ubuntu 24.04 LTS or lower: This file isn't
```

jahn-junior (approved):
Thanks, @dwilding! I left a bunch of minor wording tweaks and one minor error correction, but none of them are blocking.

Feel free to accept, ignore, or tweak these suggestions as you see fit.

dwilding:
Thanks for the reviews! I didn't have time today to finish working through the feedback. I'll come back to this after my break (from today until Aug 10).

docs/howto/manage-extensions.rst
jahn-junior: This was happening all over the Charmcraft docs, so I went ahead and fixed it in https://github.com/canonical/charmcraft/pull/2810.

lengau:
@dwilding just ping me when this is ready to merge :-)

charmcraft/templates/init-kubernetes/pyproject.toml.j2
dwilding: If we leave `ops[testing]` unconstrained, using `uv add` to change the `ops` version will set the corresponding version of `ops-scenario` in the lockfile. So you're right. I've switched to this:

```
"ops[testing]",  # uv will use the correct version based on the 'ops' dependency.
```

charmcraft/templates/init-kubernetes/pyproject.toml.j2
dwilding: Discussed internally. We'll use `~=`. I've also bumped to version 0.16, which introduces a lot more rules (and checked that the profiles don't need reformatting).

tests/integration/commands/test_init.py
dwilding: That's a good point, thanks. How about this structure to make the intent more obvious:

```py
assert "Run 'uv lock'" in output
# uv.lock is mentioned in the instructions but isn't listed as a created file.
assert "including uv.lock" in output
assert "uv.lock" not in output.replace("including uv.lock", "")
```

tests/spread/hooks/configure/task.yaml
dwilding: I don't see pushd/popd used to change directories at the very end of a `prepare` block. The existing convention seems to be to use `cd` - see `tests/spread/smoketests/destructive-mode/task.yaml` for example. I'll stick with this style.

charmcraft/application/commands/init.py
dwilding: This part of the output is shown after a list of created files, which mentions both `charmcraft.yaml` and `pyproject.toml`. So I prefer the version without "the ... files"

docs/reference/files/pyproject-toml-file.rst
dwilding: I see what you mean - thanks for raising this. I also see that "12-factor app charm" is the established wording in the Charmcraft docs. I've switched to this wording throughout.

docs/reference/files/pyproject-toml-file.rst
dwilding: I don't like "create and update" as much, as it could be read as meaning that you only need to create the file if you change the dependencies. But you always need to create it after running `charmcraft init`. I've changed the sentence to this:

```
You'll need to create the :ref:`uv-lock-file` and update it if you later change your charm's dependencies.
```

dwilding:
Thanks again for the feedback @jahn-junior & @tonyandrewmeyer! I've worked through and made changes for almost all suggestions. I'll re-request review to make sure you've had a chance to look at the latest version.

docs/reference/files/pyproject-toml-file.rst
jahn-junior: Yours is much better. Thanks!

jahn-junior (approved):
LGTM. Thanks again!

tonyandrewmeyer (approved):
Thanks for the adjustments! Looks good to me :)

dwilding:
Ready to merge @lengau

PR #2760, opened 2026-07-07
feat!: migrate from candid to Ubuntu One auth

mr-cal:
Switches from candid to Ubuntu One for authentication.

(CHARMCRAFT-667)

---


pyproject.toml
```diff
@@ -10,7 +10,7 @@ dependencies = [
     "craft-parts~=2.20",
     "craft-providers>=3.4.0,<4.0",
     "craft-platforms~=0.10",
-    "craft-store~=3.2",
+    "craft-store @ git+https://github.com/canonical/craft-store@main",
```
mr-cal: Reminder to bump before landing.

charmcraft/services/store.py
```diff
@@ -82,58 +85,78 @@ def _base_url(self) -> str:
...
-    ) -> str:
+    ) -> None:
         """Login to the store."""
-        description = self._get_description(description)
+        # packages need to be serialized
+        package_dicts = (
+            [{"type": p.package_type, "name": p.package_name} for p in packages]
+            if packages is not None
+            else None
+        )
```
mr-cal: This seems like something that should be fixed in craft-store, wdyt @lowe?

spread.yaml
```diff
@@ -305,7 +305,7 @@ suites:
 
       # Ensure that the reused charms are registered if necessary.
       if ! charmcraft status "${CHARM_DEFAULT_NAME}"; then
-        charmcraft register $CHARM_DEFAULT_NAME
+        charmcraft register "${CHARM_DEFAULT_NAME}" || charmcraft status "${CHARM_DEFAULT_NAME}"
```
lengau, line 311: I'm confused about this change.

charmcraft/services/store.py
lengau: I think we should change the `login` API here to match the craft-store API. We no longer want to use that `Package` type as we do more with the `PublisherGW` class.

tests/spread/store/ubuntu-sso-login/task.yaml
```diff
@@ -0,0 +1,18 @@
+summary: validate the Ubuntu One SSO export login flow
+
+prepare: |
+  tests.pkgs install expect
```
lengau: TIL about this package. I think it's ok if we install it in the general `prepare` so we can use it elsewhere!

mr-cal: Added!

spread.yaml
mr-cal: This is to avoid a race condition I was hitting. My guesses are either
- that the openstack runners are more likely to get this same line at the same time or
- the staging store was being slow last week and was taking a long time to register.

I added a code comment to explain the details of the race.

lengau (approved):
Thanks!

pyproject.toml
mr-cal: Bumped

tigarmo (approved):
LGTM

PR #2763, opened 2026-07-08
fix(tests/spread): fix reactive spread tests

lengau:
This PR fixes the failing reactive spread tests by making several structural and configuration changes:

1. **Test Splitting**: The single `reactive` test has been split into two distinct tests to avoid complex conditionals and cross-talk:
   - `reactive-legacy`: Runs on Ubuntu 20.04 and 22.04, testing the `2.x/stable` and `stable` charm channels. It dynamically sets its base channel to match the host system's base.
   - `reactive-resolute`: A standalone test specifically for Ubuntu 26.04, configured to strictly test the `charm/latest/stable` channel.

2. **Dependency Resolution Fix**: Explicitly adds `--binary-wheels` to the `charm-build` arguments in `charmcraft.yaml`. This prevents `charm-build` from defaulting to source builds (which fail due to PEP-517 isolated environments downloading newer `setuptools` versions that lack `pkg_resources`). Using binary wheels avoids these source build failures and speeds up the tests.

3. **Cleanup**: Removes `--upgrade-buildvenv-core-deps` and `--ignore-requires-python` flags from all `charmcraft.yaml` templates, as they were contributing to dependency resolution failures on older versions and are no longer necessary with binary wheels.


tests/spread/smoketests/reactive-resolute/reactivecharm/reactive/reactivecharm.py
lengau, line 1: These files are just copying the stuff from the old reactive task.

tigarmo (approved):

bepri (approved):

PR #2773, opened 2026-07-10
ci: ignore unfixable OSVs

lengau:
Replaces #2765 with the same fix pattern as starbase #573.


mr-cal (approved):

.github/workflows/policy.yaml
```diff
@@ -41,5 +41,10 @@ jobs:
       #    contain vulnerable versions.
       # 3. Docs contain requirements.txt files that don't specify versions.
       requirements-find-args: '! -name requirements-noble.txt ! -path "./tests/spread/*" ! -path "./docs/**"'
```
bepri: This can be removed

bepri (approved):

.github/workflows/policy.yaml
lengau:
```suggestion
```

PR #2774, opened 2026-07-11
test: speed up spread test runtime by removing redundant package setup

lengau:
## Summary
- remove a redundant apt update from top-level spread prepare
- install snapd and unzip together in one global tests.pkgs call
- remove per-task unzip installs from spread task.yaml files where unzip is already guaranteed globally

## Impact
This removes repeated package-index refreshes in spread runs, reducing runtime while preserving test coverage and assertions.


tests/spread/dependencies/strict-dependencies/task.yaml
```diff
@@ -1,8 +1,6 @@
 summary: ensure that a charm with strict dependencies can build.
 
 prepare: |
```
lengau:
```suggestion
```

PR #2784, opened 2026-07-14
docs: use microk8s 1.34-strict instead of 1.31-strict

javierdelapuente:
Use microk8s 1.34-strict/stable instead 1.31-strict/stable.

Microk8s 1.31 is EOL, and it does not work correctly in Ubuntu 26.04.

---

### DIFF
--- a/docs/reuse/tutorial/setup_edge.rst
+++ b/docs/reuse/tutorial/setup_edge.rst
@@ -54,7 +54,7 @@ Initialize LXD:
     lxd init --auto
 
 MicroK8s is required to deploy the |12FactorApp| application on Kubernetes.
-Let's install MicroK8s using the ``1.34-strict/stable`` track, add the current
+Let's install MicroK8s using the ``1.34-strict/stable`` channel, add the current
 user to the group, and activate the changes:
 
 .. code-block:: text
@@ -86,7 +86,7 @@ If successful, the terminal will output ``microk8s is running``
 along with a list of enabled and disabled add-ons.
 
 Juju is required to deploy the |12FactorApp| application.
-We'll install Juju using the ``3.6/stable`` track. Since the snap is
+We'll install Juju using the ``3.6/stable`` channel. Since the snap is
 sandboxed, we'll also manually create a directory to contain
 its files. Once Juju is ready, we initialize it by bootstrapping a
 development controller:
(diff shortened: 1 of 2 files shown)


javierdelapuente:
@lengau can you review this PR? microk8s 1.31 is not working in Ubuntu 26.04 (see for example https://discourse.charmhub.io/t/charmcraft-spring-boot-tutorial/20655/3)

lengau (approved):
Thanks!

docs/reuse/tutorial/setup_edge.rst
```diff
@@ -54,12 +54,12 @@ Initialize LXD:
     lxd init --auto
 
 MicroK8s is required to deploy the |12FactorApp| application on Kubernetes.
-Let's install MicroK8s using the ``1.31-strict/stable`` track, add the current
+Let's install MicroK8s using the ``1.34-strict/stable`` channel, add the current
```
lengau, line 57: I don't mind what's used, but this is more technically correct than what we had before. Alternatively, we could say "the `1.34-strict` track"

erinecon: No strong opinion here

erinecon (approved):
Thanks for updating the documentation! Do we need to update anything in the Spread tests? I looked quickly and don't think we need to change anything, but wanted to also ask :)

javierdelapuente:
> Thanks for updating the documentation! Do we need to update anything in the Spread tests? I looked quickly and don't think we need to change anything, but wanted to also ask :)

Spread uses 1.32 (not strict) for every tutorial.... it will be different anyway if we do not test it with the same ubuntu bases. I am not sure if we should update it now, or better do all at once for Canonical K8s (same systems, same snaps...)...

docs/reuse/tutorial/setup_stable.rst
```diff
@@ -49,12 +49,12 @@ Initialize LXD:
...
 MicroK8s is required to deploy the |12FactorApp| application on Kubernetes.
-Let's install MicroK8s using the ``1.31-strict/stable`` track, add the current
+Let's install MicroK8s using the ``1.34-strict/stable`` track, add the current
 user to the group, and activate the changes:
 
 .. code-block:: text
 
-    sudo snap install microk8s --channel 1.31-strict/stable
+    sudo snap install microk8s --channel 1.34-strict/stable
     sudo adduser $USER snap_microk8s
```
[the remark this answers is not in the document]
lengau: Amusing that it gets the channel/track distinction correct and then messes up the suggestion so badly.

PR #2808, opened 2026-07-23
docs: encourage use of dependency groups in uv migration guide

astrojuanlu:
Charms are generally not meant to be [packaged](https://discourse.ubuntu.com/t/how-does-python-packaging-really-work/84670?u=astrojuanlu). As in: uploaded to PyPI or served for other Python projects to depend on.

uv chose to use PEP 518 `[project]` table even for source trees that don't intend to be packaged. However, [the uv team admits](https://github.com/astral-sh/uv/issues/8311#issuecomment-2420772882) that this is non-standard according to [PEP 517](https://peps.python.org/pep-0517/#:~:text=If%20the%20pyproject.toml%20file%20is%20absent%2C%20or%20the%20build%2Dbackend%20key%20is%20missing%2C%20the%20source%20tree%20is%20not%20using%20this%20specification):

> If [...] the `build-backend` key is missing [from `pyproject.toml`], the source tree is not using this specification, and tools should revert to the legacy behaviour of running `setup.py` (either directly, or by implicitly invoking the `setuptools.build_meta:__legacy__` backend).

We were bitten by this in the Charmed MySQL team: we initially moved the dependencies to the `[project.dependencies]` key https://github.com/canonical/mysql-k8s-operator/pull/680, but because the team was still using Poetry back then, the tool would try to package the charm, which led to the creation of setuptools `*.egg-info` folders and other minor annoyances.

This proposes a change in the uv migration guide so that only standards-compliant `pyproject.toml` files are declared. This requires using dependency groups for all dependencies and using some uv-specific tools for declaring the minimum Python version. The upside is that users don't need to specify dummy project names and versions.

Open for discussion.

---

### DIFF
--- a/docs/howto/migrate-plugins/charm-to-uv.rst
+++ b/docs/howto/migrate-plugins/charm-to-uv.rst
@@ -38,27 +38,48 @@ List the charm's dependencies
 -----------------------------
 
 If the project directory doesn't already contain a :ref:`pyproject-toml-file`, create
-one. Next, list the charm's dependencies in the file's ``dependencies`` key.
+one. Next, list the charm's dependencies in a
+`dependency group <dependency groups_>`_ called ``main``.
 
 .. code-block:: toml
     :caption: pyproject.toml
-    :emphasize-lines: 6-9
 
-    [project]
-    name = "my-charm"
-    version = "0.0.1"
-    requires-python = ">=3.10"
-
-    # Dependencies of the charm code.
-    dependencies = [
+    [dependency-groups]
+    main = [
         "ops>=3,<4",
     ]
+
+    [tool.uv]
+    default-groups = ["main"]
+
+    [tool.uv.dependency-groups]
+    main = {requires-python = ">=3.10"}
 
 You can also add dependencies to your project's ``pyproject.toml`` file by running:
 
 .. code-block:: bash
 
-    uv add <dependency>
+    uv add --group main <dependency>
+
+.. warning::
+    Dependencies can also be placed in the ``dependencies`` key,
+    but the presence of the ``[project]`` table without a ``[build-system]``
+    is non-standard and will confuse workflow tools other than uv.
+    Use at your own risk.
+
+    .. code-block:: toml
+        :caption: pyproject.toml
+        :emphasize-lines: 6-9
+
+        [project]
+        name = "my-charm"
+        version = "0.0.1"
+        requires-python = ">=3.10"
+
+        # Dependencies of the charm code.
+        dependencies = [
+            "ops>=3,<4",
+        ]
 
 List charm library dependencies
 -------------------------------
@@ -78,27 +99,16 @@ running the following command at the root of the charm project:
 
     find lib -name "*.py" -exec awk '/PYDEPS = \[/,/\]/' {} +
 
-Next, in ``pyproject.toml``, list them in the ``dependencies`` key.
-
-.. code-block:: toml
-    :caption: pyproject.toml
-    :emphasize-lines: 4-6
-
-    # Dependencies of the charm code and PYDEPS from libraries.
-    dependencies = [
-        "ops>=3,<4",
-        "cosl",
-        "pydantic",
-        "cryptography",
-    ]
-
-Alternatively, you could list the library dependencies in a
+Next, in ``pyproject.toml``, list them in a
 `dependency group <dependency groups_>`_ called ``charmlibs-pydeps``.
 
 .. code-block:: toml
     :caption: pyproject.toml
 
     [dependency-groups]
+    main = [
+        "ops>=3,<4",
+    ]
     # PYDEPS from libraries that the charm uses.
     charmlibs-pydeps = [
         "cosl",
@@ -137,7 +147,7 @@ Add dependency groups
 ---------------------
 
 If the charm has dependency groups that should be included when creating the virtual
-environment, such as one for charm libraries, the
+environment, the
 :ref:`uv plugin's <craft_parts_uv_plugin>` ``uv-groups`` key can be set to include them:
 
 .. code-block:: yaml
@@ -151,10 +161,8 @@ environment, such as one for charm libraries, the
         build-snaps:
           - astral-uv
         uv-groups:
+          - main
           - charmlibs-pydeps
-
-Likewise, optional dependencies under the ``pyproject.toml`` key
-``project.optional-dependencies`` can be added with the ``uv-extras`` key.
 
 .. _howto-migrate-to-uv-include-extra-files:
 


astrojuanlu:
Rendered result https://canonical-charmcraft--2808.com.readthedocs.build/2808/howto/migrate-plugins/charm-to-uv/#list-the-charm-s-dependencies

tonyandrewmeyer:
If this is done, then the profiles should change accordingly as well. However, I can't support this. Charms definitely *are* packaged, into `.charm` files. The dependencies are specifying the dependencies for that packaged charm. Packaging does not [mean](https://packaging.python.org/en/latest/glossary/#term-Distribution-Package) "for PyPI" or ["for other Python packages"](https://packaging.python.org/en/latest/glossary/#term-Project), or anything specific like "wheel" or "sdist".

Charmcraft definitely should *not* "revert to using setup.py".

[Dependency groups](https://packaging.python.org/en/latest/specifications/dependency-groups/) are explicitly *not* for things included when the project is built (by charmcraft in our case), and are suitable for "internal development use-cases".

Ideally, a native-Python charmcraft would be implemented as a build backend, and you'd have the version from pyproject.toml, various other metadata there, and so on. StarCraft subsumes all, of course, so we don't live in that world and ignore some fields meant to always be used for packaging. However, we shouldn't fix poor packaging behaviour by adding more.

The problem here is poetry. Let's not have it make the uv plugin worse.

carlcsaposs-canonical:
> The problem here is poetry.

I think the problem may have been how poetry was used. Poetry has a "non-package mode" (https://python-poetry.org/docs/basic-usage/#operating-modes) for exactly this purpose and, just like uv, uses project.dependencies for non-package projects

astrojuanlu:
>> The problem here is poetry.
>
> I think the problem may have been how poetry was used. 

It is not correct that the problem here is Poetry. Any standards-compliant tool will behave in the same way.

Example with Hatch. Setup:
```
$ mkdir uv-hatch-test && cd uv-hatch-test
$ uv init --app --no-package --python (which python3)
Initialized project `uv-hatch-test`
$ uv add "ops>=3,<4"
Using CPython 3.12.3 interpreter at: /usr/bin/python3.12
Creating virtual environment at: .venv
Resolved 6 packages in 3ms
Installed 5 packages in 3ms
 + opentelemetry-api==1.44.0
 + ops==3.8.0
 + pyyaml==6.0.3
 + typing-extensions==4.16.0
 + websocket-client==1.9.0
$ cat pyproject.toml
[project]
name = "uv-hatch-test"
version = "0.1.0"
description = "Add your description here"
readme = "README.md"
requires-python = ">=3.12"
dependencies = [
    "ops>=3,<4",
]
```

Running the script with `uv run`:
```
$ uv run main.py
Hello from uv-hatch-test!
```

Running the script with `hatch run`:
```
$ uvx hatch run python main.py
Hello from uv-hatch-test!
$ ls -la
total 52
drwxrwxr-x 5 ubuntu ubuntu  4096 Jul 23 12:08 ./
drwxrwxr-x 3 ubuntu ubuntu  4096 Jul 23 12:08 ../
drwxrwxr-x 7 ubuntu ubuntu  4096 Jul 23 12:08 .git/
-rw-rw-r-- 1 ubuntu ubuntu   109 Jul 23 12:08 .gitignore
-rw-rw-r-- 1 ubuntu ubuntu     5 Jul 23 12:08 .python-version
drwxrwxr-x 4 ubuntu ubuntu  4096 Jul 23 12:08 .venv/
-rw-rw-r-- 1 ubuntu ubuntu     0 Jul 23 12:08 README.md
-rw-rw-r-- 1 ubuntu ubuntu    91 Jul 23 12:08 main.py
-rw-rw-r-- 1 ubuntu ubuntu   177 Jul 23 12:08 pyproject.toml
-rw-rw-r-- 1 ubuntu ubuntu 15957 Jul 23 12:08 uv.lock
drwxrwxr-x 2 ubuntu ubuntu  4096 Jul 23 12:08 uv_hatch_test.egg-info/  # <---- The setuptools legacy backend was invoked
```

And, again, this has been flagged by me to the uv team in two threads, and they acknowledge that their use of `[project]` for non-packaged projects is a convenience one (links above).

> Charms definitely are packaged, into .charm files.

The word "packaged" has several meanings here. I was referring to the notion of publishing/distributing a distribution package, which has very specific boundaries.

carlcsaposs-canonical:
> Any standards-compliant tool will behave in the same way.

does this behavior change when `project.dependencies` is not used and a dependency group is used instead?

from my reading of the uv thread, it doesn't make a difference since the build-backend key is still missing

tonyandrewmeyer:
> It is not correct that the problem here is Poetry. Any standards-compliant tool will behave in the same way.

Fair, it would have more correct to say the problem is using tools in inappropriate ways, first poetry and then hatch.

> Example with Hatch. Setup:

These are charms, you should setup with `charmcraft init`. Suggesting putting `tool.uv.package = false` in the profiles would be reasonable to consider, since we don't intend to have uv build or install the charm.

> Running the script with `uv run`:

Can you explain why you are trying to run the charm with uv?

> Running the script with `hatch run`:

Or with hatch?

> And, again, this has been flagged by me to the uv team in two threads, and they acknowledge that their use of `[project]` for non-packaged projects is a convenience one (links above).

They say "we're not really acting as a build frontend in this context — we're acting as a project management tool". That's exactly the case with how charms, with the uv plugin, are using uv: as a project management tool, not a build frontend. That's not because there isn't an archive being turned into a distribution package, it's because we have a different build system (charmcraft), which is not particularly compatible with the build-backend requirements.

> The word "packaged" has several meanings here. I was referring to the notion of publishing/distributing a distribution package, which has very specific boundaries.

I linked to the [definition of distribution package](https://packaging.python.org/en/latest/glossary/#term-Distribution-Package) previously. A charm is definitely a distribution package. Charmcraft is the mechanism for publishing those packages, and Charmhub the (typical) vehicle for distributing them.

Did you intend this PR to be *only* for discussion (I think an issue or, more likely, spec would be a better vehicle)? As-is, it cannot be accepted, because the uv plugin will fail to find the dependencies (it more or less runs `uv sync --no-dev --no-editable --reinstall --no-install-project`), and all Ops charms will break. You'd need to change the plugin to install your "main" group (and not other groups, which, following the recommended use-case, are not intended for inclusion in the distribution package's archive). The charmcraft versioning system always confuses me, but I think that would mean this could only be done for 28.04 forwards, or in a new plugin.

astrojuanlu:
> Did you intend this PR to be only for discussion (I think an issue or, more likely, spec would be a better vehicle)? As-is, it cannot be accepted

I definitely expected some back and forth (I did this in a haste on the web editor, without testing this thoroughly). What I didn't expect was to disagree at a fundamental level about what a distribution package is. That being the case, definitely a spec would be the appropriate venue, so I'll close this PR. This is not a hill I'm willing to die on though, so it's unlikely that I pursue a spec.

PR #2814, opened 2026-07-27
feat(profile): add secrets handling to kubernetes and machine init profiles

tonyandrewmeyer:
Extends both the `kubernetes` and `machine` profiles so that `charmcraft init` scaffolds a charm that handles Juju secrets from both sides, with unit and integration tests for it.

The profile is a strong signal to human and agentic charmers for best practices. Secrets are a good candidate for that signal because they are easy to get half-right: reading an observed secret once on `config-changed` looks correct until the operator adds a new revision and nothing happens, and rotating an owned secret looks correct until revisions pile up because nothing removes them. Both failures are silent, so the profile handles the full event set rather than the two events that are enough for a demo.

A charm also meets secrets in two roles that need different code — it observes secrets an operator grants it, and it owns secrets it generates for its workload. A charm author who has only seen one usually writes the other incorrectly.

### Changes

Both profiles in `charmcraft.yaml.j2`:

Adds an `api-token` config option of `type: secret`, with a description giving the operator the three commands they need (`juju add-secret`, `juju grant-secret`, `juju config`).

In both profiles, the observed secret:

`config-changed` and `secret-changed` share one `_configure_api_token()`, which fetches with `model.get_secret(id=..., label=API_TOKEN_LABEL)` and reads with `get_content(refresh=True)`. Attaching the label is what lets the `secret-changed` handler recognise the secret, since the event carries the label rather than the config value. A secret that isn't granted, or has gone away, leaves the unit blocked.

In both profiles, the app-managed secret:

The leader creates a workload password on `start` (machine) or `pebble-ready` (kubernetes), with monthly rotation and a 90 day expiry; creation is looked up by label first, so it's idempotent. `secret-rotate` and `secret-expired` both add a new revision — expiry doesn't remove the expiring revision, because that's the latest one and other units may still be reading it. `secret-remove` does the cleanup, once Juju reports that no unit is tracking a revision.

---

- [ ] ~In documents I changed, I [added a meta description](https://canonical-starflow.readthedocs-hosted.com/how-to/add-a-page-meta-description/) if one was missing.~
- [x] I've updated the relevant release notes.


tonyandrewmeyer:
@dwilding would you reviewing this?

charmcraft/templates/init-kubernetes/charmcraft.yaml.j2
```diff
@@ -48,6 +48,17 @@ config:
         Acceptable values are: "info", "debug", "warning", "error" and "critical"
       default: "info"
       type: string
+    # A secret config option for a user-provided API token.
+    # See https://documentation.ubuntu.com/juju/3.6/howto/manage-secrets/
```
jahn-junior:
```suggestion
    # See https://canonical.com/juju/docs/juju-cli/3.6/howto/manage-secrets/
```

docs/release-notes/charmcraft-4.5.rst
```diff
@@ -54,6 +54,22 @@ Minor features
 Charmcraft 4.5 brings the following minor changes.
 
 
+Secrets handling in the machine and Kubernetes profiles
```
jahn-junior:
```suggestion
Secret handling in the machine and Kubernetes profiles
```

docs/release-notes/charmcraft-4.5.rst
```diff
@@ -54,6 +54,22 @@ Minor features
...
+
+Charms created with the ``machine`` and ``kubernetes`` profiles now demonstrate both
+sides of Juju secrets.
+
+For a user-provided secret, the charm declares an ``api-token`` config option of type
+``secret``, resolves it, and re-reads it when the operator adds a new revision.
+
+For an app-managed secret, the leader creates a workload password with a rotation
+policy and an expiry, replaces it when Juju asks for rotation or reports expiry, and
+removes old revisions once no unit is using them.
```
jahn-junior:
```suggestion
removes unused revisions.
```

docs/release-notes/charmcraft-4.5.rst
```diff
@@ -54,6 +54,22 @@ Minor features
...
+
+Both profiles scaffold unit and integration tests for this behaviour.
```
jahn-junior:
```suggestion
Both profiles scaffold unit and integration tests for the added behavior.
```

jahn-junior (approved):
Release notes and template strings LGTM! I left a few minor suggestions, but nothing blocking. Thanks!

charmcraft/templates/init-kubernetes/charmcraft.yaml.j2
```diff
@@ -48,6 +48,17 @@ config:
...
+    # A secret config option for a user-provided API token.
+    # See https://canonical.com/juju/docs/juju-cli/3.6/howto/manage-secrets/
+    api-token:
+      description: |
+        URI of a Juju user secret containing the API token for the workload.
+        The secret must have a 'token' key.
+
+        Create the secret: juju add-secret my-api-token token=<value>
+        Grant access:      juju grant-secret <id> {{ name }}
+        Set this option:   juju config {{ name }} api-token=<id>
```
dwilding: A few suggestions that I think will help people understand the relationship between the config option and the secret. Same suggestions for machine profile.

```suggestion
    # A config option that points to a user-provided secret containing an API token.
    # See https://canonical.com/juju/docs/juju-cli/3.6/howto/manage-secrets/
    api-token:
      description: |
        URI of a Juju user secret containing the API token for the workload.
        The secret must have a 'token' key.

        Create the secret: juju add-secret my-api-token token=<value>
                           (returns a URI)
        Grant access:      juju grant-secret <uri> {{ name }}
        Set this option:   juju config {{ name }} api-token=<uri>
```

charmcraft/templates/init-kubernetes/src/charm.py.j2
```diff
@@ -49,8 +58,75 @@ class {{ class_name }}(ops.CharmBase):
...
+            secret = self.model.get_secret(id=api_token_id, label=API_TOKEN_LABEL)
+            # 'refresh' tells Juju to start tracking the latest revision.
+            content = secret.get_content(refresh=True)
+        except ops.SecretNotFoundError:
+            self.unit.status = ops.BlockedStatus("api-token secret not found or not granted")
+            return
+        token = content.get("token", "")
+        # Use 'token' to configure your workload, for example:
+        # {{ workload_module }}.configure(api_token=token)
+        logger.info("api-token resolved (%d chars); reconfiguring workload", len(token))
```
dwilding: In the example, we reconfigure by calling a function in the workload module. If a replan is needed too, this could be costly. So do you think we should set maintenance status somewhere before changing the workload configuration? Should the logging output also come before the `.configure()` line?

charmcraft/templates/init-kubernetes/src/charm.py.j2
```diff
@@ -15,6 +17,8 @@ import {{ workload_module }}
 logger = logging.getLogger(__name__)
 
 SERVICE_NAME = "some-service"  # Name of Pebble service that runs in the workload container.
+API_TOKEN_LABEL = "api-token"  # Label for the user-provided secret named in charm config.
```
dwilding: Minor: Someone might think the label `api-token` needs to match the name of the config option as defined in `charmcraft.yaml`. But making the name deliberately different might not make things clearer overall... So maybe the best practice is just to use the same name as the config option, in which case leave things as-is?

charmcraft/templates/init-kubernetes/src/charm.py.j2
```diff
@@ -15,6 +17,8 @@ import {{ workload_module }}
...
+APP_SECRET_LABEL = "workload-password"  # Label for the app-managed workload secret.
```
dwilding: Purely from reading the charm code, I think it's a bit hard to understand a couple of things:

- The relationship between API token and the app-managed secret. When `API_TOKEN_LABEL` and `APP_SECRET_LABEL` are introduced, my starting assumption is that they're connected somehow. Then later I realise they aren't connected.

- How the app-managed secret is expected to be used, and how non-leader units are expected to access it.

I think it would be really helpful to visually separate the label constants from `SERVICE_NAME`, then add a block comment before the label constants. The block comment could summarise what we're demonstrating with the secrets and confim that they are demoing different things; they aren't connected.

Also would it be helpful to add a method `_configure_workload_password()` and call it from a suitable point? And extend `_on_secret_changed()` to make sure that non-leader units can get hold of the app-managed secret?

dwilding (commented):
A few initial comments. Need to pause for today.

charmcraft/templates/init-kubernetes/src/charm.py.j2
tonyandrewmeyer: Sure, I've made those changes. I think in practice you'd want to evaluate whether the update is likely to be slow enough to warrant maintenance status, but defaulting to it seems reasonable. I did the same change in the machine profile.

charmcraft/templates/init-kubernetes/src/charm.py.j2
tonyandrewmeyer: I've tweaked the comment instead, what do you think?

charmcraft/templates/init-kubernetes/src/charm.py.j2
tonyandrewmeyer: I've done the separation, and added an example of using the workload password.

In terms of getting the password to non-leaders, I'm not sure. The proper way to do this is to pass the id over a peer relation and that means quite a lot more boilerplate. I'm already a little nervous about how much this (and the action one) adds, even though I am convinced it's valuable and the right thing to do.

What about a comment that explains what to do if the other units need it with a doc link? It[s not a complete example but would point people to the right place.

charmcraft/templates/init-kubernetes/src/charm.py.j2
dwilding: Looks good, thanks

charmcraft/templates/init-kubernetes/src/charm.py.j2
```diff
@@ -16,13 +18,25 @@ logger = logging.getLogger(__name__)
 
 SERVICE_NAME = "some-service"  # Name of Pebble service that runs in the workload container.
 
+# This charm demonstrates the two independent roles a charm plays with Juju secrets.
+# The two secrets below are unrelated: one is observed, the other is owned.
+#  - api-token: owned by the operator, granted to this charm, named in charm config.
+#  - workload-password: created and rotated by this charm for its own workload.
+API_TOKEN_LABEL = "api-token"  # Any label works; matching the config option name is clearest.
+APP_SECRET_LABEL = "workload-password"  # Label for the app-managed workload secret.
```
dwilding: The new comment is very helpful.

I think it would be OK to shorten the description of api-token - especially to remove the mention of "owned" after just saying "the other is owned" (which could be confusing).

Now we have the block comment, I think the comment after `APP_SECRET_LABEL` isn't really needed.

```suggestion
# This charm demonstrates the two independent roles a charm plays with Juju secrets.
# The two secrets below are unrelated: one is observed, the other is owned.
#  - api-token: granted by the user and referenced in a config option.
#  - workload-password: created and rotated by this charm for its own workload.
API_TOKEN_LABEL = "api-token"  # Any label works; matching the config option name is clearest.
APP_SECRET_LABEL = "workload-password"
```

charmcraft/templates/init-kubernetes/src/charm.py.j2
```diff
@@ -49,8 +63,91 @@ class {{ class_name }}(ops.CharmBase):
...
+        """Configure the workload with the app-managed password (leader only)."""
+        secret = self._ensure_app_secret()
+        # A secret's owner reads the latest revision with 'peek_content', not 'get_content'.
+        self._configure_workload(secret.peek_content()["password"])
+
+    def _configure_workload(self, password: str) -> None:
+        """Tell the workload which password it should accept."""
+        # Use 'password' to configure your workload, for example:
+        # {{ workload_module }}.configure(workload_password=password)
+        logger.info("workload password configured (%d chars)", len(password))
```
dwilding: I think it would be best to remove this method and inline the `.configure()` example into `_configure_workload_password`. Otherwise it's confusing why we call `_configure_workload` there but not in `_configure_api_token`.

Since `_configure_workload_password` doesn't touch status, I don't think we need to duplicate the "Reconfiguring the workload can be slow" comment or set maintenance status. But maybe we could add a comment that says something like:

> If this 'configure' call is slow, set maintenance status before the call.

charmcraft/templates/init-kubernetes/src/charm.py.j2
```diff
@@ -49,8 +63,91 @@ class {{ class_name }}(ops.CharmBase):
         version = {{ workload_module }}.get_version()
         if version is not None:
             self.unit.set_workload_version(version)
+        # Create the app-managed workload secret on first start (leader only).
+        if self.unit.is_leader():
+            self._configure_workload_password()
+        self.unit.status = ops.ActiveStatus()
+
+    def _on_config_changed(self, event: ops.ConfigChangedEvent):
+        """Handle config-changed event — the api-token config might name a different secret."""
```
dwilding: I don't get what "the api-token config might name a different secret" means.

charmcraft/templates/init-kubernetes/src/charm.py.j2
```diff
@@ -49,8 +63,91 @@ class {{ class_name }}(ops.CharmBase):
         version = {{ workload_module }}.get_version()
         if version is not None:
             self.unit.set_workload_version(version)
+        # Create the app-managed workload secret on first start (leader only).
+        if self.unit.is_leader():
+            self._configure_workload_password()
```
dwilding: How about extracting `self._ensure_app_secret()` and `secret.peek_content()` from `_configure_workload_password()`, and putting them here instead?

```suggestion
        # Create the app-managed workload secret on first start (leader only).
        if self.unit.is_leader():
            secret = self._ensure_app_secret()
            self._configure_workload_password()
            # A secret's owner reads the latest revision with 'peek_content', not 'get_content'.
            password = secret.peek_content()["password"]
            self._configure_workload_password(password)
        # A non-leader unit should get the secret ID from a peer relation,
        # then use 'get_content' to read the latest revision.
        # This charm doesn't demonstrate how to do that. For guidance, see
        # DOC LINK
```

Then `_configure_workload_password()` would look like:

```python
def _configure_workload_password(password: str) -> None:
    """Tell the workload which password it should accept."""
    # Use 'password' to configure your workload, for example:
    # {{ workload_module }}.configure(workload_password=password)
    # If this 'configure' call is slow, set maintenance status before the call.
    logger.info("workload password configured (%d chars)", len(password))
```

`_configure_workload()` goes away, per an earlier suggestion.

charmcraft/templates/init-kubernetes/src/charm.py.j2
```diff
@@ -49,8 +63,91 @@ class {{ class_name }}(ops.CharmBase):
...
+        if event.secret.label == APP_SECRET_LABEL:
+            self._set_new_password(event.secret)
+
+    def _set_new_password(self, secret: ops.Secret) -> None:
+        """Add a new revision of the app-managed workload password."""
+        password = secrets.token_hex(32)
+        # Don't remove the old revision here: other units might still be using it.
+        # Juju emits secret-remove once no unit is tracking it any more.
+        secret.set_content({"password": password})
+        self._configure_workload(password)
```
dwilding:
```suggestion
        self._configure_workload_password(password)
```

charmcraft/templates/init-kubernetes/src/charm.py.j2
dwilding: I think a comment plus some slight refactoring would help to give an idea of how non-leader units would work. I've tried to capture that with my suggestions - let me know if it's not clear what I have in mind.

lengau (approved):
LGTM once David approves and the conflicts are resolved.

charmcraft/templates/init-kubernetes/src/charm.py.j2
```diff
@@ -52,8 +66,92 @@ class {{ class_name }}(ops.CharmBase):
...
+            return
+        try:
+            # Attaching a label lets the secret-changed handler recognise this secret.
+            secret = self.model.get_secret(id=api_token_id, label=API_TOKEN_LABEL)
+            # 'refresh' tells Juju to start tracking the latest revision.
+            content = secret.get_content(refresh=True)
+        except ops.SecretNotFoundError:
+            self.unit.status = ops.BlockedStatus("api-token secret not found or not granted")
+            return
+        token = content.get("token", "")
```
dwilding: In `charmcraft.yaml` we say "The secret must have a 'token' key." But this point of the charm code doesn't enforce that. The API token will have length 0. It looks like this might have been intentional to reduce complexity? (It's also not covered by the unit tests)

charmcraft/templates/init-kubernetes/tests/integration/test_charm.py.j2
```diff
@@ -29,6 +29,39 @@ def test_deploy(charm: pathlib.Path, juju: jubilant.Juju):
...
+    # Act — add a new revision of the same secret. Juju notifies the charm with
+    # a secret-changed event, and the charm re-reads the content.
+    juju.update_secret(secret_uri, {"token": "a-much-longer-token-value"})
+    juju.wait(jubilant.all_active)
+
+    # Assert — the charm picked up the new revision, not the original one.
+    assert "api-token resolved (25 chars)" in juju.debug_log()
+
+    # Clean up: reset the config option so subsequent tests start clean.
+    juju.config("{{ name }}", reset=["api-token"])
```
dwilding, line 62: Very nitpicky... Resetting the config option doesn't reconfigure the workload to remove the API token. But in principle it might. Arguably it should - although I can see why we'd prefer to keep things simpler. So should we add a `juju.wait(jubilant.all_active)` here as a good practice? And/or a comment? Or is drawing attention to this going to make things more confusing?

charmcraft/templates/init-kubernetes/src/charm.py.j2
tonyandrewmeyer: I've reworded this, is it ok now?

charmcraft/templates/init-kubernetes/src/charm.py.j2
tonyandrewmeyer: Yeah, it was indeed to keep things simple. I've put in a blocking status instead.

charmcraft/templates/init-kubernetes/tests/integration/test_charm.py.j2
tonyandrewmeyer: Good catch on the wait. I've added `juju.wait(jubilant.all_active)` after the reset in both profiles. It's worth having regardless of the token-removal question: `juju.config(reset=...)` returns before the resulting config-changed has been processed, so without it the "subsequent tests start clean" comment isn't quite true.

On the comment, I've left it out. You're right that the charm should reconfigure the workload to drop the token, but a comment saying so raises a question the profile then doesn't answer, which I think lands on the "more confusing" side of your last sentence. (For what it's worth, the current behaviour on reset is that `config.get("api-token")` returns None, the isinstance narrowing returns early, and the unit stays active.)

dwilding (approved):
Latest changes look great, thanks! I confirmed that lint, unit, pack, and integration work for both charms 😁

tonyandrewmeyer:
@lengau this is ok to merge, if I understand correctly.

PR #2817, opened 2026-08-11
feat(profiles): use Ubuntu image for kubernetes integration tests

dwilding:
Currently, integration tests fail for a charm generated from the `kubernetes` profile, because `charmcraft.yaml` specifies a dummy OCI image.

This PR switches to the base [Ubuntu image](https://hub.docker.com/_/ubuntu), as a placeholder for the charmer's real image. The charm code needed to be adjusted too, so that the Pebble layer uses a placeholder command that really exists in the container.

I've verified that integration tests now pass.

---

This PR includes one small drive-by: In `charmcraft.yaml` for the `kubernetes` and `machine` profiles, we have a sample config option. The description mentions gunicorn, but nothing else in the charms is related to gunicorn. I'm making the description more generic.

---

- [ ] ~In documents I changed, I [added a meta description](https://canonical-starflow.readthedocs-hosted.com/how-to/add-a-page-meta-description/) if one was missing.~

### DIFF
--- a/charmcraft/templates/init-kubernetes/charmcraft.yaml.j2
+++ b/charmcraft/templates/init-kubernetes/charmcraft.yaml.j2
@@ -68,4 +68,4 @@ resources:
     # used by the 'canonical/charming-actions' GitHub action for automated releases.
     # The test_deploy function in tests/integration/test_charm.py reads upstream-source
     # to determine which OCI image to use when running the charm's integration tests.
-    upstream-source: docker.io/library/ubuntu:24.04  # Change this to your OCI image.
+    upstream-source: docker.io/library/ubuntu:26.04  # Change this to your OCI image.
--- a/charmcraft/templates/init-kubernetes/charmcraft.yaml.j2
+++ b/charmcraft/templates/init-kubernetes/charmcraft.yaml.j2
@@ -64,8 +64,8 @@ resources:
     type: oci-image
     description: OCI image for the 'some-container' container
     # The upstream-source field is ignored by Charmcraft and Juju, but it can be
-    # useful to developers in identifying the source of the OCI image.  It is also
+    # useful to developers in identifying the source of the OCI image. It is also
     # used by the 'canonical/charming-actions' GitHub action for automated releases.
     # The test_deploy function in tests/integration/test_charm.py reads upstream-source
     # to determine which OCI image to use when running the charm's integration tests.
-    upstream-source: some-repo/some-image:some-tag
+    upstream-source: docker.io/library/ubuntu:24.04  # Change this to your OCI image.
--- a/charmcraft/templates/init-kubernetes/charmcraft.yaml.j2
+++ b/charmcraft/templates/init-kubernetes/charmcraft.yaml.j2
@@ -43,7 +43,7 @@ config:
     # An example config option to customise the log level of the workload
     log-level:
       description: |
-        Configures the log level of gunicorn.
+        Configures the log level of the workload.
 
         Acceptable values are: "info", "debug", "warning", "error" and "critical"
       default: "info"
--- a/charmcraft/templates/init-kubernetes/src/charm.py.j2
+++ b/charmcraft/templates/init-kubernetes/src/charm.py.j2
@@ -30,6 +30,8 @@ class {{ class_name }}(ops.CharmBase):
         self.unit.status = ops.MaintenanceStatus("starting workload")
         # To start the workload, we'll add a Pebble layer to the workload container.
         # The layer specifies which service to run.
+        # The charm expects the service to have a long-running command. Don't use a command that
+        # starts the workload as a daemon then exits, otherwise self.wait_for_ready() will fail.
         layer: ops.pebble.LayerDict = {
             "services": {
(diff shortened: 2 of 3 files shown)


dwilding:
@tonyandrewmeyer please could you have a look at this one? Thanks!

charmcraft/templates/init-kubernetes/src/charm.py.j2
```diff
@@ -35,7 +35,7 @@ class {{ class_name }}(ops.CharmBase):
                 SERVICE_NAME: {
                     "override": "replace",
                     "summary": "A service that runs in the workload container",
-                    "command": "/bin/foo",  # Change this!
+                    "command": "sleep infinity",  # Change this!
```
lengau, line 41: What about using `/bin/false`? That way it's a real command but instead of hanging indefinitely it fails instantly

lengau (approved):
Thanks!

bepri (approved):
Approving assuming the feedback on `sleep infinity` is taken

charmcraft/templates/init-kubernetes/src/charm.py.j2
dwilding: We need something that doesn't exit - mimicking a server process. Otherwise, the charm detects that the service isn't running, then errors out. Maybe this deserves a comment in the code?

charmcraft/templates/init-kubernetes/charmcraft.yaml.j2
```diff
@@ -64,8 +64,8 @@ resources:
     type: oci-image
     description: OCI image for the 'some-container' container
     # The upstream-source field is ignored by Charmcraft and Juju, but it can be
-    # useful to developers in identifying the source of the OCI image.  It is also
+    # useful to developers in identifying the source of the OCI image. It is also
     # used by the 'canonical/charming-actions' GitHub action for automated releases.
     # The test_deploy function in tests/integration/test_charm.py reads upstream-source
     # to determine which OCI image to use when running the charm's integration tests.
-    upstream-source: some-repo/some-image:some-tag
+    upstream-source: docker.io/library/ubuntu:24.04  # Change this to your OCI image.
```
tonyandrewmeyer: Maybe 26.04 instead?

charmcraft/templates/init-kubernetes/src/charm.py.j2
tonyandrewmeyer: I think a small comment explaining that the service shouldn't exit (in particular, shouldn't daemonise itself) would be worthwhile. I had a quick look at the pebble docs, and couldn't see that mentioned there either. It's clear enough if you understand that Pebble is managing the services and will restart ones that fail, but might not be to someone very new to Pebble.

tonyandrewmeyer (approved):
Makes sense to me, thanks!

Nice catch with the config, too.

charmcraft/templates/init-kubernetes/charmcraft.yaml.j2
dwilding: Ah yes, changed now. I originally looked at `base` and matched that. But there's no need for them to match, is there, as potentially we could support a wide range of Ubuntus as workload images. Might as well use the latest.

Thinking about `base` - I think it's best to leave that at 24.04 for now. When `charmcraft init` gets support for a base arg, we can adjust the profile to respect that.

charmcraft/templates/init-kubernetes/src/charm.py.j2
dwilding: I've extended the comment above the layer definition:

> The charm expects the service to have a long-running command. Don't use a command that starts the workload as a daemon then exits, otherwise self.wait_for_ready() will fail.

How does that seem?

tonyandrewmeyer: Good, although it's a broader problem than the tests, since Pebble will continually try to restart it, too, or if it exits very quickly show it as failed.

dwilding: Ah I see what you mean. "self.wait_for_ready() will fail" is talking about the charm itself, but only covers the exit-within-1s case.

I could change the comment to this, if you think it would be better?

> The charm expects the service to have a long-running command. Don't use a command that starts the workload as a daemon then exits, otherwise Pebble either reports the service as stopped (if the command exits quickly) or keeps restarting the service.

tonyandrewmeyer: Yes, I like that more, I think it'll help newcomers understand more. Thanks!

dwilding:
@lengau / @bepri are you OK with where we've ended up? Keeping `sleep infinity` but with a comment explaining how `command` is expected to work.

bepri:
@dwilding I think the comment makes sense given the circumstances, I'm happy with this solution :)

charmcraft/templates/init-kubernetes/src/charm.py.j2
lengau: Thanks for the clarification. I'm happy with this!

Issue #1369, opened 2023-11-01
Snap not building on platforms without a cryptography wheel


lengau:
### Bug Description

The core22 snap isn't building on architectures that don't have a cryptography wheel

### To Reproduce

Build the snap on armhf or something :-)

### Environment

CI
https://launchpadlibrarian.net/695323802/buildlog_snap_ubuntu_jammy_ppc64el_charmcraft-edge_BUILDING.txt.gz
https://launchpadlibrarian.net/695324170/buildlog_snap_ubuntu_jammy_armhf_charmcraft-edge_BUILDING.txt.gz
https://launchpadlibrarian.net/695323868/buildlog_snap_ubuntu_jammy_s390x_charmcraft-edge_BUILDING.txt.gz

### charmcraft.yaml

```shell
N/A
```

### Relevant log output

```shell
::            Running `/tmp/pip-install-50bahshq/cryptography_46abaca5ca52471393e375cff8ae3c61/src/rust/target/release/build/cryptography-cffi-c7c4f1b7fb7b6d4d/build-script-build`
[...]

PR #2818, opened 2026-08-12
fix(snap): install python3-apt as a stage-package instead of building from source

lengau:
## Problem

The snap build fails with the current `uv` version:

\`\`\`
error: Source distribution
       \`https://launchpad.net/ubuntu/+archive/primary/+sourcefiles/python-apt/2.7.7ubuntu5/python-apt_2.7.7ubuntu5.tar.xz\`
       has a non-PEP 625-compliant filename; only \`.tar.gz\` and \`.zip\` archives are accepted
\`\`\`

`uv` now hard-rejects source distributions whose filenames aren't PEP 625
compliant (only `.tar.gz`/`.zip`), and Launchpad only serves the
`python-apt` source tarball as `.tar.xz`. There's no `uv` flag to bypass
this check, so `requirements-noble.txt` (installed via `uv pip install`
in the snap's `override-build`) can no longer be used.

We also can't switch to the prebuilt `python-apt` wheels used for local
dev (`python-apt-wheels` index in `pyproject.toml`), since the snap is
built with `PIP_NO_BINARY`/`UV_NO_BINARY` set to force building
everything from source (to avoid issues like #1369), and those wheels
aren't built in a way that's safe to ship in the snap.

## Fix

Install `python3-apt` as a normal apt stage-package in the
`charmcraft-libs` part (alongside its shared library and other runtime
deps, resolved automatically by snapcraft), and `organize` its files into
the charmcraft venv's `site-packages`, instead of trying to `pip`/`uv`
install it from source. This sidesteps the PEP 625 restriction entirely
and avoids compiling `python-apt` at build time.

- Removed `requirements-noble.txt` and the `uv pip install -r
  requirements-noble.txt` step from `snap/snapcraft.yaml`.
- Added `python3-apt` to `charmcraft-libs` stage-packages, with an
  `organize` mapping (using globs for the arch/version-specific `.so`
  files) into `lib/python3.12/site-packages/`.
- Dropped the now-unneeded `libapt-pkg-dev` build-package (only needed
  for compiling `python-apt` from source).
- Updated a stale comment in `.github/workflows/policy.yaml` referencing
  the removed `requirements-noble.txt`.

## Testing

- Built the snap locally with `snapcraft_edge pack --platform amd64`:
  succeeds.
- Installed the resulting snap and confirmed `apt`/`apt_pkg`/`apt_inst`/
  `aptsources` are present under `lib/python3.12/site-packages` and
  importable, and `charmcraft --version` works.
- Running the full `spread -reuse -resend multipass:` suite to confirm
  no regressions before taking this out of draft.

Marking as **draft** until the full spread suite finishes.


tests/unit/parts/plugins/test_charm.py
lengau, line 1: Unrelated, just fixes lint.

tests/unit/parts/plugins/test_reactive.py
lengau, line 1: Unrelated, just fixes lint.

tests/conftest.py
lengau, line 1: Unrelated, just fixes lint.

bepri (approved):

mr-cal (approved):

PR #2828, opened 2026-08-18
feat(init): default Ubuntu 26.04 framework charms to non-root

alithethird:
Scopes the non-root default to the six 12-factor framework profiles targeting Ubuntu 26.04 and higher: Django, ExpressJS, FastAPI, Flask, Go, and Spring Boot.

The generated `*-framework-26.04` charm configurations now set `charm-user: non-root`. Templates targeting Ubuntu 22.04 and 24.04 remain unchanged and continue to omit `charm-user`, so Juju runs those charms as root by default. Regression coverage renders all six profiles and verifies the behavior for both base ranges. The framework reference documentation uses synced Ubuntu-base tabs and explains that the matching Ubuntu 26.04 rock provides `/app-data` outside `/app` as an application data directory writable by the `_daemon_` user.

Depends on canonical/rockcraft#1332: https://github.com/canonical/rockcraft/pull/1332

Validation:
- `uv run pytest -q tests/integration/commands/test_init.py -k framework_profile_charm_user` — 6 passed
- `uv run --group dev ruff check tests/integration/commands/test_init.py`
- `uv run --group dev ruff format --check tests/integration/commands/test_init.py`
- `uv run --group docs sphinx-lint` on the changed RST files
- `uv run --group docs sphinx-build --fail-on-warning --keep-going -b dummy docs /tmp/charmcraft-docs-dummy`
- Pre-commit hooks on the changed files

---

- [ ] I've signed the [CLA](http://www.ubuntu.com/legal/contributors/).


docs/reuse/reference/extensions/non_root_runtime.rst
```diff
@@ -0,0 +1,10 @@
+Runtime user and writable directory
+-----------------------------------
+
+When Charmcraft generates a version 2 12-factor charm targeting Ubuntu 26.04 LTS
```
lengau: Isn't the version 1/version 2 stuff kinda internal that the user doesn't really have to be aware of except that the 26.04+ stuff behaves differently?


```suggestion
When Charmcraft generates a 12-factor charm targeting Ubuntu 26.04 LTS
```

lengau (approved):
LGTM. The docs stuff is a question that I'm fine with whichever answer you and @jahn-junior agree on.

docs/reuse/reference/extensions/non_root_runtime.rst
alithethird: Thank you for pointing it out, fixed the issue.

docs/reuse/reference/extensions/non_root_runtime.rst
```diff
@@ -0,0 +1,17 @@
+Runtime user and writable directory
+-----------------------------------
+
+.. tab-set::
+
+    .. tab-item:: Ubuntu 22.04 and 24.04
+        :sync: base-22-24
+
+        The generated ``charmcraft.yaml`` doesn't set ``charm-user``, so Juju
+        runs the charm as the root user by default.
```
erinecon, line 23: Do you expect that web app developers will understand the impact of this choice? I think it would be good to add at least one more sentence about what this means.

For example: Does this choice have security implications? Will the charm (or Juju) be able to access or modify system files? Or will the charm (or Juju) be restricted from certain actions?

docs/reuse/reference/extensions/non_root_runtime.rst
```diff
@@ -0,0 +1,17 @@
...
+
+    .. tab-item:: Ubuntu 26.04 and higher
+        :sync: base-26-plus
+
+        The generated ``charmcraft.yaml`` sets ``charm-user: non-root`` by
+        default. The matching rock provides ``/app-data`` outside ``/app`` as
```
erinecon, line 29: Flagging here to confirm that all 6 extensions have an `/app` directory

docs/reuse/reference/extensions/non_root_runtime.rst
```diff
@@ -0,0 +1,17 @@
...
+        an application data directory writable by the ``_daemon_`` user.
```
erinecon, line 30: Similar to the other tab, do we expect web app developers to understand the impact of this choice?

erinecon (commented):
Thanks for updating the documentation! Some initial questions from me

docs/reuse/reference/extensions/non_root_runtime.rst
alithethird: They do. We have a [pr](https://github.com/canonical/rockcraft/pull/1348) for that right now.

docs/reuse/reference/extensions/non_root_runtime.rst
alithethird: I think this is a very good point. There is an explanation in [here](https://discourse.charmhub.io/t/juju-3-6-0-released/16027) but  I can not confirm right now because discourse is down :(. 

After I confirm I will link it in here.

mr-cal (approved):

docs/reuse/reference/extensions/non_root_runtime.rst
erinecon: Consider cross-referencing https://canonical.com/juju/docs/charmcraft/4/reference/files/charmcraft-yaml-file/#charm-user instead of a Discourse post -- this page seems to have more information about ``charm-user`` that would be helpful here 

The target for the ``charm-user`` section is `charmcraft-yaml-key-charm-user` if you agree that we should cross-link this page.

docs/reuse/reference/extensions/non_root_runtime.rst
```diff
@@ -0,0 +1,17 @@
+Runtime user and writable directory
+-----------------------------------
+
```
erinecon, line 3: Before the tabs, should we have a lead-in sentence to provide more context + set a mental map between the `charm-user` key in the YAML file and the runtime user/writable directory? Maybe something like:
```suggestion
Runtime user and writable directory
-----------------------------------

The base (or system foundation) of your charm project
file affects how Juju interacts with your charm.

```

I'm not totally happy with this suggestion BTW -- it doesn't specifically mention the write permissions for the daemon user. I'm interested to read your thoughts and ideas here!

docs/reference/extensions/django-framework-extension.rst
```diff
@@ -13,6 +13,8 @@ keys that a user may interact with.
     If you'd like to see the full contents contributed by this extension,
     see :ref:`How to manage extensions <manage-extensions>`.
 
+.. include:: /reuse/reference/extensions/non_root_runtime.rst
```
erinecon: Nit: I wonder whether this information should be listed first on the Django page. I think that the database requirement has more impact on the user (regardless of base). WDYT?

erinecon (approved):
Documentation LGTM, I've kept my nit open about reorganizing the new section on the Django page, but it's non-blocking

lengau:
@alithethird please message me when you're ready to merge (basically, your decision about @erinecon's remaining nit :-) )

alithethird:
@lengau I did apply Erin's nitpick. we are good to go. Thank you!

PR #2853, opened 2026-08-26
test(plugins): split package install command tests

### LINKED ISSUE #2849: Refactor: Split up plugin tests
The plugin unit tests are pretty long and getting kind of unwieldy. As pointed out by @bepri at https://github.com/canonical/charmcraft/pull/2835/changes#r3832956749 we should break them up, but I don't want to break them up within that PR since it's unrelated work that makes it less obvious that we're testing the same thing.

hawkxdev:
## Description

Splits `test_get_package_install_commands` in the plugin unit tests into focused tests, one behaviour each, as discussed in #2849.

### Changes:
- Add `tests/unit/parts/plugins/conftest.py` with a `make_plugin` factory fixture and copy-command helpers, replacing the plugin construction duplicated across the three files.
- In `tests/unit/parts/plugins/test_python.py`, split the test into pip options, build-subdir resolution, copy commands per existing directory, and parent-directory exclusion.
- Apply the same split to `test_poetry.py` and `test_uv.py`.
- Replace `pytest_check` soft checks with plain `assert` now that each test covers a single phase.

Fixes #2849

### DIFF
--- a/tests/unit/parts/plugins/conftest.py
+++ b/tests/unit/parts/plugins/conftest.py
@@ -1,0 +1,80 @@
+# Copyright 2026 Canonical Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# For further info, check https://github.com/canonical/charmcraft
+"""Shared fixtures for plugin unit tests."""
+
+import pathlib
+import typing
+from collections.abc import Callable
+
+import craft_parts
+import pytest
+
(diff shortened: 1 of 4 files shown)


tests/unit/parts/plugins/conftest.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+@pytest.fixture
+def copy_command(install_path: pathlib.Path) -> Callable[[pathlib.Path], str]:
+    """Return a function building the command that copies ``source`` into the charm.
+
+    Quoted the same way as ``utils.get_charm_copy_commands`` so paths with
+    spaces compare equal.
+    """
+
```
lengau: I don't understand why we can't just use `get_charm_copy_commands` here.

lengau (commented):
Thank you! I just have one question...

tests/unit/parts/plugins/conftest.py
hawkxdev: Updated in 2f5b629d. I originally kept the expectation independent to avoid deriving it from the same helper used by the plugin. Since command formatting is already covered in tests/unit/utils/test_parts.py, expected_copies now uses get_charm_copy_commands(), while split_copy_commands() remains independent to catch extra copy commands emitted by the plugin

hawkxdev:
Hi @lengau, the changes are pushed and CI is green. Anything else?

